PR #98, opened 2026-05-13
Integrated Authentication (Kerberos / SSPI) — spec 042

oluies:
Adds Integrated Authentication support: Kerberos on POSIX via system GSSAPI, Windows SSPI deferred to a follow-up (see Phase 4 plan doc included).

End-to-end POSIX user experience:

```sh
kinit alice@CORP.EXAMPLE.COM
```

```sql
LOAD mssql;
ATTACH 'Server=sqlhost.corp.example.com;Database=YourDB;Trusted_Connection=yes;Encrypt=yes;TrustServerCertificate=yes'
    AS db (TYPE mssql);

SELECT TOP 5 name FROM db.sys.tables;
```

That's it — the extension picks up the `kinit` ticket from the credential cache, negotiates SPNEGO against the SQL Server SPN, and connects.

## What's in this PR

11 commits spanning the full spec 042 work. Each is a focused, independently-reviewable change:

| # | Commit | Type | Phase |
|---|---|---|---|
| 1 | `f2c8fa1` | docs | Spec 042 design (spec.md, plan.md, research.md, tasks.md, quickstart.md, checklists/requirements.md) |
| 2 | `3cbaeec` | refactor | Phase 1 — `IAuthenticator` interface + `AuthMethod` enum, **no behavior change** |
| 3 | `85e0f44` | feat | Phase 2 — parser surface + LOGIN7 `fIntSecurity` plumbing + `0xED` SSPI token recognition |
| 4 | `14e26b2` | feat | Phase 3 — POSIX Kerberos backend + KDC test stack (`test/kerberos/`) |
| 5 | `f4d1e9d` | fix | Phase 3 ultrareview — 11 findings addressed (LOGIN7 HostName, uint16_t overflow, secret-only Raw mode, etc.) |
| 6 | `2c2aa28` | docs | Post-merge sweep: CLAUDE.md, Kerberos.md (new, mirrors AZURE.md), README, architecture, TESTING, CHANGELOG |
| 7 | `2ce31d6` | fix | Linux `-lkrb5` linker bug + Docker build OOM cap |
| 8 | `2160626` | fix | GCC `-Wreturn-local-addr` false-positive in metadata cache |
| 9 | `93ed4b6` | feat | `mssql_kerberos_auth_test` scalar function (parallels `mssql_azure_auth_test`) |
| 10 | `f23e7e5` | security | SQL_BATCH debug-dump capped to TDS framing + operational security checklist in Kerberos.md |
[...]

PR #108, opened 2026-05-14
ci: drop cache:pip from pr-injection-scan workflow

oluies:
## Summary

Tiny fix for a workflow bug introduced in PR #100.

\`actions/setup-python@v5\` with \`cache: pip\` requires a \`requirements.txt\` or \`pyproject.toml\` in the repo. The mssql-extension repo has neither — the only Python use is the inline \`pip install llm-guard==0.3.16\` in the very next step. Without a manifest, setup-python fails immediately at:

\`\`\`
No file in /home/runner/work/mssql-extension/mssql-extension matched to
[**/requirements.txt or **/pyproject.toml], make sure you have checked
out the target repository
\`\`\`

Surfaced when PR #98 triggered the injection-scan workflow on its updated head — the check went red without the scanner ever running. Probably affects every PR that opens / synchronizes now.

## Fix

Drop \`cache: pip\`. The one \`pip install llm-guard\` is cheap enough to repeat on each run; \`llm-guard\` pulls torch+transformers as transitive deps, so the cache would be ~600 MB anyway, which dwarfs whatever pip-install overhead we'd save.

## Test plan

- [ ] This PR's own pr-injection-scan check passes (no longer dies at setup-python)
- [ ] Re-run pr-injection-scan on PR #98 after merge — should now reach the scanner step and either pass or fail on actual scan content

### DIFF
--- a/.github/workflows/pr-injection-scan.yml
+++ b/.github/workflows/pr-injection-scan.yml
@@ -54,7 +54,9 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-          cache: pip
+          # No `cache: pip` -- that mode requires a requirements.txt or
+          # pyproject.toml in the repo, which we don't have. The one inline
+          # `pip install llm-guard` is cheap enough to repeat on each run.
 
       - name: Install llm-guard
         run: |


oluies:
@VGSML — this is a chicken-and-egg situation.

The Scan PR prose for prompt injection check on this PR is failing for *exactly the reason this PR fixes*. \`pull_request_target\` triggers the workflow as it lives on \`main\` (not the PR's version), so even though the PR removes the broken \`cache: pip\` line, the check still uses the broken workflow.

The check will keep failing until the fix is merged — at which point the next PR's check will pass.

Options:
1. **Admin override merge** (recommended) — branch protection bypass for this one PR. Once merged, all subsequent PRs (including #98 and #99) will get green injection-scan checks.
2. Temporarily remove the injection-scan workflow from the required-checks list, merge this, restore it.

The diff is 3 lines (drop \`cache: pip\` + comment). Safe to merge as-is.

PR #111, opened 2026-05-15
ci(042): Kerberos integration job [WIP — needs SSSD or scope-down]

oluies:
## Status: Draft / WIP

Moves the Kerberos integration CI job from oluies/mssql-extension#13 to upstream. Now that #98 is merged, the \`test/kerberos/\` stack is on hugr-lab/main and this workflow file is the natural place to add CI coverage for it.

## What this adds

\`.github/workflows/kerberos.yml\` — path-filtered CI job that:
- Triggers on PRs touching \`src/tds/auth/**\`, the TDS protocol files, \`test/kerberos/**\`, \`test/sql/integrated_auth/**\`, \`CMakeLists.txt\`, and the workflow itself
- Also reachable via \`workflow_dispatch\`
- Spins up the compose stack with \`docker compose up --build --wait\` (healthchecks gate downstream services)
- Runs \`/run-tests.sh\` in the test-client
- Dumps KDC, SQL, and test-client logs on failure
- Tears down on \`always()\`
- Concurrency group cancels in-progress runs on new pushes

## Known blocker

The first CI runs surfaced a real failure: \`test/kerberos/sql/init.sql\` executes
\`CREATE LOGIN [EXAMPLE.COM\\testuser] FROM WINDOWS\` and SQL Server on Linux requires
SSSD/realmd to resolve Windows principals. Without that, it errors:

\`\`\`
Msg 15401, Level 16, State 1
Windows NT user or group 'EXAMPLE.COM\\testuser' not found. Check the name again.
\`\`\`

The Dockerfile fix in this PR (\`file\` package added to runtime apt list) is real — that was a separate small bug also surfaced by CI. But the init.sql blocker remains.

Two paths forward, deciding here:

**Option A — scope down (recommended)**
Drop the \`CREATE LOGIN\` from init.sql; have \`run-tests.sh\` only exercise:
- Multi-stage Docker build sanity (\`file\` ELF check etc.)
- \`mssql_kerberos_auth_test('sql.example.com', 1433)\` — exercises GSSAPI handshake to derive a service ticket, no SQL Server login needed
- \`kinit\` round-trip

Covers 80% of the regression surface (parser, GSSAPI, SPNEGO continuation, LOGIN7 wiring) at zero infra cost. The full ATTACH+SELECT stays as documented manual test.

**Option B — add SSSD to the SQL container**
~40-60 lines of Dockerfile + sssd.conf inside \`test/kerberos/sql/\`. Possibly needs swapping MIT Kerberos for Samba AD-DC on the KDC side to provide LDAP-like principal lookups. Bigger change; full ATTACH coverage.

Leaning Option A. Open to discussion.

## Commits

- \`ci(042): add Kerberos integration job\` — the workflow file
- \`ci(042): install 'file' in test-client runtime stage\` — Dockerfile fix


oluies:
Closing in favor of a fresh PR off a cherry-pick branch — the original branch was based on oluies/main pre-spec-042-upstream and produced 6 phantom conflicts that don't exist on the actual diff. Reopening with a 2-commit branch built on upstream/main.

Issue #77, opened 2026-02-18
Unable to connect to the named instance


Stan-RED:
I have multiple named instances (different MSSQL versions) installed on the same host, e.g. "localhost\SS2022". When trying to use instance name as a server for secret or directly in the connection as it usually works in other clients, I have a `Cannot resolve hostname 'localhost\SS2022'` error. Without named instance it can't connect due to the timeout error.

VGSML:
For now you can define a constant port for the SQL Server instances and attach it directly to the duckdb

sashapodgoreanu:
Hello, this would be a very useful enhancement.

VGSML:
Hi, thanks for waiting me. 

I don't have Windows PC nearby, but here the full instruction how to configure sql server named instance with to listen specific ports: https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/configure-a-server-to-listen-on-a-specific-tcp-port?view=sql-server-ver17

oluies:
Spec drafted in #115 — spec 045 *Named-Instance Resolution via SQL Browser*.

Approach (per the [MS-SQLR] wire protocol referenced from the [database engine instances](https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/database-engine-instances-sql-server) docs page):

1. Parse `host\instance` out of the `Server` field in ADO.NET, URI (`%5C`), and secret forms.
2. Send a `CLNT_UCAST_INST` packet (`0x04` + ASCII instance name + `0x00`) to `host:1434/udp`.
3. Parse the `tcp;<port>` token out of the `SVR_RESP` response.
4. Hand `(host, discovered_port)` to the existing TDS connect path. LOGIN7 `ServerName` keeps the original `host\instance` form.

Cross-platform (Linux / macOS / Windows). No new vcpkg dependencies. Resolver is stateless per-call — no port caching in v1 (rationale in `research.md` §R5: dynamic ports change on every SQL Server service restart, and a cache silently connects to the wrong port).

Two new settings:
- `mssql_browser_timeout_seconds` (default 3, range 1..30)
- `mssql_named_instance_resolution` (default true; escape hatch for environments that block outbound UDP 1434)

[...]

PR #116, opened 2026-05-16
feat(spec045): named-instance resolver + mock-browser test stack (Phases 0-2)

oluies:
## Summary

Implements spec 045 Phases 0–2 (parser + UDP resolver + docker test stack) for [#77](https://github.com/hugr-lab/mssql-extension/issues/77). **Additive only — the resolver exists but is not yet wired into `MSSQLConnectionInfo::FromConnectionString`**, so this PR introduces no user-visible behaviour change. Phase 3 (ATTACH plumbing) and Phase 4 (docs + SQL tests) land in a follow-up PR.

Per the spec's tasks.md split (recommended by @VGSML on the design review): keep PR #1 pure additive infrastructure with strong test coverage, then PR #2 is a small surgical diff that becomes trivial to review once this lands.

### Phase 0 — Pure parser (`bb1837d`)
- `src/include/connection/instance_resolver.hpp` — public types in `namespace duckdb::mssql`, no DuckDB headers (resolver is logically reusable outside DuckDB, mirroring `iauthenticator.hpp`).
- `src/connection/instance_resolver.cpp` — `ParseBrowserResponse` state machine over the SVR_RESP body (opcode `0x05` + LE u16 size + ASCII `key;value;…`). Tolerates trailing-NUL omission and missing record terminators that real-world responses sometimes exhibit. Throws `std::runtime_error` with a hex-dump diagnostic on malformed input.
- 12 unit tests covering happy paths (single + multi-instance, unknown fields ignored, TCP disabled, empty body, missing trailing `;;`) and error paths (truncated header, wrong opcode, advertised size > buffer, random garbage, missing NUL).

### Phase 1 — UDP transport + Resolve (`674d340`)
- `InstanceResolver::Resolve(host, instance, timeout) → ResolveResult`. POSIX + Winsock paths, hostname normalisation (`(local)` and `.` → `localhost`, FR-011), case-insensitive instance match (FR-004), one retry on timeout (FR-006), four distinct error kinds per research.md §R8.
- Uses `connect()`+`send`/`recv` on the UDP socket so the kernel surfaces ICMP port-unreachable as `ECONNREFUSED` instead of forcing a full timeout.
- 7 end-to-end tests against an in-process `MockBrowser` (loopback UDP listener). Covers happy path, case-insensitive match, instance-not-found, tcp-disabled, malformed-response, silent-browser timeout-and-retry (verifies ~2s wall), and `(local)` alias normalisation.
- New DuckDB settings registered: `mssql_browser_timeout_seconds` (default 3), `mssql_named_instance_resolution` (default true). Phase 3 wires the lookup.

### Phase 2 — Docker test stack (`e165008`)
- `test/named-instance/` — 9-file self-contained stack: mock browser (Python), SQL Server 2022 on the deliberately-non-default port 11433 (so a passing test proves the resolver translated `TESTINST → 11433` rather than falling back to 1433), Ubuntu test-client.
- Mock browser supports fault-injection modes (`respond`, `silent`, `garbage`, `truncate`, `slow`) via `MOCK_BROWSER_MODE` env var.
- Real-UDP healthcheck on the mock — sends a probe `CLNT_UCAST_INST` and asserts the response starts with the `SVR_RESP` opcode (`0x05`).
- Verified end-to-end on macOS Apple Silicon (`docker compose up -d --wait && docker compose exec test-client /run-tests.sh` → `ALL SMOKE TESTS PASSED`).
- Two-hostname layout (`browser.example.com` + `sql.example.com`), per research.md §R3 and @VGSML's review feedback on #115.
- CLI mode added to `test_instance_resolver` (`--resolve <host> <port> <instance> [timeout]`) so the docker run-tests.sh exercises the real C++ resolver, not a Python smoke double.

## What's NOT in this PR (Phase 3 / PR #2)

- Connection-string parsing of `host\\instance` in `MSSQLConnectionInfo::FromConnectionString`.
- LOGIN7 `ServerName` carrying `host\\instance` (line 340 of `tds_connection.cpp`'s misleading *"we don't use SQL Browser service"* comment stays for this PR; the next PR removes it).
- SQL tests under `test/sql/named_instance/` — depend on the ATTACH wiring.
- Docs + `quickstart.md` + #77 close.

## Test plan

- [x] `make test-instance-resolver` — 19/19 unit tests pass (12 parser + 7 resolver), ~3s wall (silent-browser test contributes ~2s by design).
- [x] Mock browser smoke against local Python process — happy path, garbage, silent (~4s = 2× retry), truncate all produce the expected error categories.
- [x] Full docker stack on macOS Apple Silicon — `compose up -d --wait` succeeds, `run-tests.sh` passes 5 smoke steps.
- [ ] CI green on Linux runners (Linux verification deferred to the PR's CI run).
- [ ] `make release` builds with the new `instance_resolver.cpp` in the extension TU list — deferred to CI; the file is self-contained stdlib-only and the CMake addition is a single line.

## Reviewer notes

- `instance_resolver.hpp` deliberately has zero DuckDB includes — same layering rule as `iauthenticator.hpp` (memory note: [[feedback_iauthenticator_layering]]). Errors are `std::runtime_error`; the eventual call site in Phase 3 will translate to `IOException`.
- The `MockBrowser` C++ class in `test_instance_resolver.cpp` is POSIX-only (uses raw BSD sockets); the docker stack provides the cross-platform coverage. The unit-test binary skips those tests on `_WIN32` with a clear log line.
- The Browser response parser ignores unknown fields (forward-compat with future SQL Server builds that add `rpc`, `via`, etc. fields we don't care about).

Refs #77. Closes nothing in this PR — Phase 3 + 4 follow.


VGSML (dismissed):

VGSML:
One moment: is the browser service responses in ASCII? I thought UTF16LE?

VGSML (dismissed):

oluies:
> One moment: is the browser service responses in ASCII? I thought UTF16LE?

will check https://winprotocoldoc.z19.web.core.windows.net/MC-SQLR/%5BMC-SQLR%5D.pdf

2.2 
All text strings are represented as a
multibyte character set (MBCS)  and so on.

oluies:
## Wire encoding clarification — MBCS, not ASCII

Reviewer flag (thanks @VGSML ) — every docstring and comment in this PR called the MC-SQLR wire format ASCII. The spec is actually MBCS. Quoting [MC-SQLR] v20.0 (2025-10-31) verbatim:

**§2.2 Message Syntax**:
> All integer fields are represented in little-endian format. All text strings are represented as a multibyte character set (MBCS) string [MS-UCODEREF] on the current system code page of the server and the client. (The system code page on the client and the system code page on the server are assumed to be common.) They are not case-sensitive.

**§2.2.3 CLNT_UCAST_INST**:
> INSTANCENAME (variable): A variable-length null-terminated multibyte character set (MBCS) string that does not need to be byte-aligned. It MUST be no greater than 32 bytes in length, not including the null terminator.

**§2.2.5 SVR_RESP**:
> RESP_DATA (variable): A variable-length MBCS string that does not need to be byte-aligned. The maximum size of RESP_DATA MUST be 1,024 bytes if the server is responding to a CLNT_UCAST_INST request.

### Implementation impact: none

The parser is correct under MBCS without changes because:

1. **Tokenisation** keys off `;` (0x3B), which is invariant across every MBCS codepage (0x00–0x7F is identical to ASCII in DBCS/MBCS schemes). No risk of a trail byte from a CP932 DBCS character being mistaken for a separator.
2. **Instance-name case-fold** uses `std::tolower((unsigned char))` — correct for the documented `[A-Za-z0-9_$#]{1,16}` grammar SQL Server enforces at install time, which is the *only* legal instance name in practice. §2.2.5 itself caps it: *"INSTANCENAME ... SHOULD be no greater than 16 MBCS characters."*
3. **`tcp` value parse** uses `std::stoi` over ASCII decimal digits — identical bytes in every MBCS codepage.
4. **`ServerName` and other passthrough fields** are stored as opaque `std::string` byte buffers. If a server somehow had a non-ASCII NetBIOS name under a foreign codepage, those bytes survive intact to the downstream consumer (the TCP connect's `getaddrinfo`, the Kerberos SPN derivation in Phase 3).

### What's ASCII in practice anyway

Every documented field on the wire is, in practice, ASCII:

| Field | Type | Why ASCII |
|---|---|---|
| Keys (`ServerName`, `InstanceName`, `tcp`, `np`, `Version`, `IsClustered`) | Spec literals | Fixed strings |
| `InstanceName` value | `[A-Za-z0-9_$#]{1,16}` | Microsoft install-time constraint |
| `tcp` value | Decimal port digits | ASCII |
| `IsClustered` | `Yes` / `No` | ASCII |
| `Version` | `15.0.4123.1` | ASCII |
| `ServerName` value | DNS / NetBIOS hostname | RFC 952/1123 (ASCII); NetBIOS rules predate IDN |

### Commits in this PR

- `02b6e4c` (pushed) — documentation correction across spec.md, research.md, and the code comments. No code change. 19/19 unit tests still pass.

Full citation lives in `specs/045-named-instance-resolution/research.md` §R1 for future readers.

VGSML:
So we need to encode strings. For simplicity we can limit to only ascii symbols in names (strings)

VGSML (dismissed):

VGSML (approved):

Issue #82, opened 2026-02-24
Application Name in connection string


aletome:
How to specify Application Name in the connection string?
I tried with below syntax, but I see `DuckDB MSSQL Extension` as Application Name in SQL Server Profiler.

```
ATTACH 'Server=myserver,49477;Database=mydb;User Id=myuser;Password=mypassword;Application Name=abc;' AS sqlserver (TYPE mssql);
```

Thanks.

VGSML:
Is not supported, will be added in the future releases

PR #118, opened 2026-05-17
feat(047): process-wide state cleanup + Application Name; closes #96, #82

### LINKED ISSUE #96: Catalog Error: MSSQL Error: Context 'dbalias' already exists. Use a different name or DETACH first
Hi,

We discovered an issue during an `ATTACH` statement with the `mssql` extension. In environments with a lot of contention, a new DuckDB connection that tries to `ATTACH` to a SQL Server database gets the following error:

```text
Catalog Error: MSSQL Error: Context 'dbalias' already exists. Use a different name or DETACH first
```

I am using DuckDB.Net 1.5.0, and I can reproduce the same issue with the Python library 1.5.2.

I am not sure whether the issue is related to the `mssql` extension or to DuckDB itself, but I do not get the same error when using another extension such as `postgres`.

Steps to reproduce in Python:

```python
import duckdb

sql_script = """
INSTALL mssql FROM community;
LOAD mssql;
CREATE OR REPLACE SECRET mssql_secret (
    TYPE mssql,
    HOST 'host',
    PORT 1433,
    DATABASE 'database',
    USER 'user',
    PASSWORD 'password',
    USE_ENCRYPT FALSE,
    CATALOG TRUE
);
ATTACH '' AS TO_MSSQL (
    TYPE mssql,
    SECRET mssql_secret
);
"""

for i in range(1, 10):
    con = duckdb.connect(database=':memory:')
    try:
        con.execute(sql_script)
        print(f"Iteration {i}: Script executed successfully.")
    except Exception as e:
        print(f"Iteration {i}: Error executing script: {e}")
    finally:
        con.close()
```

Additional information: if I add `con.execute("DETACH TO_MSSQL;")` before `close()`, the issue doesn't reproduce, but I cannot use that workaround in my case.

VGSML:
## Summary

Removes three extension-internal process-wide singletons, replaces them with
per-`MSSQLCatalog` ownership via `unique_ptr`, hardens the security model
around ATTACH credentials + Azure token caching, and adds three small
user-facing features that landed naturally on top.

**Closes [#96](https://github.com/hugr-lab/mssql-extension/issues/96)** —
the long-running ATTACH/DETACH-in-Python-loop crash class. **Closes
[#82](https://github.com/hugr-lab/mssql-extension/issues/82)** — custom
LOGIN7 `program_name`. Spawns [#119](https://github.com/hugr-lab/mssql-extension/issues/119)
(→ future spec 049) for lazy credential materialization.

## What's actually deleted vs kept

| Status | Singleton | Replacement |
|---|---|---|
| DELETED | `MssqlPoolManager` | per-`MSSQLCatalog` `unique_ptr<ConnectionPool>` |
| DELETED | `MSSQLContextManager` (spec 045 band-aid `70a4d90`) | direct `Catalog::GetCatalog(context, name)` lookup |
| DELETED | `MSSQLResultStreamRegistry` | per-catalog `MSSQLCatalog::RegisterStream` / `RetrieveStream` (mutex + unordered_map + UUID v4 keys) |
| KEPT + `[DEPRECATED]` | `MSSQLConnectionHandleManager` | none at process scope — the diagnostic API trio (`mssql_open` / `mssql_close` / `mssql_ping`) plus new `mssql_close_all` takes a connection string, not a catalog name. Group scheduled for removal in a future major release; this singleton retires with them. |
| KEPT + namespaced | `mssql::azure::TokenCache` | key changed from `secret_name` to `(uintptr_t(&DatabaseInstance), cache_key)` so two instances sharing a secret name no longer alias — see FR-012 below. |

Full classification (with Winsock / OpenSSL / thread_local scratch / debug-level
caches) is in
[`state_inventory.md`](specs/047-process-state-cleanup/state_inventory.md).

## Production bug classes fixed

Three patterns repeatedly hit by hugr-lab users on the old singleton-pool
codepath; all three pass post-implementation acceptance tests under
`test/cpp/`:

1. **Cross-instance singleton-pool aliasing** (`SC-001`) — two `DuckDB`
   instances ATTACHing different MSSQL targets under the same alias would
   share one pool. Fixed by per-catalog ownership. C++ test:
   `test/cpp/test_multi_instance_pool_isolation.cpp`.
2. **DETACH-then-ATTACH cross-instance pool leak** (`SC-002`) — orphaned
   sockets from instance A surviving DETACH and being handed out to instance
   B's later ATTACH. Fixed by RAII pool teardown in `~MSSQLCatalog`. Same
   test.
3. **100-iteration silent-shutdown socket leak** (`SC-003`, `SC-009`) — the
   verbatim issue #96 repro: 100 iterations of `DuckDB(nullptr) → Connection
   → LOAD mssql → ATTACH → mssql_scan` succeed indefinitely without
   `"Context 'TO_MSSQL' already exists"` regression. C++ test:
   `test/cpp/test_issue_96_attach_loop.cpp`.

## Security hardening (added during PR #118 external review)

- **FR-011 — ATTACH eager credential validation** with `lazy_validation true`
  opt-out (for container/orchestrator scenarios). Wrong password /
  unreachable host now surface as ATTACH errors instead of being deferred to
  the first query. The error message audited to never contain the password
  (`T028a` evidence: positive sentinel-substring assertion in
  `test/sql/attach/attach_validates_credentials.test`).
- **FR-012 — TokenCache per-DatabaseInstance namespacing**. Pre-047, two
  DuckDB instances each defining a secret called `mssql_secret` aliased to
  the same TokenCache row, so instance B silently authenticated with
  instance A's token. Now keyed by `(uintptr_t(&db), cache_key)`. Direct
  acceptance test:
  `test/cpp/test_token_cache_isolation.cpp` (5 sub-assertions including the
  C-never-set negative case + tenant-suffixed-key variant).
- **FR-013 — `mssql_close_all()`** shutdown helper. Single call closes
  every diagnostic-API handle; returns the count of closed handles;
  idempotent. Acceptance: `test/sql/tds_connection/close_all.test` (7
  assertions).
- **SC-005 extension — `mssql_pool_stats` credential redaction grep gate**:
  ATTACH with the SQL-auth password as a sentinel substring; `SELECT *
  FROM mssql_pool_stats()` filtered by `WHERE db LIKE '%' || PASS || '%'`
  must return 0. Vacuity guard: companion `COUNT(*) = 1` confirms the
  zero-match isn't because no rows. Test:
  `test/sql/tds_connection/pool_stats_no_credentials.test`.
- **noexcept teardown audit** (`T046k`): explicit `noexcept` + try/catch on
  the five-destructor chain `~MSSQLCatalog` / `~ConnectionPool` /
  `~TdsConnection` / `~TdsSocket` / `~TlsTdsContext` + `~TlsImpl`. A throw
  during `~AttachedDatabase` unwind would invoke `std::terminate`; this
  surfaces the contract.
- **Pool-destruction invariant** (`T046l`): debug-only
  `D_ASSERT(active_connections_.empty())` at the top of
  `ConnectionPool::Shutdown()`. Surfaces DuckDB-quiescence-contract
  violations as test failures instead of silent UB. Did NOT fire across the
  full integration suite.
- **Class doc-comment** (`T046m`): explicit destruction contract on
  `class ConnectionPool` — sockets close immediately, in-flight TDS sees
  connection-reset on next read, TCP-FIN rollback within seconds, no
  graceful ATTENTION (cross-thread socket write would be unsafe;
  cooperative cancel = follow-up spec, not 047).

## Bonus: custom Application Name (issue #82)

`Application Name=...` (and `ApplicationName=` / `App Name=` /
`application_name=` ADO.NET variants), `?applicationname=` URI query
parameter, and secret `application_name` (with `applicationname` fallback)
now propagate to LOGIN7 `program_name` — visible as `APP_NAME()` /
`sys.dm_exec_sessions.program_name`. 128-char client-side clamp matches SQL
Server's own program_name limit so what the user sees equals what we sent.
Acceptance: `test/sql/attach/application_name.test` (24 assertions across
5 groups including a per-catalog isolation pair).

**Critical deviation note (in commit body for transparency)**: tasks.md
T064 plumbing through the spec-031 `AuthenticationStrategy` framework alone
is dead code — the SQL / FEDAUTH / integrated paths in
`MSSQLCatalog::Initialize` call `TdsConnection::Authenticate*` DIRECTLY,
not via `AuthenticationStrategy::GetLogin7Options()`. Fixed by also
threading `app_name` through the three `TdsConnection::Authenticate*`
methods + both `DoLogin7*` helpers + the three factory closures + the four
`Validate*` paths. Strategy-side plumbing kept for future-proofing.

## Test plan

`make test` (unit, no SQL Server) + `make integration-test` (SQL Server +
`[sql]` groups) + the spec 047 acceptance binaries:

- [x] `make test-spec047-us1` — `test_multi_instance_pool_isolation` +
  `test_issue_96_attach_loop` (SC-001/002/003/009)
- [x] `make test-spec047-us3` — `test_result_stream_registry_isolation`
  (SC-006)
- [x] `make test-spec047-us-sec` — `test_token_cache_isolation` (SC-011)
- [x] `test/sql/attach/attach_validates_credentials.test` — SC-010 + T028a
- [x] `test/sql/tds_connection/close_all.test` — FR-013 smoke (7 asserts)
- [x] `test/sql/tds_connection/pool_stats_no_credentials.test` — SC-005
  redaction (6 asserts)
- [x] `test/sql/attach/application_name.test` — SC-012 (24 asserts across
  5 groups)
- [x] `make integration-test` — final sweep, exit 0 after every spec 047
  commit landed (304 assertions in `[sql]`; `[integration]` half also
  exit 0)
- [x] `T052` bench parity gate (SC-007) — 3× spec 047 runs of
  `test/bench/bench_codec_e2e.sh` at 1M rows + single baseline on
  `1ae9fb8` (origin/main HEAD = spec 047 fork point). **SC-007 MET WITH
  DEVIATION**: per-row codec path (`insert_values`, the only
  codec-heavy step that amortizes ATTACH) within +0.2% of main;
  ATTACH-bound steps each show +~1s regression that is EXACTLY the
  FR-011 eager-validation TCP+LOGIN7 cost (decomposed via spot-check:
  `catalog false` shaves the catalog-enum round-trip, leaving the
  validation round-trip). Operator opt-out via
  `lazy_validation true`. Full report at
  [`specs/047-process-state-cleanup/bench_results.md`](specs/047-process-state-cleanup/bench_results.md).
- [x] `T053` clang-format-14 sweep — 16 files touched, +56/−60 LOC,
  purely cosmetic; integration tests still exit 0 post-sweep.
- [x] `T054` final full-suite test gate — `make test` 113 cases / 3439
  assertions PASSED (22 env-gated skips); `make integration-test` 16
  cases / 304 assertions PASSED. Environmental note: the dev tenant's
  Azure SQL admin password expired since US3 closure; FR-011 now
  surfaces this as an ATTACH-time failure in
  `test/sql/azure/azure_lazy_loading.test` (the test passed pre-047
  only because ATTACH was lazy and the first query failed silently).
  Skipped by unsetting `AZURE_SQL_TEST_DSN` — pure environment issue,
  not a code regression.

## Files / LOC

20 commits over 4 days; spec design + 6 implementation phases + Polish.
~17 commits ahead of `origin/047-process-state-cleanup` at PR open;
**3 sources singletons gone**, **6 destructors made explicit `noexcept`**,
**~750 LOC net delta** including 8 new test files (4 C++ + 4 SQL).

## Public API surface

- Additive: `mssql_close_all()` scalar (FR-013).
- Additive: `application_name` ATTACH option / URI param / secret field
  (FR-014).
- Additive: `lazy_validation true` ATTACH option +
  `mssql_attach_validation_timeout` setting (FR-011).
- Deprecation (label-only, no behavior change): `mssql_open` / `mssql_close`
  / `mssql_ping` / `mssql_close_all` marked `[DEPRECATED]` in CLAUDE.md
  table + CHANGELOG + source comments at registration sites (FR-010).
- No removals. No behavior change for existing callers.

## Follow-ups (separate work)

- **Lazy credential materialization**
  ([issue #119](https://github.com/hugr-lab/mssql-extension/issues/119))
  → future spec 049: zeroize passwords / tokens at rest in
  `MSSQLConnectionInfo` so a `MSSQLCatalog` instance does not hold
  long-lived plaintext credentials. Spawned by the FR-011 security review.
- **Cooperative TDS cancellation** (future spec): replace
  `~ConnectionPool` immediate-close semantics with atomic flag polled by
  the owning thread so a graceful TDS ATTENTION cancel can be sent.
  Tracked in spec 047 Constraints / non-goals.
- **Removal of `mssql_open` / `mssql_close` / `mssql_ping` /
  `mssql_close_all`**: future major release retires the diagnostic API
  entirely; `MSSQLConnectionHandleManager` goes with them. Reduces
  extension-internal singleton count from 2 to 1 (TokenCache).


VGSML:
### Security review fold-in (commit `1a3087c`)

Following the external security review, the following 5 items are now folded into the spec; 2 items are explicitly out-of-scope and tracked separately.

**Folded into spec 047:**

| Review item | Spec change |
|---|---|
| #1 TokenCache cross-instance aliasing | **FR-012** — cache key changes from `secret_name` to `(database_instance_id, secret_name)`. **SC-011** — new regression test (`test/cpp/test_token_cache_isolation.cpp`). |
| #2 Password leak in ATTACH error | **FR-011 security requirement** added; **T028a** audit task; **SC-010 case 1** asserts sentinel password substring is NOT in error message. |
| #3 Credentials in `mssql_pool_stats` after FR-003 rewrite | **SC-005 extended** — grep gate on per-auth-method sentinel substring. New test: `test/sql/diagnostic/pool_stats_no_credentials.test`. |
| #4a `noexcept` on teardown chain | Constraints in plan.md; **T046k** audit task. |
| #4b (partial) Pool teardown contract | **T046l** — debug-only assert "no checked-out connections at pool destruct". **T046m** — doc-comment in `tds_connection_pool.hpp` making the close-immediately + no-graceful-ATTENTION contract explicit. |
| #6 Handle-manager retention concern | **FR-013** — new `mssql_close_all()` scalar function (also `[DEPRECATED]` per FR-010 group). Recommended shutdown hook for long-running embedding processes. |

**Out-of-scope (tracked separately):**

| Review item | Where tracked |
|---|---|
| #5 In-memory credential zeroization | [Issue #119](https://github.com/hugr-lab/mssql-extension/issues/119) — future spec 049 "Lazy credential materialization (secret-deref auth path)". Spec 047 acknowledges 2× footprint tradeoff in FR-001 + Constraints. |
| #4b (full) Graceful in-flight TDS cancellation on DETACH | Future spec covering cooperative cancellation via atomic flag + DuckDB `InterruptCheck` integration. Spec 047 Constraints / non-goals documents the contract: destructor closes sockets immediately; server-side rollback via TCP FIN. |

**Scope deltas:**

- 11 FRs → **13 FRs** (+FR-012, +FR-013)
- 10 SCs → **11 SCs** (+SC-011)
- 6 phases → **8 phases** (new Phase 7 "Security Hardening" with 13 sub-tasks T046a–T046m; old Phase 7 Polish renumbered to Phase 8)
- No changes to public API surface (one additive function `mssql_close_all`, marked `[DEPRECATED]` at registration since it serves the deprecated diagnostic family)
- No new vcpkg dependencies

PR still in DRAFT; design intent unchanged at the architectural level; security gates explicit before implementation begins.

oluies (dismissed):
## Security review — PR #118 (spec 047)

  Reviewed against `4f5f484`. Scope: the three security-hardening claims (FR-011, FR-012, FR-013), the destructor `noexcept`/`D_ASSERT` work, and the new user-controlled `application_name` plumbing.

  Severity: **HIGH** = exploitable or strong likelihood of harm; **MEDIUM** = realistic concern, narrow blast radius; **LOW/INFO** = correctness, hygiene, or design observation.

  ### Summary

  | # | Severity | Area | Title |
  |---|---|---|---|
  | H1 | HIGH | FR-013 | `mssql_close_all()` is process-wide with no authorization |
  | H2 | HIGH | FR-014 | No control-character filter in `ResolveAppName` (SQL Server log injection) |
  | H3 | HIGH | FR-014 | `application_name` ATTACH option silently dropped / hard-errors |
  | M1 | MEDIUM | T046k | Silent exception swallowing in `noexcept` destructors |
  | M2 | MEDIUM | T046l | `D_ASSERT(active_connections_.empty())` is debug-only |
  | M3 | MEDIUM | FR-012 | `OnDetach` does not invalidate tenant-suffixed TokenCache keys |
  | M4 | MEDIUM | FR-012 | TokenCache eviction docs misleading; map never shrinks via TTL |
  | M5 | MEDIUM | FR-012 | Address-reuse residual not covered by `test_token_cache_isolation` |
  | L1 | LOW | refactor | `MSSQLCatalog::CreateConnectionFactory` is dead code |
  | L2 | LOW | refactor | Per-catalog pool ownership doubles credential footprint |
  | L3 | LOW | US3 | Result-stream UUIDs are `std::mt19937`-derived |
  | L4 | LOW | diag API | `mssql_open` raw-error throws may include TDS error text |
  | L5 | LOW | FR-014 | `mssql_open` does not propagate `application_name` |

  ---

  ### HIGH

  #### H1. `mssql_close_all()` is process-wide with no authorization

  `src/connection/mssql_diagnostic.cpp:56-72` (`CloseAll`) acquires the singleton manager mutex and closes every handle from every `ClientContext` in the process. There is no per-context scoping, no permission check, and no opt-in setting. In any deployment where multiple users share a DuckDB process (long-lived server, JupyterHub kernel, Rancher pod), `SELECT mssql_close_all()` is a one-line cross-tenant DoS against in-flight diagnostic-API queries.

  The function is registered alongside the deprecated `mssql_open`/`mssql_close`/`mssql_ping` (documented for future removal), but until they retire this is a real foot-gun.

  **Fix**: scope to handles created by the calling `ClientContext`, or gate behind a `mssql_enable_close_all` setting (default off).

  #### H2. No control-character / newline filter in `ResolveAppName` — SQL Server log injection

  `src/mssql_storage.cpp:50-91`. The 128-UTF-16-code-unit clamp is correctly implemented (handles surrogate pairs and malformed UTF-8 defensively — confirmed), but it does **not** strip `\r`, `\n`, `\t`, NUL, ESC, or other C0 controls. The string lands in `sys.dm_exec_sessions.program_name`, `dm_exec_requests`, and SQL Server error log.

  An attacker who controls any ATTACH path (templated DSN, injected connection-string parameter, malicious secret) can write `Application Name=goodapp\r\n2026-05-19 fake log entry` and confuse SOC pipelines that grep the SQL Server log. Resource Governor classifiers that key on `program_name` can also be confused.

  **Fix**: strip bytes `< 0x20` and `0x7F` before the UTF-16 clamp.

  #### H3. `application_name` ATTACH option is silently dropped or hard-errors

  `src/mssql_storage.cpp:1225-1261`. The ATTACH-option loop extracts `secret`, `azure_secret`,
  `access_token`, `catalog`, `schema_filter`, `table_filter`, `order_pushdown`, `lazy_validation` — but
   **not** `application_name`/`applicationname`. The else-branch (line 1259) leaves the option in
  `options.options`, and the comment at line 1213 explicitly says DuckDB rejects unrecognized options.
  So `ATTACH '...' AS x (TYPE mssql, application_name 'Foo')` fails with "unrecognized option".

  The PR description advertises ATTACH-option support; the PR test
  (`test/sql/attach/application_name.test`) only exercises the embedded-in-connection-string, URI, and
  secret forms, so the gap is undetected.

  **Fix**: either add `application_name` extraction to the ATTACH-option loop, or remove the ATTACH-option advertisement from the docs/CHANGELOG.

  ---

  ### MEDIUM

  #### M1. Silent exception swallowing in `noexcept` destructors

  Confirmed across `~ConnectionPool`, `~TdsConnection`, `~TdsSocket`, `~TlsTdsContext`, `~TlsImpl`: every destructor wraps cleanup in `try { … } catch (...) {}`. OpenSSL `SSL_shutdown`/`SSL_free` failures, socket-close errors, and pool-internal exceptions during teardown are discarded without even a stderr line. The justification ("avoid `std::terminate` during `~AttachedDatabase` unwind") is sound, but the catches are unconditionally silent. Operationally this turns a hung TLS shutdown or a corrupted pool state into "everything looks fine" in production logs.

  **Fix**: emit at least an `MSSQL_DEBUG`-gated `fprintf(stderr, ...)` with the exception type/`what()` inside each catch.

  #### M2. `D_ASSERT(active_connections_.empty())` is debug-only

  `src/tds/tds_connection_pool.cpp:55`. The pool-destruction invariant fires only in debug builds. In release, if the DuckDB quiescence contract is ever violated (catalog destroyed while a connection is still checked out), `Shutdown()` proceeds to `socket->Close()` on a fd another thread may be in `recv()` on. On Linux/macOS POSIX guarantees `EBADF`/connection reset, which is recoverable. **On Windows with Winsock, `closesocket()` against an in-flight `recv()` is race-prone.**

  **Fix**: promote to a release-mode `if (!empty) fprintf(stderr, …)` so the contract violation is visible.

  #### M3. `OnDetach` does not invalidate tenant-suffixed TokenCache keys

  `src/catalog/mssql_catalog.cpp:737` calls `TokenCache::Invalidate(*context.db,
  connection_info_->azure_secret_name)` — bare secret name only. But `AcquireToken`
  (`src/azure/azure_token.cpp:394-398, 446`) writes the row with key `secret_name + ":" +
  tenant_id_override` whenever an interactive-auth tenant override is supplied. Result: DETACH leaves
  tenant-keyed rows behind. `FedAuthStrategy::InvalidateToken`
  (`src/tds/auth/fedauth_strategy.cpp:65-74`) correctly rebuilds the suffix; only `OnDetach` is
  asymmetric.

  Bounded — only the interactive-auth path is affected — but it is a real cache-coherence hole.

  **Fix**: in `OnDetach`, also evict any key starting with `secret_name + ":"`; or maintain a per-instance index of keys to drop.

  #### M4. TokenCache eviction documentation is misleading

  `src/include/azure/azure_token.hpp:102-106` claims "entries belonging to a dead instance become
  unreachable and are evicted naturally by TTL (~1 h max)." False — `CachedToken::IsValid()`
  (`azure_token.hpp:86-89`) only suppresses reads; the underlying `unordered_map` never shrinks except
  via `Invalidate`, `Clear` (zero callers — dead public API), or overwrite.

  Long-lived processes that ATTACH/DETACH against many distinct `(addr, secret_name)` tuples — including the issue #96 100-iteration repro the PR claims to fix — accumulate dead rows indefinitely.

  **Fix**: re-word the comment to "ignored on read; map only shrinks via Invalidate/overwrite," or add an opportunistic erase on `GetToken` when `IsValid() == false`.

  #### M5. Address-reuse residual in FR-012

  TokenCache keyed by `uintptr_t(&db)` fixes *concurrent* same-named-secret aliasing but not the
  *sequential* case. After DatabaseInstance A is destroyed, its rows persist (per M4); allocator reuses
   A's address for new instance B; B's `GetToken` lookups hit A's stale tokens. Practical
  exploitability is low (requires same-process timing + same-name secret + same-allocator-bucket
  reuse). `test_token_cache_isolation.cpp` does not cover the destroy-then-create case.

  **Fix**: evict on `Invalidate`/`OnDetach` to narrow the window, or acknowledge the residual in the header comment.

  ---

  ### LOW / Informational

  #### L1. `MSSQLCatalog::CreateConnectionFactory()` is dead code

  `src/catalog/mssql_catalog.cpp:213-228` builds a SQL-auth-only factory with no `app_name` and no eager validation. Confirmed by `grep`: no callers anywhere in `src/` or `test/`. Should be deleted — if a future refactor unwittingly calls it, FR-011 is bypassed.

  #### L2. Per-catalog pool ownership doubles credential footprint for same-DSN-multi-alias

  PR description references this — motivation for the still-unscoped issue #119 follow-up. Worth noting that PR 118 *increases* the in-memory credential exposure surface (two `MSSQLConnectionInfo` structs each hold the same password) in exchange for the lifetime / pool-isolation properties. Net is positive given the bug class it fixes, but the regression is real.

  #### L3. Result-stream UUIDs are `std::mt19937`-derived

  `src/catalog/mssql_catalog.cpp:85` uses DuckDB's `UUID::GenerateRandomUUID()` which is
  Mersenne-Twister, not a CSPRNG. The stream registry is per-catalog (post-047) and in-process, so
  brute-forcing isn't a practical attack — but a registry that hands out per-row stream IDs is the kind
   of thing static analyzers will flag. Worth a one-line comment justifying the non-CSPRNG choice.

  #### L4. `mssql_open` (diagnostic API) raw-error throws may include TDS error text

  `src/connection/mssql_diagnostic.cpp:141, 145` propagate `conn->GetLastError()` verbatim into thrown `IOException`/`InvalidInputException`. The `Validate*Connection` paths in `mssql_storage.cpp` sanitize via `TranslateConnectionError`; `mssql_open` does not. Diagnostic API is deprecated, so low priority — but worth auditing every `SetLastError` write site in `tds_connection.cpp` to confirm passwords are not stored there.

  #### L5. `MSSQLOpenFunction` does not propagate `application_name`

  Diagnostic-API path uses the default LOGIN7 program_name regardless of any ATTACH-time configuration. Minor inconsistency.

  ---

  ### Passes (worth recording)

  - TokenCache namespacing correctly isolates concurrent DatabaseInstances under the same secret name
  (FR-012 main case verified in `test_token_cache_isolation.cpp`).
  - `TranslateConnectionError` (`mssql_storage.cpp:880-935`) — every branch builds messages from
  `host`/`port`/`user`/`database` only; password never referenced.
  - `Validate*Connection` debug logs emit host/port/user/database, never password.
  - `mssql_pool_stats` output schema includes no credential field; the PR's regression test
  (`pool_stats_no_credentials.test`) holds the line.
  - 128-UTF-16-code-unit clamp in `ResolveAppName` correctly handles surrogate pairs and
  malformed-UTF-8 lead bytes.
  - `LoadAttachValidationTimeout` defaults to `mssql_connection_timeout` (default 30s); no DoS from a
  "block forever during ATTACH" misconfiguration absent an explicit `mssql_attach_validation_timeout=0`
   AND `mssql_connection_timeout=0` combo.
  - Background-cleanup thread is joined before active-connection walk in `~ConnectionPool`.
  - Plaintext password never appears in `mssql_pool_stats` output; the sentinel-based grep test guards
  against future regression.

  ---

  ### Recommended pre-merge actions

  In order of value/effort ratio:

  - [ ] **H2** — Add control-char strip in `ResolveAppName` (~10 lines, no test breakage).
  - [ ] **L1** — Delete `MSSQLCatalog::CreateConnectionFactory` (~15 lines, no caller).
  - [ ] **M4** — Re-word TokenCache header comment to accurately describe non-eviction.
  - [ ] **M3** — Invalidate tenant-suffixed variants in `OnDetach`.
  - [ ] **M1** — Add `MSSQL_DEBUG`-gated stderr emission inside destructor catches.
  - [ ] **H1** — Scope `mssql_close_all` to the calling `ClientContext`, or no-op + deprecation warning
   until the diagnostic API retires.
  - [ ] **H3** — Either add `application_name` to the ATTACH-option loop, or remove the ATTACH-option
  advertisement from docs/CHANGELOG.

  The address-reuse residual (**M5**) and credential-footprint regression (**L2**) can stay in the issue #119 follow-up; they're acknowledged as design choices in the PR.


###  Are we using smart pointers to destroy?

  Yes for lifetime, no for erasure. That distinction matters here.

  What I see in the code:

  // src/include/mssql_storage.hpp:47
  string password;     // plain std::string, default destructor

  // src/catalog/mssql_catalog.cpp:53
  connection_info_(std::move(connection_info))           // shared_ptr<MSSQLConnectionInfo>
  // :201
  connection_pool_ = make_uniq<tds::ConnectionPool>(...)  // unique_ptr<ConnectionPool>

  So:
  - shared_ptr<MSSQLConnectionInfo> ✓ ensures the struct gets deleted when the last MSSQLCatalog (and
  the pool's lambda closure) releases it.
  - unique_ptr<ConnectionPool> ✓ ensures pool teardown is RAII-deterministic.
  - But the content — std::string password, std::string access_token, std::vector<uint8_t>
  fedauth_token_utf16le — just runs the default ~string() / ~vector(), which calls delete[] on the heap
   buffer without overwriting it. The freed pages sit on the allocator's free list with the password
  verbatim until something else happens to reuse them.

  There's also a hidden multiplier: the pool factory captures the password by value into a
  std::function closure (mssql_catalog.cpp:186), so there's a second std::string copy that lives for
  the pool's lifetime alongside the canonical one in MSSQLConnectionInfo. Issue #119 calls this out.

  And I verified earlier: zero hits across src/ for OPENSSL_cleanse, explicit_bzero, SecureZeroMemory, mlock, or any zeroize primitive. Nothing today.

###   Libraries available for C++ credential handling

  In rough order of fit for this project:

  1. OpenSSL (already linked) — zero new deps.
  - OPENSSL_cleanse(void *ptr, size_t len) — non-elidable memset to zero; portable, in
  <openssl/crypto.h>. Drop-in for ~MSSQLConnectionInfo and TokenCache erasure.
  - CRYPTO_secure_malloc_init + OPENSSL_secure_malloc/free — mlock-ed secure heap, zeroed on free.
  Heavier (requires init, fixed-size heap), but defends against swap-to-disk and core dumps too.
  OpenSSL itself uses this for private keys.
  - Best fit for this codebase: already a dependency, statically linked, no vcpkg change.

  2. libsodium — cleaner API, one new vcpkg dep.
  - sodium_memzero, sodium_mlock/munlock, sodium_malloc (allocates with guard pages + canary),
  sodium_free. MIT licensed. ~300 KB static. Cross-platform mlock / VirtualLock abstraction baked in.
  Used by Signal, age, lots of crypto products.
  - Worth it only if you also pick up other libsodium primitives elsewhere.

  3. Botan / Crypto++ — heavy.
  - secure_vector<T> / SecByteBlock with lock_mem allocator. Cleaner C++ class API. Both pull in a full
   crypto stack you don't need.

  4. Platform OS APIs (for a future spec 049 "lazy materialization" angle).
  - macOS Keychain Services, Linux libsecret (D-Bus to gnome-keyring / KWallet), Windows DPAPI /
  Credential Manager. Don't store secrets in the process at all — fetch on demand. More invasive but
  most defensible.

  Don't reach for:
  - memset_s (C11) — not supported in glibc; you'll write the workaround anyway.
  - A hand-rolled volatile char * memset loop — compilers have figured out how to elide these too.

  What I'd actually change in this codebase

  Since OpenSSL is already there and proportionality matters, the minimal high-value patch is:

```
  // src/include/mssql_storage.hpp — add destructor
  struct MSSQLConnectionInfo {
      // ... fields ...
      ~MSSQLConnectionInfo() noexcept {
          if (!password.empty())     OPENSSL_cleanse(password.data(), password.size());
          if (!access_token.empty()) OPENSSL_cleanse(access_token.data(), access_token.size());
          if (!fedauth_token_utf16le.empty())
              OPENSSL_cleanse(fedauth_token_utf16le.data(), fedauth_token_utf16le.size());
      }
  };

```
  Then either stop copying the password into the pool factory closure (re-read from connection_info_ shared_ptr on each refill — kills the duplicate that issue #119 flagged), or wrap the captured strings in a move-only SecureString type that cleanses on destruction.

  For the TokenCache (src/azure/azure_token.cpp:69-72, 74-77), add OPENSSL_cleanse on the access_token field inside Invalidate and Clear before the std::string destructs. Also do it on overwrite in SetToken.

  That gets you the in-process at-rest defense without adopting a new library or rewriting the credential type system. If you later decide it's worth the bigger lift, libsodium gives you the cleanest path to mlock+munlock+canary; OpenSSL's secure-heap gets you there with no new deps but a chunkier API.

VGSML:
> H1	HIGH	FR-013	`mssql_close_all()` is process-wide with no authorization

Al was landed, l except H1 - mssql_close_all was marked as deprecated

Issue #85, opened 2026-03-10
[Bug] can't read from table with partitioned clustered index


keen85:
Hi,
I noticed some limitation when trying to query a table that has a [partitioned clustered index](https://learn.microsoft.com/en-us/sql/relational-databases/partitions/partitioned-tables-and-indexes?view=sql-server-ver17).

in MSSQL:
```sql
CREATE PARTITION FUNCTION [pf_logs_by_month] (DATETIME2(7))
AS RANGE RIGHT
FOR VALUES (
    '2026-02-01 00:00:00'
    , '2026-03-01 00:00:00'
    , '2026-04-01 00:00:00'
);
GO

CREATE PARTITION SCHEME [ps_logs_by_month]
AS PARTITION [pf_logs_by_month] ALL TO ([PRIMARY]);
GO

CREATE TABLE [dbo].[log](
    log_ts DATETIME2(7) DEFAULT (GETUTCDATE()) NOT NULL,
    message NVARCHAR(255) 
);
INSERT INTO [dbo].[log] (message) VALUES ('bar');
CREATE CLUSTERED INDEX [ix_ci_log_log_ts]
    ON [dbo].[log] (log_ts DESC) WITH (DATA_COMPRESSION = PAGE, OPTIMIZE_FOR_SEQUENTIAL_KEY = ON)
    ON [ps_logs_by_month] (log_ts);
GO
```

When trying to query it via DuckDB & mssql-extension, it fails:
```sql
SELECT * FROM attached_db.dbo.log;
```
```
Catalog Error:
Column with name log_ts already exists!
```

VGSML:
Thanks for reporting this. I will take a look

oluies:
Spec drafted: **[spec 049 — fix partitioned-table catalog](https://github.com/hugr-lab/mssql-extension/pull/120)** (PR #120, docs only).

**Root cause** confirmed in `src/catalog/mssql_metadata_cache.cpp`. Three SQL templates do `LEFT JOIN sys.partitions p ON o.object_id = p.object_id AND p.index_id IN (0, 1)`. `sys.partitions` has one row per partition, so for the partitioned clustered index on `dbo.log` (3 boundary values → 4 partitions) the join multiplies every column row 4 times. `MakeTableInfo()` then calls `info.columns.AddColumn("log_ts")` four times and DuckDB throws on the second.

[...]

PR #120, opened 2026-05-19
docs(spec): add spec 049 — fix partitioned-table catalog (issue #85)

oluies:
## Summary

Adds the design bundle for spec **049-fix-partitioned-table-catalog**, the planned fix for [issue #85](https://github.com/hugr-lab/mssql-extension/issues/85) — querying a SQL Server table with a partitioned clustered index fails with `Catalog Error: Column with name <col> already exists!`.

This PR is **spec/docs only**. No source code changes. Implementation will land in a follow-up PR per the tasks in `specs/049-fix-partitioned-table-catalog/tasks.md`.

## Root cause (investigated)

Three SQL templates in `src/catalog/mssql_metadata_cache.cpp`:

- `TABLE_DISCOVERY_SQL_TEMPLATE` (L44)
- `SINGLE_TABLE_METADATA_SQL_TEMPLATE` (L57)
- `BULK_METADATA_SCHEMA_SQL_TEMPLATE` (L79)

each do

```sql
LEFT JOIN sys.partitions p ON o.object_id = p.object_id AND p.index_id IN (0, 1)
```

`sys.partitions` has one row per partition. For an N-partition clustered table the join multiplies every column row N times. `MakeTableInfo()` in `src/catalog/mssql_table_entry.cpp:45-48` then calls `info.columns.AddColumn(...)` on each row → DuckDB rejects the second copy. The `TABLE_DISCOVERY` template doesn't crash (map dedupes by name) but silently reports the row count of an arbitrary single partition.

## Planned fix

Replace each `LEFT JOIN sys.partitions ...` with a pre-aggregated derived table:

```sql
LEFT JOIN (
    SELECT object_id, SUM(rows) AS rows
    FROM sys.partitions
    WHERE index_id IN (0, 1)
    GROUP BY object_id
) p ON p.object_id = o.object_id
```

Same shape as `src/catalog/mssql_statistics.cpp:14-22` which already aggregates correctly (this PR brings the metadata cache into alignment with code that already exists).

## User stories

| ID | Priority | Summary |
| -- | :------: | ------- |
| US1 | P1 | Query a partitioned table without the duplicate-column crash |
| US2 | P2 | `duckdb_tables().estimated_size` = SUM across partitions |
| US3 | P2 | `mssql_preload_catalog` succeeds on schemas with partitioned tables |
| US4 | P2 | Cross-cutting smoke tests: filter pushdown, INSERT, UPDATE/DELETE on PartitionedWithPK, COPY TO |
| US5 | P3 | (Out of scope, deferred) — pre-existing SQL-injection finding in identifier substitution |

## Review findings folded into the spec

1. `mssql_statistics.cpp` already uses the correct `ISNULL(SUM(p.rows), 0)` shape — fix is *converging*, not net-new SQL.
2. **Pre-existing SQL-injection surface** found in `MSSQLMetadataCache` identifier substitution (`'` not escaped, unlike `mssql_statistics.cpp::FetchRowCount`). Documented as US5 and explicitly deferred to a hardening PR so the crash fix isn't blocked.
3. **Cross-cutting gap** — issue #85 reports a read failure, but catalog metadata is the foundation of every DML/COPY path. US4 added smoke tests across filter pushdown, INSERT, UPDATE/DELETE (with a new `PartitionedWithPK` fixture), and COPY TO.

## Files

```
specs/049-fix-partitioned-table-catalog/
├── spec.md                                   (21.5K, 5 user stories, 17 FRs, 7 SCs, security notes)
├── plan.md                                   (12.9K, constitution check, structure decision)
├── research.md                               (7.4K, 5 decisions inc. alternatives matrix)
├── data-model.md                             (7.1K, before/after SQL + worked example using issue #85's table)
├── quickstart.md                             (3.1K, drop-in fix + repro/cleanup)
├── checklists/requirements.md                (3.9K)
├── contracts/metadata-query-contract.md      (5.3K, output-row shape + canonical SQL)
└── tasks.md                                  (15.2K, T001-T013 incl. T009a-d cross-cutting)
```

## Test plan

For *this* PR (spec-only):

- [x] All eight artifacts present and internally consistent (spec ↔ plan ↔ research ↔ tasks).
- [x] Spec checklist passes (`specs/049-fix-partitioned-table-catalog/checklists/requirements.md`).
- [x] Contract pins the canonical SQL shape so future refactors can be caught at unit-test time.

For the *follow-up implementation PR* (tracked in tasks.md):

- [ ] T001-T002 — partitioned-table fixture loads in dockerized SQL Server.
- [ ] T003-T005 — three SQL template fixes (single-line change each).
- [ ] T006-T007 — build + existing unit tests stay green.
- [ ] T008-T009 — regression test (`test/sql/catalog/partitioned_table.test`) passes; full integration suite stays green.
- [ ] T009a-d — cross-cutting smoke tests pass (filter pushdown, INSERT, UPDATE/DELETE, COPY).
- [ ] T010-T011 (optional) — C++ unit-test guard on the aggregation marker.
- [ ] T012 — `CLAUDE.md` "Recent Changes" entry.

Refs #85.

### DIFF
--- a/specs/049-fix-partitioned-table-catalog/checklists/requirements.md
+++ b/specs/049-fix-partitioned-table-catalog/checklists/requirements.md
@@ -1,0 +1,52 @@
+# Specification Quality Checklist: Fix Partitioned Table Catalog
+
+**Purpose**: Validate specification completeness and quality before proceeding to implementation
+**Created**: 2026-05-19
+**Feature**: [spec.md](../spec.md)
+
+## Content Quality
+
+- [x] No implementation details (languages, frameworks, APIs) in spec.md user stories — implementation details live in plan.md/contracts/.
+- [x] Focused on user value (`SELECT *` works; row counts are accurate) and business need (unblocks log/event tables in DuckDB).
+- [x] Written so a non-engineer can read US1-US3 and understand what's broken and why it matters.
+- [x] All mandatory sections completed: User Scenarios & Testing, Requirements, Success Criteria.
+
+## Requirement Completeness
+
+- [x] No [NEEDS CLARIFICATION] markers remain in spec.md.
+- [x] Requirements are testable and unambiguous (FR-001 through FR-012 each map to an SC or a test artifact).
+- [x] Success criteria are measurable (SC-001 names the query that must succeed; SC-002 names the equality the row count must satisfy).
+- [x] Success criteria are technology-agnostic to the extent meaningful (the spec is about a SQL Server interop fix, so SQL Server constructs are unavoidable; no DuckDB-internal API names appear in SC).
+- [x] All three user stories have acceptance scenarios in Given/When/Then form.
+- [x] Edge cases are identified (heap + partitions, empty table, view, columnstore, 0-row partition, concurrent split).
+- [x] Scope is bounded: three SQL templates in one file + one test fixture + one test file. Anything else is out of scope.
+- [x] Dependencies (SQL Server `sys.partitions`) and assumptions (`approx_row_count` is an estimate) are documented.
+
+## Feature Readiness
+
+- [x] Every functional requirement has a corresponding acceptance criterion or success criterion.
+- [x] User scenarios cover the primary read path (US1), the metadata enumeration path (US2), and the bulk preload path (US3).
+- [x] Success criteria match Measurable Outcomes in spec.md.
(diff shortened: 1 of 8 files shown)


VGSML (approved):
We need also check catalog loading performance before and after.

Some scenarios suppose a huge number of tables, I have tested with 200K Tables (8M columns summary).

PR #127, opened 2026-05-26
Fix dbt segfault with threads >= 2 — spec 052 thread-safe catalog entry lifetime

### LINKED ISSUE #126: Use-after-free / data race in MSSQLTableSet under concurrent reads (dbt threads>=2 segfaults)
## Summary

Confirmed via UndefinedBehaviorSanitizer in a multi-threaded stress test
that mirrors `dbt test --threads N>=2` against an attached MSSQL catalog:
**`MSSQLTableSet::GetEntry` returns raw pointers (`optional_ptr`) to
`MSSQLTableEntry` objects whose lifetime is not protected against
concurrent map mutation**, leading to use-after-free / partially-initialized
access when another thread races on `entries_` map.

## Reproduction

Branch: `main` (post-v0.2.0, commit `cc3d725`).

```bash
GEN=ninja make test-concurrent-reads
```

Test file: `test/cpp/test_concurrent_reads.cpp` (added in this report).
Scenarios:
1. 4 threads × 50 iterations of mixed `mssql_scan` reads → PASS (no shared catalog entries)
2. 8 threads × 25 iterations of `mssql_scan` → PASS
3. 4 threads concurrent ATTACH (different aliases) → PASS
4. **4 threads × 50 iterations of catalog-bound `SELECT * FROM mssql.dbo.t` → FAIL with UBSan trap**

UBSan output:
```
src/catalog/mssql_table_entry.cpp:264:4: runtime error: member access within
  address 0x6150001f0480 which does not point to an object of type
  'const duckdb::MSSQLTableEntry'
0x6150001f0480: note: object has invalid vptr
  5d 00 00 00 6e 1d 01 00 5e 00 00 00 91 53 00 00 00 00 00 00
  01 be be be be be be be 00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior
Abort trap: 6
```

`0xbebebebebe` is clang's init-poison pattern — the memory is mid-construction
when another thread reads it as an `MSSQLTableEntry`.

## Root cause

`src/catalog/mssql_table_set.cpp` `GetEntry`:

```cpp
optional_ptr<CatalogEntry> MSSQLTableSet::GetEntry(...) {
    {
        std::lock_guard<std::mutex> lock(entry_mutex_);
        auto it = entries_.find(name);
        if (it != entries_.end()) {
            return it->second.get();  // <-- raw pointer returned, lock released
        }
        ...
    }
    if (LoadSingleEntry(context, name)) {
        std::lock_guard<std::mutex> lock(entry_mutex_);
        auto it = entries_.find(name);
        if (it != entries_.end()) {
            return it->second.get();  // <-- same hazard
        }
    }
    return nullptr;
}
```

`entries_` is `unordered_map<string, unique_ptr<MSSQLTableEntry>>`. Returning
`unique_ptr::get()` while only holding the mutex during lookup means the
caller continues with a raw pointer whose lifetime is bounded by:

1. **`Invalidate()`** at `mssql_table_set.cpp:208` — `entries_.clear()` destroys
   every `unique_ptr`, dangling all outstanding raw pointers across the
   process.
2. **`LoadSingleEntry()`** if it does `entries_[name] = make_uniq<...>` —
   replaces the existing `unique_ptr`, destroying the old `MSSQLTableEntry`.

Callers of `Invalidate()` include `MSSQLCatalog::RefreshCache`,
`InvalidateMetadataCache`, `InvalidateSchemaTableSet` — fired by DDL hooks
and `mssql_refresh_cache()`. In dbt's test workflow, model materialization
triggers these hooks while parallel worker threads are mid-bind on other
queries → use-after-free.

## Affected callers

`MSSQLTableEntry::EnsurePKLoaded` at line 264 is one observed crash site
(loads PK metadata on demand). Any path that holds the `optional_ptr<CatalogEntry>`
returned from `GetEntry` across a thread boundary or across an
`Invalidate()` call is vulnerable.

## Operational impact

dbt-duckdb users with `threads >= 2` cannot reliably use this extension —
the recommended workaround in the dbt profile is `threads: 1` until fixed:

```yaml
# dbt_project.yml or profiles.yml
threads: 1
# The DuckDB community mssql extension is not safe under concurrent reads
# against an attached SQL Server -- dbt test with threads >= 2 segfaults
# the worker process mid-suite (exit 139, leaked semaphores).
```

## Proposed fix directions

1. **Switch `entries_` to `unordered_map<string, shared_ptr<MSSQLTableEntry>>`**
   and return `shared_ptr` from `GetEntry`. DuckDB's `optional_ptr` isn't
   shared-ownership-aware, so this needs a wrapper that holds the shared_ptr
   alive for the bind/execute scope.
2. **Defer invalidation** — instead of `entries_.clear()`, mark entries as
   "stale" and prune only when reference count drops to 1 (no in-flight
   users).
3. **Generation counter** — every entry has a generation; binders capture
   the generation when they look up the entry; if generation changes
   mid-query, re-resolve (or error cleanly).

Option 1 is simplest; option 3 is most DuckDB-idiomatic.

## Severity

**Critical** for any multi-threaded workload that mixes DDL with concurrent
catalog-bound reads. Single-threaded usage is unaffected.

## Out-of-scope (not investigated here)

Similar audit may be warranted for `MSSQLSchemaSet`, `MSSQLMetadataCache`,
and `MSSQLStatisticsProvider` — they share the same raw-pointer-handout
pattern with mutex-only-on-lookup. The minimal-repro test in this report
only exercises the TableEntry path.

🤖 Reproduced with [Claude Code](https://claude.com/claude-code)

VGSML:
## Summary

- Fixes the dbt-duckdb segfault under `threads >= 2` ([#126](https://github.com/hugr-lab/mssql-extension/issues/126)) — a use-after-free in `MSSQLTableSet::entries_` where concurrent first-load of the same table from multiple binders destroyed the winning entry under the loser's foot.
- Closes a second UAF in `MSSQLTableSet::Invalidate()` → `entries_.clear()` that dangled every outstanding raw pointer the binder held.
- Bonus: fixes a double-free in `MSSQLTableEntry::EnsurePKLoaded` caught by AddressSanitizer during scenario 5 stress (two threads both saw `loaded == false` and both move-assigned `pk_info_`, freeing the loser's previous `vector<PKColumnInfo>`).

Spec: [`specs/052-thread-safe-catalog-entries/`](https://github.com/hugr-lab/mssql-extension/tree/052-thread-safe-catalog-entries/specs/052-thread-safe-catalog-entries) (spec.md / plan.md / research.md / data-model.md / contracts/ownership.md / quickstart.md / tasks.md).

## Approach (per US)

| US | What | Where |
|---|---|---|
| **US1** primary dbt cause | `MSSQLTableEntry` / `MSSQLSchemaEntry` switched from `unique_ptr` to `shared_ptr` ownership; `LoadSingleEntry` / `LookupSchema` use `emplace`-only insertion (winner wins on race, loser's shared_ptr is dropped harmlessly). Per-table **singleflight** coordination via `condition_variable` + `loads_in_progress_` set so only ONE thread issues the SQL Server round trip per table — waiters re-check the cache and exit early. Different tables still load in parallel. | `src/{include/,}catalog/mssql_table_set.{hpp,cpp}`, `src/{include/,}catalog/mssql_catalog.{hpp,cpp}`, `src/include/catalog/mssql_{table,schema}_entry.hpp` |
| **US2** independent UAF | `MSSQLTableSet::Invalidate` moves retired entries into the per-catalog `table_graveyard_` (`vector<shared_ptr<MSSQLTableEntry>>`) instead of dropping them; same for `MSSQLCatalog::OnDetach` and the `schema_graveyard_`. The bind data (`MSSQLCatalogScanBindData`) gains a `shared_ptr<void> table_entry_anchor` set via `shared_from_this()` in `GetScanFunction` — keeps the underlying entry alive through the entire execute phase. Graveyard drains only at `~MSSQLCatalog` (no runtime GC — natural-bound for realistic workloads per [research § Decision 3](https://github.com/hugr-lab/mssql-extension/blob/052-thread-safe-catalog-entries/specs/052-thread-safe-catalog-entries/research.md#decision-3-graveyard-drains-on-catalog-destruction-no-runtime-gc)). | `src/include/catalog/mssql_catalog.hpp`, `src/catalog/mssql_catalog.cpp`, `src/catalog/mssql_table_set.cpp`, `src/catalog/mssql_table_entry.cpp`, `src/include/mssql_functions.hpp`, `src/mssql_functions.cpp` |
| **US3** sibling-cache audit | `MSSQLMetadataCache::GetTableMetadata` raw-pointer-handout contract documented in header (only caller, `LoadSingleEntry`, already copies immediately). `MSSQLStatisticsProvider` confirmed by-value (no pointer-handout). `MSSQLCatalog::schema_entries_` covered by US2 T012 (OnDetach routes to graveyard). | `src/include/catalog/mssql_metadata_cache.hpp`, `src/include/catalog/mssql_statistics.hpp` |
| **PK-load race bonus** | `MSSQLTableEntry::EnsurePKLoaded` gains a `pk_load_mutex_` with double-checked-load (fast path stays lock-free when `loaded == true`). | `src/include/catalog/mssql_table_entry.hpp`, `src/catalog/mssql_table_entry.cpp` |

## Test plan

All scenarios run under the default ASan-instrumented Debug build (`GEN=ninja make debug` enables UBSan + ASan via DuckDB's CMake defaults).

- [x] **Scenario 4 (canonical dbt repro)**: 4 threads × 50 catalog-bound reads — clean.
- [x] **Scenario 5 (US2 — invalidation race)**: 4 readers + 1 invalidator (`mssql_refresh_cache` @ 50 ms) for 30 s — **10007 reads × 279 invalidations**, clean.
- [x] **Scenario 6 (US3 — sibling-cache stress)**: scenario 5 + a schema-walker thread (`duckdb_schemas()` / `duckdb_tables()` loop) for 30 s — **7072 reads × 279 invals × 1111 schema walks**, clean.
- [x] **Scenario 7 (mixed writes)**: 4 writers (catalog-bound `INSERT → UPDATE → SELECT → DELETE` cycles on disjoint PK ranges) + 1 reader, all on the same table, 30 s — **3498 cycles + 3882 reads**, clean.
- [x] **Scenario 8 (pure write contention)**: 4 writers, no reader, all `INSERT` into one shared table — **14392 inserts (479/sec)** for 30 s, clean.
- [x] **T021 graveyard unit test** (`make test-catalog-graveyard`): asserts `table_graveyard_` accumulates exactly 5 entries after 5 `mssql_refresh_cache` cycles; `schema_graveyard_` +1 after `DROP SCHEMA`; `DETACH` drains via RAII without crash.
- [x] **Spec 047 regression suite**: `test-spec047-us1` / `test-spec047-us3` / `test-spec047-us-sec` / `test-token-cache-isolation` — all green.
- [x] **Full integration suite** (`make integration-test`): 304 assertions × 16 test cases, **zero regressions**.

## Notes

- **Memory**: graveyards are append-only at runtime, drained at `~MSSQLCatalog()`. Bounded by `(unique entries touched × invalidation events)`. For realistic dbt workloads single-digit MBs over a 24h server lifetime ([research § Decision 3](https://github.com/hugr-lab/mssql-extension/blob/052-thread-safe-catalog-entries/specs/052-thread-safe-catalog-entries/research.md#decision-3-graveyard-drains-on-catalog-destruction-no-runtime-gc)). No runtime GC in this spec — if production telemetry ever shows monotonic growth, a future spec can add a manual `mssql_gc_catalog()` trigger.
- **No user-visible API change**. `optional_ptr<CatalogEntry>` surface to DuckDB unchanged; bind data internals invisible to users. Per Clarification 2026-05-26 Q3, **no** `mssql_pool_stats` graveyard column added — lifetime mechanism is purely internal.
- **C++11 ABI compatible**. No structured bindings, no C++14 init captures. Uses `duckdb::make_shared_ptr` (not `std::make_shared`) to match DuckDB's `shared_ptr` wrapper.
- **Commit-per-US** for easy bisect / review:
  - `1bb8fdb` docs(spec): spec + plan + research + data-model + contracts + quickstart + tasks
  - `bd78fce` feat(052/US1): shared_ptr ownership + emplace-only + singleflight
  - `a2a64b4` feat(052/US2): invalidation→graveyard + bind anchor + PK-load race bonus
  - `355b426` feat(052/US3): sibling-cache audit + scenario 6
  - `87e5f60` chore(052/polish): graveyard unit test + CHANGELOG
  - `cbedadd` test(052): concurrent-write scenarios 7+8

Closes #126


oluies:
## Code review — findings

Review of the spec 052 catalog-entry lifetime fix. Approach is sound and tests are substantial; one HIGH finding plus a few smaller items below.

### HIGH — `EnsurePKLoaded` fast-path is a data race (UB)

`src/catalog/mssql_table_entry.cpp:256-275`:
```cpp
if (pk_info_.loaded) { return; }        // unsynchronized read
std::lock_guard<std::mutex> lock(pk_load_mutex_);
if (pk_info_.loaded) { return; }
```
The header comment says *"torn read at worst forces a redundant mutex acquire."* That would be true only if ``loaded`` were ``std::atomic<bool>`` with acquire/release semantics. As-is, ``loaded`` is a plain ``bool``. Under the C++ memory model, observing ``loaded == true`` without an acquire fence does **not** establish happens-before with the writer's publication of ``pk_info_.columns`` — a reader can see ``loaded == true`` and stale / partially-published ``columns``. Exactly the bug class the rest of the spec is trying to eliminate. TSan would flag this.

This is the canonical double-checked-locking trap that motivated ``std::atomic`` in C++11.

**Fix:** make ``PrimaryKeyInfo::loaded`` ``std::atomic<bool>``, write with ``store(true, std::memory_order_release)`` **after** publishing ``columns``, read in the fast path with ``load(std::memory_order_acquire)``. Or simply drop the fast path — it's one uncontended mutex acquire per call, cheap.

### MEDIUM — ``Scan()`` invokes user callback while holding ``entry_mutex_``

``src/catalog/mssql_table_set.cpp:122-149`` — pre-existing pattern, but with the new singleflight design ``LoadSingleEntry`` now blocks on ``entry_mutex_`` to grab the load slot. If the callback transitively triggers ``LoadSingleEntry`` on another table in the same TableSet (DuckDB binder + virtual columns + ``duckdb_tables()`` walkers do this), you can deadlock or starve. Recommend snapshotting ``entries_`` under the lock and iterating outside.

### MEDIUM — ``GetOrCreateSchemaEntry`` comment is misleading

``src/catalog/mssql_catalog.cpp:387-391``:
```cpp
std::lock_guard<std::mutex> lock(schema_mutex_);
auto it = schema_entries_.find(schema_name);
if (it != schema_entries_.end()) { return *it->second; }
// "On race the late-arriving shared_ptr is discarded ..."
auto entry = make_shared_ptr<MSSQLSchemaEntry>(*this, schema_name);
auto insert_result = schema_entries_.emplace(schema_name, std::move(entry));
```
``schema_mutex_`` is held across find + emplace — no race is possible here. The comment reads like the ``LoadSingleEntry`` lock-free pattern but doesn't apply. Reword or drop.

### MEDIUM — Graveyard bound worth quantifying

Spec accepts append-only at runtime, drained at ``~MSSQLCatalog``. Worst case for a long-lived dbt-style process: 1K tables × hourly ``mssql_refresh_cache`` × 30 days ≈ 720K ``shared_ptr<MSSQLTableEntry>``. Each carries a ``vector<MSSQLColumnInfo>`` — not "single-digit MB" at that scale. Options:
- Debug log warning when graveyard crosses a threshold so users see it before OOM
- A doc note in CHANGELOG suggesting periodic ``DETACH``/``ATTACH`` for very-long-lived processes
- Add internal observability (not user-facing) — e.g. a debug-only ``mssql_pool_stats`` column gated behind ``MSSQL_DEBUG``

### LOW

- **``AppendToTableGraveyard`` is ``noexcept`` but calls ``reserve()``** — bad-alloc inside ``noexcept`` calls ``std::terminate``. Defensible for ENOMEM but worth dropping ``noexcept`` or wrapping.
- **``shared_ptr<void>`` anchor comment** — "shared_ptr_cast" is misleading; the conversion uses the aliasing/converting constructor, no cast.
- **No focused singleflight unit test** — scenarios 4-6 prove the end-to-end UAF is gone but don't directly assert *"N threads racing on the same cold table issue exactly 1 SQL Server round trip."* A counter in the test (incremented on ``GetTableMetadata`` entry) would lock the invariant against future regressions.
- **No TSan run** — the contracts in ``specs/052/contracts/ownership.md`` list TSan as a gate but the PR description only mentions ASan. Given the HIGH finding above, a TSan pass on scenarios 5/6 would be the highest-leverage missing check.

### Approve-after-discussion

Everything else is correct: ``shared_ptr`` ownership, ``emplace``-only insertion, singleflight (waiter wake/exception/owner-died paths all handled), graveyard lock-order avoids inversion (``schema_mutex_`` → ``entry_mutex_`` released → ``graveyard_mutex_``), bind data ``Equals`` correctly ignores anchors, C++11-ABI compatibility respected (``make_shared_ptr``, no structured bindings).

Recommendation: address the HIGH ``EnsurePKLoaded`` race; the MEDIUM items are negotiable.

oluies:
Pushed `64b2139` to address the review findings:

- **HIGH — atomic `pk_loaded_`**: hoisted out of `PrimaryKeyInfo` (which must stay move-assignable) into a separate `std::atomic<bool>` on `MSSQLTableEntry`. Fast path is `load(acquire)`, publication is `store(release)` as the last write. Also patches `GetVirtualColumns`, which had the same racy read.
- **MEDIUM — `Scan()` callback outside `entry_mutex_`**: phase-1 populate entries under the mutex into a `vector<shared_ptr<MSSQLTableEntry>>` snapshot, phase-2 callback after releasing. Snapshot also defends against a concurrent `Invalidate()` between the two phases.
- **LOW — `noexcept` dropped** from `AppendToTableGraveyard` / `AppendToSchemaGraveyard` so `bad_alloc` from `reserve()` / `push_back()` propagates instead of calling `std::terminate`. Size accessors stay `noexcept`.
- **MEDIUM — misleading comment** in `GetOrCreateSchemaEntry` reworded to note that `schema_mutex_` is held across find+emplace (no race to mention).

Skipped the local 30+ min ASan/UBSan rebuild — relying on CI's concurrent-reads + catalog-graveyard runs for verification.

Diff: +85 / -54 across 7 files (4 cpp, 3 hpp). No test changes — existing scenarios 4-8 cover the same surfaces; the atomic fix doesn't change observable behaviour, only makes the synchronization sound under the C++ memory model.

oluies:
@VGSML where is the concurrent tests executed?

oluies (dismissed):

oluies:
Ran spec 052's new C++ tests locally against the patched branch (commit `d33fb47`), since neither `make test-concurrent-reads` nor `make test-catalog-graveyard` is invoked by CI (only the load-only smoke test runs on PR pushes; `integration-test` is `workflow_dispatch`-only).

**Environment**: macOS arm64, Debug build (ASan/UBSan via DuckDB defaults), SQL Server 2022 in Docker (`make docker-up`).

### `test-catalog-graveyard` — PASSED

```
=== Graveyard grows on Invalidate; drains on DETACH ===
  after 5 refresh cycles: table_graveyard_size = 5 — PASSED
=== Schema graveyard on DROP SCHEMA ===
  schema_graveyard_size after DROP = 1 — PASSED
```

### `test-concurrent-reads` — PASSED (all 8 scenarios)

| # | Scenario | Result |
|---|---|---|
| 1 | 4 threads × 50 mixed reads | 200/200 in 372ms |
| 2 | 8 threads × 25 mixed reads | 200/200 in 189ms |
| 3 | 4 concurrent ATTACHes | 1110ms |
| 4 | 4 threads × 50 catalog-bound reads | 200/200 in 398ms |
| 4b | 8 threads × 25 catalog-bound reads | 200/200 in 202ms |
| 5 | 4 readers + invalidator @ 50ms for 30s | **11680 queries × 297 invalidations** |
| 6 | + schema-walker for 30s | **8119 reads × 295 invals × 1286 schema walks** |
| 7 | 4 writers + 1 reader, INSERT/UPDATE/DELETE cycles | **3750 cycles + 4200 reads** |
| 8 | 4 pure writers, INSERT only | **17100 inserts (569/sec)** |

All sanitizer-clean. Throughput numbers slightly above what the PR author reported (different hardware) — no regression from the review fixes.

**Recommendation**: add a `make test-concurrent-reads` invocation to a CI workflow (gated on Docker availability) so this stress suite doesn't bit-rot. Out of scope for this PR but worth a follow-up issue.

oluies (dismissed):

oluies (dismissed):

oluies (dismissed):

oluies (dismissed):

oluies (dismissed):

oluies (dismissed):

oluies (dismissed):

oluies:
*Analysis posted by Claude (Opus 4.7) on behalf of @oluies.*

The `terminate called without an active exception` in the Concurrency stress (scenarios 1-8) job is **not** the root cause — it's a secondary failure that hides the real bug. The CI run is [this one](https://github.com/hugr-lab/mssql-extension/actions/runs/26522394065/job/78116807748?pr=127).

## Error chain

1. **Real bug (scenario 5 leak):** at the end of `scenario_concurrent_catalog_reads` (8 threads × 25 iters), one connection is still in `active_connections_` when `~MSSQLCatalog` → `~ConnectionPool` runs. The first thing `ConnectionPool::Shutdown()` does (`src/tds/tds_connection_pool.cpp:58-65`) is print the `[MSSQL POOL] WARNING: ~ConnectionPool 'mssql' called with 1 connection(s) still checked out` line.

2. **Secondary bug (terminate):** the very next line is `D_ASSERT(active_connections_.empty())` at `src/tds/tds_connection_pool.cpp:66`. In debug builds, `D_ASSERT` does **not** call `assert(3)` — it expands to `DuckDBAssertInternal(...)` which **throws `InternalException`** (see `duckdb/src/common/assert.cpp:13`). The throw unwinds out of `Shutdown()` *before* it reaches:

   ```cpp
   if (cleanup_thread_.joinable()) cleanup_thread_.join();   // ← SKIPPED
   ```

   `~ConnectionPool` is `noexcept` and catches the exception, but `cleanup_thread_` is still joinable when its `std::thread` member destructor runs — and `~std::thread` on a joinable thread calls `std::terminate()`. That's the `terminate called without an active exception` followed by `Aborted (core dumped)`.

The comment at `src/tds/tds_connection_pool.cpp:53-57` says "debug builds emit both the warning and the assert hit before std::terminate" — that expected `assert(3)`-style abort semantics, not the DuckDB-debug throw-then-skip-cleanup semantics.

## Fix sketch (two independent things)

- **Reorder `Shutdown()`:** signal `shutdown_flag_`, join `cleanup_thread_`, close idle/active connections, **then** assert. The warning is already printed, so the operator-visibility goal is met even without the trailing assert; if you want to keep the assert, run it at the end so it doesn't strand the cleanup thread.
- **Hunt the leak:** scenario 5 runs a `Connection cleanup(db); cleanup.Query("SELECT mssql_exec('mssql', 'DROP TABLE IF EXISTS dbo.concurrent_read_test')");` after the workers join (`test/cpp/test_concurrent_reads.cpp:343-346`). The `mssql_exec` path does its own pool `Acquire` — if it returns without `Release` on the cleanup path (or the guard is elided/optimized differently under ASan/UBSan), that's the candidate holdout. Worth instrumenting `Acquire` / `Release` with thread-local counters and dumping the mismatch on Shutdown to confirm.

Mac repro of the warning (without the terminate): I ran spec 051's `parallel_queries.test` against the local docker SQL Server yesterday and saw the same `1 connection(s) still checked out` warning at teardown — but on macOS clang the release build's `D_ASSERT` is a no-op, so the cleanup_thread_.join() still ran and the process exited cleanly. The Linux CI debug build is what flips the warning into an abort.

oluies (dismissed):

PR #128, opened 2026-05-27
security: wipe bearer credentials on destruction (password, access_token, FEDAUTH token)

oluies:
## Summary

Add user-declared destructors that `OPENSSL_cleanse` sensitive credential storage so bearer secrets are not left in heap-recycled memory after the owning struct dies.

- **`MSSQLConnectionInfo::~MSSQLConnectionInfo()`** wipes `password` (SQL auth) and `access_token` (pre-provided Azure AD JWT — spec 032). Both are bearer credentials parsed from connection strings or DuckDB secrets.
- **`~MSSQLCatalog()`** wipes `fedauth_token_utf16le_` — the UTF-16LE Azure FEDAUTH token vector held on the catalog after `AcquireToken` returns (spec 047).
- **`OPENSSL_cleanse`** rather than `std::fill` / `memset`: defeats dead-store elimination so the wipe is not optimised away. OpenSSL is already linked extension-wide via vcpkg; the call is microsecond-cheap.
- **Rule-of-five compliance**: copy/move ctors and assigns on `MSSQLConnectionInfo` are explicitly `= default`-ed. User-declaring the destructor would otherwise silently disable move generation and fall back to copy.
- **`<openssl/crypto.h>` stays out of public headers** — destructor implementations live in `mssql_storage.cpp` and `mssql_catalog.cpp`.

## Threat model

Defence in depth against post-mortem memory inspection: process core dumps, container restart races, OOM-kill survivors, swap-file scrapes. Doesn't change the in-use threat surface — while a `MSSQLConnectionInfo` lives, its `password` is plaintext in memory exactly as before, but a leaked allocator slab after teardown contains zeros instead of the secret.

## Files

| File | Change |
|---|---|
| `src/include/mssql_storage.hpp` | Declare `~MSSQLConnectionInfo()` + explicit rule-of-five defaults |
| `src/mssql_storage.cpp` | Implement destructor, `#include <openssl/crypto.h>` |
| `src/catalog/mssql_catalog.cpp` | Replace `~MSSQLCatalog() = default` with body wiping `fedauth_token_utf16le_` |
| `CHANGELOG.md` | `[Unreleased]` Security entry |

## Test plan

- [ ] CI green on all platforms (Linux GCC + Clang, macOS Clang, Windows MSVC + MinGW). `<openssl/crypto.h>` is already a transitive include in TDS/TLS code, so no new vcpkg deps.
- [ ] No behaviour change at runtime — wipe only runs at destruction. Existing integration suite suffices to confirm no regression.
- [ ] Optional: ASan run on `make integration-test` to confirm no use-after-free introduced by the explicit destructor (the rule-of-five defaults guard against this, but worth a check).

## Notes

- `MSSQLConnectionInfo` is held via `shared_ptr<MSSQLConnectionInfo>` throughout (`FromSecret`, `FromConnectionString`, `MSSQLCatalog::connection_info_`), so wipes run exactly once per logical credential, on the last reference drop.
- This does NOT wipe transient copies the OS / vcpkg / TDS layer may make of these strings during LOGIN7 encoding or TLS handshakes. Those are out of scope for a single-PR change; widening that surface would require codec-level audit (spec-class work). The struct-destructor wipe handles the long-lived references that matter for forensic exposure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **Wipe bearer credentials on destruction.** `MSSQLConnectionInfo` gains a
+  user-declared destructor that `OPENSSL_cleanse`s `password` and
+  `access_token`; `~MSSQLCatalog` wipes the cached `fedauth_token_utf16le_`
+  byte vector. `OPENSSL_cleanse` defeats dead-store elimination, so secrets
+  do not linger in heap-recycled memory after the owning
+  `shared_ptr<MSSQLConnectionInfo>` or `MSSQLCatalog` is destroyed.
+  Rule-of-five compliance: copy/move ctors and assigns on
+  `MSSQLConnectionInfo` are explicitly `= default`-ed so that user-declaring
+  the destructor doesn't silently disable move generation.
+
 ## [0.2.0] - 2026-05-20
 
 Major release: integrated authentication (Kerberos + Windows SSPI),
--- a/src/catalog/mssql_catalog.cpp
+++ b/src/catalog/mssql_catalog.cpp
@@ -1,4 +1,6 @@
 #include "catalog/mssql_catalog.hpp"
+#include <openssl/crypto.h>
+
 #include "azure/azure_token.hpp"
 #include "catalog/mssql_ddl_translator.hpp"
 #include "catalog/mssql_schema_entry.hpp"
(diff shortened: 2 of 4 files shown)


oluies:
[Why does OPENSSL_cleanse look so complex and thread-unsafe?](https://stackoverflow.com/a/30236735/203968) — good rundown of why securely zeroing memory is genuinely hard in portable C: the optimizer will happily delete a "dead" `memset`, `volatile` doesn't reliably prevent it, and `memset_s` isn't available on glibc. Also covers how OpenSSL 1.1.0 finally simplified it down to a `volatile` function pointer to `memset`.

oluies:
## Rationale

OWASP recommends zeroing memory once a secret is no longer needed, so it doesn't linger where it could be recovered — see [Secrets Management Cheat Sheet §2.5 — Handling Secrets in Memory](https://cheatsheetseries.owasp.org/cheatsheets/Secrets_Management_Cheat_Sheet.html#25-handling-secrets-in-memory):

> **Zeroing Memory:** After a secret has been used, the memory it occupied should be zeroed out to prevent it from lingering in memory where it could potentially be accessed.

VGSML (approved):

PR #139, opened 2026-06-04
fix(auth): fragment integrated-auth LOGIN7 across TDS packets (#138)

### LINKED ISSUE #138: macOS ARM64: TLS handshake fails with "Connection reset by peer" during LOGIN7 (Kerberos ccache mode)
Relates to #62 — Kerberos support was added in v0.2.0 but appears broken on macOS ARM64 due to TLS.

### Environment

- OS: macOS ARM64 (Apple Silicon)
- DuckDB: 1.5.3 (Python bindings via INSTALL mssql FROM community)
- mssql extension: 0.2.0 (SELECT mssql_version() returns 0.2.0)
- SQL Server: 2019 RTM-CU32-GDR (15.0.4470.1), Standard Edition, Windows Server 2022
- Auth: Kerberos ccache mode (kinit already run, valid service ticket confirmed via klist)

### Steps to reproduce
```
INSTALL mssql FROM community;
LOAD mssql;
ATTACH 'Server=<SERVER>,1433;Database=master;Trusted_Connection=yes;Encrypt=yes;TrustServerCertificate=yes'
    AS db (TYPE mssql);
```

### Error
```InvalidInputError: MSSQL connection validation failed: Failed to receive LOGIN7 response (integrated):
TLS receive failed: Receive failed (SSL_ERROR_SYSCALL, ret=-1): Connection reset by peer (errno=54)
```

### What works

- nc -z sqlhost 1433 — TCP connection succeeds
- klist — shows valid TGT + service ticket
- pyodbc with ODBC Driver 18 (Trusted_Connection=yes;TrustServerCertificate=yes) — connects and queries fine
- mssql_version() — returns 0.2.0

### What doesn't work (all produce the same TLS error)

- Encrypt=yes;TrustServerCertificate=yes
- Encrypt=no
- FQDN vs short hostname
- lazy_validation true (ATTACH succeeds but pool connections all fail)

### Analysis
The server resets the TCP connection during the TDS prelogin TLS exchange. Since Encrypt=no also fails, the extension still performs TLS prelogin (as required by TDS protocol) and the server rejects the handshake.

ODBC Driver 18 on macOS uses Secure Transport (Apple's TLS stack), which negotiates successfully. The extension's bundled OpenSSL (from vcpkg) may be offering cipher suites or TLS parameters that this SQL Server rejects.

Happy to test macOS.

oluies:
Fixes #138.

## Problem

Kerberos integrated auth fails against domain-joined SQL Server with:

```
Failed to receive LOGIN7 response (integrated): TLS receive failed:
Receive failed (SSL_ERROR_SYSCALL, ret=-1): Connection reset by peer (errno=54)
```

The report attributed this to OpenSSL TLS cipher negotiation on macOS, but that is a red herring. The error fires at `tds_connection.cpp` **after** `DoPrelogin()` already completed the TLS handshake, the local GSSAPI token was generated, and the encrypted LOGIN7 was sent. The server TCP-resets while we read the *response*.

## Root cause

`AuthenticateIntegrated()` sent LOGIN7 as a single `TdsPacket` via `SendPacket()` with **no TDS-packet fragmentation** (contrast the SQL_BATCH / BulkLoad paths, which fragment). LOGIN7 is sent *before* packet-size negotiation, so SQL Server enforces the default **4096-byte** packet size. A real Active Directory Kerberos AP-REQ wraps the service ticket containing the **PAC** (group SIDs), routinely 2–8 KB+, so LOGIN7 exceeds 4096 and the server resets.

This is **not** macOS-specific — Linux against the same server fails identically. It passed CI because `test/kerberos/`'s KDC issues minimal PAC-less tickets (LOGIN7 stays under 4096). `pyodbc`/ODBC Driver 18 works on the same host because MS's driver fragments LOGIN7 correctly.

## Fix

- Add `TdsProtocol::SplitIntoPackets(message, max_packet_size)` — fragments an already-built message to the (pre-negotiation default 4096-byte) packet size, preserving packet type and setting EOM only on the last packet. Mirrors the existing `BuildSqlBatchMultiPacket` / `BuildBulkLoadMultiPacket` logic. Messages that already fit are returned unchanged, so the common path is byte-identical.
- Apply it to both the initial LOGIN7 and the SSPI continuation sends in `AuthenticateIntegrated()`.

## Test

`test/cpp/test_login7_encoding.cpp` test `[8]` builds a LOGIN7 with a 10000-byte SSPI blob and asserts: multi-packet fragmentation, per-packet serialized size ≤ 4096, EOM only on the last packet, and byte-identical reassembly. Small blobs still produce exactly one EOM packet.

Run: `make test-login7-encoding` (no SQL Server required).

```
[8] LOGIN7 SSPI fragmentation (issue #138)...
ALL TESTS PASSED
```

## Verification still needed from reporter

This was diagnosed by static analysis + unit test; I don't have a domain-joined server to repro against. Asked @renre-axw on the issue to confirm with `MSSQL_DEBUG=3` that the LOGIN7 blob exceeds 4096 bytes, and that the connection now succeeds.


src/tds/tds_connection.cpp
```diff
@@ -38,6 +38,28 @@ std::string GetClientHostname() {
...
+// negotiation completes, so they must be fragmented to the pre-negotiation
+// default of 4096 bytes (issue #138). MSSQL_LOGIN7_MAX_PACKET is a TEST-ONLY
+// override that lowers that boundary so the multi-packet path can be exercised
+// end-to-end against a real SQL Server without an AD-sized Kerberos PAC -- the
+// server reassembles the fragments regardless of how small they are. Unset (or
+// out of range) in production -> the standard 4096-byte default. The clamp
+// guarantees room for the 8-byte TDS header plus the 94-byte LOGIN7 fixed
+// region with payload to spare.
+size_t Login7FragmentPacketSize() {
+	const char *env = std::getenv("MSSQL_LOGIN7_MAX_PACKET");
```
VGSML: I am not sure that the env is a right place for this, use duckdb settings instead

VGSML (dismissed):

VGSML (approved):

PR #154, opened 2026-06-07
fix(catalog): invalidate catalog cache after DDL run via mssql_exec() (#151)

### LINKED ISSUE #151: Raw DDL via mssql_exec() does not invalidate the catalog cache → stale CREATE TABLE IF NOT EXISTS and "Invalid object name"
### Summary

DDL executed as raw T-SQL through `mssql_exec()` (e.g. `DROP TABLE`) does **not** invalidate the extension's catalog metadata cache. A subsequent catalog-level operation that depends on the cached existence state — notably `CREATE TABLE IF NOT EXISTS` — then acts on stale metadata: it sees the table as still present, **skips creation**, and the next read of the table fails server-side with `Invalid object name`.

Surfaced by `test/sql/ctas/ctas_if_not_exists.test:79` (the test mixes `mssql_exec('DROP TABLE …')` with catalog DDL).

### Reproduce

Deterministic in the test harness (the sequence of prior catalog ops warms the cache):

```sql
ATTACH '…' AS mssql (TYPE mssql);

CREATE TABLE IF NOT EXISTS mssql.dbo.T AS SELECT 1 AS id, 'first' AS name;   -- creates, cached as existing
CREATE OR REPLACE TABLE mssql.dbo.T AS SELECT 4 AS id, 'fourth' AS name;     -- replaced

SELECT mssql_exec('mssql', 'DROP TABLE dbo.T');                              -- raw drop: cache NOT invalidated

CREATE TABLE IF NOT EXISTS mssql.dbo.T AS                                    -- cache says T exists -> SKIPPED
    SELECT i AS id, 'row_'||i::VARCHAR AS name FROM range(100) t(i);

SELECT COUNT(*) FROM mssql.dbo.T;                                            -- Invalid object name 'dbo.T'
```

### Root cause (confirmed via MSSQL_DEBUG)

After the raw `mssql_exec` `DROP`, the catalog cache still holds the `dbo.T` entry. The debug trace shows no create for the `IF NOT EXISTS` statement (it is skipped because the cache reports the table as existing), and the following `SELECT COUNT(*)` generates `SELECT [id] FROM [dbo].[T]` against a table that no longer exists server-side:

```
[MSSQL FN]  mssql_exec: context=mssql, sql=DROP TABLE dbo.IfNotExistsTest
[MSSQL TE]  GetScanFunction: dbo.IfNotExistsTest ...        <- cache still has it
[MSSQL TABLE_SCAN] generated query = SELECT [id] FROM [dbo].[IfNotExistsTest]
Invalid Input Error: SQL Server error [208, severity 16]: Invalid object name 'dbo.IfNotExistsTest'.
```

The reverse direction has the same shape: a table CREATEd via raw `mssql_exec` is invisible to the catalog until the cache is refreshed.

### Impact / triage

- **Severity:** Medium. Correctness/coherence bug, but only when **mixing raw `mssql_exec` DDL with catalog DDL/reads** in the same session. Pure-catalog or pure-`mssql_exec` workflows are unaffected. The CTAS data load itself is fine (verified: rows land correctly); this is purely stale existence metadata.
- **Default `mssql_catalog_cache_ttl = 0`** (manual refresh) makes it sticky — the stale entry never expires on its own.

### Workarounds

- Use catalog DDL instead of raw exec for schema changes: `DROP TABLE mssql.dbo.T;` (stays coherent — verified).
- Or call `mssql_refresh_cache('mssql')` after any raw `mssql_exec` DDL.

### Suggested fix (options)

1. Detect DDL verbs (`DROP`/`CREATE`/`ALTER`/`TRUNCATE` … `TABLE`/`VIEW`) in `mssql_exec()` and point-invalidate the affected catalog entry (or the schema) after execution.
2. Have existence-sensitive catalog ops (`CREATE … IF NOT EXISTS`, drop, etc.) verify against the server rather than trusting a possibly-stale cache.
3. At minimum, document that `mssql_exec` DDL bypasses the catalog cache and requires `mssql_refresh_cache`.

### Notes

- Found while correcting stale `CALL mssql_exec` test syntax in #150. The affected tests (`ctas_if_not_exists`, `ctas_auto_tablock`) were left untouched in that PR pending this fix, since editing their assertions would mask this bug.

oluies:
## Summary

DDL executed as raw T-SQL through `mssql_exec()` (e.g. `DROP TABLE`) did **not** invalidate the extension's catalog metadata cache. A subsequent catalog operation then acted on stale metadata:

- `CREATE TABLE IF NOT EXISTS` saw the (already-dropped) table as still present, **skipped creation**, and the next read failed with `Invalid object name`.
- A table **created** via raw `mssql_exec` was invisible to the catalog until a manual `mssql_refresh_cache()`.

Fixes #151.

## Root cause

`mssql_exec()` runs arbitrary T-SQL on a pooled connection and never touched the catalog cache. Reads/`CREATE … IF NOT EXISTS` decisions are driven by the cached `MSSQLMetadataCache` **and** the bound `MSSQLTableEntry` objects held in each schema's table set — both of which went stale after a raw DDL.

Note: invalidating only the metadata cache (`MSSQLMetadataCache::InvalidateAll()`) is **not** enough — once a table has been read, its `MSSQLTableEntry` is bound into the schema's table set and persists, so `CREATE … IF NOT EXISTS` still sees it. The fix must also invalidate the schema table sets.

## Fix

After a successful `mssql_exec()` whose SQL may change schema (`CREATE`/`DROP`/`ALTER`/`TRUNCATE`/`RENAME`/`EXEC`), call `MSSQLCatalog::InvalidateMetadataCache()`. This is the **lazy** path:
- marks the metadata cache stale, **and**
- iterates every schema entry calling `GetTableSet().Invalidate()` (evicting bound table entries via the spec-052 graveyard),
- with reload deferred to next access — no eager round trip (unlike `RefreshCache()`).

DML verbs (`INSERT`/`UPDATE`/`DELETE`) do **not** trigger it, so transaction-pinned DML through `mssql_exec()` is unaffected (verified: `transaction_commit` stays green). Detection is a conservative keyword check — over-detection only costs one lazy cache refresh; `EXEC` is included because a stored procedure may run DDL.

## Tests

- **`test/sql/catalog/exec_ddl_cache_invalidation.test` (new)** — reproduces the DROP-via-exec → `CREATE IF NOT EXISTS` → read scenario plus the create-via-exec visibility case. **Verified it fails without this fix** (`Invalid object name`) and passes with it (23 assertions). It also fails with an `InvalidateAll`-only fix, confirming the table-set invalidation is the essential part.
- **`ctas_if_not_exists` / `ctas_auto_tablock`** — corrected stale `CALL mssql_exec` (scalar fn → `SELECT`) and `already exists` → `already an object named`. Both were failing on this exact cache bug and **only pass with this fix**, so they double as real-world #151 coverage (19 and 18 assertions).

## Verification

Built `make release` against current `main` and ran against a live SQL Server 2022 (TLS): new test 23, `ctas_if_not_exists` 19, `ctas_auto_tablock` 18, `transaction_commit` 37 — all pass.

Closes #151

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,6 +138,12 @@ duckdb --unsigned -c "INSTALL mssql FROM local_build_debug; LOAD mssql;"
 - SQL test context names: unique per file, prefixed by operation (e.g., `txtest`, `mssql_upd_scalar`)
 - Test tables: PascalCase (`TestSimplePK`, `TxTestOrders`) or snake_case for simple ones (`tx_test`)
 
+## Documentation
+
+- **`DATAMODEL.md`** is the layered-architecture reference (TDS → connection pool → catalog → cache → codec) with Mermaid diagrams. **On every PR, consider whether `DATAMODEL.md` needs updating** — if the change adds/removes a layer component, alters an invariant (lifetime, cache invalidation, threading, ownership), or changes an end-to-end flow, update the relevant layer section (and its diagram) in the same PR.
+- Keep architecture/internals diagrams in `DATAMODEL.md`; keep `README.md` user-facing (link into `DATAMODEL.md` rather than duplicating internals).
+- Use GitHub-rendered ```mermaid``` fenced blocks for diagrams.
+
 ## Key Architecture Concepts
 
 - **Custom TDS implementation**: No external TDS/ODBC library. TDS v7.4 protocol with PRELOGIN, LOGIN7, SQL_BATCH, ATTENTION packet types.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -182,6 +182,7 @@ duckdb --unsigned -c "INSTALL mssql FROM local_build_debug; LOAD mssql;"
 | `mssql_metadata_timeout` | 300 | Metadata query timeout in seconds (0 = no timeout) |
 | `mssql_attach_validation_timeout` | 0 | ATTACH-time eager credential validation timeout in seconds (0 = inherit `mssql_connection_timeout`). Spec 047 FR-011. |
 | `mssql_catalog_cache_ttl` | 0 | Metadata cache TTL (0 = manual via `mssql_refresh_cache()`) |
+| `mssql_exec_invalidate_cache` | true | Auto-invalidate the catalog cache after DDL run via `mssql_exec()` (CREATE/DROP/ALTER/TRUNCATE/RENAME/EXEC). Set false to preserve a large preloaded cache and invalidate manually with `mssql_invalidate_cache()`. Issue #151. |
 | `mssql_insert_batch_size` | 1000 | Rows per INSERT statement |
 | `mssql_insert_max_rows_per_statement` | 1000 | Hard cap per INSERT |
 | `mssql_insert_max_sql_bytes` | 8MB (8388608) | INSERT SQL size limit |
@@ -215,7 +216,8 @@ Available in: ATTACH options, ADO.NET connection strings (`SchemaFilter`/`TableF
 | `mssql_scan(context, query)` | Table | Execute raw T-SQL, stream results |
 | `mssql_exec(context, sql)` | Scalar | Execute T-SQL, return affected row count |
 | `mssql_pool_stats([context])` | Table | View connection pool statistics |
-| `mssql_refresh_cache(context)` | Scalar | Refresh metadata cache |
+| `mssql_refresh_cache(context)` | Scalar | Refresh metadata cache (eager full reload) |
+| `mssql_invalidate_cache(context [, schema [, table]])` | Scalar | Lazy point invalidation — whole catalog / one schema / one table (keeps other tables' cached columns) |
 | `mssql_preload_catalog(context [, schema])` | Scalar | Bulk-load all metadata in one round trip |
 | `mssql_open(conn_string)` | Scalar | [DEPRECATED] Open standalone diagnostic connection (spec 047 FR-010 — prefer ATTACH + catalog-bound functions) |
 | `mssql_close(handle)` | Scalar | [DEPRECATED] Close diagnostic connection (spec 047 FR-010) |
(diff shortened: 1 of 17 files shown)


oluies:
Added two enhancements from review feedback (commit 5509ee4):

**1. `mssql_exec_invalidate_cache` setting (BOOLEAN, default `true`)** — gates the auto-invalidation. Set `false` to preserve a large `mssql_preload_catalog` cache and invalidate manually, instead of discarding the cache on every `mssql_exec()` DDL.

**2. `mssql_invalidate_cache(catalog [, schema [, table]])`** — lazy point invalidation at three granularities:
- catalog → `InvalidateMetadataCache()`
- schema → `InvalidateSchemaTableSet()`
- **table → `InvalidateTableEntry()`** (new) — re-fetches just that table's columns + re-checks its existence, **keeping every other table's cached columns**. Backed by two new primitives: `MSSQLMetadataCache::InvalidateSchemaTableList()` (table-list/existence only, no column drop) and `MSSQLTableSet::InvalidateEntry(name)` (evict one bound entry).

DML verbs still never invalidate, so transaction-pinned DML is unaffected. Default `true` keeps #151 fixed out of the box.

New test `test/sql/catalog/exec_invalidate_cache_setting.test` (25 assertions): function returns `true` at each granularity; with the setting **off** the #151 stale-cache scenario reproduces and table-level `mssql_invalidate_cache()` recovers it. Regression batch green (transaction_commit 37, ctas_if_not_exists 19, ctas_auto_tablock 18, filter_pushdown 164). Docs updated in README / CLAUDE.md / DATAMODEL.md (incl. a per-table granularity table).

VGSML (dismissed):
I propose to set the invalidate settings flag to false by default, the same behavior is in the Postgres extension - it requires manual schema cache invalidation after postgres_execute.

oluies:
@VGSML done — flipped `mssql_exec_invalidate_cache` to **default `false`** (commit 3f8b8c6), matching `postgres_execute`: by default `mssql_exec()` no longer touches the catalog cache; callers invalidate manually with `mssql_invalidate_cache()` (or `SET mssql_exec_invalidate_cache = true` to opt into auto-invalidation).

Tests that mix raw `mssql_exec()` DDL with catalog DDL now `SET … = true` (exec_ddl_cache_invalidation, ctas_if_not_exists, ctas_auto_tablock); the setting test reorders the manual invalidate before the catalog read. Verified against live SQL Server (24/25/20/19 + transaction_commit 37, filter_pushdown 164). Docs updated (README / CLAUDE.md / DATAMODEL.md).

VGSML (approved):

PR #156, opened 2026-06-09
docs: auto-updating download metrics (README chart + DuckDB-Wasm interactive page)

oluies:
Adds a bot that keeps the README's published version and download metrics current, plus an interactive in-browser companion chart. All data is fetched/aggregated by **DuckDB itself** (`httpfs` + `read_json`) — the same queries the [Community Extensions metrics page](https://duckdb.org/community_extensions/download_metrics) documents.

## What's included

**Static (renders inside the README on GitHub):**
- Badge row above the title (h3-duckdb convention): CI · DuckDB v1.4.1+ · release · downloads/week · MIT · stars
- Inline SVG download-trend chart + summary line + a link to the interactive version
- `scripts/update_download_metrics.py` — DuckDB-backed fetch/aggregate → renders SVG + rewrites two managed README regions (`METRICS-BADGES`, `METRICS-CHART`). Idempotent: re-runs produce no diff.
- `.github/workflows/update-metrics.yml` — weekly cron + manual dispatch; installs the DuckDB CLI, commits only on change.

**Interactive (GitHub Pages):**
- `web/index.html` — boots DuckDB-Wasm and runs the *same* `read_json()` queries live in the browser (no backend), rendering an Observable Plot trend chart with hover tooltips + summary cards.
- `.github/workflows/pages.yml` — deploys `web/` to GitHub Pages.

## ⚠️ Required setup for the Interactive (GitHub Pages) part

1. **Enable Pages**: repo **Settings → Pages → Build and deployment → Source = "GitHub Actions"**, or the `pages.yml` deploy step will fail. Once enabled it serves at `https://hugr-lab.github.io/mssql-extension/`.
2. **Merge to `main`**: the scheduled metrics workflow and the Pages workflow only trigger from the default branch — they won't run while this is on the feature branch.
3. **CORS caveat**: the interactive page depends on the metrics endpoint allowing cross-origin browser requests. The DuckDB metrics page does the same in-browser queries, so it should work — worth a glance once Pages is live (the page shows a clear error if CORS blocks it).

## Notes
- Data layer dogfoods DuckDB end-to-end (CLI for the README bot, DuckDB-Wasm for the page).
- No new build/runtime dependencies for the extension itself; the bot installs the DuckDB CLI in CI only.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,37 @@
+name: Deploy metrics page
+
+# Publishes web/ (the DuckDB-Wasm interactive download-metrics chart) to
+# GitHub Pages. Requires Settings -> Pages -> "Build and deployment" Source
+# set to "GitHub Actions". Served at https://<owner>.github.io/<repo>/.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "web/**"
+      - ".github/workflows/pages.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/configure-pages@v5
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: web
+      - id: deployment
+        uses: actions/deploy-pages@v4
--- a/.github/workflows/update-metrics.yml
+++ b/.github/workflows/update-metrics.yml
@@ -28,9 +28,16 @@ jobs:
           python-version: "3.x"
 
       - name: Install DuckDB CLI
+        # Pinned version + checksum (not "latest"): this workflow holds
+        # contents:write to the default branch, so the CLI it executes must be
+        # exactly the reviewed bytes. Update both values together when bumping.
+        env:
+          DUCKDB_VERSION: v1.5.3
+          DUCKDB_SHA256: 35caef1fecbc8d7e2c07de4fd2cdefc5189ec9ba9e1cca228fb1a1c48cc52a8a
         run: |
           curl -fsSL -o /tmp/duckdb.zip \
-            https://github.com/duckdb/duckdb/releases/latest/download/duckdb_cli-linux-amd64.zip
+            "https://github.com/duckdb/duckdb/releases/download/${DUCKDB_VERSION}/duckdb_cli-linux-amd64.zip"
+          echo "${DUCKDB_SHA256}  /tmp/duckdb.zip" | sha256sum -c -
           unzip -o /tmp/duckdb.zip -d "$HOME/.local/bin"
           "$HOME/.local/bin/duckdb" --version
           echo "$HOME/.local/bin" >> "$GITHUB_PATH"
--- a/.github/workflows/update-metrics.yml
+++ b/.github/workflows/update-metrics.yml
@@ -1,0 +1,54 @@
+name: Update download metrics
+
+# Keeps the README badges, latest published version and the download-metrics
+# chart in sync with the DuckDB Community Extensions metrics endpoint.
+# Scheduled weekly; also runnable on demand. Scheduled runs only fire on the
+# repository's default branch.
+
+on:
+  schedule:
+    - cron: "17 6 * * 1" # Mondays 06:17 UTC (the weekly archive refreshes daily ~04:00Z)
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+concurrency:
+  group: update-metrics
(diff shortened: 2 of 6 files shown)


oluies:
## 🔒 Security Review

**Scope:** `.github/workflows/pages.yml`, `.github/workflows/update-metrics.yml`, `scripts/update_download_metrics.py`, `web/index.html`, `docs/assets/download-metrics.svg`, `README.md`

### Findings

**No high-confidence security vulnerabilities identified.**

### Attack surfaces examined

| Surface | Result |
|---|---|
| SQL passed to DuckDB CLI subprocess (`scripts/update_download_metrics.py`) | Safe — `subprocess.run` list form (no shell); SQL built exclusively from constants (`EXTENSION = "mssql"`, URL templates) and `range()` integers; `url_literal()` quote-escapes; remote JSON re-enters Python only as `int()`-coerced or `CAST(... AS DATE)`-constrained values |
| SVG generation (`build_svg`) | Safe — interpolated values are coordinate floats/ints, integer formatting, or `esc()`-escaped date strings; GitHub additionally serves README images via camo |
| README markdown rewrite (`build_chart_block`) | Safe — only non-numeric remote value is the repo's own release `tag_name` (maintainer-controlled by definition); rendered as inert markdown text |
| GitHub Actions workflows | Safe — `schedule`/`workflow_dispatch` triggers only (no PR-triggered untrusted code); `${{ }}` interpolation limited to `secrets.GITHUB_TOKEN` and `github.repository` into `env:` (not inline in `run:`); `contents: write` scoped to committing two generated files; `pages.yml` deploys static files with standard OIDC permissions |
| DOM sinks in `web/index.html` | Safe — all remote-data writes use `textContent` / `replaceChildren` with Plot-constructed elements; tooltip `title` renders as SVG text (not HTML) from `Date.toISOString()` / `toLocaleString()` outputs; no `innerHTML`, `eval`, or `document.write`; DuckDB-Wasm SQL runs sandboxed in the visitor's browser against public data |

### Noted but excluded (hardening, not vulnerabilities)

- `update-metrics.yml` downloads `duckdb_cli-linux-amd64.zip` from the `latest` GitHub release over HTTPS without checksum/version pinning (trusted upstream; supply-chain hardening)
- `web/index.html` loads CDN ESM modules without SRI, and Observable Plot floats on minor version `@0.6` (standard lack-of-hardening, no concrete exploit path)

oluies:
## 🔒 Hardening follow-up (b39edd3)

Applied the two items the security review flagged as "excluded but worth tightening", with the reasoning for each:

### 1. `update-metrics.yml`: DuckDB CLI pinned to v1.5.3 + sha256 verified

**Before:** `curl .../releases/latest/download/duckdb_cli-linux-amd64.zip` — the workflow executed whatever binary GitHub served at run time, with no integrity check.

**Why it matters here specifically:** this workflow holds `contents: write` to the default branch and auto-commits its output. That makes the downloaded CLI the one piece of *unreviewed, runtime-fetched* code in a privileged context — if the `latest` release asset were ever compromised (or simply changed behavior), the blast radius is a push to `main`. Pinning the exact tag + verifying the sha256 turns "trust whatever is served today" into "execute only the bytes reviewed once":

```yaml
env:
  DUCKDB_VERSION: v1.5.3
  DUCKDB_SHA256: 35caef1fecbc8d7e2c07de4fd2cdefc5189ec9ba9e1cca228fb1a1c48cc52a8a
run: |
  curl -fsSL -o /tmp/duckdb.zip ".../download/${DUCKDB_VERSION}/duckdb_cli-linux-amd64.zip"
  echo "${DUCKDB_SHA256}  /tmp/duckdb.zip" | sha256sum -c -
```

**Trade-off:** bumping DuckDB now means updating two adjacent lines (version + hash) — Dependabot won't do it. Both values live in one `env:` block with a comment saying to update them together.

### 2. `web/index.html`: Observable Plot pinned to exact `0.6.17`

**Before:** `@observablehq/plot@0.6` — a floating minor, so jsDelivr could silently serve any future `0.6.x` to page visitors. Now pinned to the exact current release (`0.6.17`, verified resolving on jsDelivr). `@duckdb/duckdb-wasm@1.29.0` was already exact.

### What was deliberately *not* done, and why

- **No SRI / import-map integrity on the CDN modules:** bare `import` statements can't carry `integrity`, and DuckDB-Wasm dynamically fetches its worker JS + `.wasm` bundle at runtime via `getJsDelivrBundles()` — those fetches can't be SRI-protected from the page, so an import map would cover only the entry module while leaving the actually-executing wasm unverified. Partial SRI here is security theater.
- **No self-hosting of the wasm bundle:** that's the airtight option (everything served from the Pages origin, zero third-party runtime deps), but it costs ~40 MB of vendored artifacts plus manual bumps — disproportionate for a static, no-auth, no-secrets page charting public data. Worst case of a CDN compromise is malicious JS on a page with nothing to steal. Happy to do it anyway if you'd rather the page be CDN-independent for availability reasons.

Net: the fix with real blast radius (#1) is airtight; the low-stakes one (#2) got the proportionate one-line treatment.

VGSML (approved):

PR #163, opened 2026-06-15
fix(auth): lazy dlopen GSSAPI/krb5 on Linux so extension loads without the Kerberos runtime (#161)

### LINKED ISSUE #161: mssql extension needs unused krb5 so file.
### What happens?

> _duckdb.IOException: IO Error: Extension "/root/.duckdb/extensions/v1.5.3/linux_arm64/mssql.duckdb_extension" could not be loaded: libgssapi_krb5.so.2: cannot open shared object file: No such file or directory

it seems to expect this `libgssapi_krb5.so.2` shared object in my image and I don't even need kerberos auth, which should make it optional.

I could install the debian `libgssapi-krb5-2` package, I would be installing something I'm not using.

### To Reproduce

```Dockerfile
FROM ghcr.io/astral-sh/uv:python3.11-trixie-slim AS uv
RUN uv init && uv add duckdb
RUN uv run python -c "import duckdb; conn = duckdb.connect(':memory:'); conn.execute('INSTALL mssql FROM community'); conn.close()"
```

as mentioned in the original https://github.com/duckdb/duckdb/issues/23257#issuecomment-4697233728 ideally would be to disable Kerberos by default for Linux community binaries or make it a runtime dependency that only gets loaded as needed.

oluies:
Fixes #161.

## Problem

On Linux the community binary hard-links `libgssapi_krb5.so.2` + `libkrb5.so.3` (a `DT_NEEDED` entry, introduced by integrated auth in spec 042). The dynamic loader demands those libraries **at `LOAD mssql` time**, before any code runs — so the extension fails to load on minimal images even for users who never use Kerberos:

> IO Error: Extension `.../linux_arm64/mssql.duckdb_extension` could not be loaded: `libgssapi_krb5.so.2`: cannot open shared object file

## Fix (spec 053)

Make GSSAPI/krb5 a **lazy runtime** dependency on Linux instead of a link-time one.

- **New `src/tds/auth/gssapi_runtime.{hpp,cpp}`** — function-pointer tables resolved via `dlopen`/`dlsym` on the first `authenticator=krb5` use only (thread-safe `std::call_once`, `RTLD_NOW|RTLD_LOCAL`). Pointer types come from `decltype(&gss_*)`/`decltype(&krb5_*)` so signatures always match the platform.
- **CMake** keeps the GSSAPI/krb5 *include dirs* (headers needed at build time) but **drops `target_link_libraries`** on Linux and links `${CMAKE_DL_LIBS}`. macOS keeps `-framework GSS` (system framework, always present) and fills the table with direct `&gss_*` addresses — call-site code is identical on both platforms.
- `krb5_authenticator.cpp` / `krb5_test_function.cpp` route **every** `gss_*`/`krb5_*` call through the tables — no direct symbol references remain.
- When Kerberos is requested but the runtime is absent, `Krb5RuntimeUnavailable` is thrown naming the missing `.so` and the install package (`libgssapi-krb5-2` / `krb5-libs`) — a clear connection-time error, not a load failure.

Windows (SSPI) is unaffected.

## Verification

Verified in a Debian aarch64 container:
- `readelf -d` on a binary built from the shim shows **no gssapi/krb5 in `DT_NEEDED`** (links with only `-ldl`).
- Runtime **present** → `dlopen` loads it → auth path works.
- Runtime **deleted** (the #161 image) → binary **still runs** (no load failure) and throws the clear "install libgssapi-krb5-2" error.
- macOS: compiles + unit test passes via `-framework GSS`.

## CI (new — guards the regression)

- **`cpp-unit-tests`** job runs on every push/PR (Linux + macOS): builds/runs `test/cpp/test_gssapi_runtime.cpp` and, on Linux, asserts the shim binary has no gssapi/krb5 `DT_NEEDED`.
- **`build`** job installs `libkrb5-dev` (so the shipped artifact contains the Kerberos path) and `readelf`-guards the real `mssql.duckdb_extension`, failing if libgssapi_krb5/libkrb5 ever reappear in `DT_NEEDED`.

This closes the gap that let #161 ship: all build/smoke runners *have* the Kerberos runtime, so the hard dependency was invisible there.

## Spec

Full spec/plan/research/contracts under `specs/053-lazy-gssapi-linking/`.

**Still to verify on a full build (not in this branch's local checks):** clean-image `LOAD` of the real extension, and the `test/kerberos/` end-to-end stack (CredCache/keytab/raw) — best run via the manual `workflow_dispatch` build + the kerberos docker stack.


VGSML (dismissed):

oluies:
## Code review summary

**Verdict: looks good to merge.** Correct, well-isolated, conventions-clean, and verified end-to-end in CI.

### Verified
- **No `DT_NEEDED` on the Kerberos runtime** — the new `cpp-unit-tests` job `readelf`-asserts the shim binary links only `-ldl`; confirmed locally in a stripped Debian container too (extension runs with `libgssapi_krb5`/`libkrb5` deleted, throws a clear "install libgssapi-krb5-2" error).
- **No Kerberos regression** — the **Kerberos end-to-end (KDC + SQL Server)** job passed (22m), exercising the lazy-`dlopen` path through CredCache/keytab/raw.
- All required checks green (CodeQL, Concurrency stress, Docker stack, unit tests ×2, secret scan, lints).

### Design notes
- `decltype(&gss_*)` pointer types sidestep MIT/Heimdal signature drift; macOS keeps `-framework GSS` and fills the table with direct addresses (one uniform call path).
- Thread-safe lazy init via `std::call_once`; on failure the flag stays unset so a later attempt retries after the user installs the package.
- libkrb5 is the same mapping gssapi pulls in transitively, so raw/keytab share one krb5 context state.

### Minor (non-blocking)
- The shim resolves **all** GSSAPI symbols eagerly, incl. `gss_acquire_cred_from` — so a pre-MIT-1.11 (2012) runtime lacking it would fail even CredCache-default mode. Universally present in practice; documented in a code comment (7e23d9f).
- The `readelf` guard on the **real shipped artifact** runs in the `workflow_dispatch`-gated `build` job; routine PRs get the per-PR shim-binary guard (strong proxy). This PR's Kerberos e2e built+loaded the real extension, so it was covered here.

No changes requested beyond the comment already pushed.

VGSML (approved):

Issue #178, opened 2026-07-07
MSSQL extension crashes on concurrent queries from Python notebook (and DuckDB.NET) while single-thread works


lucap-irion:
**Environment**

- DuckDB version: 1.5.2 / 1.5.3 / 1.5.4 (tested with 1.5.x)
- Python: 3.11 (Windows Store build)  
- Client: DuckDB Python in Jupyter / VS Code notebook [duckdb](https://duckdb.org/docs/lts/clients/python/overview.html)
- Extension: `mssql` community extension (TDS-based SQL Server connector) [[duckdb](https://duckdb.org/community_extensions/extensions/mssql.html)](https://duckdb.org/community_extensions/extensions/mssql.html)
- OS: Windows 11 x64  
- SQL Server: on localhost

The same basic pattern also reproduces with DuckDB.NET (C#) using multiple concurrent tasks against the MSSQL extension, but works correctly when only one thread/task is active. 

***

**Repro script (Python, notebook cell)**

This is the minimal script I’m using in a Python notebook that consistently crashes the Python kernel when using the `mssql` extension. The same pattern works fine when querying only native DuckDB tables, so it appears specific to the MSSQL extension:

```python
import duckdb
from threading import Thread

con = duckdb.connect(":memory:")

# Setup once on the parent connection
con.execute(r"""
LOAD mssql;

CREATE OR REPLACE SECRET secret(
    TYPE mssql,
    HOST 'localhost',
    PORT 1433,
    DATABASE 'database',
    USER 'my_user',
    PASSWORD 'my_password',
    USE_ENCRYPT FALSE,
    CATALOG TRUE,
    SCHEMA_FILTER '^(dbo)$'
);

ATTACH '' AS TO_MSSQL (
    TYPE mssql,
    SECRET secret
);
""")

sql = "SELECT * FROM TO_MSSQL.dbo.MyTable1 LIMIT 10;"
```
```python
def run_query(parent_con, name):
    print(f"Starting {name}")
    child_con = parent_con.cursor()
    try:
        # same pattern as DuckDB docs: per-thread cursor on shared connection
        for _ in child_con.execute(sql).fetchall():
            pass
    finally:
        child_con.close()
    print(f"Finished {name}")


def test():
    t1 = Thread(target=run_query, args=(con, "task1"))
    t2 = Thread(target=run_query, args=(con, "task2"))

    t1.start()
    t2.start()
    t1.join()
    t2.join()
```
```python
test()
```

[...]

PR #179, opened 2026-07-08
fix(catalog): TSan/ASan-confirmed concurrency races behind concurrent-read crashes (#178)

VGSML:
## Summary

Fixes the crash-grade concurrency bugs found while investigating #178 (process crash under concurrent queries against the same attached catalog). Full root-cause analysis and repro history: `docs/proposals/issue-178-concurrent-crash-analysis.md` (included in this PR).

### The crash (ASan-proven on pre-fix main)

`MSSQLMetadataCache` guarded `schemas_` with **two different mutexes**: `Refresh()` (and `HasSchema`/`HasTable`) used `mutex_`, while every hot-path reader/writer (`GetTableMetadata`, `EnsureSchemasLoaded`, `ForEachTableInSchema`, `LoadAllTableMetadata`, …) used `schemas_mutex_`. There was **no mutual exclusion** between a `mssql_refresh_cache()` freeing the whole map and a concurrent bind iterating it:

```
==ERROR: AddressSanitizer: heap-use-after-free ... READ of size 17
    #2 MakeTableInfo src/catalog/mssql_table_entry.cpp:46
    #4 MSSQLTableEntry::MSSQLTableEntry(...)
```

Reproduced on unmodified main with the spec-052 harness scenario 6 (readers + `mssql_refresh_cache` invalidator + `duckdb_tables()` walker): **heap-use-after-free abort in 2 of 3 runs**. Any workload mixing reads with refresh/invalidation/catalog enumeration (IDE introspection, notebooks, dbt) can hit this — the most plausible mechanism for the segfaults reported in #178 on extension v0.2.1.

### Fixes

1. **One cache-wide `mutex_`** in `MSSQLMetadataCache` (merges `mutex_`/`schemas_mutex_`). Also removes the per-schema `load_mutex` in `EnsureTablesLoaded` (it excluded nothing — the containers it "protected" are read under the map mutex) and the unlocked double-checked fast paths; `GetDatabaseCollation` returns by value (a reference would outlive the lock).
2. **`MSSQLTableSet::Scan`**: the `known_table_names_` insert now runs under `names_mutex_` like every other access site (it mutated the set under `entry_mutex_` only — TSan caught it racing `Invalidate()`'s clear). Lock order `names_mutex_ → entry_mutex_` matches `GetEntry`'s existing nesting.
3. **Invalidation epoch guard**: `Scan`'s unconditional trailing `names_loaded_/is_fully_loaded_ = true` stomped a concurrent `Invalidate()`, leaving TRUE flags over just-cleared containers — `GetEntry` then answered *"Table … does not exist"* for existing tables (observed in harness scenario 6 once the memory races were fixed). `Invalidate`/`InvalidateEntry` bump an epoch counter; `Scan` publishes the flags only if the epoch is unchanged.
4. **Hygiene (TSan-clean)**: `ttl_seconds_`/`metadata_timeout_ms_` become atomics — they are rewritten by `EnsureCacheLoaded` on every catalog lookup while loaders read them mid-query with the cache mutex held. Default comes from `tds::DEFAULT_METADATA_TIMEOUT`; the `mssql_metadata_timeout` setting overrides it at runtime exactly as before. The `static int level = -1` lazy-init in 29 debug loggers is replaced with thread-safe magic statics.

### Verification (Linux arm64 sanitizer lab, SQL Server 2022)

| Check | pre-fix main | this branch |
|---|---|---|
| spec-052 harness scenarios 1–8, **ASan+UBSan** | **heap-use-after-free abort** (2/3 runs, scenario 6) | clean ×2, all PASS |
| same suites under **TSan** | 5 + 55 race reports (incl. the two crash-grade ones) | **0 reports**, all PASS |
| new `test/sql/integration/issue178_concurrent_same_table.test` (8-connection `concurrentloop`, same 500k-row table, LIMIT-abandoned streams) | n/a | PASS (release, ASan, TSan) |
| `make test` (unit) + release build (macOS + Linux) | — | PASS |

### Not in this PR (follow-ups from the analysis doc)

- `stability = VOLATILE` for side-effecting scalar functions (`mssql_invalidate_cache` trips a debug-only ExpressionExecutor assert and is constant-folded at plan time in release builds)
- `GetTableMetadata` copy-out-under-lock (retire the raw-pointer-into-cache contract)
- Windows concurrency CI job (the #178 report is Windows-only evidence; no Windows concurrency coverage exists today)
- Reporter follow-up on #178: DuckDB 1.5.2/1.5.3 community channels serve pre-spec-052 binaries (0.1.18 / 0.2.0) containing the known #126 crash; need `mssql_version()` confirmation from the crashing 1.5.4 environment

DATAMODEL.md updated with the new cache locking invariant (per project policy).

Refs #178

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -147,7 +147,7 @@ duckdb --unsigned -c "INSTALL mssql FROM local_build_debug; LOAD mssql;"
 ## Key Architecture Concepts
 
 - **Custom TDS implementation**: No external TDS/ODBC library. TDS v7.4 protocol with PRELOGIN, LOGIN7, SQL_BATCH, ATTENTION packet types.
-- **Connection pool**: Thread-safe pool, **owned per `MSSQLCatalog` via `unique_ptr<ConnectionPool>` (spec 047)** — pool lifetime is bounded by catalog lifetime; no singleton, no cross-instance sharing. Two ATTACHes against the same DSN under different aliases get independent pools; DETACH tears down the pool deterministically via RAII. Idle timeout, background cleanup, configurable limits as before.
+- **Connection pool**: Thread-safe pool, **owned per `MSSQLCatalog` (spec 047; since the issue #178 review a `shared_ptr` whose SOLE strong reference is the catalog — result streams hold `weak_ptr` handles for safe destructor-time release)** — pool lifetime is bounded by catalog lifetime; no singleton, no cross-instance sharing. Two ATTACHes against the same DSN under different aliases get independent pools; DETACH tears down the pool deterministically via RAII. Idle timeout, background cleanup, configurable limits as before.
 - **Transaction support**: Connection pinning maps DuckDB transactions to SQL Server transactions via 8-byte ENVCHANGE descriptors. See `docs/transactions.md`.
 - **Catalog integration**: DuckDB Catalog/Schema/Table APIs with incremental metadata cache (lazy loading, TTL-based expiration, point invalidation), primary key discovery for rowid support.
 - **DML**: INSERT uses batched VALUES; UPDATE/DELETE use rowid-based VALUES JOIN pattern with deferred execution in transactions.
--- a/DATAMODEL.md
+++ b/DATAMODEL.md
@@ -30,7 +30,7 @@ flowchart TD
         tx[MSSQLTransaction]
     end
     subgraph L2["Layer 2 — Pool (spec 047)"]
-        pool["ConnectionPool<br/>(per-catalog, unique_ptr)"]
+        pool["ConnectionPool<br/>(per-catalog; sole strong ref,<br/>streams hold weak_ptr)"]
     end
     subgraph L1["Layer 1 — Protocol (TDS)"]
         conn[TdsConnection]
@@ -150,7 +150,7 @@ classDiagram
         +LookupSchema()
         +RegisterStream(stream) uuid
         +RetrieveStream(uuid) stream
-        -connection_pool_ : unique_ptr~ConnectionPool~
+        -connection_pool_ : shared_ptr~ConnectionPool~ (sole strong ref)
         -metadata_cache_ : unique_ptr~MSSQLMetadataCache~
         -statistics_provider_ : unique_ptr~MSSQLStatisticsProvider~
         -schema_entries_ : map~string,shared_ptr~MSSQLSchemaEntry~~
--- a/DATAMODEL.md
+++ b/DATAMODEL.md
@@ -276,7 +276,7 @@ classDiagram
     note for TokenCache "Spec 047 FR-012:<br/>two DatabaseInstances<br/>sharing a secret name<br/>do NOT alias"
 ```
 
-- `MSSQLMetadataCache` is incremental and lazy. Hand-out contract is documented in the header: the caller (`MSSQLTableSet::LoadSingleEntry`) copies the metadata immediately, so the cache's raw-pointer return is safe.
+- `MSSQLMetadataCache` is incremental and lazy. `GetTableMetadata` **copies** the metadata out under the cache mutex — the previous raw-pointer return escaped the lock and raced `Refresh` / bulk reloads freeing the map node (issue #178 review finding).
 - **Locking invariant (issue #178)**: ONE cache-wide `mutex_` guards `schemas_` and everything reachable through it (tables, columns, load states), plus `state_` / `database_collation_`. Loads hold it across their SQL round trip so partial state is never visible — concurrent metadata loads serialize by design. `ttl_seconds_` / `metadata_timeout_ms_` are atomics (written per-lookup by `EnsureCacheLoaded`, read by loaders mid-query while the mutex is held). The pre-#178 split (`mutex_` for Refresh/HasSchema, `schemas_mutex_` for everything else) let `Refresh()` free the whole map under a reader — TSan-confirmed UAF.
 - `MSSQLStatisticsProvider` returns stats by value; no raw-pointer hand-out.
 - `TokenCache` is the only remaining process-wide static, but it is **namespaced by `DatabaseInstance*`** (spec 047 FR-012) so two embeddings can use the same Azure secret name without aliasing.
--- a/DATAMODEL.md
+++ b/DATAMODEL.md
@@ -277,6 +277,7 @@ classDiagram
 ```
 
 - `MSSQLMetadataCache` is incremental and lazy. Hand-out contract is documented in the header: the caller (`MSSQLTableSet::LoadSingleEntry`) copies the metadata immediately, so the cache's raw-pointer return is safe.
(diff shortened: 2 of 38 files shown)


VGSML:
## Self-review round (multi-agent, high effort)

Ran an adversarial workflow-backed review of the first commit — **10 verified findings (5 root causes)**, all fixed in 8b3d600:

| # | Finding | Fix |
|---|---|---|
| 1 | **Residual UAF**: `GetTableMetadata`'s raw-pointer return escaped the cache mutex; `LoadSingleEntry` dereferenced it while `Refresh`/`LoadAllTableMetadata`/TTL reloads freed the map node | metadata **copied out under the mutex** (`bool GetTableMetadata(..., out_meta)`) |
| 2 | **Stale resurrection**: epoch guard covered only Scan's trailing flag stores — Scan's fill and `LoadSingleEntry`'s publication could re-insert pre-invalidation entries after `Invalidate()` | both publications epoch-guarded; dirty epoch → entry served to the in-flight call only (anchored), nothing persists |
| 3 | **Negative-cache poisoning**: GetEntry step 4 read `names_loaded_` before `names_mutex_`; racing Invalidate could latch a permanent `attempted_tables_` negative for an existing table | flag re-checked under the lock |
| 4 | **Lock-hold latency**: Scan held `names_mutex_`+`entry_mutex_` while blocking on the cache mutex (held across metadata round trips) → SELECTs on cached tables stalled | Scan restructured into 3 passes; table-set locks and cache mutex never held together |
| 5 | **`~MSSQLResultStream` race** (TSan-caught in re-verification): destructor ran `MetaTransaction::Get` on a worker thread racing the client thread's `TransactionContext::Commit` | owning catalog + pinned flag captured at construction; destructor never touches a `ClientContext`. Also: `Refresh()` failure no longer latches `schemas_load_state_=LOADED` over an emptied map; dead `EnsureNamesLoaded` deleted |

**Re-verification (Linux arm64 lab):** spec-052 harness scenarios 1–8 — TSan **0 reports**, ASan ×2 clean; issue-178 concurrentloop under TSan clean; `make test` + full `test/sql/*` sqllogictest sweep green.

oluies (dismissed):

oluies:
1. `~MSSQLResultStream` traded a fail-safe by-name catalog lookup (try/catch, worst case drops the connection) for a raw `MSSQLCatalog*` dereference. The lifetime argument rests on DuckDB serializing DETACH against active queries plus `Shutdown()` asserting quiescence — and asserts compile out in release. If the argument is ever wrong, the old code leaked a connection; the new code is UB. A `weak_ptr` to the pool (or nulling registered streams' catalog pointer during catalog teardown) would keep the fix while restoring the safe failure mode.

2. The destructor now calls `SetNeedsReset(true)` unconditionally on every non-pinned release, where `ConnectionProvider::ReleaseConnection` previously decided. If that adds a reset round trip on next reuse for every streamed query — including cleanly drained streams — that's a per-query cost beyond the race fix. Check whether reset can stay conditional on `conn_state != Idle`.

3. metadata loads now serialize globally since `mutex_` is held across SQL round trips (acknowledged in the comments — cold-start with dbt `threads=8` binds sequentially; fine for a crash fix, candidate for per-schema condition-variable load states later). `GetDatabaseCollation` by value is correct — the return copy is initialized before the lock_guard destructs. Scan's Pass B calls `PreloadRowCount` under the cache mutex, establishing cache→stats lock order; appears safe as stats is a leaf, but note it. The test's `Issue178Big` table (500k rows) persists in the target DB via the `IF OBJECT_ID` guard — deliberate for rerun speed presumably, but a comment saying so would prevent someone "fixing" it with a DROP that makes concurrent CI runs race on setup.

VGSML:
@oluies thanks — addressed in 4272806:

**1. Raw `MSSQLCatalog*` → `weak_ptr` (fixed as suggested).** The pool is now a `shared_ptr` on `MSSQLCatalog` with the catalog holding the **sole strong reference** (spec 047's deterministic DETACH teardown is preserved — weak handles can't extend the pool's lifetime, except transiently deferring `~ConnectionPool` to a concurrent in-flight `Release` that won the `lock()`, which is strictly better than either UB or a leak). `~MSSQLResultStream` captures the `weak_ptr` at construction; a failed `lock()` degrades to dropping the connection. The `nulling registered streams` alternative wouldn't cover scan-local streams (owned by `MSSQLScanGlobalState`, unreachable from the catalog), hence `weak_ptr`.

**2. `SetNeedsReset(true)` — no behavior change, and no extra round trip.** The pre-existing autocommit release path has always set it unconditionally: `ConnectionProvider::ReleaseConnection` → *"Autocommit mode, flagging connection for reset"* → `conn->SetNeedsReset(true)` on every non-pinned release, including cleanly drained streams. The destructor mirrors that exactly. Cost-wise, `RESET_CONNECTION` is a status bit in the TDS header of the **next** SQL_BATCH (MS-TDS 2.2.3.1.2) — the server resets session state while processing that batch; there is no separate reset round trip. Added a comment at the call site so the next reader doesn't re-derive this.

**3. Test table persistence** — documented in the test file: the `IF OBJECT_ID` guard is deliberate (500k-row population dominates runtime; a DROP/re-create would make concurrent CI runs against a shared server race on setup).

Also noted: the metadata-load serialization under the single cache mutex is acknowledged as the crash-fix tradeoff; per-schema load states with a condition variable are the natural follow-up if cold-start bind latency with high thread counts becomes a real complaint. The cache→stats leaf lock order from Scan's Pass B is now the only cross-component nesting on that path (table-set locks are never held with the cache mutex after the 3-pass restructure).

Re-verified after the change: TSan harness **0 reports**, ASan clean, issue-178 concurrentloop under TSan clean; `make test` + `integration-test` green.

oluies (dismissed):
LGTM 

if a stream destructor’s lock() wins the race against a concurrent DETACH, ~ConnectionPool runs on the worker thread, inside the stream destructor, after Release(), so Release() must tolerate being called on a pool that OnDetach/Shutdown has already quiesced (drop, not enqueue), and the pool destructor must be safe to run off the client thread (it joins the background cleanup thread from whatever thread drops the last reference). 

if Release-after-Shutdown enqueues into a dead pool, that’s a real bug in the rare path.

VGSML:
@oluies good edge case — traced it through, and both requirements hold; hardened anyway in 8ea61de.

**The DETACH-race sequence is safe by construction:**
1. Stream destructor's `lock()` wins → it holds a strong pool reference.
2. `~MSSQLCatalog` drops the catalog's strong ref → pool stays **alive and not shut down** (refcount 1, and `Shutdown()` has no callers besides `~ConnectionPool` — verified by grep).
3. `Release()` runs against a live pool — the connection is enqueued into `idle_connections_` normally.
4. The stream drops the last ref → `~ConnectionPool` → `Shutdown()` on the worker thread: signals the flag, **joins `cleanup_thread_`** (legal from any thread; the cleanup thread holds no pool references, so self-join is impossible), then closes everything in `idle_connections_` — including the connection Release just enqueued. Nothing leaks, `active_connections_` is empty so the quiescence warning/assert stays silent.

So "Release against an already-quiesced pool" cannot happen from the stream path: any `Release` caller on this path holds a strong ref for the duration of the call, and `Shutdown` only ever runs from the destructor, i.e., strictly after all such refs are gone. `Release`'s first line also already checks `shutdown_flag_` and returns without enqueueing (the dropped `shared_ptr` closes the socket) — that covers the legacy raw-reference callers inside queries, where DuckDB's DETACH quiescence is the pre-existing contract.

**Hardening added regardless** (8ea61de): `shutdown_flag_` is now re-checked **under `pool_mutex_`** in `Release` — if any future code path ever races a shutdown past the unlocked check, the connection is closed instead of enqueued into a quiesced pool. Both contracts are documented at the check site and in `Shutdown()`'s new thread-affinity note (it may run on whatever thread drops the last reference).

Re-verified: TSan harness scenarios 1–8 — 0 reports; `make test` + issue-178 concurrentloop green.

oluies (commented):
LGTM

PR #187, opened 2026-07-15
fix(copy): BCP COPY to VARCHAR(n) rejects valid strings (#181)

### LINKED ISSUE #181: BCP COPY to VARCHAR(n) rejects valid strings
`COPY ... FORMAT 'bcp'` seems to reject valid VARCHAR(n) values.

Example:
SQL Server:
```
CREATE TABLE dbo.duckdb_bcp_varchar_test (
    s VARCHAR(10) NOT NULL
);
```

DuckDB:
```
COPY (
    SELECT 'aaaaaa' AS s
) TO 'ms.dbo.duckdb_bcp_varchar_test'
  (FORMAT 'bcp', CREATE_TABLE false);
```

Expected: this should load. VARCHAR(10) should accept 6 ASCII chars.

Actual: it fails with a length/capacity error.

It looks like the BCP path encodes strings as UTF-16LE, then checks that byte length against the VARCHAR(n) limit. 
That makes VARCHAR(10) behave like it only allows 5 ASCII chars.

VARCHAR(n) should be checked as SQL Server VARCHAR capacity, not as UTF-16 bytes.

oluies:
Fixes #181.

## Root cause

The BCP path transmits all character data as NVARCHAR (UTF-16) and lets SQL Server convert to the target column type. For an **existing** target table, `TargetResolver::GetExistingTableColumnMetadata` passed `sys.columns.max_length` through unchanged as the declared NVARCHAR wire length:

- `varchar(10)` → `sys.columns.max_length = 10` (single-byte characters)
- declared on the wire as NVARCHAR with `max_length = 10` **UTF-16 bytes** = 5 characters
- `INSERT BULK` declaration became `nvarchar(5)`

So the server rejected any value longer than n/2 characters — `'aaaaaa'` (6 ASCII chars) into `VARCHAR(10)` failed.

## Fix

In `SQLServerTypeMaxLength` (src/copy/target_resolver.cpp), split the char-type branch:

- **varchar/char**: declare `2 × sys.columns.max_length` UTF-16 bytes; when `n > 4000` (would exceed the 8000-byte inline NVARCHAR limit), fall back to PLP `nvarchar(max)`.
- **nvarchar/nchar**: unchanged — `sys.columns.max_length` is already in UTF-16 bytes.

The 2× declaration is an upper bound on what the client may send per row; the server still enforces the target column's real capacity at conversion time, so genuinely over-long values keep failing with the normal truncation error (covered by the test).

## Test

`test/sql/copy/copy_varchar_length.test` (integration, requires SQL Server):
- the exact #181 repro (6 chars into `VARCHAR(10)`, `CREATE_TABLE false`)
- exact-capacity boundary (10 chars into `VARCHAR(10)`, `CHAR(10)`, `NVARCHAR(10)`)
- over-capacity value still rejected
- `VARCHAR(8000)` with an 8000-char value exercising the PLP fallback

Not run locally (fresh worktree, no local build) — relying on CI integration tests.

### DIFF
--- a/src/copy/target_resolver.cpp
+++ b/src/copy/target_resolver.cpp
@@ -717,12 +717,25 @@ static uint16_t SQLServerTypeMaxLength(const string &type_name, int16_t max_leng
 		return 8;
 	} else if (type_lower == "smallmoney") {
 		return 4;
-	} else if (type_lower == "varchar" || type_lower == "nvarchar" || type_lower == "char" || type_lower == "nchar") {
+	} else if (type_lower == "varchar" || type_lower == "char") {
 		// For (max), max_length is -1 in sys.columns
 		if (max_length == -1) {
 			return 0xFFFF;	// MAX indicator
 		}
-		// For nvarchar/nchar, max_length is in bytes (2 bytes per char)
+		// varchar/char data is transmitted as NVARCHAR over the BCP wire, so the
+		// declared length must be UTF-16 bytes: sys.columns reports single-byte
+		// characters, hence 2x. varchar(n) with n > 4000 exceeds the 8000-byte
+		// inline NVARCHAR limit and must be sent as nvarchar(max) (PLP).
+		if (max_length > 4000) {
+			return 0xFFFF;	// MAX indicator (PLP)
+		}
+		return static_cast<uint16_t>(max_length * 2);
+	} else if (type_lower == "nvarchar" || type_lower == "nchar") {
+		// For (max), max_length is -1 in sys.columns
+		if (max_length == -1) {
+			return 0xFFFF;	// MAX indicator
+		}
+		// For nvarchar/nchar, sys.columns max_length is already in bytes (2 bytes per char)
 		return static_cast<uint16_t>(max_length);
 	} else if (type_lower == "text" || type_lower == "ntext") {
 		return 0xFFFF;	// MAX indicator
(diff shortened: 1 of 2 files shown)


oluies:
**Question: should varchar with a Unicode (UTF-8) collation be treated the same as nvarchar here?**

While working on this we considered whether `varchar(n)` columns using a UTF-8 collation (`*_UTF8`, SQL Server 2019+) should take the nvarchar path (pass `sys.columns.max_length` through undoubled) instead of the 2× rule. Our conclusion was **no** — the 2× rule is correct for every varchar collation — but flagging it for review since it's subtle:

The declared NVARCHAR wire length only needs to be an upper bound on the UTF-16 size of any value that legitimately fits the target column; the server enforces the target's real capacity at conversion time. For `varchar(n)` the worst case per collation is:

| Target charset | worst-case content | UTF-16 expansion vs stored bytes |
|---|---|---|
| single-byte (Latin1, …) | n 1-byte chars | 2× |
| DBCS (e.g. CP932) | 2-byte char → 2 UTF-16 bytes | 1× |
| UTF-8 collation | ASCII 1→2 bytes; 2/3-byte seq → 2 bytes; 4-byte → 4 (surrogate pair) | ≤ 2× |

So 2× is tight for the ASCII case (the #181 repro) and a safe over-estimate otherwise. Treating UTF-8-collated varchar like nvarchar (1×) would reintroduce the bug for ASCII data: `varchar(10)` with a UTF-8 collation can hold 10 ASCII chars = 20 UTF-16 bytes, but the pass-through would declare 10.

The one place collation awareness *could* buy something is the PLP cutoff: `varchar(n)` with `4000 < n ≤ 8000` now always falls back to `nvarchar(max)` (PLP), even though actual row values might fit inline. We can't know value sizes ahead of time, so the conservative fallback seemed right — but if the per-row PLP overhead matters for wide-varchar-heavy workloads, an alternative would be checking actual encoded lengths per batch. Opinions welcome.

VGSML (approved):

Issue #188, opened 2026-07-15
COPY ... TO (FORMAT 'bcp') can fail on existing VARCHAR/CHAR targets due to NVARCHAR length metadata mismatch


dhaddad-accumtech:
### Summary
When using `COPY ... TO ... (FORMAT 'bcp')` to write to an existing MSSQL table with `varchar/char` columns, the operation can fail with either:

```
Invalid Input Error: MSSQL COPY: Failed to execute INSERT BULK: Incorrect syntax near the keyword 'with'.
If this statement is a common table expression, an xmlnamespaces clause or a change tracking context clause,
the previous statement must be terminated with a semicolon.
```

or

```
Invalid Input Error: MSSQL: NVARCHAR column 'MemberCode' overflow: value is 9 UCS-2 code units (18 UTF-16LE bytes) but column max is 7 code units (15 bytes)
```

A regular `INSERT INTO schema.target_table (...) SELECT ...` succeeds for the exact same rows.

### How To Reproduce
1. In MSSQL, create a target table with `varchar` columns (example: `Id int identity(1, 1) not null, MemberCode varchar(15)).
2. In DuckDB, attach an MSSQL database/catalog.
3. Create a local temp table with string data that should fit those target column sizes.
4. Run:
   ```sql
   COPY my_table TO 'mssql_db.schema.target_table' (FORMAT 'bcp', CREATE_TABLE false, FLUSH_ROWS 500000, TABLOCK true);
   ```

### Expected Result
`COPY ... FORMAT 'bcp'` succeeds for values that fit the target `varchar(N)` columns, matching normal `INSERT` behavior.

### Actual Result
`COPY` fails with either:
* An `INSERT BULK` syntax error near `WITH`.
* A false `NVARCHAR` overflow where the reported max length is effectively half of the expected character capacity.

### Root Cause
In the BCP metadata path (`src/copy/target_resolver.cpp`), `varchar/char` targets are normalized to NVARCHAR wire metadata (intentional), but length handling is inconsistent:
- NVARCHAR declaration rendering could derive invalid lengths from odd byte counts (e.g., ending up as `nvarchar(0)`).
- Existing-table `varchar(N)` max length is not converted to UTF-16 byte capacity (`2N`) when used with NVARCHAR metadata, causing false overflow checks on the client side.

### Evidence
- The same data loads successfully via regular  `INSERT INTO ... SELECT ...`.
[...]

PR #190, opened 2026-07-16
fix(scan): catalog scans silently truncated wide VARCHAR/CHAR and TEXT

oluies:
> ## ⚠️ Behavior change — please read first (@VGSML)
>
> **This reverses a documented, tested decision and needs your sign-off.**
>
> `VARCHAR(n > 4000)` → `NVARCHAR(4000)` truncation was deliberate: it's in the README type-mapping
> table, in the Limitations list, in spec 026 (*"Key limitation (VARCHAR >4000 truncation) is
> documented and acceptable per user requirements"*), and asserted by test **T027**. This PR
> reverses it, so wide columns return their full value instead of a silently shortened one.
>
> **Why I think the original decision was based on a false premise.** spec 026's `research.md` says:
>
> > VARCHAR(8000) at max → **must** cap at NVARCHAR(4000) with silent truncation
>
> There is no such "must". A column wider than 4000 *does* have a valid NVARCHAR target —
> `NVARCHAR(MAX)` — and the extension **already** CASTs `VARCHAR(MAX)` → `NVARCHAR(MAX)` on this exact
> code path. I verified `NVARCHAR(MAX)` round-trips all 8000 characters against SQL Server 2022. The
> cap never bought anything; it only cost data.
>
> If you'd rather keep the documented truncation, the TEXT half below stands on its own and I'll
> split it out — TEXT was never in spec 026's scope and is an unambiguous bug.

## The bug

Reading a non-UTF8 `CHAR`/`VARCHAR` column through the catalog CASTs it to NVARCHAR so it decodes as
UTF-8 (spec 026). `GetNVarcharLength` derived that CAST length from `sys.columns.max_length` and
capped it at 4000. Measured against SQL Server 2022 — server value vs. what a catalog scan returns:

| column | server | catalog scan (before) |
|---|---|---|
| `varchar(4000)` | 4000 | 4000 ✓ |
| `varchar(4001)` | 4001 | **4000** |
| `varchar(8000)` | 8000 | **4000** |
| `char(5000)` | 5000 | **4000** |
| `varchar(max)` | 9000 | 9000 ✓ |
| **`text`** | 9000 | **16** |

**No error is raised — the query just returns a shorter string.** `mssql_scan()` is unaffected (it
applies no CAST), so the same column read two ways disagreed:

```
┌──────────────────┬─────────┐
│ catalog scan     │ 4000    │
│ mssql_scan raw   │ 8000    │
└──────────────────┴─────────┘
```

`TEXT` is the worst case and was **never in spec 026's scope** (the spec doesn't mention it at all).
`sys.columns.max_length` for `text` is **16** — the size of the in-row pointer, not the data (which
runs to 2 GB) — so the scan emitted, verbatim via `MSSQL_DEBUG=2`:

```sql
SELECT CAST([v8000] AS NVARCHAR(4000)) AS [v8000], CAST([t] AS NVARCHAR(16)) AS [t] FROM [dbo].[i188_trunc]
```

Only non-UTF8 collations are affected (`is_utf8` columns skip conversion). Present since **v0.1.15**
(5fbe3c1, 2026-02-02) — every release including v0.2.1.

## The fix

`GetNVarcharLength` now takes the column so it can see the type name:

```
max_length == -1   -> "MAX"   (unchanged)
text               -> "MAX"   (was NVARCHAR(16))
max_length > 4000  -> "MAX"   (was "4000", truncated)
else               -> n       (unchanged)
```

`TEXT` also now follows `mssql_convert_varchar_max` like other LOBs, so turning that setting off
skips the conversion rather than truncating.

This mirrors the `>4000 → PLP nvarchar(max)` fallback #187 just added to the BCP **write** path in
`SQLServerTypeMaxLength` — same limit, same reasoning, opposite direction.

## Testing

Run against SQL Server 2022 locally (**note: CI cannot run these — see below**).

- **New** `test/sql/catalog/scan_wide_varchar_length.test` — all six columns above, content (not just
  length) preserved, and catalog scan vs. `mssql_scan` agreement.
- **Rewritten** T027 in `varchar_encoding.test` — asserted truncation to 4000; now asserts the full
  5000, plus a `RIGHT()` check that the tail survives.
- Both **fail without the code change** and pass with it.
- `test/sql/catalog/*`: **30/30 pass** (1314 assertions).
- `test/sql/copy/*`: identical 4 failures before and after this change (verified by rebuilding
  unmodified `main`) — **no regressions**. Those 4 are pre-existing and unrelated; see below.

## Two things this PR deliberately does NOT fix

1. **CI never runs any of these tests.** The `integration-test` job is `workflow_dispatch`-only, and
   even when dispatched it only pipes `test/sql/smoke_test.sql` through the CLI — it never invokes
   the SQLLogicTest runner or sets `MSSQL_TESTDB_DSN` (that comes from the Makefile). Every
   `.test` file is green-by-skip, `require-env` skips silently. That's how this bug survived
   5 months with a test suite that covers the exact column. Filed separately.

2. **`copy_varchar_length.test` (from #187) still fails**, now at line 69 instead of 63 — the
   `VARCHAR(8000)` assertion this PR fixes now passes, but a COPY that fails client-side validation
   leaks its pooled connection, leaving the `INSERT BULK` transaction open and the target table
   locked until the process exits, so the test's own cleanup `DROP` times out. Separate bug, filed
   separately, needs a destructor on `MSSQLCopyGlobalState`.

## Context

Found while verifying that #188 (duplicate of #181) was genuinely fixed by #187. #187's write path
is correct — the server stores all 8000 characters; it was the read-back that truncated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1674,8 +1674,12 @@ VARCHAR and CHAR columns with non-UTF8 collations (e.g., `Latin1_General_CI_AS`)
 | VARCHAR Length | Converted To | Notes |
 |----------------|--------------|-------|
 | VARCHAR(n) where n ≤ 4000 | NVARCHAR(n) | Full data preserved |
-| VARCHAR(n) where n > 4000 | NVARCHAR(4000) | Data may be truncated to 4000 characters |
+| VARCHAR(n) where n > 4000 | NVARCHAR(MAX) | Full data preserved. NVARCHAR has no inline length above 4000 characters, so these columns are sent as PLP. |
+| CHAR(n) | same as VARCHAR(n) | Full data preserved (blank-padded to `n` by SQL Server) |
+| TEXT | NVARCHAR(MAX) | Full data preserved; converted by default, disable with `mssql_convert_varchar_max = false` |
 | VARCHAR(MAX) | NVARCHAR(MAX) | Converted by default; disable with `mssql_convert_varchar_max = false` |
+
+> **Note**: Before v0.2.2, `VARCHAR(n > 4000)` was converted to `NVARCHAR(4000)` and `TEXT` to `NVARCHAR(16)`, both of which silently truncated the value on read. Both now convert to `NVARCHAR(MAX)` and return the full value.
 
 **VARCHAR Encoding Setting:**
 
@@ -1728,7 +1732,6 @@ The codec validates the input first and falls back to a slower scalar implementa
 - XML columns in INSERT/UPDATE are limited to 4096 bytes per value — use COPY TO with BCP protocol for larger documents
 - Very large DECIMAL values may lose precision at extreme scales
 - Connection pool statistics reset when all connections close
-- VARCHAR columns >4000 characters with non-UTF8 collations are truncated when queried via catalog (see VARCHAR Encoding above)
 - NVARCHAR(max) and VARCHAR(max) columns are not supported in the Microsoft Fabric Warehouse
 
 ## Third-Party Licenses
--- a/README.md
+++ b/README.md
@@ -1676,10 +1676,12 @@ VARCHAR and CHAR columns with non-UTF8 collations (e.g., `Latin1_General_CI_AS`)
 | VARCHAR(n) where n ≤ 4000 | NVARCHAR(n) | Full data preserved |
 | VARCHAR(n) where n > 4000 | NVARCHAR(MAX) | Full data preserved. NVARCHAR has no inline length above 4000 characters, so these columns are sent as PLP. |
 | CHAR(n) | same as VARCHAR(n) | Full data preserved (blank-padded to `n` by SQL Server) |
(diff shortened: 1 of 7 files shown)


oluies:
Related: #191 (COPY connection leak — why `copy_varchar_length.test` still fails at line 69) and #192 (CI never runs the `.test` files — how this bug survived 5 months with a test covering the exact column). Closed #188 as a duplicate of #181; #187's write path verified correct against SQL Server 2022.

oluies:
Companion PR #193 fixes #191 (the BCP connection leak). Independent and disjoint from this one — together they turn `copy_varchar_length.test` (shipped red in #187) green; it needs both halves.

VGSML (approved):

VGSML:
Thanks @oluies, that's pass.

Issue #165, opened 2026-06-18
CI: Windows MSVC build fails in DuckDB bundled fmt/format.h (C2143/C2447) on windows-latest


oluies:
## Summary

The **Windows MSVC** build job (`ci.yml` → `build-windows-msvc`, manual `workflow_dispatch` with `run_windows_build=true`) is failing in **DuckDB's own bundled `fmt`**, not in extension code. The compile dies very early (step **8/712**), while building DuckDB core third-party libraries — long before any `src/` or extension code is reached.

This is **pre-existing and repo-wide** (any branch built on today's `windows-latest` MSVC hits it), not introduced by any recent feature PR. Surfaced while validating #162; that PR does **not** touch the DuckDB submodule, `CMakeLists.txt`, or `extension-ci-tools`.

## Error

```
duckdb\third_party\fmt\include\fmt/format.h(337): error C2143: syntax error: missing ';' before '{'
duckdb\third_party\fmt\include\fmt/format.h(337): error C2447: '{': missing function header (old-style formal list?)
ninja: build stopped: subcommand failed.
```

## Evidence

- Failing run: https://github.com/hugr-lab/mssql-extension/actions/runs/27768275954
- Failing job: **Build Windows (MSVC)** — https://github.com/hugr-lab/mssql-extension/actions/runs/27768275954/job/82160915235
- **MinGW build in the same run: success** — this is **MSVC-only**.
- Linux/macOS Build + Smoke + Integration: success (run 27766244734).

## Likely cause

Incompatibility between the **current `windows-latest` MSVC toolchain** (cl.exe / Windows SDK on the GitHub-hosted image) and the **`fmt` version bundled in the pinned DuckDB submodule** (`14eca11bd9d4a0de2ea0f078be588a9c1c5b279c`). A `C2143` at `fmt/format.h:337` is the classic signature of a newer MSVC parsing an older bundled-`fmt` construct that newer cl.exe rejects.

## Candidate fixes (need triage)

1. **Bump the DuckDB submodule** to a SHA whose bundled `fmt` compiles under current MSVC (preferred if a compatible SHA exists).
2. **Pin the MSVC toolchain / Windows SDK** version in `ilammy/msvc-dev-cmd` to the last-known-good combination, so hosted-image drift doesn't move under us.
3. **Patch/upgrade bundled `fmt`** in the DuckDB submodule (least desirable — touches vendored third-party).

## Notes

[...]

PR #196, opened 2026-07-16
chore(deps): move OpenSSL 3.4.1 → 3.5.4 (LTS) ahead of 3.4 EOL

oluies:
## Why now

**OpenSSL 3.4 is not LTS and goes end-of-life on 2026-10-22 — 97 days away.** After that it gets no security fixes at all. 3.5 is the LTS line, supported through 2030-04-08.

We're pinned to **3.4.1** (Feb 2025), five patch releases behind 3.4.6.

## Reachability — read this before treating it as urgent

I checked what the extension actually calls rather than counting CVEs. The scary ones **do not apply**:

- The High-severity 3.4.x CVEs are **CVE-2026-45447** (heap UAF in `PKCS7_verify`) and **CVE-2025-15467** (stack overflow in CMS `(Auth)EnvelopedData` parsing). Both are in CMS/PKCS7 code.
- The extension's OpenSSL surface is a **pure TLS client** — `SSL_read`, `SSL_get_error`, `SSL_new/free`, `BIO_set_retry_*`, custom BIO callbacks.
- Grepped `src/` for `PKCS7`, `CMS_`, `PKCS12`, `SM2`, `QUIC`, `OCB`, `GCM_SIV`, `CRMF`, `_CMP_`, `BIO_f_linebuffer`: **zero hits.**

What *is* plausibly reachable is ASN.1 certificate parsing (CVE-2026-34180 over-read, CVE-2026-7383 multibyte overflow) plus a couple of TLS-path issues — all rated **Low/Moderate**.

**So this is EOL-driven hygiene, not incident response.** Please don't merge it in a panic; merge it when the MinGW job is green.

## Why the baseline has to move (you can't just bump the override)

vcpkg's version database **only ever contained openssl `3.4.0` and `3.4.1`** for that series — 3.4.2–3.4.6 were never packaged. And the current baseline is not arbitrary:

| commit | date | what it is |
|---|---|---|
| `5ee5eee` | 2025-02-12 | current baseline — *literally* the `[openssl] Update to 3.4.1` commit |
| `52c9e08` | 2026-06-24 | this PR's baseline (vcpkg 2026.06.24) |

The old baseline's version DB tops out at 3.4.1, so a supported OpenSSL requires moving the baseline.

## Keeping the blast radius to exactly one thing

Moving the baseline would also drag **simdutf 6.1.1 → 8.2.0** (two major versions, right through the spec 043/044 codec layer). So simdutf is **explicitly pinned at 6.1.1**. Verified both are resolvable at the new baseline (`simdutf` uses the `version-semver` scheme; per the vcpkg docs plain `"version"` is the correct non-deprecated override key for every scheme).

**The only thing this PR actually changes is the OpenSSL version.**

`vcpkgGitCommitId` is deliberately **not** touched, and I verified why rather than assuming. run-vcpkg's bundle does:

```js
gitTool.arg(['clone', this.vcpkgUrl, '-n', '.']);        // no --depth → FULL clone
gitTool.arg(['checkout', '--force', this.vcpkgGitCommitId]);
```

A full clone, then checkout. So the Jan-2026 pinned checkout (`65604f5`) still has the Jun-2026 baseline's objects locally and resolves it with no network fetch. Leaving the vcpkg tool where it is keeps OpenSSL the single variable under test.

## This needs the MinGW job to be green before merge

**The pin is load-bearing.** `3d1277e` (vgsml, Jan 2026): *"Pin OpenSSL to 3.4.1 for MinGW compatibility — to resolve MinGW (x64-mingw-static) build failures in community-extensions CI."* MinGW is exactly where OpenSSL builds break, so 3.5.4 could reintroduce it.

`vcpkg.json` is part of the MinGW cache key (`vcpkg-windows-mingw-x64-mingw-static-${{ hashFiles('vcpkg.json') }}`), so this bump **forces a real OpenSSL rebuild** rather than a cache hit — the job genuinely tests the thing we care about.

Dispatched against this branch: run [29540869000](https://github.com/hugr-lab/mssql-extension/actions/runs/29540869000). I'll report the result here.

Note the dispatch also runs the MSVC job, which is independently red per #165 — ignore that one.

## If MinGW fails

Fallback order:
1. Try `3.5.x` at a different patch level.
2. Bump `vcpkgGitCommitId` to match the baseline (newer vcpkg may carry MinGW fixes for the OpenSSL port) — costs coherence with `release.yml`, which pins the same SHA in two places.
3. If 3.5 can't build on Rtools 4.2 MinGW at all, that's a real decision to surface: MinGW support vs. a supported TLS library after October.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -6,11 +6,17 @@
     "openssl",
     "simdutf"
   ],
-  "builtin-baseline": "5ee5eee0d3e9c6098b24d263e9099edcdcef6631",
+  "$comment-builtin-baseline": "vcpkg 2026.06.24. Moved off 5ee5eee (openssl 3.4.1) because that baseline's version DB tops out at 3.4.1 -- vcpkg never packaged 3.4.2-3.4.6, so reaching a patched OpenSSL requires moving the baseline, not just the override.",
+  "builtin-baseline": "52c9e08cdf8580d2d9762f547d22b96fd81e82f2",
+  "$comment-overrides": "Both deps are pinned so that moving the baseline changes exactly one thing: OpenSSL. simdutf stays at 6.1.1 (the old baseline's version) even though the new baseline offers 8.2.0 -- a two-major-version codec jump does not belong in a security bump and would confound the MinGW result this change exists to test. openssl 3.5.4 is the newest 3.5.x in vcpkg; 3.5 is LTS (supported to 2030-04-08) whereas 3.4 goes EOL 2026-10-22.",
   "overrides": [
     {
       "name": "openssl",
-      "version": "3.4.1"
+      "version": "3.5.4"
+    },
+    {
+      "name": "simdutf",
+      "version": "6.1.1"
     }
   ]
 }


oluies:
## MinGW: green ✅ — and it's a real test, not a cache hit

Run [29540869000](https://github.com/hugr-lab/mssql-extension/actions/runs/29540869000) — **all jobs passed**, including the two that matter here:

| Job | Result |
|---|---|
| Build Windows (MinGW) | ✅ success |
| Smoke Test Windows (MinGW) | ✅ success |
| Build Windows (MSVC) | ✅ success |
| Smoke Test Windows (MSVC) | ✅ success |
| Lint, C++ Unit Tests (linux + macos) | ✅ success |

### Proof OpenSSL 3.5.4 was genuinely compiled for MinGW

A green build would be worthless if vcpkg had silently reused a cached 3.4.1, so I checked the log rather than trusting the checkmark:

```
Cache not found for input keys: vcpkg-windows-mingw-x64-mingw-static-e76ee8a4...
Downloading https://github.com/openssl/openssl/archive/openssl-3.5.4.tar.gz
Building openssl:x64-mingw-static@3.5.4...
Elapsed time to handle openssl:x64-mingw-static: 6.9 min
-- Found OpenSSL: .../lib/libcrypto.a (found version "3.5.4")
-- TLS using OpenSSL version: 3.5.4
```

Cache miss → full 6.9-minute compile from source under Rtools 4.2 / `x64-mingw-static` → linked → **and the smoke test loads the extension**. The MinGW incompatibility that motivated the 3.4.1 pin in `3d1277e` does not reproduce on 3.5.4.

### The blast-radius pin held

vcpkg's resolved install plan confirms simdutf did **not** get dragged to 8.2.0:

```
The following packages will be built and installed:
    openssl:x64-mingw-static@3.5.4
    simdutf:x64-mingw-static@6.1.1
```

OpenSSL is the only thing that moved, as intended.

### Correction to my PR description

I wrote *"the MSVC job is independently red per #165 — ignore that one."* **That was wrong** — MSVC built and smoke-tested clean on this run.

#165 ("Windows MSVC build fails in DuckDB bundled fmt/format.h") is still **open**, and the comment in `ci.yml` still says *"Windows / Azure stay manual (cost, and #165 keeps MSVC red)"*. Neither matches what CI just did. This PR can't have fixed it — #165 is about DuckDB's bundled `fmt`, nothing to do with vcpkg deps — so the likely explanation is a DuckDB submodule bump since. Worth verifying and closing #165 / updating that comment separately; I'm not claiming it's fixed off one green run.

### Unrelated latent issue spotted in the log

```
warning: The 'x-gha' binary caching backend has been removed.
  on expression: clear;x-gha,readwrite
```

vcpkg has **removed** the `x-gha` binary-caching backend, so vcpkg binary caching is silently doing nothing on the Windows jobs (that's why this run paid the full 6.9 min for OpenSSL). Pre-existing, not caused by this PR, and it's what made the test honest here — but it's costing CI time generally. Also relevant to the `ACTIONS_CACHE_URL` / `ACTIONS_RUNTIME_TOKEN` export step in `codeql.yml`, which exists to feed that same removed backend. Separate cleanup.

### Verdict

Ready to merge from my side. This gets you onto the LTS line before 3.4 goes EOL on **2026-10-22**, with MinGW proven green.

VGSML (approved):

PR #198, opened 2026-07-17
ci: make vcpkg binary caching actually work (x-gha is gone)

oluies:
## TL;DR

**vcpkg binary caching has been doing nothing on every platform.** Two independent bugs, both silent — CI stayed green the whole time and rebuilt every dependency from source on every run.

## Bug 1: `x-gha` was removed from vcpkg 15 months ago

`codeql.yml` set it explicitly; the Windows jobs in `ci.yml`/`release.yml` get it implicitly from `run-vcpkg`.

vcpkg **removed** the provider in [vcpkg-tool#1662](https://github.com/microsoft/vcpkg-tool/pull/1662) (2025-04-29), after GitHub shut down the internal Actions Cache API it was consuming. From that PR:

> Recent changes in GitHub Actions Cache's API have rendered the `x-gha` binary caching provider useless. […] Their internal APIs were not intended to be consumed the way our program does.

The failure mode is nasty: `clear` drops the default `files` provider, and `x-gha` then resolves to nothing — so the net configuration is **zero binary sources**. vcpkg emits a warning and carries on:

```
%VCPKG_BINARY_SOURCES%: warning: The 'x-gha' binary caching backend has been removed.
  on expression: clear;x-gha,readwrite
```

## Bug 2: the `actions/cache` steps cached the wrong directories

They cached `vcpkg/installed`, `vcpkg/packages`, `vcpkg/buildtrees`. But in **manifest mode** vcpkg installs to `build/release/vcpkg_installed` (as `ci.yml` itself already knows — it references that path when locating simdutf) and looks for prebuilt archives in its *binary cache*. None of those three paths is ever consulted.

**This isn't a theory.** Runs [29339658347](https://github.com/hugr-lab/mssql-extension/actions/runs/29339658347) and [29345077972](https://github.com/hugr-lab/mssql-extension/actions/runs/29345077972) — consecutive, same branch, `vcpkg.json` unchanged, so a guaranteed cache hit — both logged:

```
Cache restored from key: vcpkg-x64-linux-7603ea2f...    <- HIT
Building openssl:x64-linux@3.4.1...                      <- rebuilt anyway
Elapsed time to handle openssl:x64-linux: 1.9 min
```

A green cache hit that rebuilt everything regardless. Same story on Windows, worse: MinGW spends **6.9 min** compiling OpenSSL every run.

## The fix

Use vcpkg's `files` provider pointed at `$GITHUB_WORKSPACE/vcpkg_bincache`, and let `actions/cache` persist *that* directory.

This is **migration option 2** from vcpkg-tool#1662. Option 1 (GitHub Packages via NuGet) is the vcpkg team's recommendation but needs a PAT and `packages:write` for no benefit at this scale. Option 2 is what the GitHub Actions team endorses, and pointing it at the *binary cache* rather than the install tree keeps per-port granularity — the downside vcpkg warns about ("the whole installed tree gets cached as a single artifact") doesn't apply.

For the Windows jobs, defining `VCPKG_BINARY_SOURCES` in the workflow is the **supported opt-out** from run-vcpkg's default. Straight from its bundle:

```js
// Allow users to define the vcpkg's binary source explicitly in the workflow,
// in that case do not override it.
this.baseUtils.setVariableIfUndefined(globals.VCPKG_BINARY_SOURCES, VcpkgRunner.VCPKG_BINARY_SOURCES_GHA);
```

### Why the prefix `restore-keys` are safe

vcpkg names archives by **ABI hash**. A stale or partial restore can never produce a wrong binary — vcpkg simply doesn't find a matching hash and rebuilds. So a loose restore-key is a pure win: it seeds a run whose `vcpkg.json` changed with whatever still matches.

## Scope

All **8** vcpkg-using jobs, across four workflows:

| Workflow | Jobs |
|---|---|
| `ci.yml` | build (linux+osx), build-windows-msvc, build-windows-mingw |
| `release.yml` | build (linux amd64/arm64 + osx), build-windows-msvc, build-windows-mingw |
| `codeql.yml` | analyze |
| `concurrency-tests.yml` | concurrency-tests |

`concurrency-tests.yml` had the same broken paths and would have been missed if I'd only grepped for `x-gha` — it's fixed too. `named-instance.yml` and `lint-extra.yml` don't use vcpkg.

Also removes `codeql.yml`'s `github-script` step that re-exported `ACTIONS_CACHE_URL` / `ACTIONS_RUNTIME_TOKEN`. It existed solely to feed x-gha and now feeds nothing.

## Interaction with #195

Worth flagging: with binary caching working, CodeQL's tracer won't see OpenSSL's translation units on a warm cache, so vendored OpenSSL alerts may vanish on their own — but **non-deterministically** (a cold cache compiles OpenSSL and they'd reappear). #195's SARIF filter is what makes that deterministic. The two are complementary, not redundant.

The `ccache` reasoning in `codeql.yml` ("ccache short-circuits compiler invocations and CodeQL's tracer would miss those translation units") does **not** transfer: vcpkg binary caching only elides *dependency* compilation, never the extension's own TUs, which are what the analysis is about.

## Verification

In progress — and this needs real verification precisely because the bug being fixed is one that looked green for 15 months. A cold run is populating the cache now ([29543742860](https://github.com/hugr-lab/mssql-extension/actions/runs/29543742860)); I'll then re-run and confirm the warm run **restores OpenSSL instead of rebuilding it**. Results to follow. Not merging on "the YAML parses".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,7 +239,22 @@ jobs:
         uses: actions/cache@v6
         with:
           path: vcpkg_bincache
-          key: vcpkg-bincache-${{ matrix.platform.vcpkg_triplet }}-${{ hashFiles('vcpkg.json') }}
+          # The primary key MUST be unique per run. actions/cache skips the
+          # post-run save on an exact primary-key hit (saveImpl.ts: "Cache hit
+          # occurred on the primary key ..., not saving cache."). With a stable
+          # key the cache would freeze after the first save: once the ABI hash
+          # drifts -- a runner-image compiler bump, or vcpkg master churn, since
+          # the clone above is unpinned -- vcpkg would rebuild every run and the
+          # fresh archives would never be persisted. That is the same silent
+          # no-op this whole change exists to kill.
+          #
+          # hashFiles('vcpkg.json') is kept only as a readable namespace marker;
+          # it is NOT sufficient on its own, because vcpkg.json's `version` is
+          # the extension's own version (rotates for the wrong reason) and it
+          # says nothing about compiler/triplet/vcpkg-tool (fails to rotate for
+          # the right one). The prefix restore-key below is what actually seeds
+          # each run, and it matches across hashes.
+          key: vcpkg-bincache-${{ matrix.platform.vcpkg_triplet }}-${{ hashFiles('vcpkg.json') }}-${{ github.run_id }}
           restore-keys: |
             vcpkg-bincache-${{ matrix.platform.vcpkg_triplet }}-
 
@@ -519,13 +534,17 @@ jobs:
       # Caches the binary-cache dir named by VCPKG_BINARY_SOURCES above, not
       # the install tree -- see the linux/macos build job for why the old
       # vcpkg/installed + vcpkg/packages paths never elided a rebuild.
+      # Namespace is the vcpkg triplet, deliberately NOT job- or workflow-scoped:
+      # release.yml's MSVC job uses the same triplet and shares this namespace,
+      # so a Windows build dispatched on main can seed the release build. See
+      # the linux/macos job for why the primary key carries github.run_id.
       - name: Cache vcpkg binary packages
         uses: actions/cache@v6
         with:
           path: vcpkg_bincache
-          key: vcpkg-bincache-windows-msvc-x64-windows-static-release-${{ hashFiles('vcpkg.json') }}
+          key: vcpkg-bincache-x64-windows-static-release-${{ hashFiles('vcpkg.json') }}-${{ github.run_id }}
           restore-keys: |
-            vcpkg-bincache-windows-msvc-x64-windows-static-release-
+            vcpkg-bincache-x64-windows-static-release-
 
       - name: Setup Visual Studio Environment
         uses: ilammy/msvc-dev-cmd@v1
@@ -627,9 +646,9 @@ jobs:
         uses: actions/cache@v6
         with:
           path: vcpkg_bincache
-          key: vcpkg-bincache-windows-mingw-x64-mingw-static-${{ hashFiles('vcpkg.json') }}
+          key: vcpkg-bincache-x64-mingw-static-${{ hashFiles('vcpkg.json') }}-${{ github.run_id }}
           restore-keys: |
-            vcpkg-bincache-windows-mingw-x64-mingw-static-
+            vcpkg-bincache-x64-mingw-static-
 
       - name: Build Release
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,13 @@ jobs:
             docker_available: false
             vcpkg_triplet: arm64-osx
 
+    env:
+      # Send vcpkg's binary cache to a directory actions/cache persists (see
+      # the "Cache vcpkg binary packages" step). Without this, vcpkg falls back
+      # to its default files provider under $HOME/.cache/vcpkg/archives, which
+      # nothing caches -- so every run rebuilt OpenSSL from source.
+      VCPKG_BINARY_SOURCES: 'clear;files,${{ github.workspace }}/vcpkg_bincache,readwrite'
+
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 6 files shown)


oluies:
## Verified: the cache now actually elides the rebuild ✅

Two runs on this branch, same `vcpkg.json`, all jobs green.

**Cold** ([29543742860](https://github.com/hugr-lab/mssql-extension/actions/runs/29543742860)) — populates:
```
Restored 0 package(s) from .../vcpkg_bincache in 23 us
Building openssl:x64-linux@3.4.1...
Elapsed time to handle openssl:x64-linux: 2 min
Cache saved with key: vcpkg-bincache-x64-linux-7603ea2f...
```

**Warm** ([29560310309](https://github.com/hugr-lab/mssql-extension/actions/runs/29560310309)) — reuses:
```
Cache restored from key: vcpkg-bincache-x64-linux-7603ea2f...
Restored 5 package(s) from .../vcpkg_bincache in 373 ms
Elapsed time to handle openssl:x64-linux: 4.84 ms
```

| | Cold | Warm |
|---|---|---|
| Packages restored | 0 | **5** |
| OpenSSL | **2 min** (compiled) | **4.84 ms** (restored) |
| `Building openssl` in log | 1 | **0** |

`2 min → 4.84 ms`, and zero `Building openssl` lines. Contrast with the old setup, where a green cache hit was *followed* by a full rebuild — which is the whole point.

---

## Review #365 found a real bug in this PR — fixed in 67a44c0

An automated review flagged four findings against 72c7a04. The headline one was **correct and important**: my primary cache key never rotated.

`actions/cache` skips the post-run save on an exact primary-key hit. From `src/saveImpl.ts`:
```js
if (utils.isExactKeyMatch(primaryKey, restoredKey)) {
    core.info(`Cache hit occurred on the primary key ${primaryKey}, not saving cache.`)
```

So with a stable key, the first ABI-hash drift (a runner-image compiler bump, or churn in the unpinned vcpkg master clone) makes vcpkg rebuild every run while the fresh archives are **never persisted** — no self-healing until `vcpkg.json` happens to change. That is the *same silent no-op this PR exists to kill*, reintroduced one layer down.

**The warm run above caught it live**, which is the nicest possible confirmation:
```
Cache hit occurred on the primary key vcpkg-bincache-x64-linux-7603ea2f..., not saving cache.
```

Fixed: primary keys now carry `${{ github.run_id }}`, with the existing prefix `restore-keys` seeding each run.

### Also fixed

- **MEDIUM — namespaces never written on the default branch.** `concurrency-tests.yml` runs only on `pull_request`/`workflow_dispatch`, `release.yml` only on tags. GitHub scopes cache reads to the current ref + default branch, and a cache written from `refs/pull/N/merge` is restorable only by re-runs of *that same PR* — so both were structurally cold forever. All namespaces are now the vcpkg triplet, and `release.yml`'s matrix gained `vcpkg_triplet` to match `ci.yml` exactly, so the weekly main cron and codeql's push-to-main can seed them.
- **LOW** — `vcpkg_bincache/` added to `.gitignore` (verified with `git check-ignore`).
- **LOW** — the `vcpkg-setup` comment justified its toolchain-file check with a `vcpkg/installed` cache restore this PR deletes. Reworded; guard kept (strictly more robust than a directory check).

The reviewer also correctly caught that my codeql.yml comment — *"vcpkg.json pins the dependency versions, so it is the right cache key"* — was wrong twice: `version` is the **extension's** version (rotates for the wrong reason), and it captures nothing about compiler/triplet/vcpkg-tool (fails to rotate for the right one). Rewritten.

### Two caveats I did not paper over

- `concurrency-tests` runs on `ubuntu-latest` while ci/codeql pin `ubuntu-22.04`, so seeded archives may still miss on ABI and rebuild. No worse than today, and still correct (archives are ABI-named). Aligning the runner images would make the seeding land — separate PR.
- GitHub's docs do **not** clearly state that a tag-triggered run can restore a default-branch cache (they only document that tags can't read *other tags'* caches). So I'm not claiming the release-side benefit as certain; the namespace unification is right regardless.

Both recorded as comments in the workflows.

**Verification of the fixes:** `yamllint 1.35.1` with the exact CI ruleset exits 0 across `.github/`; all 12 workflows parse; all 8 vcpkg jobs confirmed `run_id`-unique and namespace-aligned; the Makefile change is comment-only.

VGSML (approved):

Issue #191, opened 2026-07-16
COPY (FORMAT 'bcp') leaks its connection on client-side failure, locking the target table until the process exits


oluies:
### Summary

A `COPY ... (FORMAT 'bcp')` that fails **client-side validation** leaks its pooled connection. The
`INSERT BULK` transaction is never aborted and the connection is never returned to the pool, so the
target table stays **locked for the lifetime of the DuckDB process**. Every subsequent statement
touching that table — including `DROP TABLE` — blocks until the query timeout fires.

### Reproduction

Both statements in the **same** DuckDB process (a new process works fine, which is the tell — the
lock dies with the connection):

```sql
ATTACH 'Server=...;Database=TestDB;User Id=sa;Password=...' AS db (TYPE mssql);
SELECT mssql_exec('db','CREATE TABLE dbo.repro_fail (s VARCHAR(10) COLLATE SQL_Latin1_General_CP1_CI_AS NOT NULL)');

-- fails client-side, as it should: 11 chars into VARCHAR(10)
COPY (SELECT '12345678901' AS s) TO 'mssql://db/dbo/repro_fail' (FORMAT 'bcp', CREATE_TABLE false);

-- now this times out
SELECT mssql_exec('db','DROP TABLE IF EXISTS dbo.repro_fail');
```

Output:

```
Invalid Input Error: MSSQL: NVARCHAR column 's' overflow: value is 11 UCS-2 code units (22 UTF-16LE bytes) but column max is 10 code units (20 bytes)
Invalid Input Error: MSSQL execution error: SQL Server error 0: Query timeout
```

The first error is correct and expected. The second is the bug.

At teardown the pool confirms it:

```
[MSSQL POOL] LEAKED active conn id=1 spid=0 state=3 use_count=1 pool='db'
[MSSQL POOL] WARNING: ~ConnectionPool 'db' shut down with 1 connection(s) still checked out — DuckDB quiescence contract violated.
```

### Root cause

`MSSQLCopyGlobalState` (`src/include/copy/copy_function.hpp:69`) holds the connection:

```cpp
struct MSSQLCopyGlobalState : public GlobalFunctionData {
	std::shared_ptr<tds::TdsConnection> connection;
	...
};   // <-- no destructor
```

`BCPCopyInitGlobal` has a `release_connection_on_error` helper, but it only covers failures **during
init**. The validation that throws here (`ValidateNVarcharLength`, `src/codec/string_codec.cpp`) runs
[...]

Issue #177, opened 2026-07-03
COPY TO (bcp) throws NotImplementedException on HUGEINT (e.g. SUM() output)


oluies:
## Summary

`SUM()` (and other aggregates) over an integer column in DuckDB returns **`HUGEINT`** (int128). Writing that to SQL Server via `COPY ... TO ... (FORMAT 'bcp')` **fails at row-encode time** with:

```
Not implemented Error: MSSQL: BCP encoding for HUGEINT is not yet implemented (spec 045 Phase 6)
```

The table is created successfully first (see below), then the data write throws — so it's a partial/confusing failure.

## Reproduction

```sql
ATTACH '<dsn>' AS db (TYPE mssql);
CREATE TABLE t AS SELECT range AS x FROM range(1000);   -- BIGINT column

-- SUM(x) -> HUGEINT
COPY (SELECT SUM(x) AS total FROM t) TO 'db.dbo.sum_out' (FORMAT 'bcp');
-- => Not implemented Error: MSSQL: BCP encoding for HUGEINT is not yet implemented
```

Also reproduces via CTAS with BCP (the default): `CREATE TABLE db.dbo.sum_out AS SELECT SUM(x) FROM t;`

## Root cause

`src/codec/integer_codec.cpp` — `EncodeToBcp` (both the `Vector` and `Value` overloads) explicitly throws for `HUGEINT` / `UHUGEINT`:

```cpp
case LogicalTypeId::HUGEINT:
case LogicalTypeId::UHUGEINT:
    // Deferred to spec-045 Decimal family migration (Phase 6 / T069).
    throw NotImplementedException("MSSQL: BCP encoding for %s is not yet implemented (spec 045 Phase 6)", ...);
```
(Vector overload ~L151-155, Value overload ~L189-192.)

This is a **left-behind TODO** from the spec-045 codec migration. Notably, the *other* two HUGEINT paths in the same file are already implemented:

- `FormatDdlTypeName` maps `HUGEINT`/`UHUGEINT` → **`DECIMAL(38,0)`** (integer_codec.cpp L259-261) — so CTAS/COPY **creates** the table fine.
- `FormatSqlLiteral` serializes HUGEINT as `SerializeDecimal(HugeIntValue::Get(v), 38, 0)` (L228-229) — so the **INSERT** path already works.

Only the **BCP row-encode** path is missing, and it's exactly the path CTAS/COPY default to (`mssql_ctas_use_bcp=true`).

## Impact

[...]

PR #209, opened 2026-07-28
spec 054: materialization baseline + phase-0 quick wins (encode 1.6–2.8×, decode 1.5–2.6×, #177 fix)

VGSML:
## Summary

Phase 0 of the batch/SIMD materialization series (design: `docs/proposals/simd-chunk-materialization-design.md`, spec: `specs/054-simd-baseline-and-quick-wins/spec.md`). Baseline-first: all measurement infrastructure and recorded baselines land BEFORE the optimization commits, every conversion-touching change is gated by a byte-identity differential harness.

### Infrastructure
- `make bench-build` — release build with tpch injected via `EXTRA_EXTENSION_CONFIGS` (bench-only, never ships in `extension_config.cmake`)
- `make bench-materialize` — micro harness (`test/cpp/bench_materialize.cpp`): string-decode / fixed-decode / bcp-encode groups incl. dictionary/constant vector inputs
- `test/bench/bench_tpch_e2e.sh` — TPC-H e2e (write via BCP + scan back), release-comparison protocol via `MSSQL_BENCH_LOAD_SQL`/`MSSQL_BENCH_UNSIGNED`
- `test/bench/diff_check.sh` — 13-query byte-identity gate across two builds (every type family, NULL-heavy, PLP/MAX, embedded NUL, collations, 0/1/>2048 rows, BCP write-back)
- D4 debug counters at `MSSQL_DEBUG>=2`: per-stream / per-BCP-writer close summaries + thread-local UTF-16 fallback counter
- All baselines + deltas recorded append-only in `test/bench/bench_results_simd_baseline.md`

### Fixes
- **#177**: HUGEINT (e.g. `SUM()`) → BCP encodes as DECIMAL(38,0) with a client-side range guard, in BOTH the Integer and Decimal codec paths (COPY CREATE_TABLE reads target metadata back and dispatches via Decimal); COLMETADATA precision was 0/0
- CTAS mid-stream BCP failure no longer leaks a locked connection (#191 pattern)
- Pool teardown: cleanup thread now waits on a CV instead of a blind 1 s sleep — CLI round-trip 2.06 s → 0.03 s
- Latent: uuid/binary codecs read via `FlatVector::GetData` — wrong for dictionary/constant inputs; now format-based

### Quick wins (each gated by diff_check byte-identity + before/after micro)
- **W1+W2** (`EncodeChunk`): UnifiedVectorFormat + family encoder + NULL wire kind hoisted once per column per chunk (was 2× ToUnifiedFormat + 9-way switch per CELL). bigint 18.3→8.2 ns/value; dictionary penalty eliminated (2.8×)
- **W3+W4**: single-validation NVARCHAR encode, exact-size append, per-chunk accumulator reserve; unaligned destinations stay on simdutf (was: ~half of string values silently on the scalar legacy path)
- **R1**: string decode straight into the vector slot (`EmptyString` + valid-convert) — len32 2.6×, non-ASCII ~2×

### Gates
- Acceptance criterion 6 (no e2e step regresses >3%): **PASS** — every cross-session flag refuted by same-session interleaved A/B
- Pre-merge release comparison at SF 2 (lineitem 11,997,996 rows), 3 interleaved pairs vs the v0.2.2 release artifact: **every step improved** — copy_part −24.8%, copy_customer −32.7%, copy_lineitem −7.8%, Q1 −11.6% (short steps additionally reflect the pool fix: 2 s floor → instant)

Open item: Linux x86_64 micro baseline run (needs a real x86_64 host — QEMU-on-ARM timing is meaningless).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -347,6 +347,27 @@ jobs:
           LD_LIBRARY_PATH=build/release/src DYLD_LIBRARY_PATH=build/release/src ./build/test/test_datetime_codec
           echo "::endgroup::"
 
+          # spec 054: run the remaining per-family codec unit tests. Only
+          # datetime ran in CI before, so the #177 HUGEINT guard and the W1
+          # dictionary/constant fixes had no CI coverage. Same link set as
+          # `make test-codec-<fam>`, plus bcp_row_encoder.cpp — integer/decimal
+          # EncodeToBcp call BCPRowEncoder::EncodeDecimal, defined there.
+          for fam in boolean integer float decimal string binary uuid money; do
+            echo "::group::test_${fam}_codec"
+            "${CXX:-c++}" -std=c++17 -pthread -Wno-deprecated-declarations \
+              -I src/include -I duckdb/src/include -I "$PREFIX/include" \
+              "test/cpp/codec/test_${fam}_codec.cpp" \
+              src/codec/*.cpp \
+              src/tds/encoding/utf16.cpp src/tds/encoding/datetime_encoding.cpp \
+              src/tds/encoding/decimal_encoding.cpp src/tds/encoding/guid_encoding.cpp \
+              src/tds/encoding/bcp_row_encoder.cpp \
+              -L "$PREFIX/lib" -lsimdutf \
+              -L build/release/src -lduckdb \
+              -o "build/test/test_${fam}_codec"
+            LD_LIBRARY_PATH=build/release/src DYLD_LIBRARY_PATH=build/release/src "./build/test/test_${fam}_codec"
+            echo "::endgroup::"
+          done
+
       # Spec 053 (#161): the shipped Linux extension must NOT hard-link the
       # Kerberos runtime, or it fails to LOAD on images without it. GSSAPI/krb5
       # are dlopen'd lazily, so they must be absent from DT_NEEDED.
--- a/DATAMODEL.md
+++ b/DATAMODEL.md
@@ -122,6 +122,7 @@ classDiagram
         -cleanup_thread_ : thread
         -pool_mutex_ : mutex
         -available_cv_ : condition_variable
+        -cleanup_cv_ : condition_variable
     }
     class PoolConfiguration {
         +connection_limit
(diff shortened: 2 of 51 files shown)


oluies:
Reviewed the correctness-critical paths (codec encode/decode, UTF-16 primitives, `EncodeChunk`, pool teardown, the #177 fix). This is careful, well-commented work. A few notes, and I pushed some test follow-ups directly to the branch; see the end.

## What holds up

- `EncodeChunk` moving the `0xD1` ROW-token ownership is consistent: `BuildRowToken` is deleted and `WriteRows` delegates, so no double or missing token.
- The W3 buffer-sizing invariant is sound: `Utf16LEByteLengthView` and `Utf16LEEncodeValidDirect` are the paired simdutf `utf16_length_from_utf8` / `convert_valid_utf8_to_utf16le`, and UTF-16 length in code units never exceeds UTF-8 byte length, so the pre-sized buffer and the unaligned scratch are always large enough.
- The pool CV fix is correct: `shutdown_flag_` is set under `pool_mutex_` before `notify_all`, closing the missed-wakeup race, and `wait_for` releases the mutex during the wait so Acquire/Release are not blocked.
- The #177 metadata (`precision=38, scale=0, max_length=17`) matches `EncodeDecimal`, and the CTAS mid-stream leak fix wraps `AddChunkBCP` so a guard throw releases the connection.

## One correctness edge: HUGEINT min slips the range guard

`CheckHugeintFitsDecimal38` (and `CheckMantissaFitsPrecision`) computed `abs = value.upper < 0 ? Hugeint::Negate(value) : value`. `Hugeint::Negate` on HUGEINT min (`-2^127`) is not representable in int128, so that one 39-digit value either threw a raw overflow instead of the column-named error, or wrapped back to a negative and passed the guard, sending a value that aborts the BCP stream mid-batch (the exact thing #177 set out to prevent). The existing guard test covers `int128_max` but not `int128_min`.

Fix pushed: bound both ends directly (`value > max || value < -max`); `±(10^38−1)` are both representable, so no `Negate` on the input and the message is preserved for every value.

## Test coverage

The #177 coverage is thorough. Two gaps, both now filled on the branch:

1. Dictionary/constant vectors. The W1 format-based access and the uuid/binary `FlatVector::GetData`-to-`UnifiedVectorFormat` fix exist specifically because table functions may emit CONSTANT/DICTIONARY vectors, but the unit tests only built flat vectors, and the only dictionary/constant assertion lived in the micro-bench (not a CI gate). Added `TestEncodeToBcpDictionaryConstantVectors` (integer) and `TestEncodeConstantDictionaryVectors` (uuid). Both encode a CONSTANT and a reversing DICTIONARY vector and assert byte-equality with the flat encode.
2. HUGEINT min: added `TestEncodeToBcpHugeintMinEdge` alongside the guard fix.

While wiring those I found two infrastructure issues and fixed them:

- `make test-codec-integer` / `test-codec-decimal` could not link: `BCPRowEncoder::EncodeDecimal` lives in `bcp_row_encoder.cpp`, which was missing from `CODEC_TEST_ENCODING_SOURCES`. Added it.
- CI only compiled `test_datetime_codec`. The other nine codec suites, including the new #177 tests, never ran in CI. The build job now runs every per-family codec test.

## CI note

The `Build` job is `workflow_dispatch` / `schedule` gated, so it skips on the PR and the codec tests (old and new) do not run on a normal PR push. I dispatched a build on the branch to exercise the newly wired tests; happy to leave the gating as-is or discuss a `paths`-triggered lane for `src/codec/**` separately.

Minor: literal `9` for the family count in the `bcp_writer.cpp` / `mssql_result_stream.cpp` debug loops where `TYPE_FAMILY_COUNT` exists.

All of the above is on `spec/054-simd-baseline-quick-wins` as one commit; revert it if you would rather take the tests differently.

PR #211, opened 2026-07-29
fix(scan): serialize catalog-scan bind data so subplans stay distinct

VGSML:
## The bug

Two scans of **different** tables in one query could silently return the same table's rows:

```sql
SELECT 'A' AS t, min(c) AS v FROM db.dbo.T1
UNION ALL
SELECT 'B' AS t, min(c) AS v FROM db.dbo.T2;
-- both rows carried T1's data
```

No error, no warning — just wrong data. Reproduced on the released **v0.2.2** artifact as well as on `main`.

## Root cause

DuckDB's common-subplan optimizer (`src/optimizer/common_subplan_optimizer.cpp`, added in v1.5.4) finds equal subplans and materializes them once as a CTE. It identifies them by **serializing** each operator and hashing the bytes.

For a `LOGICAL_GET` over a table function, `FunctionSerializer::Serialize` writes the bind data only when the function has **both** `serialize` and `deserialize` callbacks. `mssql_catalog_scan` had neither, so the signature contained only the function name, signature, returned types, column names and filters — and `logical_get.cpp:248-255` then falls back to the parameter list, which is empty for a catalog scan.

Schema and table names live **only** in `MSSQLCatalogScanBindData`, so two same-shaped `mssql_catalog_scan` operators hashed identically. One was materialized as a CTE and both consumers read it.

`MSSQLCatalogScanBindData::Equals` was already correct, but is never consulted here — this optimizer compares serialized bytes, not `FunctionData::Equals`.

## Scope

| | affected |
|---|---|
| catalog scan (`db.schema.table`) | **yes** |
| `mssql_scan('db', '...')` | no — the query string is a table-function *parameter*, which IS serialized on the no-callback path |
| mssql 0.2.1, 0.2.2 (DuckDB 1.5.4 / 1.5.5) | **yes** |
| mssql 0.1.18, 0.2.0 (DuckDB 1.5.2 / 1.5.3) | no — predate the optimizer |

Triggering needs same-shaped scans (identical projected column names and types) in one query: `UNION ALL`, scalar subqueries, and n-way unions all hit it. Tables with different column names or types produce different signatures and were never affected.

## The fix

Add `serialize` / `deserialize` to `mssql_catalog_scan`, modelled on DuckDB's own `TableScanSerialize` (`src/function/table/table_scan.cpp:851`).

Serialized: `context_name`, `schema_name`, `table_name`, plus the pushdown state we build ourselves — `complex_filter_where_clause`, `order_by_clause`, `top_n`. That last part matters independently: two scans of the *same* table with different pushed-down predicates must not collapse either. Everything else in the bind data (column lists, PK/rowid metadata) is derived from the table and is therefore implied by schema+table.

`deserialize` rebuilds through the catalog entry, so column and PK metadata come from the live catalog rather than from serialized copies.

**Legitimate deduplication is preserved** — verified on the plan: two scans of the same table with the same filters still collapse into one `MSSQL_CATALOG_SCAN` with two `CTE_SCAN` consumers, which is a real win for a remote source. Different tables now produce two scan nodes and no CTE.

## Tests

New `test/sql/query/multi_table_subplan.test` (28 assertions) covering `UNION ALL`, scalar subqueries, a three-way union, plain `UNION ALL` without aggregation, and the same table twice with different pushed-down filters.

Full local suite against SQL Server 2022 in Docker: **250 test cases, 7352 assertions, all passing** (46 skipped for missing Azure/Kerberos/TLS env).

## How it was found

While building a live-server read/write benchmark, a verification query that compared several fixture tables in one `UNION ALL` reported identical contents for tables that demonstrably held different data.

### DIFF
--- a/src/table_scan/table_scan.cpp
+++ b/src/table_scan/table_scan.cpp
@@ -6,8 +6,13 @@
 #include <algorithm>
 #include <cctype>
 #include <cstdlib>
+#include "catalog/mssql_table_entry.hpp"
 #include "connection/mssql_settings.hpp"
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/table_catalog_entry.hpp"
 #include "duckdb/common/common.hpp"		   // For COLUMN_IDENTIFIER_ROW_ID
+#include "duckdb/common/serializer/deserializer.hpp"
+#include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/common/table_column.hpp"  // For TableColumn, virtual_column_map_t
 #include "duckdb/planner/operator/logical_get.hpp"
 #include "mssql_compat.hpp"		// For FlatVector header relocation (spec 051 M2)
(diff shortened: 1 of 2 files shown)


oluies:
Reviewed the fix and the analysis. This is a correct, well-scoped fix for a nasty silent-wrong-results bug, and the writeup of the root cause is accurate.

## Correctness

- The diagnosis holds: `HasSerializationCallbacks()` requires both callbacks before `FunctionSerializer` writes bind data, which is why both are needed even though only `serialize` runs today.
- `top_n` is `int64_t` in the struct and `deserialize` reads `ReadProperty<int64_t>`, so no width or tag mismatch. Field IDs 100 through 105 line up across both halves.
- The signature is complete. Walking the struct, the only identity-bearing fields that can vary independently at COMMON_SUBPLAN time are `context/schema/table` and `complex_filter_where_clause`, and all are serialized. `order_by_clause` and `top_n` are set later by the EXTENSION optimizer (still empty at signature time), so serializing them is for round-trip fidelity, not distinctness. Everything else (`return_types`, `column_names`, `all_*`, PK/rowid metadata) is table-derived or projection-implied, so it is already covered by schema+table or by the LogicalGet signature. `rowid_requested` changes the projected types, which changes the LogicalGet hash on its own.
- The `!= "mssql"` guard uses the right literal and matches the existing pattern in `mssql_refresh_function.cpp`, `mssql_preload_catalog.cpp`, and `target_resolver.cpp`. Rebuilding through `entry.GetScanFunction` instead of trusting serialized column/PK copies is the correct choice.

## Minor notes

- `deserialize` is unexercised today (as the comment says), so the guard, the rebind, and the field restore have no runtime coverage. The code reads correctly, but if a future DuckDB feature ever round-trips a plan, this path runs for the first time untested.
- Inside `deserialize`, `Catalog::GetCatalog` / `GetEntry` throw `CatalogException` when the alias is gone, which is inconsistent with the `SerializationException` wrapping right next to them. Only matters if the path is ever exercised; a catalog-existence check before the type guard would make the failure modes uniform.

## Test coverage

The regression test is solid and covers the scenarios by data (different tables must return different rows), including the same-table-different-complex-filter case that `complex_filter_where_clause` protects.

Two things worth knowing about what actually runs:

1. The `.test` needs `MSSQL_TEST_DSN`, so it is an integration test. It does not run on a normal PR, and the integration job currently selects zero `.test` files, so this regression is guarded only by a local run today. That is a pre-existing infrastructure gap, not this PR's problem.
2. A standalone C++ unit test of the serialize distinctness is not practical: `CatalogScanSerialize` is file-static in `table_scan.cpp`, whose translation unit pulls in the query executor, catalog, and pool, so it cannot be linked in isolation the way the codec unit tests are.

What would add real value on top of the data test is a plan-shape assertion. The PR's central claim (different tables no longer collapse, same table still does) is currently verified by hand, not by the test, and DuckDB's sqllogictest can assert it:

```
# different tables: no CTE materialization
query II
EXPLAIN SELECT min(c) FROM subplan_db.dbo.SubplanA
UNION ALL
SELECT min(c) FROM subplan_db.dbo.SubplanB;
----
physical_plan	<!CTE_SCAN>

# same table twice: dedup preserved
query II
EXPLAIN SELECT min(c) FROM subplan_db.dbo.SubplanA
UNION ALL
SELECT min(c) FROM subplan_db.dbo.SubplanA;
----
physical_plan	<REGEX>:.*CTE_SCAN.*
```

I left this out rather than commit it, because I could not run the suite to confirm the exact operator label and match directives against your local SQL Server. If you add it, it turns "verified on the plan" into an assertion.

## Build

The change touches `table_scan.cpp`, and the `Build` job is dispatch-gated, so the PR checks do not actually compile it. I dispatched a build on the branch to confirm it compiles on Linux and macOS and will follow up with the result.

Verdict: correct, well-reasoned, right-sized. Good to merge once the compile is confirmed.

oluies:
Compile confirmed: dispatched a build on this branch ([run 30445572058](https://github.com/hugr-lab/mssql-extension/actions/runs/30445572058)) — `Build` and `Smoke Test` both green on linux_amd64 and osx_arm64. (The `Build` job is `workflow_dispatch`/`schedule` gated, so it doesn't run on the PR itself.)

oluies:
@VGSML — the two CI-coverage gaps I mentioned in the review (this PR's `table_scan.cpp` change isn't compiled by the PR checks, and the `.test` regression only runs in the integration suite, which currently selects zero files) are now tracked in #212. That issue has the concrete fixes: a `paths`-triggered compile lane on `pull_request`, and correcting the `test/sql/*` selection plus a hard guard so a zero-test run fails instead of reporting PASSED. Nothing to do on this PR — just so the plan-shape test I suggested has somewhere to actually run once #212 lands.

Issue #197, opened 2026-07-16
NTEXT and IMAGE columns cannot be read through the catalog (no CAST, no decoder for wire types 0x63 / 0x22)


oluies:
### Summary

`NTEXT` and `IMAGE` columns cannot be read through the catalog at all. The scan emits no CAST for them, so SQL Server sends their native LOB wire types, which have no decoder — the query fails instead of returning data.

### Reproduction

```sql
-- SQL Server
CREATE TABLE dbo.nt_probe (nt NTEXT NULL, im IMAGE NULL);
INSERT INTO dbo.nt_probe VALUES (N'hello ntext', 0x0102);
```
```sql
-- DuckDB
ATTACH '...' AS db (TYPE mssql);
SELECT nt FROM db.dbo.nt_probe;   -- IO Error: Connection closed while waiting for COLMETADATA
SELECT im FROM db.dbo.nt_probe;   -- IO Error: Connection closed while waiting for COLMETADATA
```

Verified against SQL Server 2022 with the extension built from `main` @ 2c1c4cf.

### Root cause

Both are in `IsKnownSQLServerType` (`src/catalog/mssql_column_info.cpp`), so `is_cast_required` is **false** and neither picks up the `is_cast_required` → `CAST(... AS NVARCHAR(MAX))` rewrite that unsupported types get. Then, in `BuildColumnExpression` → `NeedsNVarcharConversion` (`src/table_scan/table_scan.cpp`):

- **NTEXT** short-circuits on the `col.is_unicode` early return (`IsUnicodeType` lists `ntext`), so no CAST.
- **IMAGE** is not a text type (`IsTextType` doesn't list it), so no CAST.

With no CAST, the native wire type reaches the codec and nothing handles it: `TDS_TYPE_NTEXT` (0x63) and `TDS_TYPE_IMAGE` (0x22) have no case in `codec::FamilyFromTdsType` (`src/codec/type_family.cpp`), and `TypeConverter::GetLogicalType` throws for them too.

`TEXT` is the odd one out only because it *is* a text type and non-unicode, so it goes through the NVARCHAR CAST path — which is why #190 could fix it there.

### Suggested fix

Give the legacy LOBs an explicit CAST in `BuildColumnExpression`, before the `is_unicode` / `IsTextType` checks that currently skip them:

- `NTEXT` → `CAST([c] AS NVARCHAR(MAX))`
[...]

Issue #89, opened 2026-03-31
INTERNAL Error: Expected vector of type INT128, but found vector of type VARCHAR


wahmedswl:
Hi,

Getting following error when trying to run queries in DuckDB via this extension.

```
SQL Error: INTERNAL Error: Expected vector of type INT128, but found vector of type VARCHAR
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
```

It's just simple select statement 

```
select * from ms.dbo.vm_SaviCardPointTransaction;
```

Following is the schema retrieved via DuckDB catalog 

```
vm_SaviCardPointTransaction	Id	BIGINT
vm_SaviCardPointTransaction	Transaction #	VARCHAR
vm_SaviCardPointTransaction	Location	VARCHAR
vm_SaviCardPointTransaction	MID	VARCHAR
vm_SaviCardPointTransaction	Serial #	VARCHAR
vm_SaviCardPointTransaction	Date	TIMESTAMP
vm_SaviCardPointTransaction	Method	VARCHAR
vm_SaviCardPointTransaction	Name	VARCHAR
vm_SaviCardPointTransaction	Company	VARCHAR
vm_SaviCardPointTransaction	Brand	VARCHAR
vm_SaviCardPointTransaction	Last 4	VARCHAR
vm_SaviCardPointTransaction	Amount	DECIMAL(19,4)
vm_SaviCardPointTransaction	Auth Code	VARCHAR
vm_SaviCardPointTransaction	Status	VARCHAR
vm_SaviCardPointTransaction	Parent Transaction #	VARCHAR
vm_SaviCardPointTransaction	User	VARCHAR
vm_SaviCardPointTransaction	Plan ID	VARCHAR
vm_SaviCardPointTransaction	Currency	VARCHAR
vm_SaviCardPointTransaction	Tip	VARCHAR
vm_SaviCardPointTransaction	Plan Name	VARCHAR
vm_SaviCardPointTransaction	Order ID	VARCHAR
vm_SaviCardPointTransaction	Invoice #	VARCHAR
vm_SaviCardPointTransaction	Purchase #	VARCHAR
vm_SaviCardPointTransaction	Tax Amount	DECIMAL(19,4)
vm_SaviCardPointTransaction	Refund Reason	VARCHAR
vm_SaviCardPointTransaction	Front End	VARCHAR
[...]

Issue #212, opened 2026-07-29
CI does not compile PRs or run the integration suite (build job dispatch-gated; test/sql/* selects zero files)


oluies:
## Summary

Two CI-coverage gaps let C++ and SQL regressions reach `main` behind green checks. Neither is a product bug; both keep re-surfacing during review (three times in the last week: #206, #209, #211), so they are worth tracking and fixing once.

1. A pull request never compiles the extension. The `Build` job is manual-dispatch only, so a C++ change can merge without CI ever building it.
2. The integration suite reports success while running zero `.test` files. The runner's glob matches nothing and the script treats that as a pass. This was #192, closed by #200 without the selection actually being fixed.

## Gap 1: PRs do not compile the extension

`.github/workflows/ci.yml`, the `Build` job:

```yaml
if: ${{ github.event_name == 'workflow_dispatch' && inputs.run_build }}
```

`Integration Test`, `Smoke Test`, and the Windows/Azure jobs are gated the same way (dispatch or the weekly `schedule`). On a normal `pull_request` push, all of them skip. The only C++ that compiles at PR time is the standalone unit-test job, which builds a handful of libduckdb-free TDS files (`test_login7_encoding`, `test_gssapi_runtime`, and similar). It never touches `table_scan.cpp`, the codec sources, or anything else in the extension proper.

So a change to the extension can go green and merge without CI compiling it once. To confirm that #206, #209, and #211 even built, each needed a manual `workflow_dispatch` run. #165 (Windows MSVC) is the same blind spot from the other direction: it stays red where nobody looks.

Proposed fix: run the release build plus smoke on `pull_request` when the relevant paths change, keeping the heavy matrix manual.

```yaml
on:
  pull_request:
    paths:
      - 'src/**'
      - 'CMakeLists.txt'
      - 'vcpkg.json'
      - '.github/workflows/**'
```

Scope it to `linux_amd64` (and `osx_arm64` if the budget allows) and include the codec unit tests. The cost is one build per C++ PR, which is the manual step made automatic. Windows, Azure, and the full integration matrix stay dispatch-gated.

## Gap 2: the integration suite runs zero `.test` files

`scripts/ci/integration_test.sh`:

```bash
if "$UNITTEST_BIN" "test/sql/*"; then
    echo "=== Integration Test PASSED ==="
    exit 0
```

[...]

Issue #185, opened 2026-07-14
fuzz_tds_tokens: per-unit >35s timeout in TokenParser (unbounded recursion + O(n²) compaction)


oluies:
### Summary

`fuzz_tds_tokens` (the post-login TDS token-stream harness) trips libFuzzer's **per-unit timeout (`timeout after 35 seconds`)** — a single input drives `TokenParser::TryParseNext()` into a pathologically slow / deeply recursive path. This is a **hang / DoS-class** signal, not a memory-safety crash ("finished running without reportable crashes" but ClusterFuzzLite fails the run on the timeout).

Pre-existing and **not a regression**: `main`'s scheduled ClusterFuzzLite is green (2026-07-06, 2026-07-13). It surfaced under PR-mode fuzzing on #184, which does not touch `tds_token_parser.cpp`. It is **nondeterministic** — PR-mode generated a slow input on two runs and not on a third — and **no reproducer file was saved**, so it is not (yet) a corpus regression.

### Evidence

From ClusterFuzzLite PR runs [29334386184](https://github.com/hugr-lab/mssql-extension/actions/runs/29334386184) and [29339634826](https://github.com/hugr-lab/mssql-extension/actions/runs/29339634826):

```
==52== ERROR: libFuzzer: timeout after 35 seconds
    #4 ... duckdb::tds::TokenParser::ConsumeBytes(unsigned long)  src/tds/tds_token_parser.cpp:56
    #5 ... duckdb::tds::TokenParser::TryParseNext()               src/tds/tds_token_parser.cpp:184
    #6 ... LLVMFuzzerTestOneInput                                 fuzz/fuzz_tds_tokens.cc:33
```

### Likely root cause (two compounding issues)

**1. Unbounded recursion in `TryParseNext()`.** The FEDAUTHINFO branch consumes and then recurses instead of looping:

```cpp
// src/tds/tds_token_parser.cpp ~184-186
ConsumeBytes(5 + token_length);
...
return TryParseNext();   // <-- recursion per token
```

A stream that is a long run of minimal FEDAUTHINFO tokens (`EE` + 4-byte length `0` = 5 bytes each) recurses once per token — for a ~MB input that is hundreds of thousands of frames deep.

**2. O(n²) buffer compaction.** Each consume calls `CompactBuffer()`, which `erase()`s from the front once `buffer_pos_` passes the halfway mark:

```cpp
// src/tds/tds_token_parser.cpp ~56-67
[...]

PR #213, opened 2026-07-30
spec 055: column-major read staging + batch decode (read −14…−47%, wide row −23%; #197)

VGSML:
Spec 055 — the read path is column-major and batch-decoded. Every per-value
conversion is gone from `scan`; what is left is one batch call per column per
chunk.

## What changed

The shipped path decomposed each ROW token into a `vector<vector<uint8_t>>`,
one heap buffer per column, then walked that a second time dispatching on the
TDS type for every cell. Two passes over the same bytes and a type switch in
each.

Now the row is walked once. Per-column dispatch resolves after COLMETADATA, so
the walk carries a single switch over a column-invariant arm; a type whose wire
form is already DuckDB's lands straight in the output vector with no
intermediate buffer at all, and everything else stages column-major and
finalizes in one call per column.

Landed in order: staged row walk (T5) · one UTF-16 conversion per string column
(D5/T10) · three copies out of the socket read path · VARBINARY · UUID ·
temporal · DECIMAL/MONEY · NCHAR and single-byte CHAR/VARCHAR · receive-buffer
resize removal · PLP/MAX · the issue-#89 divergence (which removed the last
per-value arm) · TEXT/NTEXT/IMAGE (#197) · one placement rule for batch
finalizers · per-column delimiter search · D10 counters · staging addresses
bound once.

## Results

Three builds interleaved in ONE session with the pass order rotated, medians of
3, 500k rows, threads=1. Client CPU ns/value, tip vs the last commit on the
per-value path (`1f63bad`) / vs the branch's starting point:

| family | 1f63bad | tip | delta | vs pre-055 |
| bigint | 33.2 | **23.4** | −29.5% | −76.0% |
| int | 40.8 | **21.6** | −47.1% | −75.1% |
| double | 35.8 | **25.6** | −28.5% | −74.9% |
| date | 27.4 | **21.6** | −21.2% | −74.4% |
| dec38 | 39.0 | **33.6** | −13.8% | −75.2% |
| uuid | 45.8 | **35.2** | −23.1% | −69.8% |
| blob | 56.0 | **41.2** | −26.4% | −66.4% |
| str16 | 94.8 | **67.6** | −28.7% | −55.8% |
| str16u (non-ASCII) | 89.0 | **53.4** | −40.0% | −62.4% |
| str16max (PLP) | 118.0 | **84.4** | −28.5% | −56.4% |
| strnull (50% NULL) | 56.4 | **37.2** | −34.0% | −69.4% |
| **wide, 15 cols** | 74.3 | **57.2** | **−23.0%** | **−36.8%** |

The three sides do not overlap on any pass of any load-bearing step, and the
local-table controls are flat to within 0.2 ns, so the machine held for the run.
The write path shows no signal and none is expected — this PR does not touch it;
its per-pass spread is ±50%, so no write delta is claimed. `str200` gains least
(−14.5%) and should: 400 wire bytes per value is bandwidth.

Full table and caveats in `test/bench/bench_results_live_server.md`.

## Behaviour changes

- **Invalid UTF-16 now gets standard U+FFFD replacement.** The legacy decoder
  swallowed the code unit after an unpaired high surrogate and dropped a
  trailing one — proven end-to-end against a live server. Needs a release note.
- **TEXT / NTEXT / IMAGE are readable** (#197), through the catalog *and* raw
  `mssql_scan`: the server-side CAST the spec asked for, plus native decode of
  the legacy-LOB wire form, since raw scan has no metadata to cast from.
- **New setting `mssql_tds_packet_size`** (default 4096 → 16384). The server
  caps the grant at its own maximum, so this is a ceiling, not a demand.

## Gates

- `diff_check.sh` **13/13 byte-identical** at every conversion-touching commit,
  with the invalid-UTF-16 change as the one reviewed, intended diff
- full suite **254 cases / 7528 assertions** on the rebased tree
- `clang-format-14 --dry-run --Werror` clean over `src`
- new tests: `legacy_lob_scan.test`, `unpaired_surrogates.test`,
  `test_column_staging.cpp`

Rebased onto `main` after #211; no conflicts, gates re-run on the rebased tree.
Full CI (incl. Windows) has NOT been dispatched yet — the tree touches
mangling-sensitive codec headers, so MSVC/MinGW should run before merge.

## Deferred, deliberately

`specs/058-read-framing/spec.md` is a design draft for what the D10 counters
now point at: the batch kernels are 3–5 ns/value against 21–34 ns/value for the
whole path, so the decode is no longer the read path's cost centre — the
framing is. It carries one measurement (a build calling `SkipRow` twice and
discarding the result puts the second walk at 2.2–3.1 ns/value, 4–13%) and
three corrections worth reading, chief among them that **exceptions are not the
cost**: zero-cost EH means a `throw` on a cold branch costs nothing, and what
actually cost here were out-of-line CALLS that happen to throw.


oluies (dismissed):
Reviewed at `17ab611`, diffed against `origin/main` @ `1306683`. The architecture is sound and unusually well documented — the invariants that matter (positional NULL slots, delimiter-based boundary recovery, column-invariant dispatch, the "only the lower bound may be used to skip" rule) are all written down where the next reader will find them. This review is scoped to two defects and the test coverage around the new hot path.

**I have not built or run anything.** Everything below is from reading the diff.

---

## 1. NBC row + `0xFFFF` length prefix on a binary/char column → out-of-bounds read

`src/codec/staging/row_stager.cpp:519` (`StageNBCRow`, `P2StageBinary`) reads the 2-byte length prefix and unconditionally does `AppendVar(p + 2, length)`. The non-NBC arm at `row_stager.cpp:411` checks `length == 0xFFFF` first; the NBC arm does not.

The bounds proof the staged walk depends on comes from `RowReader::SkipNBCRow`, and `src/tds/tds_row_reader.cpp:858` returns **2** for a `0xFFFF` prefix ("Should not happen in NBC for non-NULL"). So a row whose bitmap bit is clear but whose prefix is `0xFFFF` is admitted as a 2-byte value, and the stager then `memcpy`s 65535 bytes out of `receive_buffer_` — past the row, and past the allocation for any row near the buffer tail. `p` also advances past `end`, so every remaining column in that row parses garbage, and `D_ASSERT(p == end)` is compiled out in release.

Affected arms: `varbinary(n)`, `binary(n)`, `char(n)`, `varchar(n)` — every non-PLP form.

`P2StageString` survives only by accident: `0xFFFF & 1` is odd, so `ThrowOddUtf16Length` fires first. That's luck, not design — worth an explicit check rather than leaving it resting on the parity test.

Not producible by a conforming SQL Server, but it is exactly the malicious/corrupt-stream model this repo already fuzzes and has already shipped fixes for (#162, #184, #185).

## 2. `FinalizeChunk`'s validity memcpy discards NULLs the kernel set

`row_stager.cpp:598-606` publishes validity by overwriting the whole mask from `st.validity_words` whenever `null_count > 0`. But `src/codec/datetime_codec.cpp:409` sets rows NULL *inside* the kernel for issue #168 (a datetime2 beyond the target variant's int64 range).

So for a datetime2 column holding both a staged NULL and an out-of-range value, the memcpy clears the kernel's `SetNull` and the row publishes as valid — with `result[row]` never written, i.e. a stale slot from the previous chunk.

Reachability is narrow: `Datetime2WireToNativeUnit` only returns false for a `TIMESTAMP_NS` target, and `IsSupportedRetarget` (`src/codec/staging/column_ops.cpp:189`) is the only thing that admits that target. When `null_count == 0` the code `continue`s and the `SetNull` survives, so it is the mixed case only.

Either way the two contracts contradict each other, and the next kernel that wants to emit NULL will break silently. `mask.GetData()[w] &= st.validity_words[w]`, or publishing validity before the kernel runs, resolves it.

---

## Tests I'd add

**A differential framing test — the highest-value one.** The staged walk's memory safety rests entirely on `RowStager`'s framing consuming byte-for-byte what `RowReader::SkipValue` / `SkipValueNBC` consumed. Those are two independent switch statements over the same wire types in different files, and nothing pins them together. A table-driven test over every TDS type × {bare, P1, P2, PLP, LOB} × {ROW, NBCROW} — each case followed by a sentinel column whose decoded value proves the preceding column consumed exactly the right number of bytes — catches finding 1 and every future drift between the two. Built under ASan it also catches the over-read directly.

**Fuzz the raw-row path.** `fuzz/fuzz_tds_tokens.cc` never calls `SetRawRowMode(true)` and never constructs a `RowStager`, so the new parser — the one that has no per-value bounds checks *by design* — is entirely unfuzzed, while the path it replaces is fuzzed weekly. A `fuzz_row_stager` harness needs libduckdb linked, unlike the existing `src/tds`-only harnesses, but ASan finds finding 1 immediately.

**Gaps in `test/cpp/codec/test_column_staging.cpp`.** It covers staging mechanics well, but nothing downstream of them:

- `SplitWithEmbeddedNuls` — a value containing its own U+0000. No coverage, and it is the subtlest function in the PR.
- The mixed-script trap the comment at `src/codec/string_codec.cpp:316-321` names explicitly: a value whose UTF-8 length falls strictly between `u` and `2u`. Every bench fixture is single-script, so this passes the whole matrix.
- Both `FindDelimiter` instantiations — the `MEMCHR_THRESHOLD_BYTES` decision at `string_codec.cpp:406` picks one per column, and a given fixture only ever exercises one.
- `Utf8UpperBound` at its worst case: an all-CJK column at 3 bytes/unit exactly filling the reservation.
- NCHAR trailing-space trim on *invalid* UTF-16 — the stated reason the trim moved to the output side.
- A chunk closed early by `STAGING_CHUNK_PAYLOAD_BUDGET_BYTES` (short `DataChunk`).
- The `needs_value_fallback` / `FinalizeKernel::Text` path (issue #89).

**No end-to-end test crosses a chunk boundary.** Both new `.test` files use ≤ 5 rows. Arena reuse across chunks, `BeginChunk` re-taking `direct_dst` after `DataChunk::Reset`, offset/count reset, and the whole watermark policy only engage past 2048 rows — and none of it is exercised by any SQL test. A `> 2048` row scan over a mixed-type table (one Direct column, one Var column with NULLs) is cheap and covers all of it.

**No test produces an NBCROW.** SQL Server emits NBCROW only for rows with enough NULLs, and neither new test has the column count to trigger it — so `StageNBCRow`, half the new hot path and where finding 1 lives, has no coverage at all.

## CI wiring

`test/cpp/codec/test_column_staging.cpp` is the only new C++ test in this PR and **it does not run in CI**. `.github/workflows/ci.yml:355` iterates a fixed family list (`boolean integer float decimal string binary uuid money`) plus datetime; `make test-column-staging` exists but nothing invokes it. Adding it to that step is a one-line change.

Both new `.test` files land in the integration suite, which is `workflow_dispatch`-gated and has historically reported PASSED while running zero `.test` files (#212) — so a green check here says nothing about either of them. Worth stating in the PR description rather than leaving it implied.

## Minor

- `RowStager::Invalidate()` (`src/include/codec/staging/row_stager.hpp:106`) is never called. A second COLMETADATA throws at `src/query/mssql_result_stream.cpp:416`, so there's no stale-ops bug today, but the API is dead — wire it or drop it.
- `test/sql/query/unpaired_surrogates.test` requires `MSSQL_TEST_DSN`; 85 of the other SQL tests use `MSSQL_TESTDB_DSN`. Both are exported by the Makefile so it does run, but it's the minority spelling and will silently skip for anyone exporting only the common one.
- `ApplyNegotiatedFraming` → `SetReceiveFraming` zeroes `receive_len_`/`receive_pos_`, discarding anything buffered. Safe at login (the login path drains to a message boundary), but it's an unguarded data-loss path — an assert costs nothing.
- `NextPacket` (`src/tds/tds_socket.cpp:722`) accepts a header length below `TDS_HEADER_SIZE`: `receive_pos_` then advances by less than a header, or not at all. Both callers reject it and return false, so it terminates — but the cursor is left mis-framed, where the pre-055 `TdsPacket::Parse` validated before anything advanced.
- `TDS_RECV_BUFFER_TARGET_BYTES`' comment claims a larger frame "does not raise per-connection memory". The buffer is `scratch + packet_size`, so 4096 → ~68 KB and 32767 → ~96 KB. Bounded, not flat.
- `case AppendArm::Unsupported:` falls through to `Skip` in both walks. Correct, since `ThrowUnsupportedType` is `[[noreturn]]`, but it will trip `-Wimplicit-fallthrough` on some toolchains.

oluies:
## Follow-up: finding 1 reproduced under ASan, plus two more from actually building the branch

I built and ran this branch after my review above. Environment: macOS arm64, Apple clang, `duckdb` submodule at `d8cdaa33`, PR head `17ab611`, `make release` clean.

Three things came out of it. The first confirms finding 1 as a real memory-safety bug; the other two are new and were not in my review.

---

### 1. The NBC `0xFFFF` over-read, reproduced

I wrote a differential framing test (fixture: `<column under test>, <canary INT>`, row copied into an exact-sized `new uint8_t[n]`, built with `-fsanitize=address,undefined`). Feeding a `varbinary(50)` an NBCROW whose bitmap says *present* but whose length prefix is `0xFFFF`:

```
=================================================================
==25255==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000182f7
READ of size 65535 at 0x6020000182f7 thread T0
    #0 __asan_memcpy
    #1 duckdb::mssql::codec::staging::ColumnStaging::AppendVar(unsigned char const*, unsigned int)
         column_staging.hpp:410
    #2 duckdb::mssql::codec::staging::RowStager::StageNBCRow(unsigned char const*, unsigned long, unsigned long long)
         row_stager.cpp:521

0x6020000182f7 is located 0 bytes after 7-byte region [0x6020000182f0,0x6020000182f7)
allocated by thread T0 here:
    #0 _Znam
```

A 65535-byte read starting one byte past a 7-byte row — the row `SkipNBCRow` bounded at 1 (bitmap) + 2 (sentinel) + 4 (canary). Reproduces for `varbinary`, `binary`, `varchar` and `char`; `nvarchar`/`nchar` escape only because `0xFFFF` is odd and trips `ThrowOddUtf16Length` first.

Two lines on each of the NBC var arms fix it — treat `0xFFFF` as NULL, matching `SkipValueNBC` (`tds_row_reader.cpp:858`) and the non-NBC arms:

```cpp
const uint32_t length = static_cast<uint32_t>(p[0]) | (static_cast<uint32_t>(p[1]) << 8);
if (length == 0xFFFF) {
    staging_[c]->AppendNull();
    p += 2;
    break;
}
```

With that applied the whole framing suite passes (68 wire forms × 3 row shapes: ROW, NBCROW, NBCROW-with-bitmap-NULL — bare fixed, all P1 widths, DECIMAL at all four wire widths including the short-mantissa zero-extend, temporal at scale 0/3/7, GUID, P2, PLP declared/UNKNOWN/two-chunk/empty/NULL, XML, and the legacy LOBs). Reverting just the fix reproduces the trace above, so it is a real gate rather than a test that passes either way.

I'd suggest putting the `nvarchar` check in explicitly too — right now the parity test is load-bearing by accident, and it would stop being so the moment anyone reorders those two branches.

---

### 2. `test_column_staging` fails on this branch

`make test-column-staging` is red at `17ab611`, on files this comment does not otherwise touch:

```
[7] ColumnOps: per-column classification (D4)...
FAIL: declared width is carried through for preallocation — got 34, expected 32
FAIL: fallback columns stage as Var (mismatched.kind == StagingKind::Var)
FAIL: unknown wire type falls back (unknown.needs_value_fallback)
3 assertion(s) failed.
```

This is exactly the CI gap from my review — the target exists but nothing invokes it, so the failures are invisible.

Two of the three read as stale assertions rather than code bugs:

- **`max_value_bytes` 34 vs 32** — `column_ops.cpp:391` deliberately adds `+2` for the staged U+0000 delimiter, and the comment explains that omitting it is what made the D10 grow counter fire. The test predates that change.
- **`unknown wire type falls back`** — the `catch` in `ResolveAppend` intentionally sets `Unsupported` *without* the fallback flag, and says so. The meaningful assertion next to it (`does not throw`) passes.

The third one looks like a real bug.

---

### 3. The issue-#89 fallback never reaches its kernel for fixed-width wire types

`ResolveAppend`'s trailing `if (width > 0)` block sets `kind` and `stride` but never assigns an `arm`, so `arm` keeps its default of `Unsupported`. Probing the exact case the test asserts:

```
diverging INT->VARCHAR: arm=17 (Unsupported=17, Skip=18) kind=4 kernel=text fallback=1 stride=4
```

`kernel=text` and `fallback=1` are right, but `arm=Unsupported` means `StageRow` hits

```cpp
case AppendArm::Unsupported:
    ThrowUnsupportedType((*metadata_)[c]);
```

on the first row, so `FinalizeFallbackColumn` is never reached. A view with an inline CAST over an `INT` — the scenario issue #89 is about — would get:

> MSSQL: column 'x' arrives as TDS type 0x38, which this extension cannot decode. CAST it in the query to a type it can…

for a plain integer, where the pre-staging path rendered it as text. Same for every other fixed-width wire type whose catalog type diverges (`BIGINT`, `DATE`, `DATETIME2`, `UNIQUEIDENTIFIER`, DECIMAL, MONEY…).

If the intent is what `needs_value_fallback`'s own comment says — *"The column still stages by its WIRE framing like any other; only the kernel changes"* — then that block wants an arm (`RawStageFixed`/`P1StageFixed` by framing, or `Var`), and the test's `kind == Var` expectation is the one to reconcile against.

---

### Note on my approval

I approved this above at the maintainer's instruction, before building it. Findings 1 and 3 are both things I'd normally hold on, and #3 in particular is a functional regression that this PR's own test catches. Reading the approval as "reviewed and clean" would be wrong; treat the review body and this comment as the actual state.

Happy to open the framing test (`test/cpp/codec/test_row_stager_framing.cpp` + a `make test-row-stager-framing` target + the CI wiring for both staging tests) as a PR against this branch if that's useful — say the word.

oluies:
## Retraction — the comment above is stale

My previous comment reports three problems as live. All three had already been fixed in `5b05262` and `8ce1412` before I posted it, and the PR was merged. Apologies for the noise.

What happened: I built and measured `17ab611`, which was the head when I started the review. By the time I wrote the results up, the branch had moved two commits and merged. I checked the PR state before reviewing but not before commenting.

Corrected status against the merged code:

| Reported | Actual |
| --- | --- |
| NBC `0xFFFF` read as a length → 65 KB over-read | **Fixed.** Both NBC var arms now guard the sentinel — four `length == 0xFFFF` checks where there were two. |
| `FinalizeChunk` memcpy discarding kernel-set NULLs | **Fixed.** Now `published[w] &= st.validity_words[w]`. |
| issue-#89 fallback leaving `arm = Unsupported` | **Fixed.** The trailing `if (width > 0)` block assigns an arm, with a comment naming the failure mode precisely. |
| `test_column_staging` red, three assertions | **Fixed**, and the target is now wired into CI at `ci.yml:374` — which closes the coverage gap from my review. |

The ASan trace and the `arm=17` probe in that comment were real, but they describe `17ab611` and no longer describe `main`. Nothing there needs acting on.

One item from the original review is still open, and it is the reason the NBC divergence was possible at all: nothing pins `RowStager`'s framing to `RowReader::SkipRow`'s. Those are two independent switches over the same wire types in different files, and a disagreement is a heap over-read rather than a wrong answer. I have a differential test for it (68 wire forms × ROW / NBCROW / NBCROW-with-bitmap-NULL, under ASan+UBSan) and will open it as its own PR against `main` rather than continue here.

VGSML:
working on it (almost done)

VGSML:
Thanks for the retraction — and no need to apologise, reviewing a moving branch is how it goes.

On the item you left open: **that differential test exists and is about to land.** I have been building it as spec 059 (`spec/059-staged-path-test-coverage`) since #213 merged, for exactly the reason you name — `RowStager` and `RowReader::SkipRow` are two independent switches over the same wire types in different files, and a disagreement between them is a heap over-read rather than a wrong answer.

What is in the branch, all in `test/cpp/codec/test_row_stager.cpp` and wired into CI beside `test_column_staging`:

- **The framing matrix**: 47 wire forms (bare / P1 / P2 / PLP incl. the UNKNOWN-total form / legacy LOB, every `*N` width, every scale of TIME / DATETIME2 / DATETIMEOFFSET, all four DECIMAL widths plus a trimmed mantissa, empty values) × ROW / NBCROW. Each case asserts that `RowStager` and `RowReader::SkipValue` / `SkipValueNBC` consumed the **same byte count**, and a sentinel column behind the case asserts they consumed the **same bytes** — agreeing on a number and consuming the right bytes are different claims. Built without `-DNDEBUG` on purpose, so `D_ASSERT(p == end)` is live and is itself part of the check.
- **NBCROW coverage** (D1): the bitmap convention asserted directly, every append arm both present and NULLed by the bitmap, the `0xFFFF`-inside-NBC case on both P2 arms, and an `nbc_rows` counter so a fixture that drifts back to plain ROWs cannot pass silently. Plus `test/sql/query/nbc_row.test` against a live server — and note the NBCROW recipe is byte economics, not a column count: two NULL integers among eleven columns is a *tie* against the 2-byte bitmap and the server sends a plain ROW. My first fixture tested the wrong walk while passing.
- **A fuzz harness for the raw-row path**: `fuzz/fuzz_row_stager.cc`, parser in raw-row mode feeding a configured `RowStager` with real output vectors. It is out of the default `TARGETS` because it is the only harness that links libduckdb; `MSSQL_DUCKDB_INC=... MSSQL_DUCKDB_LIB=... TARGETS=row_stager fuzz/build.sh` builds it under libFuzzer + ASan + UBSan.

**I took your third row form.** My matrix had ROW-present, ROW-NULL and NBCROW-present but not NBCROW-with-the-bitmap-bit-set, which is the one where `SkipValueNBC` is never consulted at all. Added, credited in the code, green across the matrix.

Four defects fell out, all now fixed in the same branch:

| found by | defect |
| --- | --- |
| the string-kernel tests (D5) | A value containing `U+0000` mis-split **silently**, and the damage cascaded — every later row in the chunk took the wrong delimiter. SQL Server does store `U+0000` in NVARCHAR (verified live). The boundary walk skipped to each value's lower bound, so a value *ending* in `U+0000` had that zero taken for its delimiter and the real one jumped over inside the next value's skipped span; zeros consumed still equalled values, so the "did the walk end on the last byte" check saw nothing. Fixed by not skipping, which makes that check exact again. Six cheaper schemes were measured and are worse — details in the spec-055 doc. |
| the fuzzer, 5k execs | `DateTimeEncoding::ConvertDatetime` multiplies a full signed 32-bit day count by 86'400'000'000 — signed overflow, UB. **Pre-existing, shared by both read paths**, not something #213 introduced; the conversion layer had simply never been fuzzed (`fuzz_tds_tokens` stops at framing and does not even link `datetime_encoding.cpp`). One range check on the type's own 1753..9999 span. |
| the fuzzer, 132k execs | The issue-#89 fallback in `ResolveColumnOps` took `stride = width` on trust, and `AppendStagedFixed`'s `default:` copies **seventeen** bytes — while `HasOneBytePrefix` omitted MONEYN / DATETIMEN / UNIQUEIDENTIFIER, so a `DATETIMEN` declaring 52 bytes was read as a *bare* value. The parser consumed the row's one-byte NULL prefix, the walk read 17. `D_ASSERT` catches it in debug; release reads past the row. Fixed both ways: the width must be one the staged append actually has a case for, or the column resolves to Unsupported. |
| the fuzzer, 191k execs | `DecimalEncoding::ConvertMoney` assembles its 64-bit value as `(int64_t(high) << 32) | uint32_t(low)`. **`high` is negative for every negative amount** — MONEY -0.0001 *is* -1, so its high word is -1 — and shifting a negative value left is UB. This one fires on ordinary data, not on anything a hostile server has to send; it has been there since the type landed. Assembled through `uint64_t` now. Also pre-existing and shared by both read paths. |

So the open item is closed by this branch rather than left. Rather than two harnesses, I would suggest folding whichever of your 68 forms mine is missing into the same file when the branch lands — same fixture builders, same CI target, one place for the next person to add a form. Happy either way, but say so before you open yours so we do not spend the effort twice.

VGSML:
Correction to the table above, on the DATETIME row: **it is not a range check.**

I wrote it as one — reject days outside 1753..9999 with a named error — then measured it: +0.2 ns on every legal value of every DATETIME column, about a fifth of that kernel, to improve the wording of a case only a corrupt stream can reach. The maintainer's call, and the right one: the conversion now assembles through `uint64_t` exactly like `ConvertMoney`, where overflow is defined. A day count no server can send yields a wrapped, meaningless timestamp instead of an error.

Measured interleaved against `main`: 0.92-1.02 ns/value either side, i.e. free. The `ConvertMoney` fix is free in the stronger sense — identical instruction sequence, `orr x1, x8, x9, lsl #32` unchanged.

The regression test changed with it: it no longer asserts an error, it asserts that such a row still **frames** as eight bytes and leaves the column behind it untouched, which is the property that actually matters.

PR #216, opened 2026-07-30
perf(codec): publish a uniform column-chunk as a CONSTANT vector (spec 056)

VGSML:
> Stacked on #214 — base retargets to `main` when that merges. The diff here is the single commit `08be67b`.

A column-chunk whose values are all the same is published as a **CONSTANT** vector: the batch decode collapses from N values to one. A column-chunk that is entirely NULL runs no kernel at all.

## Measured

Medians of three interleaved runs, ns/value, 2048-row chunks:

| column | uniform | distinct | delta |
| --- | --- | --- | --- |
| DECIMAL(18,2) | 4.81 | 5.59 | **−14%** |
| NVARCHAR(16) | 6.72 | 7.88 | **−15%** |
| BIGINT | 2.26 | 2.27 | 0 |

**The miss is free.** A mixed-NULL column exits on a counter comparison without looking at a value; a column of distinct values exits at row 1.

## Only columns that decode are scanned

The first version scanned every arm. A direct-write column has no decode to collapse — its values were stored into the output vector as they arrived — so scanning it buys a downstream constant and nothing else: **+0.26 ns/value** on a uniform BIGINT for zero scan-side saving. Restricted to the staged families, the change is a pure win where it applies and costs nothing where it does not.

The all-NULL case still covers every column, direct ones included: two counters compared, no value read, and the validity mask never published.

(An earlier detector used `memcmp` per value — a call into libc where a word compare belongs, +1.7 ns/value on BIGINT. Typed loads brought that to +0.26 before the restriction removed it.)

## Why constants and not dictionaries

Spec 056 was closed on the finding that detection costs 13–21 ns/value. That number is a property of the prototype's generic hashing, not of the problem: when the value fits a 64-bit word — INT, BIGINT, any string ≤ 8 bytes — the value *is* the key, and detection measures 0.5–0.9 ns.

Dictionaries still do not ship, for the corrected reason: the deciding half is downstream, where GROUP BY / DISTINCT / JOIN are 2–3.5× faster but **`ORDER BY` is 3× slower**, and the scan cannot know which it is feeding. Emitting one is a bet on the consumer. A constant vector has no such asymmetry — there is no operator for which it is a pessimisation — which is why this half ships and that half waits for an end-to-end measurement.

## Two defects it surfaced

- **`string::DecodeFromTds` treated NTEXT as single-byte text.** It lists NCHAR, NVARCHAR and XML as UTF-16 and never got NTEXT — invisible while nothing called it for that type, since the legacy path could not read NTEXT at all before #197. The constant path publishes row 0 through that entry point and produced `O\0k\0` for `Ok`.
- **The unit-test fixture reused vectors across chunks without a reset**, a state production never reaches because `FillChunk` calls `DataChunk::Reset`. The new `D_ASSERT` in `BeginChunk` caught it on the first run. The fixture now holds a real `DataChunk`, so the reuse path under test is the one that ships.

## Also checked, and not needed

`pk_direct_to_rowid` looked like it would have to be excluded. It does not: that mode writes into a plain output vector and `PopulateRowIdVector` returns without touching it, and the composite paths copy through `VectorOperations::Copy`, which is vector-type aware. The real hazard was `CountChunkForDebug`, which reads every row through `FlatVector::GetData` and would count whatever is left in the buffer of a constant vector — debug-only, fixed here.

## Gates

17 unit blocks; the full local integration suite (132 cases, 3879 assertions); and `diff_check.sh` **byte-identical across 13 queries** against a build without the constant path — NULLs, empty values, embedded `U+0000`, PLP and BCP write-back included.

### DIFF
--- a/specs/056-chunk-analyzer-dictionary/spec.md
+++ b/specs/056-chunk-analyzer-dictionary/spec.md
@@ -207,3 +207,157 @@ detector exits at the second distinct value) and fixed-width dictionaries
 (0.4 ns detection). Both are small wins; neither justifies a phase of its own.
 **Recommendation: fold the CONSTANT/fixed-width part into whatever phase next
 touches the scan, and close this spec.**
+
+---
+
+# Reopened for CONSTANT only (2026-07-30)
+
+The verdict above stands for dictionaries but rested on one number that turns
+out to be a property of the prototype rather than of the problem: **13-21 ns per
+value to detect**. That prototype hashed values generically. When the value fits
+in a 64-bit word — INT, BIGINT, and any string of at most 8 bytes — the value
+*is* the key: no hashing of bytes, no `memcmp`, no pointer chasing, just open
+addressing on a power-of-two table.
+
+Re-measured on that shape, 2048-row chunks (ns per value):
+
+| detector | 1 distinct | 16 | 128 | 1024 | 2048 |
+| --- | --- | --- | --- | --- | --- |
+| CONSTANT check, int32 | 0.49 | ~0 | ~0 | ~0 | ~0 |
+| CONSTANT check, int64 | 0.28 | ~0 | ~0 | ~0 | ~0 |
+| dictionary build, int32 | 0.46 | 0.47 | 0.74 | 0.88 | 0.88 |
+| dictionary build, int64 | 0.46 | 0.47 | 0.78 | 0.93 | 0.89 |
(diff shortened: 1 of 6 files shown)


oluies (approved):

VGSML:
Superseded: GitHub refuses to retarget this PR because it is part of a stack (its base is the #214 branch), and the stack blocks #214 from merging via API **and** web. Reopening the same work as a fresh PR against `main` from `spec/056-constant-vectors-v2`, which carries these commits unchanged plus the merge of #214's branch.

PR #223, opened 2026-07-31
fix(catalog): partitioned tables were unreadable (spec 049, #85)

### LINKED ISSUE #85: [Bug] can't read from table with partitioned clustered index

VGSML:
Closes #85. Implements spec `049-fix-partitioned-table-catalog`, which had been sitting in Draft since 2026-05-19 with the diagnosis already written down.

## The bug

`SELECT * FROM attached.dbo.<table>` against a table with a partitioned clustered index fails at catalog binding:

```
Catalog Error: Column with name log_ts already exists!
```

Three metadata queries joined `sys.partitions` directly to pick up `p.rows`. That view holds **one row per partition**, so on a four-partition table every object and every column came back four times. A plain clustered index reads fine — only partitioning triggers it. Because it fails at bind time, even `LIMIT 1` and `DESCRIBE` fail; the only escape was `mssql_scan()`.

The same join caused a second, silent bug: `approx_rows` took whichever partition the join happened to surface — typically an empty one — so the planner saw a partitioned table as nearly empty, and disagreed with the statistics provider, which already summed. Joining a pre-aggregated subquery fixes both.

## Tests

`test/sql/catalog/partitioned_table.test`, 75 assertions, covering US1.1-1.6 and US2.1-2.2: columns reported once, table listed once, rows read once, projection pushdown, a partitioned **heap** (`index_id = 0` multiplies exactly as `1` does), a partitioned clustered **columnstore** (`index_id = 1`, `type = 5`), the **bulk preload path**, a single-partition control so the common path is pinned, and `estimated_size = 3` across four partitions — a value no single partition holds, so the assertion cannot pass by picking one.

`test/cpp/test_index_kind.cpp` pins the `sys.indexes.type` mapping. It runs in the light **C++ Unit Tests** job on every PR — self-contained header, no submodule checkout, nothing to link.

**Note on evidence.** Integration `.test` files sit behind the `workflow_dispatch`/schedule gate (#212), so the 206-line regression test does **not** run on this PR. The evidence for this change is that it fails with the exact issue-#85 error when the fix is reverted, and that reintroducing the defect in the bulk template alone makes preload report `1 tables, 8 columns` for a two-column table — not a green check here.

Full local suite green: 123 SQL cases / 3689 assertions, 246 C++ cases / 7378 assertions; integration 1098 assertions in 36 cases.

## Beyond the fix

The same aggregated subquery now also reports the object's physical shape — `index_kind` (a typed `MSSQLIndexKind`, not a bool: a clustered **columnstore** reports `index_id = 1` exactly as a rowstore one does and wants the opposite TABLOCK answer) and `partition_count` on `MSSQLTableMetadata` — at no extra round trip.

The write path needs both, and measurement is why: TABLOCK is decisive for a heap under parallel load and *serialises* a clustered index, and sorted input only pays into a clustered index. A clustered index that is not a primary key is invisible to primary-key discovery (it filters on `kc.type = 'PK'`), so this is the only place the extension can learn of one. Verified against `sys.indexes` ground truth for a heap, a single-partition clustered table and a four-partition clustered table.

Adding those trailing columns outgrew the short-row guards in all five parsers, so they are widened too — reading `values[13]` behind a `size() >= 12` check would have been an out-of-bounds read on exactly the malformed input the guard exists to catch.

## Divergences from the plan

- **T001/T002** (seed fixtures via `docker/init`) not done — the test creates and drops its own partition function, scheme and tables, so it needs no container rebuild.
- **T009a-d** (separate pushdown/insert/DML/COPY files) not written — verified manually instead: INSERT, `COPY ... (FORMAT 'bcp')`, filter pushdown and the cardinality estimate all work against a partitioned table. UPDATE/DELETE are rejected there for an unrelated pre-existing reason: they need a primary key for the rowid path, and a clustered index is not one.
- **T010/T011** (C++ test asserting the SQL literals) not done — it needs the file-local templates exposed through accessors, i.e. production surface added for a test weaker than the SQL-level regression test already present.

## Review follow-ups (4ebc9e3)

Chasing the reviewer's "nothing consumes these fields" point turned up a **second broken path**: `mssql_preload_catalog()` runs through `BulkLoadAll`, a third parser the test file never reached, and it carried the identical multiplication. US1.6 now asserts its counters — which come from the parse loop, not the cache — against a dedicated single-table schema.

Also: the five per-site comments said `index_id` where the code reads `sys.indexes.type`; the shape parse publishes both values together so a malformed `partition_count` no longer discards a good `index_kind`; and the uncorrelated derived table was measured on a 20,000-object / 730,000-column catalog — **+2%** over the unfixed direct join, while the `OUTER APPLY` alternative is **2.3x slower** because it re-runs the aggregate per outer row.


oluies (commented):
Reviewed the diff at `fix/049-partitioned-table-catalog` (5 files, +425/−12). I fetched the branch to read surrounding code but did not build or run anything.

## What's right

- **The diagnosis and the fix match**, and one change closes both symptoms — the column multiplication and the arbitrary-partition `approx_rows`. That is the correct shape for this bug.
- **`SUM()` cannot double-count.** `WHERE p.index_id IN (0, 1)` selects heap *or* clustered index, never both, so each group aggregates partitions of a single structure. The same reasoning makes `MAX(ISNULL(i.type, 0))` safe: every row in a group shares one `index_id`, hence one `i.type`.
- **All five short-row guards were widened, and the indices line up.** I audited each against its SELECT list — `<12` / `<14` / `<14` / `>=5` / `>=5`, reading `[10][11]` / `[12][13]` / `[12][13]` / `[3][4]` / `[3][4]`. The sixth guard (`>=8`) belongs to the untouched column-only query. The claim that widening was mandatory holds: `values[13]` behind `size() >= 12` would have been exactly the out-of-bounds read the guard exists to prevent.
- **Enum over bool is the right call**, for the stated reason — a clustered columnstore reports `index_id = 1` like a rowstore one but wants the opposite TABLOCK decision.
- **The test picks the two cases most likely to be skipped**: a partitioned *heap* (`index_id = 0` multiplies identically) and a single-partition control pinning the 99% path. `estimated_size = 3` across partitions holding 2/0/1/0 cannot pass by picking one partition.

## Issues

### Medium — `index_kind` and `partition_count` have no consumers

`git grep` across `src/` outside `mssql_metadata_cache.cpp` returns nothing for either field or for `MSSQLIndexKind`. They are added for a spec-057 write path that does not exist yet, so this lands as dead code whose behaviour nothing exercises and nothing would notice regressing.

Either split them into the write-path PR, or assert them now so the values are pinned before a consumer starts depending on them.

### Medium — four inline comments say `index_id` where the code reads `sys.indexes.type`

Each parser carries a variant of:

```
// index_id 1 is a clustered index, 0 a heap; partition_count > 1 marks a partitioned object.
```

but `values[10]` is `index_type`, from `MAX(ISNULL(i.type, 0))`. For a clustered columnstore `index_id = 1` while `type = 5` — which is precisely the divergence the header comment cites as the reason this field is not a bool. So the inline comments contradict the design rationale sitting a couple of hundred lines above them. The code is right; the comments describe a different column.

### Medium — the clustered-columnstore case is untested

The enum-not-bool decision is justified entirely by CCI, and CCI is the one shape the test does not cover — clustered rowstore, heap and single-partition are all there. A `CREATE CLUSTERED COLUMNSTORE INDEX` fixture asserting `index_kind = 5` would close the gap between the rationale and the evidence, and it is the only case where the enum earns its existence.

### Low — the catch clobbers a good parse

```cpp
try {
    table_meta.index_kind = IndexKindFromSysIndexesType(values[10]);
    table_meta.partition_count = static_cast<idx_t>(std::stoll(values[11]));
} catch (...) {
    table_meta.index_kind = MSSQLIndexKind::HEAP;   // resets a value that parsed fine
    table_meta.partition_count = 0;
}
```

`IndexKindFromSysIndexesType` swallows its own exceptions, so `stoll` is the only thrower here — meaning a malformed `partition_count` discards a correctly-parsed `index_kind`. Parse into locals and assign after, or scope the `try` to the `stoll`.

### Low — the derived table is not correlated

The subquery aggregates all of `sys.partitions` with no schema or object predicate. For `WHERE o.object_id = OBJECT_ID('%s')` the optimizer can usually push the predicate through the `GROUP BY` (the grouping key is the join key), but that is not guaranteed, and the schema-wide and bulk variants aggregate the whole catalog regardless.

The ~1000 µs / ~600 µs A/B in the description measures the `sys.indexes` join, not how the aggregation scales with object count. Worth a measurement against a catalog with a few thousand objects before treating it as free; `OUTER APPLY` correlates naturally if it does show up.

### Low — the regression test will not run on this PR

Integration `.test` files sit behind the `workflow_dispatch`/schedule gate (#212), so 206 lines and 37 assertions land unrun. "Verified to fail with the exact issue-#85 error when the fix is reverted" is the real evidence for this change — a green check on the PR is not, and it is worth saying so in the description rather than leaving it implied.

### Info — `%s` interpolation stays unescaped

`StringUtil::Format(TEMPLATE, schema_name)` does no quoting, so a schema or table name containing `'` breaks these queries. Pre-existing and not introduced here, but these are the templates being edited.

## Style and conventions

Clean. `MSSQLIndexKind` correctly carries the `MSSQL` prefix for a type in namespace `duckdb` per the CLAUDE.md namespace rule; comment density matches the file; the enum values are SQL Server's own, so the mapping stays checkable against `sys.indexes` without a translation table. The CLAUDE.md Recent Changes entry is thorough.

`DATAMODEL.md` is untouched. It does not enumerate `MSSQLTableMetadata`'s fields and no invariant changed, so I do not think an update is required — flagging it only against the repo's own "consider on every PR" rule.

## Verdict

The core fix is correct and well-targeted, and the test design shows real care about what could pass vacuously — the heap case and the single-partition control are both things a weaker test would have missed.

My reservations are all about the scope beyond the fix: two fields nothing consumes, whose motivating case is untested and whose inline documentation describes a different column than the one being read. I would want the comment correction plus either a columnstore test or the fields split out; the rest is polish.

oluies (dismissed):

VGSML:
Thanks — the "no consumers" point led somewhere I did not expect, so it earned more than the fix it asked for.

Addressed in 4ebc9e3.

### The review found a second broken path

Chasing "what would notice this regressing" turned up a parser the test file never reached: `mssql_preload_catalog()` goes through `BulkLoadAll`, which has its own SELECT list and its own guard. `duckdb_columns()` drives the per-table and per-schema loaders instead, so **every assertion in the original test could pass with preload still multiplying**.

It was. Reintroducing the defect in the shared bulk template alone — drop the `GROUP BY`, keep the column count, so the query still returns 14 columns and fails by multiplication rather than by error — makes preload report:

```
Preloaded schema 'partsch': 1 tables, 8 columns
```

for a two-column table on four partitions. Those counters come out of `BulkLoadAll`'s parse loop, not the cache, so US1.6 asserts them exactly against a dedicated single-table schema and cannot pass vacuously. With the fix in place: `1 tables, 2 columns`.

### Medium — no consumers

Pinned rather than split, since the write path is the next spec on this branch's roadmap.

`MSSQLIndexKind` and its mapping moved to a self-contained header, `src/include/catalog/mssql_index_kind.hpp` — no DuckDB or TDS includes — and `test/cpp/test_index_kind.cpp` pins 0/1/5, the unknown-type degradation, and the malformed-input contract. It is wired into the light **C++ Unit Tests** job, so unlike the `.test` file it runs on every PR: `c++ -std=c++17 -I src/include`, no submodule checkout, nothing to link.

What that does **not** pin is the end-to-end value, because `index_kind` has no SQL-visible surface until a consumer exists. The test says so in as many words rather than implying coverage it does not have.

### Medium — comments said `index_id`

Correct, and there were five, not four. All now name `sys.indexes.type` and spell out 1 / 5 / 0, which is exactly where a clustered columnstore diverges from a rowstore one.

### Medium — clustered columnstore untested

US1.5 adds a partitioned CCI. Ground truth confirmed on SQL Server 2022 before writing the assertions: `index_id = 1`, `type = 5`.

### Low — the catch clobbers a good parse

Right. Both values now parse into locals and publish together, in one `ParseTableShape` helper that replaced the five copies of that block.

### Low — the derived table is not correlated

Measured rather than argued, on a synthetic catalog of **20,000 objects / 730,000 columns** (~1/11 of the 200K-table, 8M-column shape this preload path was built for), buffer pool warmed, three alternating passes:

| bulk query shape | pass 1 | pass 2 | pass 3 |
|---|---|---|---|
| shipped (uncorrelated derived table) | 407 ms | 404 ms | 403 ms |
| pre-fix (direct join) | 394 ms | 401 ms | 395 ms |
| `OUTER APPLY` (correlated) | 932 ms | 935 ms | 919 ms |

The aggregation costs about **+2%** over the unfixed join and is stable. `OUTER APPLY` is **2.3x slower** — it re-runs the aggregate per outer row, so correlating it would be a real regression at exactly the scale the concern was about. The single-table variant is indistinguishable across all three shapes at 2,000 objects (45-52 ms per 200 executions, sign flipping between passes): the optimizer does push the predicate through the `GROUP BY`.

### Low — the regression test will not run on this PR

Agreed, and now stated in the PR description rather than left implied. The evidence for this change is the revert check, not a green check.

### Info — `%s` interpolation stays unescaped

Pre-existing and untouched; not smuggling a quoting change into a bugfix. Worth its own issue.

Verified: `partitioned_table.test` 75 assertions, integration 1098 in 36 cases, sql 304 in 16, lint clean, `make test-index-kind` green.

Issue #224, opened 2026-07-31
mssql_scan() returns invalid UTF-8 for non-UTF8 collation char/varchar columns


VGSML:
## Summary

`mssql_scan()` with raw T-SQL puts **code-page bytes into a DuckDB VARCHAR without transcoding or validation**. DuckDB VARCHAR is UTF-8 by contract, so the result is an invalid string: it looks like data was read, and every downstream function silently produces garbage.

Catalog scans are **not** affected — they wrap non-Unicode string columns in a server-side `CAST(... AS NVARCHAR(n))` (`BuildColumnExpression`, `src/table_scan/table_scan.cpp`), so UTF-16LE arrives on the wire. `mssql_scan()` does not go through that path.

## Reproduction

```sql
ATTACH '<dsn>' AS d (TYPE mssql);

SELECT mssql_exec('d', $$
CREATE TABLE dbo.L1 (id INT, v VARCHAR(50) COLLATE SQL_Latin1_General_CP1_CI_AS);
INSERT INTO dbo.L1 VALUES (1, 'caf' + CHAR(233)), (2, 'na' + CHAR(239) + 've');
$$);

-- correct: the catalog path casts server-side
SELECT id, v FROM d.dbo.L1 ORDER BY id;
--   1 | café
--   2 | naïve

-- WRONG: raw scan, no cast
SELECT id, v, length(v) FROM mssql_scan('d', 'SELECT id, v FROM dbo.L1 ORDER BY id');
--   1 | caf<0xE9>  | 4
--   2 | na<0xEF>ve | 5

-- and the corruption propagates
SELECT id, upper(v) FROM mssql_scan('d', 'SELECT id, v FROM dbo.L1 ORDER BY id');
--   1 | CAF澀
--   2 | NA          <- the rest of 'naïve' is gone
```

`0xE9` is `é` in CP1252 and is not a valid UTF-8 sequence on its own.

## Scope

| path | column collation | result |
| --- | --- | --- |
| catalog scan (`d.dbo.t`) | any | correct — server-side CAST to NVARCHAR |
| `mssql_scan()` | UTF-8 (`*_UTF8`) | correct — bytes are already UTF-8 |
| `mssql_scan()` | NVARCHAR/NCHAR | correct — UTF-16LE decoded |
| **`mssql_scan()`** | **any other char/varchar/text** | **invalid UTF-8** |

## Decision

[...]

PR #227, opened 2026-07-31
Advertise UTF8SUPPORT in LOGIN7 — and three bugs it uncovered (#225)

### LINKED ISSUE #225: Advertise UTF8SUPPORT in LOGIN7: copy UTF-8 collation columns straight into the vector
## What

LOGIN7 currently sends feature extensions only for FEDAUTH (`src/tds/tds_protocol.cpp`); it never advertises **UTF8SUPPORT** (MS-TDS feature id 0x0A).

The consequence is measurable in the protocol. A column declared `VARCHAR(50) COLLATE Latin1_General_100_CI_AS_SC_UTF8` arrives on the wire as **type 0xE7 (NVARCHAR), i.e. UTF-16LE**, not as UTF-8 bytes — the server transcodes it for us because we did not say we could take UTF-8. Verified by dumping the TDS collation word per column:

| column | declared | wire type | collation flags |
| --- | --- | --- | --- |
| `VARCHAR(50) COLLATE ..._UTF8` | varchar | 0xE7 NVARCHAR | 0x4D (`fUTF8` set) |
| `CHAR(20) COLLATE ..._UTF8` | char | 0xE7 NVARCHAR | 0x4D |
| `VARCHAR(50) COLLATE SQL_Latin1_General_CP1_CI_AS` | varchar | 0xA7 BIGVARCHAR | 0x0D |

## Why it is worth doing

DuckDB strings are UTF-8. A UTF-8-collation column delivered as UTF-8 needs **no conversion at all** — the staged bytes can be copied straight into the vector, skipping the UTF-16 batch decode entirely (spec 055's fastest measured string path is still ~2.1 ns/value; a plain copy is cheaper than that).

It also roughly **halves the wire bytes** for ASCII-heavy data, since UTF-16 spends two bytes on every ASCII character. That reduces recv() work, TDS framing and staging pressure on the whole read path, not just the decode.

UTF-8 collations have been available since SQL Server 2019 and are the natural choice for anyone storing multilingual data in `varchar`, so this is not an exotic case.

## Scope

- send the UTF8SUPPORT feature extension in LOGIN7 and honour the server's acknowledgement (it may decline, in which case nothing changes);
- route columns whose collation has `fUTF8` to a direct-copy finalize instead of the UTF-16 decode;
- measure: the string cells in `test/cpp/bench_materialize.cpp` plus an end-to-end read of a UTF-8-collation table, against the same table read without the feature.

Related: #224 (non-UTF8 collations on the raw scan path).

VGSML:
Closes #225.

Advertises the TDS `UTF8SUPPORT` feature extension in LOGIN7 so that columns whose collation is a UTF-8 one arrive as UTF-8 instead of being transcoded to UTF-16 for us. On the client that means the bytes are copied straight into the vector — `BIGCHAR`/`BIGVARCHAR` already resolve to `P2StageBinary`/`PlpStageBinary`, one allocation and one copy per column — so the UTF-16 batch decode is skipped entirely.

The request goes out on every connection; `mssql_utf8_support` exists to turn it **off**. Asking costs nothing on a server that lacks the feature: it omits the acknowledgement and keeps sending UTF-16.

## Measured

SQL Server 2022 in docker, 1M rows of ~41-character values, wire bytes from the extension's own counter:

| column | feature off | feature on |
| --- | --- | --- |
| `varchar(100)` with a UTF-8 collation | 85.8 MB, 0.44 s | **43.9 MB, 0.25 s** |
| `nvarchar(100)` | 85.8 MB | unchanged — the feature does not touch it |

Live Fabric Warehouse, 10,000 rows × 3 varchar columns: **1,020,060 → 540,030 wire bytes** (1.89×). Fabric stores UTF-8 natively and its database collation is `Latin1_General_100_BIN2_UTF8`, so every string column takes the "already UTF-8" branch with no cast at all.

## Three bugs this turned up

**1. `FEATUREEXTACK` was being misparsed — and had been since Azure AD shipped.** `ParseLoginResponse` walks the login reply with a chain of token cases and an "unknown token: skip a USHORT length" fallback. `FEATUREEXTACK` (0xAE) is not length-prefixed — [MS-TDS] 2.2.7.11 defines it as a run of `FeatureId(1) | FeatureAckDataLen(4, LE) | FeatureAckData(n)` ending at 0xFF — so the fallback read the first two bytes of the first entry as a length. For the UTF8 ack that is `0x010A` = 266 bytes. It lands past the end of the buffer after everything the parser needs has already been read, so nothing observably broke: all four ENVCHANGEs and the negotiated frame size parse identically either way. That is luck, not design, and the same fallback has been misreading the **FEDAUTH** ack (0x02) all along. The token now has its own bounds-checked walk.

**2. The client's `UTF8_SUPPORT` request must carry no data — `FeatureDataLen` is 0.** Only the server's acknowledgement has a byte. Sending one is tolerated by SQL Server 2022 (it acks and switches the wire form anyway) but makes **Azure SQL reject the login** with error 18456, "Authentication failed for user ..." — indistinguishable from a wrong password. Isolated by probing a live endpoint: an empty feature list connects, `AZURESQLSUPPORT` with a data byte connects, `UTF8_SUPPORT` with a data byte does not. Nothing local reproduces it. Without running the Azure suite this would have shipped and broken every Azure SQL login.

**3. `BuildLogin7WithSSPI` measured its field lengths in UTF-8 bytes** while the protocol counts UTF-16 code units. For ASCII the two agree, which is why nothing caught it; for a non-ASCII value every later offset and the record's own `Length` came out too large and the server was handed a LOGIN7 claiming more bytes than were sent. Two of those fields are user-supplied — the database name, and `Application Name`, which spec 047 made configurable — so `Database=БазаДанных` with Kerberos was enough. Fixed by routing the builder through `EncodeLogin7VarField`, the helper spec 043 introduced for `BuildLogin7`.

## Also: a VARCHAR CTAS target was a silent data-loss switch

`mssql_ctas_text_type='VARCHAR'` created the column with no `COLLATE`, so it inherited the database's code page and SQL Server replaced everything outside it with `?` **on insert** — no error, and nothing downstream can tell, because `?` is valid UTF-8. On a CP1252 database:

```
'plain ascii' -> 0x706C61696E206173636969   intact
'Привет мир'  -> 0x3F3F3F3F3F3F203F3F3F     ?????? ???
'漢字とかな'   -> 0x3F3F3F3F3F               ?????
'euro €'      -> 0x6575726F2080             € became the CP1252 byte, not UTF-8
```

The target now asks for a UTF-8 collation and the same rows round-trip byte-for-byte. Which one is `mssql_utf8_collation`, defaulting to `Latin1_General_100_CI_AS_SC_UTF8`: the default is right for the usual Latin1-based databases and wrong as an assumption, so it is overridable (verified with `Cyrillic_General_100_CI_AS_SC_UTF8`), and setting it empty adds no `COLLATE` clause at all — correct when the database default is already UTF-8, and the documented way back to the old behaviour. `CI_AS` rather than `BIN2` deliberately: unlike a projection cast, this collation is stored in the schema and governs every later comparison against the column, so it must not quietly make it case- and accent-sensitive. Gated on the server having granted the feature — a server without it has no UTF-8 collations either and the DDL would simply fail. When the database default is already UTF-8 (Fabric), no `COLLATE` is emitted; inheriting is cleaner than imposing a Latin1 one.

The ack travels the way `is_fabric_endpoint` does: recorded on `MSSQLConnectionInfo` by the ATTACH-time validation login, which every non-lazy attach performs anyway, so it costs no round trip. Tri-state, because "declined" and "never asked" are different answers — `lazy_validation` skips the login entirely, and that case falls back to borrowing an already-logged-in pooled connection once.

## How each path was verified

| path | evidence |
| --- | --- |
| SQL authentication | local docker, full `.test` suite |
| ADAL | live Azure SQL **and** live Fabric Warehouse |
| SSPI / Kerberos | `test/kerberos` stack, including a non-ASCII database name |
| FedAuth | byte-layout unit test only — it has no production caller (all three Azure entry points go through ADAL) |

Three of the four builders cannot be reached from the local test server, so `test/cpp/test_login7_encoding.cpp` pins their bytes instead: the record's declared `Length` must equal the bytes produced — that is what catches an arithmetic slip — the feature list must parse the way a server would, `FEDAUTH` must still lead the list where it was already present, and a non-ASCII database name must not disturb any of it.

`docs/proposals/utf8-wire-path.md` works out when to go further and ask the server to cast, which is **not** a free win: the cast costs the same regardless of content, so ASCII-ish `nvarchar` gains 35% and CJK loses 35%. It also records the traps found while measuring — `COLLATE` placed after `CAST` converts through the database code page first and replaces Cyrillic/CJK/emoji with `?`; an undersized target truncates on a character boundary with no error and valid UTF-8 output; and Fabric does not support `COLLATIONPROPERTY` at all, closing the connection rather than raising, which killed an earlier idea to put it in the shared metadata queries.

## Deliberately not in this PR

- **The read-side fallback.** A UTF-8-collated `varchar(max)` read *without* the feature is pathological: the server does the downgrade at the protocol layer and burns ~10 s of CPU and 11.9M logical reads per million rows, against 0.69 s when the same conversion is requested as an explicit `CAST` in the SELECT list (server DMV: `cpu_time == elapsed_time`, `wait_type` empty throughout; the identical query via sqlcmd, which negotiates UTF-8, costs the server 353 ms). This PR removes the case wherever the feature is granted, but the fallback branch — cast UTF-8 columns like code-page ones when the ack is absent — is a separate change.
- **The write path.** BCP still maps every char type to `TDS_TYPE_NVARCHAR` unconditionally, so a UTF-8 target is still fed UTF-16 and converted server-side. Same symmetry as the read side, and the same size of win.
- **COPY's auto-create** always emits `nvarchar(max)`, independent of the CTAS setting; unchanged here.

## Verification

- LOGIN7 layout unit test: all pass
- Full `.test` suite as CI runs it (`unittest "test/sql/*"`): **3984 assertions in 131 test cases**
- Azure suite: 62 assertions, 2 live cases
- Fabric Warehouse: connects via ADAL, acknowledges the feature, 1.89× fewer wire bytes
- `test/kerberos` stack: all steps complete; LOGIN7 parsed and answered with both ASCII and non-ASCII database names
- lint clean


oluies (dismissed):

VGSML:
> Do we have dual byte tests?

We did not — good catch. Every string fixture here was Cyrillic, CJK or emoji stored under a UTF-8 or UTF-16 collation, and both of those are wide by design, so the case where one `varchar` character occupies two bytes was untested. That is the shape most likely to break an encoding change.

Added in 9001e75, after checking the behaviour on a live server rather than assuming it.

**Catalog path — correct, and unaffected by the feature.** `Chinese_PRC_CI_AS` (code page 936) stores 汉字测试 in 8 bytes; `Japanese_CI_AS` (932) stores 日本語テスト in 12. Read with the feature on and off, both return identical text:

```
utf8_support=true   汉字测试 | 日本語テスト | 汉字测试
utf8_support=false  汉字测试 | 日本語テスト | 汉字测试
```

That is by construction: `UTF8SUPPORT` only changes the wire form of collations whose `fUTF8` bit is set, and what keeps a double-byte column readable is the scan's own server-side `CAST` to `NVARCHAR` (spec 026), which the feature does not touch. The test asserts the text **and** the character counts (4 and 6), since a code-page round trip that lost anything would not land on those.

**Raw `mssql_scan` — broken, and it predates this PR.** With no `CAST` in the projection the CP936 bytes come back reinterpreted as UTF-8:

```
SELECT v_cn FROM dbo.DbcsProbe   ->  ���ֲ���
CAST(x AS BLOB)                  ->  Conversion Error: Invalid byte encountered in STRING -> BLOB
```

DuckDB rejects its own value, which is a fair verdict. That is #224. I deliberately did not pin it in a test — asserting the mojibake would enshrine it, and the fix belongs to that issue.

One more thing the question is worth flagging for: `docs/proposals/utf8-wire-path.md` claims that for double-byte East Asian code pages a server-side cast to UTF-8 *inflates* relative to UTF-16, which is why that branch is gated. That claim is reasoning, not measurement — it belongs to the not-yet-implemented cast branch, and I have marked it as such rather than let it read as verified.

oluies:
Measured this locally against SQL Server 2019, 2022, 2025 and Azure SQL Edge (Docker, `mssql_utf8_support` on vs off, 50 000 rows × 40 characters per script). Two things worth folding into the PR.

## 1. The win is ASCII-only, and at CJK the feature makes the wire *larger*

Per-row wire bytes on a UTF-8-collated `varchar`, identical on **2019, 2022, 2025 and Edge**:

| script | UTF-8 B/char | feature **on** | feature **off** | ratio |
| ASCII | 1 | **42** | 82 | **0.51×** |
| Cyrillic | 2 | 82 | 82 | **1.00×** |
| CJK | 3 | **122** | 82 | **1.49× — worse** |
| emoji (astral) | 4 | 162 | 162 | **1.00×** |

These match the server-side ground truth exactly (`DATALENGTH` of 40/80/120/160, plus the 2-byte length prefix), so it is arithmetic rather than a measurement artefact: UTF-8 beats UTF-16 only where a character costs one byte, ties at two, and loses at three.

The PR's headline 1.89–2× comes from fixtures that are ASCII-dominant — "~41-character values" and Fabric's `Latin1_General_100_BIN2_UTF8` data. `docs/proposals/utf8-wire-path.md` states this asymmetry plainly for the **cast** case ("ASCII-ish nvarchar gains 35% and CJK loses 35%"), but the PR's measured table carries no equivalent caveat for the **native** case — and unlike the cast, this one is on by default for every connection.

Not arguing against the default: ASCII-dominant text is the common case, `mssql_utf8_support=false` exists, and a CJK-heavy `varchar` with a UTF-8 collation is a narrower population than it might sound. But a reader of that table will reasonably expect ~2× and a CJK workload will get 1.49× *more* traffic, so it is worth a row in the table or a sentence under it. If you would rather ship as-is and revisit, happy to open an issue instead — it did not seem to warrant one while the PR is still open.

Correctness is unaffected: all four width classes round-trip identically with the feature on and off, on every version, including through the DBCS (`Japanese_CI_AS`, CP932) and CP1252 columns, which take the server-side `CAST` and are untouched by the feature.

## 2. The measured MB figures are single-chunk samples — #233

While measuring I hit a pre-existing counter defect, now filed as #233: the `MSSQL_DEBUG=2` stream counters report **one 2048-row chunk** of a multi-chunk scan. A query correctly returning 50 000 rows prints `rows=2048 chunks=0 wire_in=86016B`, where 86016 = 2048 × 42 exactly.

`src/query/mssql_result_stream.cpp` is not touched by this PR, so this is not yours — but it does bear on the description. The "85.8 MB → 43.9 MB" and "1 020 060 → 540 030 wire bytes" figures were presumably taken the same way; if so the **ratios stand** (both sides sampled identically) while the **absolutes understate by roughly `rows / 2048`**. Might be worth either re-deriving them from `DATALENGTH` × row count, or noting how they were obtained.

## Method

Both findings came out of `test/compat/mssql_version_matrix.sh` (#232), which stands a container up per server version and drives the extension rather than sqlcmd. It also confirms something this PR asserts but CI cannot reach: **on SQL Server 2017, which has zero UTF-8 collations, advertising `UTF8SUPPORT` is harmless** — login succeeds and reads are normal.

VGSML:
Thanks — both findings are right, and the version matrix is the part I am most glad to have.

## On the CJK row: accepted, with the ledger's other side

The arithmetic is not arguable. UTF-8 beats UTF-16 at one byte per character, ties at two, loses at three, and your numbers land exactly on `DATALENGTH + 2`. A CJK-heavy `varchar` under a UTF-8 collation gets 1.49× more traffic with the feature on, and the PR's table does not say so.

Two things frame how much that costs, and I want both on the record rather than only the second.

**It is a schema choice, and the natural choice for CJK points the other way.** A UTF-8 collation on a `varchar` is something someone opted into. For predominantly CJK text the type that fits is `nvarchar` — two bytes per BMP character, which is the tight encoding for that script — and `UTF8SUPPORT` does not touch `nvarchar` at all: it arrives `0xE7` UTF-16LE with the feature on and off alike, and takes the same batch decode either way. So the population your row describes is "CJK stored in a column typed for a Latin-ish workload", not "CJK users".

Fabric is the exception worth naming, since it is the one place a user does not get to choose: its database collation is `Latin1_General_100_BIN2_UTF8`, so every string column there is a UTF-8-collated `varchar` whether the data is English or Chinese. A CJK warehouse on Fabric lands in your row wholesale.

**The wire table measures one side of the trade.** With the feature off, a UTF-8-collated column is not simply cheaper on the wire — the server downgrades it to UTF-16 at the protocol layer for every value, and the client converts it back to UTF-8 for the vector. (Code-page `varchar` reaches UTF-16 by a different route: our own `CAST` in the projection, spec 026, which the feature does not affect.) With the feature on, both conversions disappear and the column is a copy at each end.

So at CJK the trade is +49% bytes against two conversions removed. Which way that nets out is not settled by byte counts: on loopback or a fast link the conversions likely dominate, on a constrained link the bytes do. I have a data point only at the pathological end — a UTF-8-collated `varchar(max)` read without the feature cost the server ~10 s of CPU and 11.9M logical reads per million rows, against 0.69 s for the same conversion requested as an explicit `CAST` — and that is `MAX`/PLP specifically, so it does not transfer to your sized `varchar(40)` fixture. For the sized case the server-side cost is unmeasured. I would rather write the caveat as "the wire grows 1.49×; whether the run is slower depends on the link, and has not been measured" than as "CJK is worse", because the latter is a stronger claim than either of us has evidence for.

**The default stays on.** There is no universal answer between the two encodings, but the asymmetries that do exist point one way for this project: DuckDB is UTF-8 internally, so UTF-8 on the wire is the form that needs no conversion at the destination, and Fabric is UTF-8-only by construction. `mssql_utf8_support=false` is there for the workload that measures otherwise.

**Where the caveat goes.** `docs/proposals/utf8-wire-path.md` is the durable home — it already carries the equivalent asymmetry for the cast branch, and it is inconsistent that the native branch has none. Your table belongs there verbatim, with attribution and the version list, since "identical on 2019/2022/2025/Edge" is worth more than the ratios alone. The `mssql_utf8_support` entry in the settings table gets a sentence, and the merged PR body a pointer. No issue needed unless you would rather have one to track it.

## #233 — ours, and thank you for isolating it

`rows=2048 chunks=0 wire_in=86016B` for a 50 000-row scan is unambiguous, and `86016 = 2048 × 42` leaves no room to argue. Confirmed as a defect on our side; it will be fixed in the next PR (the BCP write path), which touches that instrumentation anyway.

Your reading of the consequence is the right one: the ratios in this PR stand, because both sides were sampled the same way, and the absolutes are low by roughly `rows / 2048`. I will re-derive them from `DATALENGTH` × row count when the counter is fixed rather than leave two sets of numbers in circulation.

It reaches further than this PR, which is the more useful part of the report — several read-path figures in `test/bench/bench_results_live_server.md` are derived from `wire_in`, and at least one downstream conclusion (a per-row byte estimate used to bound the parser's buffer copy) is computed from it. Those get re-checked against the fixed counter before anything is built on them.

## #232

The per-version harness is the piece I would not have built and now would not want to be without — particularly the SQL Server 2017 case. The PR asserts that advertising `UTF8SUPPORT` against a server with no UTF-8 collations is harmless; that assertion could not be exercised anywhere in CI, so it was reasoning until you ran it.

Separately: the documentation is outgrowing a handful of Markdown files in the repo root and proposals directory. A proper docs site is on the cards before it gets worse.

PR #232, opened 2026-08-02
test(compat): local SQL Server version × charset matrix

oluies:
Local dev tool, **not wired into CI**. One new file, no production code.

## Why

CI runs exactly one SQL Server version (2022), so everything version-dependent is verified there and nowhere else. Two of those things carry real risk:

- **The LOGIN7 feature list.** A server that dislikes what we advertise rejects the login with 18456 — indistinguishable from a wrong password. That has already shipped once: #225 found `UTF8_SUPPORT` carrying a data byte, which 2022 tolerates and Azure SQL rejects outright.
- **UTF-8 collations and UTF8SUPPORT**, both introduced in 2019. On 2017 there are none, and that is a branch no `.test` file can reach with a single-version container.

## What it does

Stands a container up per version, probes what the server can actually do, seeds accordingly, then exercises **the extension** — `build/release/duckdb`, not sqlcmd — for login, charset round-trip, and per-row wire cost.

Covers **2017 / 2019 / 2022 / 2025 / azure-sql-edge** × four UTF-8 width classes (ASCII 1-byte, Cyrillic 2, CJK 3, astral 4) × each storage shape the extension frames differently:

| shape | path through the extension |
| --- | --- |
| `VARCHAR` + UTF-8 collation | raw bytes into the vector |
| `NVARCHAR` | UTF-16 batch decode |
| `VARCHAR` + DBCS collation (CP932) | server-side `CAST` |
| `VARCHAR` + single-byte code page (CP1252) | server-side `CAST` |

```
test/compat/mssql_version_matrix.sh --versions 2019,2022 --port 14333 --json out.json
```

## Verified

| ver | product | sqlcmd found at | UTF-8 colls | nvarchar | dbcs(jp) | cp1252 | utf8col |
|---|---|---|---:|---|---|---|---|
| 2017 | 14.0.3540.1 | `/opt/mssql-tools/bin` | **0** | pass | pass | pass | n/a (skipped) |
| 2022 | 16.0.4245.2 | `/opt/mssql-tools18/bin` | 1553 | pass | pass | pass | pass |

The 2017 row is the one worth having: it confirms on a real server that **advertising UTF8SUPPORT to a version that lacks it is harmless** — the extension logs in and reads normally.

## Details that had to be discovered rather than assumed

- **`sqlcmd` is not in a fixed place.** 2017/2019 have `/opt/mssql-tools/bin` and *reject* `-C`; 2022+ have `/opt/mssql-tools18/bin` and *require* it. **Azure SQL Edge ships no `sqlcmd` at all**, so it is driven from a `mssql-tools` sidecar sharing the container's network namespace.
- **Readiness polls the service, never the port.** A published port can accept a TCP connection and still reset every byte — and SQL Server accepts connections before recovery finishes. `nc -z` proves nothing here.
- **Seeding branches on UTF-8 collation availability.** Asking 2017 for one is a hard error, not something to skip past.
- **Every literal is `N'...'`.** A bare `'literal'` is parsed in the *database* code page before it is stored, so the CJK rows would already be `?` by the time they reached the column — whatever that column's own collation is.
- **Wire cost is reported per row.** The D4 stream counters report a single chunk rather than the whole scan (`rows=2048 chunks=0` on a query that correctly returns 50 000), so only a per-row figure is comparable between runs.
- **Degrades gracefully on a `main` build.** `has_utf8_setting()` probes for `mssql_utf8_support`, so the UTF-8 cases skip with a stated reason rather than failing — useful before #227 lands, not only after.

## Why not CI

~8 GB of images, and on Apple silicon four of the five run under amd64 emulation (only `azure-sql-edge` is multi-arch with a native `linux/arm64`). Minutes per version. It is a thing you run deliberately before touching the login handshake or the string decode path.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -12,7 +12,8 @@ This guide provides comprehensive instructions for testing the DuckDB MSSQL Exte
 6. [Test Data Reference](#test-data-reference)
 7. [Troubleshooting](#troubleshooting)
 8. [Kerberos Tests (spec 042)](#kerberos-tests-spec-042)
-9. [CI/CD Integration](#cicd-integration)
+9. [Version × Charset Matrix (local only)](#version--charset-matrix-local-only)
+10. [CI/CD Integration](#cicd-integration)
 
 ---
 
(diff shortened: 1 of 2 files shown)


oluies:
## Full matrix run — all five versions green

| ver | product | UTF-8 colls | nvarchar | dbcs(jp) | cp1252 | utf8col |
| --- | --- | ---: | --- | --- | --- | --- |
| 2017 | 14.0.3540.1 | **0** | pass | pass | pass | n/a (skipped, with reason) |
| 2019 | 15.0.4480.2 | 1553 | pass | pass | pass | pass |
| 2022 | 16.0.4245.2 | 1553 | pass | pass | pass | pass |
| 2025 | 17.0.4050.1 | 1585 | pass | pass | pass | pass |
| edge | 15.0.2000.1574 | 1553 | pass | pass | pass | pass |

Wire cost per row on a UTF-8-collated `varchar`, feature on/off — identical on 2019, 2022, 2025 and Edge:

| script | on | off |
| --- | ---: | ---: |
| ASCII | 42 | 82 |
| Cyrillic | 82 | 82 |
| CJK | 122 | 82 |
| emoji | 162 | 162 |

## Three defects the run found in the script itself

Edge initially returned `n/a` in every column. That reads as "not applicable to this version" and actually meant "the seed failed and nothing said so" — the same silent-skip shape worth refusing to accept anywhere. Fixed in 593eb1c:

1. **`seed()` suppressed all output.** It now captures diagnostics, returns non-zero, and the version is reported `SEED_FAIL` with the reason rather than blank columns. The capture also keeps the line *after* `Msg NNN`, which is where `Incorrect syntax near ...` actually lives.

2. **A multi-line batch passed as a `-Q` argument gets mangled through `docker run`.** Edge returned `Msg 102` near a run of mojibake which, byte-swapped from UTF-16, decodes to `EXECUTABLE_LOCATION=/opt/mssql-tools/bin/sqlcmd` — the sidecar's own environment landing inside the SQL. `docker exec` survived the identical string, which is precisely why only the sidecar path (Edge) broke while 2017–2025 looked fine. Batches now arrive on stdin via `-i /dev/stdin`, so there is no argv to mangle.

3. **`CREATE DATABASE` returns before the database is usable**, so the next `-d MatrixDB` could fail. It now waits for `state_desc='ONLINE'` and afterwards asserts the fixture really has its five rows, instead of trusting that no news is good news.

Plus a fourth of the same family: the row-count check used `USE MatrixDB; SELECT COUNT(*)`, and `USE` emits `Changed database context to 'MatrixDB'.` — which `head -1` returned as the count. Now passes `-d` and keeps only a bare number, the same discipline `roundtrip()` needed after the identical mistake (grepping ASCII `|` when DuckDB's box renderer draws U+2502).

Every one of these produced a *plausible-looking* result rather than an error, which is the argument for the assertions now guarding them.

Issue #233, opened 2026-08-02
Stream debug counters under-report: rows=2048 chunks=0 on a 50000-row scan (MSSQL_DEBUG=2)


oluies:
The D4/D10 per-stream counters (`MSSQL_DEBUG=2`, spec 055 / #213) report a **single chunk** of a multi-chunk scan. Any wire-byte or timing figure taken from them is a ~2048-row sample, not a total.

Pre-existing on `main` — `src/query/mssql_result_stream.cpp` is not touched by #227, where I ran into it.

## Reproduction

50 000 rows, one `varchar(400)` column of 40 ASCII characters:

```sql
SELECT sum(length(v)) FROM db.dbo.W_ascii;
```

Query result is **correct**: 50 000 rows, 2 000 000 characters. The counter is not:

```
[MSSQL COUNTERS] stream close: rows=2048 chunks=0 nulls=0 wire_in=86016B str_out=0B
                               plp=0 utf16_fallback=0 fill=0us (parse=0us read=0us process=0us)
[MSSQL COUNTERS]   values/family: string=2048
```

- `rows=2048` — exactly one `STANDARD_VECTOR_SIZE` chunk of the 50 000 read.
- `wire_in=86016` = 2048 × 42 exactly (40 payload + 2 length-prefix bytes), so it is one chunk's bytes, not a partial total.
- `chunks=0` **while `rows=2048`** — the two are incremented in the same function (`rows_read_++` at `mssql_result_stream.cpp:337`, `counters_.chunks++` at `:452`, both inside `FillChunk`), so a stream that read 2048 rows should report at least one chunk.
- `fill=0us` with all phases zero, despite `timing_enabled_` being set at `MSSQL_DEBUG>=1`.
- Exactly **one** `stream close` line is printed, so this is not several parallel streams each reporting a share.

Reproduces identically across four tables and both settings of `mssql_utf8_support`, and is unchanged by `SET threads=1`.

I did not root-cause it. Every `goto exit_loop` path reaches the counter block and the only early `return 0`s are before the row loop, so the divergence is not obvious from reading `FillChunk`.

## Why it matters

Anything measured with `MSSQL_DEBUG=2` is a single-chunk sample presented as a total:

- **Ratios survive** — comparing two runs of the same shape is still valid, because both are sampled the same way.
[...]

Issue #153, opened 2026-06-07
Should COPY TO (bcp) allow compatible numeric coercion (e.g. BIGINT->INT) into an existing table?


oluies:
### Summary

`COPY … TO (FORMAT 'bcp', CREATE_TABLE false)` into an existing table now **rejects any column type difference**, including a compatible numeric narrowing such as DuckDB `BIGINT` → SQL Server `INT`:

```
Invalid Input Error: MSSQL COPY: Column 'id' type mismatch: target expects INT,
source provides BIGINT. Use REPLACE=true to recreate the table with the new schema.
```

[`test/sql/copy/copy_type_mismatch.test` (line ~57)](https://github.com/hugr-lab/mssql-extension/blob/51f4781b44b953e9451b6437a44ecc8b337f7d1e/test/sql/copy/copy_type_mismatch.test#L56) expects this case to **succeed** ("COPY should succeed (BIGINT to INT is compatible)") and load the rows, so the test and the implementation now disagree.

### The question

Is the strict reject-on-any-difference behavior intended, or should the COPY validator allow safe/compatible conversions (e.g. integer widenings/narrowings that SQL Server itself would accept) and only reject genuinely incompatible types (e.g. VARCHAR → INT)?

For reference, `copy_connection_leak.test` only ever expected the **incompatible** `VARCHAR → INT` case to fail with `type mismatch`, which suggests the original intent was: compatible numeric conversions allowed, incompatible ones rejected.

### Options

1. **Keep strict** (current): reject any non-exact type match, require `REPLACE=true`. Then update `copy_type_mismatch.test:57` to expect the error (a `statement error`), and document the behavior.
2. **Allow compatible coercions**: widen the validator to accept conversions SQL Server can perform losslessly/safely (and let the server reject true mismatches), restoring the BIGINT→INT case. Keep the test asserting success.

### Impact / triage

- **Severity:** Low–Medium, behavioral/usability. Affects appending to existing tables whose column types differ from the DuckDB source by a compatible numeric type — common when the DuckDB side defaults to `BIGINT`/`HUGEINT`.

Found while correcting stale `CALL mssql_exec` test syntax in #150; `copy_type_mismatch.test` left untouched there pending this decision. Related: #125 (BCP column handling).

PR #234, opened 2026-08-03
spec 057: BCP write path rebuilt — 6.6x, and the defects that shipped silently

VGSML:
Spec 057 — the BCP write path, rebuilt from measurement. 47 commits, 65 files, +9915/-530.

The premise going in was that the write path was client-CPU-bound. It is not: the bound is SQL Server's **ingest** rate, and the server parallelises across sessions. Most of what follows comes from that, and from the correctness defects the work uncovered — none of which surfaced as an error. They all surfaced as "works, but wrong".

The last third of the branch is a seven-axis review of the first two thirds, and its findings are listed separately below. It found four more silent data corruptions, so the review earned its place.

## Results

44 columns x 1M rows, clean builds, same server:

| | wall | peak RSS |
|---|---|---|
| v0.2.2 | 20.39 s | 510 MB |
| this branch | **3.11 s** | 343 MB |
| this branch, annotated types, no MAX | **2.05 s** | — |

38M x 44 from Parquet: heap 85.6 s (~444k rows/s), clustered columnstore 137.3 s. For scale, the [blog comparison](https://blog.arpe.io/import-parquet-into-sqlserver-duckdb-vs-fasttransfer) measured a DuckDB extension at 553 s.

CTAS, 5M x 44: 44.93 s -> 12.33 s. CTAS inside a transaction, 2M rows: 4.13 s -> 2.32 s (it was capped at one writer there).

## Correctness defects found

Each of these shipped, and each produced wrong or slow behaviour silently.

**Wrong data, no error on either side:**

- **Every integer widening the validator advertised was broken** (#153).
- **`DOUBLE -> real` and `FLOAT -> float` never worked** — the float encoder ignored the target column entirely and sent 8 bytes into a 4-byte column.
- **COPY into an existing decimal column moved the decimal point.**
- **`decimal(19)` stored the wrong SIGN.** The wire magnitude is unsigned 8 bytes and reaches 10^19-1; the kernel picked its arithmetic type from the target's byte size, so 2^63 came back as -2^63. Every smaller value round-trips, which is why the existing precision sweep missed it.
- **`UBIGINT -> bigint` stored -1** for 18446744073709551615 — same width both sides, so direct copy took it, and a round-trip range check cannot catch a same-width unsigned/signed pair.
- **`-1::TINYINT` stored 255.** SQL Server's `tinyint` is the one UNSIGNED integer it has. This one was known and deliberately deferred as a "wire-compatibility question" — but there was no compatibility to keep: the catalog reads such a column back as UTINYINT, so writing -1 never returned -1 either.
- **Parallel COPY lost rows silently** — 205376 of 1000000 arrived. Two locks that did not compose.
- **`TIME` rounded on one encode path and truncated on the other**, and had no end-of-day guard: `23:59:59.999999` into `time(0)` emitted 86400, outside the type's domain. Same value, outcome decided by whether some *other* column forced the row path.
- **`varbinary(n)` truncated on the columnar path only** — the row path sent the whole value and the server rejected the batch.
- **`DATE -> datetime2` crashed.**

**Wrong behaviour:**

- **auto-TABLOCK never fired** (issue #45). The flag it tested was set from `TryGetCurrentSetting` succeeding, which it always does — both auto branches were dead since spec 030.
- **TABLOCK on a clustered columnstore pinned the server to one core.** 8.92 s at ~99% CPU with the hint, 5.23 s at 305% without, and **identical compression either way**. The policy could not be evaluated until the loads were concurrent.
- **A columnstore target compressed nothing** at the default batch size: 102400 rows is SQL Server's own threshold, and the default was 100000. One OPEN rowgroup and 53 MB, versus nine COMPRESSED and 7 MB.
- **Two of three `datetime2` scales never reached the kernel** — the width was being overridden instead of read from `sys.columns`.
- **`money`/`smallmoney` declared a width nothing writes**, which dropped the *whole chunk* to the row path: on a 44-column table two money columns took encode from 190 ms to 394 ms — the cost was paid by the other 42 columns.
- **`mssql_ctas_drop_on_failure` never fired on a source-side error.** It hung off catch blocks in Sink and Finalize, both of which need the *sink* to be what failed; a cast in the SELECT aborts the pipeline without either running. That is also the commonest way a CTAS fails.
- **The CTAS `WITH (...)` clause was parsed, accepted, and dropped.**
- **The counters killed the query they measured** (#233).
- **Two data races**, on COPY's `error_message` (double free) and CTAS's `phase`.
- **A research hook shipped**: `MSSQL_BCP_EXTRA_HINTS` spliced an environment variable into T-SQL verbatim.

**Tests and CI:**

- **CI was red and unnoticed** — a TABLOCK policy change left a unit test asserting the old answer, on a job that runs unguarded on every PR.
- **Four COPY test files had never run.** They gate on `MSSQL_TEST_SERVER`, which neither the Makefile nor CI has ever set.
- **The parallel tests never pinned a thread count.** A writer is claimed per thread, so `parallel_writers = 4` is a ceiling — on a one-core runner both the "one writer" and "four writers" sections tested the same serial path.
- **A float assertion subtracted a column from itself**, so it was 0 for any encoder.

## Performance work

- Columnar scatter on a resolved kernel, replacing the per-row path, family by family: direct-copy, decimal/GUID, temporal (3.5x), strings, then PLP/MAX (1.7x).
- Conversions as a resolved arm: 14.31 -> 1.69 ns/value.
- ASCII `nvarchar` encodes with no simdutf call at all — string encode -50%.
- TDS frames written straight from the accumulator, drained every two chunks: 12.24 -> 10.07 s and 506 -> 332 MB.
- Parallel writers for COPY and CTAS: 10.55 s -> 3.24 s on a heap.
- `UTINYINT` targets stopped dropping the chunk to row-major. The catalog maps SQL Server `tinyint` to UTINYINT, so one such column sent whole tables row-major in an mssql -> mssql load.

## Behaviour changes worth a reviewer's attention

- **CTAS no longer participates in an explicit transaction, at all.** This reverses a change made earlier in this same branch, and the reversal is the considered answer rather than a retreat. Routing CTAS onto the transaction's pinned connection did make `ROLLBACK` discard the rows — but one connection cannot stream a result set and receive a bulk load at once, so `BEGIN; CREATE TABLE t AS SELECT * FROM <the same catalog>` failed outright and left an empty table behind, and every in-transaction CTAS was capped at one writer. The guarantee was half a guarantee anyway: `CREATE TABLE` is DDL on a pool connection and autocommits, so a rollback left an empty table rather than nothing. Undoing a table the statement *created* means dropping it, which is complete and needs no shared transaction — that is `mssql_ctas_drop_on_failure`. `ctas_transaction.test` asserts the new semantics, including that `ROLLBACK` now leaves the rows.
- COPY is unchanged here and deliberately differs: it keeps the pinned connection and one writer inside a transaction, because it may be loading into a table it cannot undo.
- **`-1::TINYINT` now errors** against an existing `tinyint` column instead of storing 255, and a table *created* from a signed `TINYINT` gets a `smallint` — the same widening `USMALLINT -> int` already did. Loads holding only 0..127 are unaffected.
- `mssql_copy_flush_rows` default moved 100000 -> 102400, and `mssql_copy_tablock` is now `auto` (heap on, anything clustered off). Both have their measurements in the setting docs.
- The parallel-writer cap stays at 8 deliberately. Raising it to 16 measured better on a 12-core server, but the writer count should respect DuckDB's thread limit rather than grow a ceiling of our own.

## Docs

- `DATAMODEL.md` had no write path at all. New section: who owns which connection (the two operators now deliberately differ), the writer model, the fact that **one `RowFallback` takes the whole chunk row-major**, and the failure/cleanup sequence.
- README stopped promising an error where strings are now truncated — the only doc divergence that cost a user data rather than surprise — and its tuning section no longer recommends what this branch measured as harmful.

## Next, deliberately not here

- `docs/proposals/columnar-write-close-the-gaps.md` — one NULL in one fixed-width column sends the chunk to the cursor path, where decimal/uuid/datetime are still called **per value** across a TU boundary (+0.23 s CPU per 16M values; wall clock does not move on an ingest-bound server). The design is recorded: the kernels need no NULL branch at all, because a NULL row's cursor advances by 1 and the next column overwrites the garbage. `HUGEINT` is the cheap first move — one such column takes the whole table row-major today, and `SUM()` over integers returns one.
- Deduplicating the parallel-writer machinery between COPY and CTAS. No measured speed, real drift; worth reviewing as a refactor.
- Environment note that is not a code change: a database in FULL recovery gets no minimal logging, and a bulk load into one writes the log a second time — measured 3228 MB of log for 3290 MB of data, against 47 MB in SIMPLE.


VGSML:
It almost includes your PR #233.

oluies:
Ran the integration suite locally against this branch. Scope note: I reviewed **the tests**, not the 32 source files — this is evidence and one actionable gap, not a review of the write path.

Environment: SQL Server 2022 (Developer, 16.0.4245.2) in Docker on port 14333, `TestDB` seeded with this repo's own `docker/init/init.sql` + `init-transaction-tests.sql`.

## All 23 new test files pass

```
test cases:  172 |  171 passed | 1 failed | 21 skipped
assertions: 4889 | 4888 passed | 1 failed | 21 skipped
```

All 23 executed (verified individually — 0 not-executed), plus `counters_uniform_chunk` passes separately with `MSSQL_COUNTERS=1`. The 21 skips are all env-gated paths I have no credentials for (Azure ×6, TLS ×4, `MSSQL_TEST_CONNECTION_STRING` ×4, Kerberos, WinSSPI, named instance).

The single failure is **not from this PR and not a defect**: `copy/copy_varchar_collation.test:74` builds `Server=${MSSQL_TEST_HOST}` with no port, so it hit `localhost:1433` — a different container on my machine — and failed auth. That test silently ignores `MSSQL_TEST_PORT`; pre-existing.

## `make integration-test` still runs none of them

This is the one thing I'd change before merge, and it is a one-line edit to a recipe this PR already touches.

The recipe is unchanged on this branch:

```make
build/release/test/unittest "[integration]" --force-reload
build/release/test/unittest "[sql]" --force-reload
```

Measured, on this branch:

| filter | result |
| --- | --- |
| `"[sql]"` | 8 top-level files, 304 assertions, exit 0 — **none of this PR's tests** |
| `"[mssql]"` (what all 23 are tagged) | `No test cases matched '[mssql]'` → `No tests ran` → **exit 0** |
| `"test/sql/*"` | **172 cases, 4889 assertions** — everything |

So twelve defects that shipped silently get twelve regressions that neither CI nor `make integration-test` executes. The group filters only ever matched the 8 files at the top level; the 164 in subdirectories need the path glob.

**This PR already has the fix — in the wrong target.** The new `counters-test` uses `"test/sql/*"`. Carrying that to `integration-test` closes it.

Worth saying that this PR independently found and fixed two adjacent instances of the same bug: `MSSQL_TEST_SERVER` added to the Makefile ("which nothing has ever set — not this Makefile and not CI — so [four tests] have been skipping silently since they were written"), and `counters-test` gated on `MSSQL_COUNTERS=1` because "a suite that never turns them on cannot catch that class again". The group-vs-glob one is the same bug one level up, and it is the one that hides all 23 new files.

Separately, both jobs that could run these are `workflow_dispatch`/`schedule` gated, so a dispatch on this PR before merge would be worth it regardless.

## One pre-existing seed bug

`docker/init/init.sql` line 258 seeds `9999-12-31 23:59:59.9999999` into a `datetime2`; line 276 then runs `DATEADD(day, id, col_datetime2)` over it and the whole `INSERT ... SELECT` dies with `Msg 517, Adding a value to a 'datetime2' column caused an overflow`. `dbo.AllDataTypes` ends up with **3 rows instead of 6**. The identical line is on `main` (268), so it predates this PR — flagging it because it affects anyone running the suite, and a short fixture is the kind of thing that later reads as a code defect.

## On the tests themselves

Notably strong, and several are built to defeat the assertion that would have passed anyway:

- `parallel_writers.test` asserts count **+ DISTINCT count + checksum**, with the reasoning stated: *"a row written twice and a row lost cancel out in a count."* That is exactly the trap a plain `count(*)` falls into on a lost-rows bug.
- `ctas_transaction.test` asserts CTAS does **not** roll back — the surprising answer — and records that both intuitive alternatives were tried and rejected. The previous version asserted a comment.
- `compatible_type_conversion.test` names the exact error the defect produced and covers widening *and* narrowing.

The TABLOCK handling is the part I would point other work at. The SQL test says outright that *"TABLOCK is NOT observable from DuckDB — nothing here can assert which hint went out, and pretending otherwise would make this file lie"*, then defers to `MSSQLResolveTablock` pinned in `test/cpp/test_index_kind.cpp` — which runs in `cpp-unit-tests`, on every PR. Extracting the undecidable-from-SQL decision into a pure function and gating that is the right answer to a genuinely hard testing problem, and it is the only part of this PR's test surface that CI currently runs.

oluies (dismissed):
Approving on the **test** dimension, which is what I actually examined.

What I verified:

- All 23 new `.test` files execute and pass against SQL Server 2022 — 172 cases / 4889 assertions overall, the single failure being a pre-existing test that ignores `MSSQL_TEST_PORT`, not this PR.
- Each of the twelve claimed defects has a regression, and the ones where a naive assertion would pass anyway are built to defeat it (`parallel_writers`' count + DISTINCT + checksum; `ctas_transaction` asserting the surprising behaviour).
- The TABLOCK decision — untestable from SQL — is factored into `MSSQLResolveTablock` and pinned in `test_index_kind.cpp`, which runs on every PR.

What I did **not** review: the 32 source files. The columnar scatter, the resolved conversion arms, the encoders and the parallel COPY sink are ~8900 lines I have not read, and this approval should not be taken as covering them. Someone should look at that half on its own terms.

One non-blocking item from the comment above, worth folding in since the file is already being edited: `make integration-test` still filters on `"[integration]"`/`"[sql]"`, which match 8 top-level files and none of these 23. `"test/sql/*"` matches all 172, and this PR already uses that glob in the new `counters-test` target.

VGSML:
All three fixed and verified. Thank you — the `integration-test` one is the most useful finding on this PR, and I'd been stepping around it all week without seeing it: I ran the suite as `unittest "*mssql*"` because that was "the filter that works", never asking why the Makefile's filter didn't.

**1. `make integration-test` ran none of them.** Confirmed exactly as you measured, including the part that makes it invisible:

```
"[sql]"       ->   8 files,  304 assertions, exit 0
"[mssql]"     ->   No test cases matched -> No tests ran -> exit 0
"test/sql/*"  -> 172 cases, 4988 assertions
```

The exit 0 is the whole problem — a filter matching nothing is indistinguishable from a suite that passed. Carried the glob over from `counters-test`.

While in there, your remark about "two adjacent instances" prompted an audit of **every** `require-env` in the suite against what the Makefile and CI actually set. It found a third: `MSSQL_TEST_CONNECTION_STRING` gates `copy_auto_tablock`, `ctas_auto_tablock`, `ctas_if_not_exists`, `ddl_if_not_exists` — nothing has ever set it. Two of those cover auto-TABLOCK, the behaviour this PR *rewrote*, so its own tests had never executed. All four pass; skips drop 21 → 17.

**2. The seed bug — and it had been baked into assertions.** `dbo.AllDataTypes` had 3 rows where the seed always intended 6, exactly as you diagnosed. Fixing it broke five assertions in `filter_pushdown.test` that had been written against the half-populated table. Re-derived every `AllDataTypes` query in that file from the server rather than patching failures one at a time. The boundary row keeps `9999-12-31 23:59:59.9999999` instead of being shifted — it is worth having.

**3. `copy_varchar_collation.test:74`.** Only ATTACH in the suite building a DSN without the port. Adding the port alone does not fix it, which is presumably why it was written that way: the harness substitutes only variables the file **declares**, so `${MSSQL_TEST_PORT}` came through literally. The `require-env` line was the actual fix. `winsspi_basic.test` had the same omission twice — gated, so never run, but it would have hit the wrong server the moment anyone did.

`make integration-test` now exits 0 with **4988 assertions across 155 executed cases**.

On scope: your point that the 32 source files are unreviewed is fair and I'm not treating this as covering them. That half had a seven-axis review earlier in the branch — which is what turned up the two silent data corruptions in the columnar path — but that was self-review, and it is not the same thing as another pair of eyes. Happy to split the source half out if that would make it reviewable.

VGSML:
@oluies — a scoped follow-up ask, because you were right to decline 8900 lines and I think I described the PR badly.

**It is not 8900 lines of source.** That figure is the whole diff. Measured:

| | lines |
|---|---:|
| whole diff | 9915 |
| `src/` added | 3937 |
| …of which comment / blank | 1695 |
| **actual source code** | **2242** |

And it does not spread evenly across the 32 files:

| area | code | independent? |
|---|---:|---|
| **columnar core** — `write_column_ops.{cpp,hpp}`, `bcp_row_encoder.cpp`, `decimal_codec.cpp`, `datetime_codec.cpp`, `string_codec.cpp` | **1409** | no — one interlocking change |
| concurrency + connection ownership — `copy_function.cpp`, `dml/ctas/*` | 471 | yes |
| everything else — metadata, settings, integer/uuid/binary codecs, `bcp_writer.cpp` | 244 | yes |

**The ask: read the first row only.** Six files, 1409 lines. The other 715 can ride on the tests you already verified — I am not asking you to read them.

That is not an arbitrary boundary. All four silent data corruptions the branch's own review found were in those six files: `decimal(19)` storing the wrong sign, `UBIGINT -> bigint` storing -1, TIME rounding on one encode path and truncating on the other, `varbinary(n)` bounded on one path only. None were in the concurrency half. The defect density has already told us where to look.

A reading order, most load-bearing first:

1. **`codec/write_column_ops.cpp`** (304 lines) — resolves each column to a `ScatterArm` from the (source type, target column) pair. Everything downstream trusts it. Two failure modes: a wrong arm writes wrong bytes, and a needlessly conservative arm sends the **whole chunk** row-major — one column's mistake costs the table. Both have happened.
2. **`tds/encoding/bcp_row_encoder.cpp`** (658) — assembles the row. Two layouts: `ScatterBlock` (fixed stride, taken when every column is fixed-width and no NULLs) and `CursorBlock` (per-row positions, taken when either is false). The width invariant is the thing to be suspicious of: a kernel writing a different number of bytes than the chunk was sized from puts every later column where the server is not looking.
3. **`codec/{decimal,datetime,string}_codec.cpp`** (1096 combined) — the kernels. Each has a strided form and must agree byte-for-byte with the row-major encoder. That agreement is asserted by tests rather than guaranteed by structure, which is the honest weak point.

Also, separately: my fixes for your three findings dismissed your approval (`dismiss_stale_reviews` is on). Nothing in them is new functionality — the `integration-test` glob, the `MSSQL_TEST_CONNECTION_STRING` export, the seed fix with the five stale assertions it exposed, and the two ATTACHes missing the port. If you are willing to re-approve the test dimension on that basis, it would unblock the PR independently of the source review above.

VGSML:
Added the map I should have written before asking: **[`docs/columnar-encode-core.md`](https://github.com/hugr-lab/mssql-extension/blob/spec/057-write-recon/docs/columnar-encode-core.md)** (`ca8e000`).

It covers the six files as a system rather than a diff — the three encode paths and what selects each, the full `ScatterArm` vocabulary, and the part that is actually worth reading: **the four invariants in the order they have historically broken**, each named with the defect that broke it. If you only read one section, that is the one; it is the review checklist.

The design idea to hold onto while reading: **compatibility IS `arm != Unsupported`**. It used to be a separate table of type *names* that disagreed with what the encoders could execute — every widening it advertised died in the encoder, while a decimal scale mismatch was waved through and moved the decimal point (#153). Collapsing the two is what makes an unexecutable conversion unadvertisable.

The weak points are in there too, stated rather than argued away — a guide that only explains the design is worse than none:

- the "one implementation, so the two paths cannot differ" property is **partial**. It holds for the copy and convert arms; it does not for decimal/uuid/datetime/string, where the row encoder has its own code. That gap is precisely where TIME and `varbinary(n)` diverged, and it is why the byte-equality tests exist at all;
- the cursor path still calls three kernels per value across a TU boundary (+0.23 s CPU per 16M values against a `bigint` control that costs nothing);
- `HUGEINT` and `UBIGINT` columns take the whole table row-major, though both already travel as DECIMAL where a kernel exists. `SUM()` over integers returns HUGEINT, so that is an ordinary shape.

Writing it also turned up two stale doc comments **in the file I asked you to read first**: `Datetime` still claimed TIME and DATETIMEOFFSET "are not here yet" (this branch added both), and `RowFallback` still described whole families taking the row path when what lands there now is specific pairs. Both would have actively misled you. Fixed in the same commit.

oluies:
Four documentation-accuracy findings on `ca8e000` (`docs/columnar-encode-core.md` + the `write_column_ops.hpp` comments). I checked each against the code rather than passing them along — line references corrected where they were off.

Nothing here is a code defect; the guide is genuinely useful. These are places where it states as settled something the code has not reached yet.

### Medium — the doc drops a hedge the header deliberately keeps

`docs/columnar-encode-core.md:75` — *"`Unsupported` | no kernel for this pair — **this is the compatibility answer**"* — and `:81`, *"Making compatibility mean `arm != Unsupported` **is** what stops a conversion being advertised that cannot be performed."*

Both present-tense. The header is careful to say the opposite (`write_column_ops.hpp:34`):

> Compatibility is **meant to become** `arm != Unsupported`

Verified: neither half holds today.

- `IsTypeCompatible` is alive and is still the gate — defined `src/copy/target_resolver.cpp:648`, called `:787`.
- `ResolveWriteColumnOps` can never return `Unsupported`. The only producer is `DirectCopyArm`'s `default:` (`src/codec/write_column_ops.cpp:59`), reachable only for a width outside {1,2,4,8}, and the caller has already excluded those. Every other exit assigns a concrete arm.

Worth fixing mainly because of where it sits: the file has a *"Weak points, stated rather than hidden"* section that lists two other gaps and not this one — and this is the one a reviewer would most want flagged. Matching the header's tense, and adding a bullet there, would do it.

### Low — "HUGEINT and UBIGINT take the whole table row-major" is true on one path only

`docs/columnar-encode-core.md:134`, unqualified.

It holds for generated metadata (CTAS / replace), where `col.duckdb_type = source_types[i]` (`target_resolver.cpp:1149`) so `DirectCopyTargetWidth` returns 0 and the column falls back.

For COPY into an **existing** table the metadata comes from the catalog, where the column is `DECIMAL(38,0)`. `PhysicalType::INT128` is in the resolver's `src_ok` set and `target.max_length == GetDecimalByteSize(38)`, so a HUGEINT source resolves to `ScatterArm::Decimal` (`write_column_ops.cpp:106-121`). Only UBIGINT (`UINT64`, absent from `src_ok`) is unconditionally row-major.

### Low — the row-major condition is not exhaustive

The path table presents its condition as complete, but two live routes are missing:

- **Family mismatches** — a non-int-storage source into DECIMAL, TIMESTAMP into DATE, non-UUID into `uniqueidentifier` (`write_column_ops.cpp:110-119, 134-175, 181-190`).
- **A `PlanColumn` failure on a string column**, which abandons the whole chunk and is not `RowFallback` at all: `bcp_row_encoder.cpp:654-656` (the review said `:659-663`; the actual `return false` is at `:655`).

### Low — `VarString` also serves PLP/MAX

Described as *"2-byte-length string or binary"*, but `nvarchar(max)` / `varchar(max)` / `varbinary(max)` resolve to the same arm under `WireKind::Plp` with 8-byte framing (`write_column_ops.cpp:89-97`) — as that code's own comment says, *"Same arm: the plan owns the framing difference."* A reader taking the table literally will not expect MAX columns there.

---

These came from a roborev pass I ran on `ca8e000`; I verified them before posting and am leaving the wording to you, since it is your prose and the hedging is the substance of the Medium one.

oluies (commented):
Reviewed the **source** this time — the ~10.5k lines I explicitly declined to approve on last week. Concurrency and the scatter core in depth; the codec families more selectively. Tests and docs were covered in earlier comments.

Short version: the two structurally dangerous parts — the parallel sink and the strided scatter — are correct, and correct for reasons that are written down. One latent narrowing worth guarding, and two comment/code mismatches.

## The parallel sink is right, and the fix is the right fix

The original defect was real ("two locks that did not compose"), and the repair is the minimal correct one: `WriteRows` and `FlushToServer` now happen under a **single continuous** `gdata.write_mutex` hold, with the comment stating the invariant — *"Already held from the append above — the two must not be separable."* That closes the window where a flush could send and clear the accumulator between another thread's append and its flush check.

Verified the obligations that fix creates:

- **One `FlushToServer` call site** (`copy_function.cpp:800`), so "caller holds the lock" has exactly one place to be honoured, and it does.
- **No reverse lock order.** The order is `gdata.write_mutex` → `BCPWriter::write_mutex_` (`bcp_writer.cpp:181`). `BCPWriter` holds no reference to `MSSQLCopyGlobalState` and never calls back, so the reverse acquisition does not exist and cannot deadlock.
- **The per-writer branch shares nothing** but atomics — own `writer`, own `connection`, `rows_sent`/`rows_confirmed`/`batches_flushed` via `fetch_add`. No mutex needed and none taken.
- **`BCPCopyCombine` flushes thread-local rows** via `FinishLocalWriter`, which is the bug this shape usually has. First-error-wins under `error_mutex`, connection released by the local state's destructor (the #191 contract).

Separating `error_mutex` from `write_mutex` is also right, and the comment explains why — the failing writer's `unique_lock` unwinds before the catch runs.

## The strided scatter is memory-safe, and non-obviously so

`ScatterBlock` computes `col_off += 1 + widths[c]` while the planner computes `stride += 1 + widths[c]` for fixed columns and `stride += 2` for variable ones. Those disagree — which would overrun — **except** the strided path is guarded by `if (all_valid && !has_variable)` (`bcp_row_encoder.cpp:744`), so `widths[c]` there is always a real fixed width and `col_off` tracks `stride` exactly. Buffer is `row_count * stride`. Correct, but it rests on a guard three hundred lines from the arithmetic it protects; worth a line at `ScatterBlock` saying it may only be called when `!has_variable`.

Dispatching the arm outside the row loop, and putting the width in the type rather than a variable the loop re-reads, is the right shape and the measurements back it.

### Latent: `widths` is `uint8_t`, `wire_width` is `uint32_t`

`bcp_row_encoder.cpp:662` — `widths[c] = static_cast<uint8_t>(ops[c].wire_width);` — is an unguarded narrowing.

It is **safe today**: `CanScatter()` rejects `RowFallback`/`Unsupported` before this line, and every remaining fixed arm is ≤17 bytes (DirectCopy ≤8, Decimal ≤17, Guid 16, Datetime ≤10, Int/FloatConvert ≤8). The several `ops.wire_width = target.max_length` assignments — which for `binary(300)` would exceed 255 — all land on arms that never reach here.

But the safety is a coincidence of the current arm set, not a stated invariant. Add a strided fixed-width arm for `binary(n)`/`char(n)` — a natural next optimisation, and `target.max_length` is already being assigned in those branches — and `widths[c]` truncates silently, `stride` comes out too small, and the kernel writes the full width into it. Heap overflow, no diagnostic.

Given this PR's own premise is twelve defects that shipped silently, an explicit guard seems in keeping:

```cpp
if (ops[c].wire_width > std::numeric_limits<uint8_t>::max()) {
    return false;   // row path; a strided arm this wide is not representable here
}
```

or make `widths` a `vector<uint32_t>` and let the cost be nothing.

## Two comment/code mismatches

- `copy_function.cpp`, `BCPCopyCombine` — *"Recorded rather than thrown from here: finalize owns the error path"* sits directly above a `throw;`. It records **and** rethrows. The behaviour looks intentional; the comment describes a different one.
- Same family as the `columnar-encode-core.md` findings in my previous comment (the `Unsupported` hedge), so I suspect a single pass over the new prose would catch both.

## What I did not review

`datetime_codec.cpp` (+386), `string_codec.cpp` (+347), `decimal_codec.cpp` (+333) and `integer_codec.cpp` (+106) I read only where the scatter path enters them. The per-family kernels and their row-major twins — which the doc itself flags as the place "the byte-equality tests exist because the structure does not enforce it" — deserve their own pass. `IntConvert` does carry a real range check (`ScatterConvTyped` returns false, then `ThrowIntRange` against `numeric_limits<DST>`), which was the claim I most wanted to confirm.

Not blocking on any of the above. The narrowing is the one I would take before merge, because it costs a line and removes a whole class of future failure.

oluies (dismissed):
Approving. Scoping it, because this PR is large and an unqualified approval would claim more than I checked.

**Verified in depth:**

- **The parallel sink.** Single continuous `write_mutex` hold across `WriteRows` + `FlushToServer`; one call site honouring that contract; lock order `gdata.write_mutex` → `BCPWriter::write_mutex_` with no reverse path possible, since `BCPWriter` holds no reference to global state and never calls back; per-writer branch sharing only atomics; `BCPCopyCombine` flushing thread-local rows. The lost-rows defect is genuinely closed, and closed for stated reasons.
- **The strided scatter's memory safety.** `col_off` tracks `stride` exactly under the `all_valid && !has_variable` guard; buffer sized `row_count * stride`. Correct.
- **The tests.** All 23 new `.test` files executed and passed locally against SQL Server 2022 (172 cases / 4889 assertions overall); each of the twelve defects has a regression; the untestable-from-SQL TABLOCK decision is factored into `MSSQLResolveTablock` and pinned in a test that runs on every PR.

**Not reviewed by me:** the per-family kernels — `datetime_codec.cpp` (+386), `string_codec.cpp` (+347), `decimal_codec.cpp` (+333), `integer_codec.cpp` (+106) — read only where the scatter path enters them. The doc itself flags these as where "the byte-equality tests exist because the structure does not enforce it", so the columnar/row-major twins are the place a divergence would hide. That is a real gap in this approval, not a formality.

**Outstanding, and approved over rather than resolved:** the `uint8_t` narrowing at `bcp_row_encoder.cpp:662`. Safe today only because every scatterable fixed arm is ≤17 bytes; a future strided `binary(n)` arm truncates `widths[c]` silently and overruns. One line, and worth taking whenever the write path is next opened even if not now.

Also two comment/code mismatches noted earlier (the `BCPCopyCombine` "recorded rather than thrown" above a `throw;`, and the `Unsupported` hedge in `columnar-encode-core.md`) — prose, not behaviour.

VGSML:
All three taken, in `646b39b` — and thank you for reading the source half. That was the ask and it landed.

**The narrowing.** Took the second option: `widths` is now `vector<uint32_t>`, matching `wire_width`, so the truncation cannot exist rather than being caught. A guard that is unreachable today reads as a check for its own sake, while the type change removes the class and costs nothing — `widths` is indexed per column, not per value. Your reasoning about *why* it is currently safe went into the comment, since "safe by coincidence of the arm set" is the part that would otherwise be rediscovered.

**`ScatterBlock`'s precondition** is now stated at the function, with the mechanism: its `col_off += 1 + widths[c]` tracks the planner's `stride += 1 + widths[c]` only because a variable column cannot be present — the planner adds `stride += 2` for one while setting `widths[c] = 0`. You are right that a guard three hundred lines from the arithmetic it protects should not be the only record of it.

**The `BCPCopyCombine` comment** was simply wrong and is rewritten: it records *and* rethrows, and the halves are needed for different reasons — the record makes first-failure-wins work across N writers, the rethrow is what actually fails the statement, Combine being the last place that thread can.

---

One thing your *previous* comment turned up that is worth flagging, because it went further than the Low finding it was filed as. You noted HUGEINT is row-major only with generated metadata, since into an existing `decimal(38,0)` it resolves to `ScatterArm::Decimal`. **Four tests used exactly that fixture — `decimal(38,0)` fed `1::HUGEINT`, into an existing table — specifically to force the row path.** They never forced anything: instrumenting the path choice prints `COLUMNAR`. So `time_rounding_both_paths`, `smalldatetime_write` and `string_bound_truncation`'s row-path section were comparing the columnar path with itself, and their assertions could not fail.

Fixed in `b6c21d3`. The working lever is a signed `TINYINT` into SQL Server's unsigned `tinyint` — verified, one `ROW-MAJOR` chunk each. They all still pass, so the paths did agree; it just was not being tested. The guide now says that a both-paths test needs its **lever** checked, not only its assertions.

Also found while verifying: **`INTERVAL` into an `nvarchar` target raises an INTERNAL Error.** The resolver's comment claims such a render-as-text source "keeps the row path, which formats it first" — it does not; `string::EncodeToBcp` reads the vector as VARCHAR and throws. Nothing covers it. That is pre-existing and out of scope here, so I will open it separately rather than widen this PR.

Suite is green: `make integration-test` exits 0 at 4988 assertions / 155 executed cases.

Last thing, and it is awkward: pushing these fixes dismissed your approval again (`dismiss_stale_reviews`). Nothing in `646b39b` is behaviour beyond the three items above. I am not going to ask you to re-approve after every round — if you are comfortable with the shape, say so and we will merge on the substance of your two reviews rather than the badge.

Issue #189, opened 2026-07-15
Allow sharing temp tables across connections


p-j-smith:
Thanks for making this extension!

I'm currently using C# to query MSSQL but am interested in switching to DuckDB and this extension. In C#, I use a long-lived connection to create a set of global temp tables that I then query from multiple threads concurrently, writing the results to parquet files.

However, I've run into a problem when testing out switching to use this extension - there is currently no connection lifetime a shared temp table can anchor to:

- session-scoped temp tables require a pinned transaction, so they're confined to a single DuckDB connection i.e. the queries that use the temp tables them cannot run concurrently
- global temp tables don't survive pooling. The creating connection is flagged for reset when released back to the pool, so the tables are dropped whenever that connection is next reused. Creating them inside a pinned transaction doesn't help either: the tables hold schema locks until COMMIT, so other connections block, and the commit releases the pin anyway.

I can build the temp tables once in a pinned transaction, read them into a local DuckDB catalogue, and BCP-COPY them into `#` temp tables for each session / worker. This works fine, but it means I need to reupload the same data into every session. I could instead create the temp tables in each session, but that puts more load on the server and means e.g. `CURRENT_TIMESTAMP` would be slightly different across workers.

## Suggested fix

A user-controllable leased connection, excluded from pool reuse/reset until released, would be great:

```sql
SELECT mssql_lease('my_catalogue');   -- pin one TDS connection for this DuckDB connection, outside any transaction
SELECT mssql_exec('my_catalogue', 'SELECT ... INTO ##mytable FROM ...');  -- runs on the leased connection
-- ##mytable is now visible to all other pooled connections while the lease is held
SELECT mssql_release('my_catalogue'); -- reset + return to pool (or DETACH does it)
```

I did look at the `mssql_open`/`mssql_close` handle API, but besides being deprecated, the handle only supports `mssql_ping`/`mssql_close` — there's no way to execute SQL on it, so it can't create the temp tables anyway. A lease on the catalogue's own pool (with `mssql_exec`/`mssql_scan` routing through the leased connection while it's held) also seems to be more in line with keeping everything catalogue-scoped (which I saw in spec 047).

Happy to make a PR if you think this is useful, and open to alternative approaches.

VGSML:
Hi, this is interesting question. You can use a duckdb table instead a sql server tmp table for this, can't you?
I get some time to think about (week)

VGSML:
Hi, my thoughts:

[...]

PR #237, opened 2026-08-04
Proposal to add a user-controllable leased connection

p-j-smith:
Related to #189

Proposal to add support for leased connections:

- pooled connections get reset between statements, so `#`/`##` temp tables vanish before multiple queries or workers can share them
- propose new `mssql_lease`/`mssql_release` functions that pin one specific connection so its temp tables survive across statements and workers.
- only affects `mssql_exec`/`mssql_scan` calls made outside a transaction. An active transaction always takes priority.
- reuses existing pinning and pool-lifetime patterns already in the codebase
- alternative considered: write to an ordinary table in tempdb instead. This is simpler, but needs a `CREATE TABLE` grant on `tempdb`

Feel free to close if it's not something you want to support


p-j-smith:
closing in favour of a much simpler idea https://github.com/hugr-lab/mssql-extension/issues/189#issuecomment-5179935893

PR #245, opened 2026-08-06
bench: final v0.2.2→v0.2.3 performance report + dictionary-vector CI test

VGSML:
Closing report for the phase-R item of `docs/proposals/simd-chunk-materialization-design.md`: what a user actually gets upgrading community mssql 0.2.2 → 0.2.3, measured end to end on one wide real-shape table (44 columns × 38M rows, all type families), two interleaved rounds, stock CLI + community 0.2.2 as the baseline.

## Headlines

- **Write, like for like** (same statement, 1 writer, NVARCHAR(MAX)): **1.34×**; **out of the box** (default derives 8 parallel writers): **≈6×**; with sized strings: **≈9.7×**.
- **The PLP tax**: giving MAX-typed strings a length (`SET mssql_default_string_length=200`) takes **2.45×** off the load on identical data.
- **Columnstore**: 4.5× smaller tables, rowgroups land COMPRESSED directly.
- **Reads**: 1.22–1.40× wall; client cost drops much harder — **16× fewer instructions** on an 8-column drain (1.42 T → 87 G), 6.7× less peak RSS on write, 4× fewer packets (16K TDS frames).
- **Finding F0**: v0.2.2 **segfaults** on COPY from any real columnar source at scale (parquet reads and native-table scans emit dictionary vectors; the old row encoder indexed them as flat — `EncodeBinaryPLP` dereferences string content as a pointer). Fixed structurally by the spec-057 unified-format encode; now pinned in CI by `test/sql/copy/vector_encodings_bcp.test` (dictionary + constant vectors, content round-trip both directions).
- Content verified at full scale: 44/44 per-column hash signatures match for plain-heap, varchar-UTF-8 and columnstore writes.

## Contents

- `test/bench/bench_results_v023_report.md` — the report (methodology, honesty notes incl. the macOS `.timer` user/sys unreliability that forced `/usr/bin/time` for CPU claims).
- `test/bench/v023_report/` — runnable harness (variant runner, generated flat-vector source for the 0.2.2 baseline, verify script) + raw CSVs of both rounds.
- `test/sql/copy/vector_encodings_bcp.test` — new integration test, 15 assertions, green.

No production code changes.


oluies (commented):
Reviewed the new CI test and the harness scripts. No production code here, so this is about whether the test pins what the report claims it pins.

## What holds up

**The assertions in `vector_encodings_bcp.test` are strong.** `EXCEPT ALL` in *both* directions catches missing rows, extra rows and corrupted content — not `count(*)`, which a duplicate-plus-a-loss slips straight past. The constant-vector case checks `count(DISTINCT a)` and `count(b)` rather than just survival.

**The test is honest about its limit**, and says so where a reader will find it:

> There is no SQL-visible assertion "this vector was dictionary-encoded", so the test constructs sources that are dictionary-encoded by construction

Same discipline as `tablock_by_target_shape.test` declining to assert a hint it cannot observe.

**The harness does not repeat the `bench_058_t0.sh` bug** that the #244 review just turned up. None of the five scripts set `-e`, so the `rc=$?` at `run_variant.sh:79` is genuinely reachable — in #244 the identical check sat under `set -euo pipefail` and could never run. Worth noting since these landed days apart.

**`verify.sh` actually verifies.** Per-column `sum(hash(col))` signatures, `diff`ed, `exit 1` on mismatch with the differing column names printed. Not a rubber stamp.

## The one thing worth changing

**The dictionary coverage is by construction, unverified, and can go vacuous without failing.**

The test depends on DuckDB's parquet reader and storage layer *choosing* dictionary encoding for those columns. That is true today. If a future DuckDB emits flat vectors there — a cardinality heuristic changing, a storage format revision — every assertion still passes, and nothing about dictionary handling is exercised. F0 is specifically a dictionary-vector crash, so the one property this file exists to hold is the one that can quietly stop being tested.

This is the same shape as the tautology roborev found in my own matrix script last week: coverage that silently stops covering, with a green check either way.

There is a precedent in the repo for the robust version. `test_index_kind.cpp` took a decision that could not be asserted from SQL, extracted it, and pinned it in a test that runs in `cpp-unit-tests` **on every PR**. The equivalent here is a C++ test that constructs a `Vector` in `DICTIONARY` and `CONSTANT` form explicitly and runs the encoder over it:

- it pins the actual invariant rather than a proxy for it;
- it needs no SQL Server;
- it runs on every PR, whereas this `.test` file sits behind the `workflow_dispatch`/`schedule` gate and will not.

I would keep the SQL test — the end-to-end round-trip is worth having — and add the C++ one beside it.

## Smaller notes

- **~950k rows of BCP** join the integration suite (400k parquet + 400k native + 150k constant). Fine while that suite is gated; worth knowing it is now a slower gate.
- **F0's segfault claim I did not verify** — that needs a v0.2.2 build. The mechanism described (`EncodeBinaryPLP` dereferencing string content as a `string_t` pointer on a dictionary vector) is plausible and the spec-057 fix is structural, but I am taking it on your word rather than confirming it.
- Since F0 is **a crash in a released version**, it may deserve its own issue for the changelog and release notes, separate from a bench PR — users on 0.2.2 hitting a segfault on any parquet-sourced COPY would want to find that by searching issues.

Nothing blocking.

VGSML:
Took all three actionable points:

**The vacuous-coverage risk → `test/cpp/test_vector_encodings.cpp`** (54166d3). Exactly the `test_index_kind` move you described: DICTIONARY vectors built explicitly (`Slice` over a selection, NULL inside the dictionary) and CONSTANT vectors (incl. constant NULL), asserted **byte-identical** to a flat encode of the same logical content — on both encoder paths (`EncodeChunk`'s scatter and `EncodeRow`), with the PLP families (nvarchar(max)/varbinary(max)) the 0.2.2 crash lived in. The SQL test stays as the end-to-end round trip.

One correction to the "runs on every PR" expectation: the per-PR `cpp-unit-tests` job is deliberately libduckdb-free (it compiles self-contained TUs only), and this test needs `Vector`/`DataChunk` plus the encoder's dependency cone — `BCPColumnMetadata`'s methods live in `target_resolver.cpp`, which pulls catalog/connection/TLS. So it links the full static archives like `make test-cpp` and runs in the build job's unit-test step: every merge to main and every full CI, not every PR. Still strictly better than the SQL twin, which sits behind the SQL Server gate.

**F0 as its own issue → #246**, with the stack, the ~1M threshold, the LIMIT-hides-it trap, and the workaround statement (none; upgrade).

**The ~950k-row cost note** — acknowledged, leaving the row counts as they are: 400k is what makes the parquet dictionary pages real, and the suite is already server-gated.

(The red checks on this PR are the current GitHub Actions major outage — every job died with "not acquired by Runner", including yamllint. Will re-run when Actions recovers.)

oluies:
A roborev pass over `origin/main..a9600020c` raised six findings. I checked each against the branch before relaying — all six reproduce. Two are worth acting on; the rest are small.

Complements rather than repeats my earlier comment: that one was about the dictionary coverage being by-construction, this is the harness and the arithmetic.

## Medium — the native-table section verifies in one direction only

`test/sql/copy/vector_encodings_bcp.test:85-95`. The parquet section asserts `count(*)` **and** `EXCEPT ALL` in both directions. The native-storage section that follows the second COPY asserts only server-minus-source.

That direction alone proves the server multiset is a *subset* of the source. Silently **dropped** rows — a chunk lost at a batch boundary, exactly the class of bug #234's parallel-writer work was about — pass it. Given this file is the F0 guard, and F0 is a data-corruption crash, the asymmetry is worth closing: add the reverse `EXCEPT ALL` or a `count(*)`, mirroring the section above it.

## Medium — the baseline binary is detected by path substring

`run_variant.sh:25`:

```bash
if [[ "$BIN" == *stock* ]]; then LOAD_LINE="LOAD mssql;"; fi
```

The README documents `MSSQL_BENCH_BASELINE` as a free-form path. Any baseline CLI whose path lacks the literal `stock` — `~/duckdb-1.5.5/duckdb`, say — silently skips the `LOAD`, and since community extensions are not autoloaded, every baseline run then dies at `ATTACH ... (TYPE mssql)`. A candidate path that happens to contain "stock" gets a spurious `LOAD`.

Better driven off an explicit env flag, or by probing (`"$BIN" -c "LOAD mssql"` succeeding ⇒ stock).

## Low — three reported midpoints do not match the committed CSV

Recomputed from `results_r1r2.csv`:

| figure | r1 | r2 | midpoint | report says |
| F1 0.2.2 COPY (`gen-w` stock) | 928.351 | 937.873 | **933.112** | 933.2 |
| F2 `mssql_default_string_length=200` (`w-maxlen`) | 95.305 | 91.961 | **93.633** | 93.7 |
| F5 grouped aggregation, 0.2.2 (`r-grp` stock) | 38.600 | 34.468 | **36.534** | 36.6 |

Each is +0.1 rather than nearest-rounded, so it looks like a systematic aggregation difference rather than three typos — worth saying which aggregation was used. In fairness the derived headline barely moves: 230/93.633 = 2.456 vs the quoted 2.45×, so that claim stands.

## Low — `campaign.sh` block B never runs `w-plain-1`, and block G does not exist

`campaign.sh:23` is `for n in 2 4 8`, yet the committed CSV's very first row is `r1,w-plain-1,release/duckdb,664.808` — so the published round-1 single-writer number came from outside the committed harness. Worth adding to block B so the report is reproducible from what is checked in.

Separately, the comment at `campaign.sh:31` says the old/new CTAS pair is "measured from the generated source in block G", but the case statement handles only `B|C|B2|C2` and the usage string says the same. Either add G or point the comment at `gen_run.sh`.

## Low — the password reaches the console on failure

`run_variant.sh:18` builds the DSN with `Password=${MSSQL_TEST_PASS}` and `:75` interpolates it raw into `ATTACH '$DSN'`. A password containing `'` truncates the literal and produces a parser error that echoes the offending line — and `:83` then does `tail -5 "$LOG" >&2`, printing the DSN, password included. That contradicts the "never printed" guarantee in the header and README.

Escaping (`${MSSQL_TEST_PASS//\'/\'\'}`) fixes the syntax half; scrubbing the DSN from the failure `tail` fixes the disclosure half.

## Low — assorted

`gen_run.sh:33` has `min_rt=3; [ -n "$LOAD_LINE" ] && min_rt=3 || min_rt=2`, where the first assignment and the truthy branch are both no-ops. `run_variant.sh` defines a `w-plain0` variant nothing references. On a failed `r-parquet` run the script exits before its `rm -f`, leaving a multi-GB `r_export_out.parquet` in the tree with no ignore entry (same for `results.csv` / `sig_*.txt`; only `*.log` is covered today) — a `trap` would handle it.

---

Nothing here is blocking. The two Mediums are the ones I would take: the one-directional `EXCEPT ALL` because this file is the F0 regression guard, and the substring detection because it fails in the direction that looks like a broken extension rather than a broken harness.

VGSML:
All six taken in 3b8c6d9:

**Native-table `EXCEPT ALL` now runs both directions** (16 assertions, green against the live server) — you're right that server-minus-source alone blesses dropped rows, and dropped-at-a-batch-boundary is exactly the class this file guards.

**Binary detection is a probe now**: `mssql_version()` resolving *without* `LOAD` ⇒ static candidate; otherwise the script emits `LOAD mssql;`. Smoke-tested both directions. Fails-toward-broken-harness rather than broken-extension, as you put it.

**The midpoints** — good catch, and it was systematic: I averaged the already-rounded per-round values instead of the raw CSV. A full cross-check against `results_r1r2.csv` found **seven** such +0.1 drifts, not three (also w-plain-1/-4/-8 and w-cci). All recomputed from raw values; the header now states the aggregation exactly. Derived ratios are unchanged (2.45×, 4.52×, ≈6×).

**campaign.sh**: block B now runs `w-plain-1` (the committed CSV is reproducible from the committed harness), the "block G" comment points at `gen_run.sh`, and `w-plain0` carries a comment naming its purpose (the F0 bisect at LIMIT sizes).

**Password**: quote-escaped in all four scripts' DSNs, and the failure `tail` scrubs `Password=…` before echoing.

**Assorted**: `min_rt` is a real branch, a `trap` cleans the r-parquet export on any exit, and run artifacts (logs, CSVs, signatures, export) got a `.gitignore`.

Issue #164, opened 2026-06-15
Authentication failed for user - Login and user are not the same


Viroxken:
I'm not able to connect (SQL authentication) to a sql server database in which the login `suser_sname()` and the database user `user_name()` is not the same. In my case the login is mapped to `dbo` user.

**Error:**
`_duckdb.InvalidInputException: Invalid Input Error: MSSQL connection validation failed: Authentication failed for user 'xxx' - check username and password`

I'm able to connect to another database in the same sql server instance. In that case login and user are the same.

VGSML:
Thanks for reporting this. @oluies could you take a look on it?

oluies:
NeXT week

oluies:
Thanks @Viroxken — I dug into this and I don't think it's actually a password problem.

**Root cause of the confusing message.** SQL Server rejects a `LOGIN7` with **error 18456** for many *distinct* reasons — bad password, an inaccessible default/initial database, a disabled login, etc. The only field that tells them apart is the ERROR token's **State** byte. Our parser was reading that byte and immediately throwing it away, and then `TranslateConnectionError` mapped *every* 18456 to `"check username and password"`. So a database-access failure (correct password) looked identical to a wrong password — which is exactly your case: the same credentials work against another DB in the same instance.

The differing variable between the working DB and the failing one is the **initial database** sent in `LOGIN7`. The usual 18456 states here are:
- **State 40** — could not open the database specified in the login (initial catalog problem)
- **State 38** — database offline / inaccessible for that login
- **State 5/6** — login/user name or case mismatch

`dbo`-mapping itself is normal and works in SSMS/ADO.NET, so something about how the initial catalog is resolved for that login is the trigger — but we couldn't confirm which without the state.

**What I changed** (branch `fix/164-login-error-state`, PR incoming):
- The parser now captures the State byte, and login failures surface as `error 18456, state N` in `MSSQL_DEBUG` output.
- Added unit + fuzz coverage for the login-response parser.
- (Bonus hardening #183: clamped the ERROR message length to the token extent.)

**Could you help confirm the exact state?** Either:
1. Reproduce with `MSSQL_DEBUG=3` and paste the `DoLogin7` line, **or**
2. Pull the 18456 line from the SQL Server **errorlog** (it logs the state there even though the wire only ever shows "Login failed for user 'x'").

Also, two quick questions:
- Are you passing `database=` in the ATTACH string, and if so, what is it?
[...]

PR #255, opened 2026-08-11
test(tds): #164 follow-ups to #254 — the quiet casualties and the fuzz corpus

oluies:
Rebased onto #254, which landed the same DONE-length fix (plus a second site I had missed, `ParseDoneForAttentionAck`). The parser change is dropped; what remains is the coverage #254 did not include.

## Four cases

| Case | Why it wasn't covered |
|---|---|
| `TestPacketSizeEnvChangeAfterDone` | **The quiet casualty.** A PACKETSIZE ENVCHANGE behind a DONE was lost to the same desync — but the login still *succeeds*, so nothing failed and nothing errored. It just pinned the connection at the 4096 default, and spec 055's measured −28% CPU / −43% wall silently did not happen. |
| `TestPlainDoneBeforeLoginAck` | The Synapse capture is all DONEINPROC/DONEPROC. Plain `0xFD` shares the branch but was covered by assumption only. |
| `TestDoneExactlyAtBufferEnd` | `ptr + 12 <= end` has two directions and #254 holds one. An off-by-one here would reject the commonest token in every login response. |
| `TestTruncatedDoneDoesNotOverread` | 9 payload bytes — enough for the old 8-byte skip, four short of the real 12 — so the bounds check is exercised on malformed input, not just well-formed. |

## Fuzz corpus

`fuzz/corpus/login_response/` was untouched by #254. The three seeds carrying a DONE encoded the **same pre-7.2 8-byte payload as the bug** — a shape SQL Server never sends. The corpus therefore never started from a valid DONE-bearing response, and coverage past a DONE was reachable only by mutation luck.

Same class as the `fuzz_tds_tokens` mode byte in #247: seeds silently testing something other than what they were committed for, with a green CI throughout.

- `error_18456_state40`, `error_18456_state8`, `loginack_ok` re-baselined to the 7.2+ layout
- new `regression_doneinproc_before_loginack` carrying the Synapse shape — it reaches LOGINACK only with #254's fix in place

All 7 seeds verified to parse as intended:

```
error_18456_state40                    success=0 err=18456 state=40 msg='Login failed for user 'app_user'.'
loginack_ok                            success=1 err=0 state=0
regression_doneinproc_before_loginack  success=1 err=0 state=0     <- reaches LOGINACK past the DONE run
regression_envchange_zerolen           success=0 err=0 state=0 msg='No LOGINACK token in response'
```

## Verification

- 16 tests / 41 checks green under `-fsanitize=address,undefined`
- Real-server check on SQL Server 2022 (RTM-CU24): ATTACH + query round-trip, PACKETSIZE negotiated to 16192 — the path `TestPacketSizeEnvChangeAfterDone` pins


VGSML (approved):

oluies:
### Verification status

**Unit** — 16 tests / 41 checks green under `-fsanitize=address,undefined`. All 7 corpus seeds verified to parse as intended:

```
error_18456_state40                    success=0 err=18456 state=40 msg='Login failed for user 'app_user'.'
error_18456_state8                     success=0 err=18456 state=8
error_overlong_msglen                  success=0 err=18456 state=40 msg=''
loginack_ok                            success=1 err=0 state=0
regression_doneinproc_before_loginack  success=1 err=0 state=0     <- reaches LOGINACK past the DONE run
regression_envchange_zerolen           success=0 err=0 state=0 msg='No LOGINACK token in response'
regression_fedauth_offset_overflow     success=0 err=0 state=0 msg='No LOGINACK token in response'
```

**Real server** — SQL Server 2022 (RTM-CU24, 16.0.4245.2). ATTACH + query round-trip, and the PACKETSIZE ENVCHANGE this PR pins is negotiated live:

```
[MSSQL PROTO] ParseLoginResponse: ENVCHANGE type=1, len=27
[MSSQL PROTO] ParseLoginResponse: ENVCHANGE type=7, len=8
[MSSQL PROTO] ParseLoginResponse: ENVCHANGE type=2, len=23
[MSSQL PROTO] ParseLoginResponse: ENVCHANGE type=4, len=21
[MSSQL CONN] DoLogin7: authentication successful, spid=0, packet_size=16192
```

**Integrated auth (Kerberos)** — `AuthenticateIntegrated` is the fourth caller of `ParseLoginResponse` and the one with the most distinct response shape (SSPI continuation rounds), so #254's change is worth exercising there rather than assuming. Ran `test/kerberos/` (KDC + SQL Server + client; the Dockerfile does `COPY . /src/`, so it compiled this branch). All 5 steps completed, and step 4 is the informative one:

```
[MSSQL CONN] AuthenticateIntegrated: LOGIN7 payload=1016 bytes -> 3 TDS packet(s) (max_packet=512)
[MSSQL PROTO] SSPI token: blob_len=186
... MSSQL Kerberos auth rejected by server (error 18452): Login failed. The login is from an
    untrusted domain and cannot be used with Integrated authentication.
[run-tests] step 4: OK -- multi-packet LOGIN7 reassembled & answered by SQL Server
```

The ERROR token is **parsed and surfaced with its number** rather than lost to a desync — which is the property at issue. (18452 is this stack's expected outcome: it has no AD trust, so the step asserts the multi-packet LOGIN7 was reassembled and answered, not that the login succeeds.)

**Not covered:** no Azure SQL or Synapse instance available to me, so the Synapse shape is exercised only through the captured bytes in the fixture and the new corpus seed, not against a live gateway.

PR #257, opened 2026-08-11
spec 068: login routing for every auth path (handoff)

VGSML:
Spec for the routing follow-up to #254 — @oluies picking this up from here (per #255 discussion; the DONE-size fix landed in #254, this is the second half).

## What the spec settles

- **D1 — the contract**: `has_routing` outranks `success`, uniformly. Today we get it wrong both ways: route-without-LOGINACK on the fedauth path dies as "auth failed" (the `!success` return at `tds_connection.cpp:~618` precedes the routing capture at `~633`), and route-with-LOGINACK on SQL auth "succeeds" against a gateway that told us to leave.
- **D2 — one hop driver**: extract the loop from `AuthenticateWithFedAuth` (it already does everything right: `host\instance:port` normalization, SNI to the routed host, per-hop state resets, MAX_ROUTING_HOPS=5) into a private `RunWithRoutingHops(attempt)`; all three auth paths become thin wrappers. No SQL Browser on a hop.
- **D3 — authenticator factory for integrated auth**: `AuthenticateIntegrated` takes a `(host, port) → IAuthenticator` callable instead of a pre-built instance — a hop needs a Kerberos/SSPI ticket for the **routed host's SPN**, not a retry of the gateway's. Explicit `service_principal_name=` survives hops verbatim. Callers (`mssql_catalog.cpp:194`, `mssql_storage.cpp:1338`) already hold `MSSQLConnectionInfo` by value, so the change is local.
- **D4 — testing without a routing server**: parser pins (route±LOGINACK, ROUTING behind a DONEINPROC run — composition with #254), a loopback fake-TDS-server hop test (two listeners, `use_encrypt=false`), Azure Test + Kerberos e2e dispatched on the branch as smoke.

Full text: `specs/068-login-routing-unification/spec.md`.

Known risk called out in the spec: the SPN-over-hop design mirrors go-mssqldb but has no live AD+MI routing environment to verify against — the error path names the SPN it tried so a field report is diagnosable in one round.

### DIFF
--- a/specs/068-login-routing-unification/spec.md
+++ b/specs/068-login-routing-unification/spec.md
@@ -1,0 +1,180 @@
+# Spec 068 — Login routing for every auth path, not just FEDAUTH
+
+**Status**: DRAFT
+**Branch**: `spec/068-login-routing`
+**Follows**: PR #254 (the TDS 7.2+ DONE-token size fix, merged e2d2822). That PR
+fixed the parse desync that #88/#164 actually hit; this spec closes the gap it
+exposed on the way: TDS login-time routing (ENVCHANGE type 20) is honoured on
+exactly one of our three authentication paths.
+
+## 1. Reconnaissance — what exists and where it stops
+
+`ParseLoginResponse` parses the ROUTING ENVCHANGE completely and correctly
+(`tds_protocol.cpp` ~832): `has_routing`, `routed_server`, `routed_port` land on
+`LoginResponse`, with a level-1 debug line. The parser is not the problem.
+
+**The one working consumer** is `AuthenticateWithFedAuth`
+(`tds_connection.cpp:243–391`). It owns a hop loop with everything a hop needs,
+built and battle-tested for the Azure SQL / Fabric gateways:
+
+- `MAX_ROUTING_HOPS = 5`, full PRELOGIN + TLS + LOGIN7 per hop;
+- routed-target normalization: Fabric sends
+  `hostname.pbidedicated.windows.net\INSTANCE:port`, so it extracts a trailing
+  `:port` (digits only), keeps `hostname\instance` (no port) as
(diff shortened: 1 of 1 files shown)


VGSML:
Superseded by #258, which carried the identical spec.md and is now merged (e2d2822 → main). The routing work continues in #263.

oluies:
Closing as superseded — the spec text landed in #258 rather than separately.

`specs/068-login-routing-unification/spec.md` on `main` is byte-identical to this branch (SHA-256 `421378dc75e1…`), and it is the only file this PR touched, so nothing is lost. It arrived that way because the implementation branch took the spec verbatim from here before building on it.

Where the spec ended up, for the record:

- **D1** (`has_routing` outranks `success`) — implemented on all three auth paths, expressed as a tri-state `LoginAttemptOutcome` rather than the bool-plus-side-channel the spec implied, since that side channel *was* the original bug.
- **D2** (one hop driver) — `TdsConnection::RunWithRoutingHops`, which also owns every state transition; the per-attempt reset covering the first attempt turned up a latent reused-connection packet-id bug, and the hop reconnect was found to be ignoring `mssql_connection_timeout`.
- **D3** (authenticator factory) — shipped, and **its acknowledged risk is now closed**: `test/kerberos` gained a routing gateway (#261), and CI shows a real MIT KDC issuing a second service ticket for the routed host.
- **D4** (testing without a routing server) — parser pins plus a loopback fake-TDS-server suite, now running on Linux, macOS and MSVC.

Two things the spec did not anticipate, both fixed: `BuildLogin7` mirrored a single argument into both HostName and ServerName, so the routed-ServerName fix initially moved HostName with it; and the empty-server-name ROUTING edge is already refused by the parser, so no hop to an empty target is possible.

Thanks for the handoff and for the independent verification on #258 and #263.

PR #258, opened 2026-08-11
spec 068: login routing on every auth path (implements #257)

### LINKED ISSUE #256: SQL-auth and integrated-auth logins ignore the ROUTING ENVCHANGE (Azure gateway redirect)
Split out of #164, where it came up as a candidate cause and was **ruled out** — the failure there was the DONE-length desync fixed in #254. This gap is real but separate, and to my knowledge nobody has hit it yet.

## What

`ParseLoginResponse` parses the ROUTING ENVCHANGE (type 20) and sets `LoginResponse::has_routing` / `routed_server` / `routed_port` for **every** caller. Only one caller reads it.

| Auth path | Function | Follows routing? |
|---|---|---|
| Azure AD / FEDAUTH | `AuthenticateWithFedAuth` | yes — up to `MAX_ROUTING_HOPS` (5), re-does PRELOGIN + TLS against the routed host, uses the routed name for SNI |
| SQL authentication | `DoLogin7` | **no** — checks `login_response.success` only |
| Integrated (Kerberos / SSPI) | `AuthenticateIntegrated` | **no** |

Verified on `e2d2822`: neither function mentions `has_routing`.

## Why it matters

Azure SQL Database's gateway can answer LOGIN7 with a ROUTING token and **no LOGINACK**, expecting the client to reconnect to the node named in the token. That is normal operation under the Redirect connection policy, and it is per-database — databases live on different nodes — so it would present as *this database works, that one doesn't* with identical credentials.

On the two paths above we would fall straight through to:

```
Authentication failed: No LOGINACK token in response
```

which is the same user-facing message #164 produced for a completely different reason. That collision is its own argument for closing the gap: the next person to see that string should not have to re-run the #164 investigation to find out which of the two it is.

## Suggested fix

Hoist the hop loop out of `AuthenticateWithFedAuth` into something the three paths share, rather than copying it twice. It already handles the parts that are easy to get wrong — capping hops, re-running PRELOGIN and TLS against the new host, and using the routed hostname for SNI rather than the original gateway (per `go-mssqldb`).

Failing that, at minimum: if `has_routing` is set and we do not follow it, say so — `"server requested routing to <host>:<port>, which this authentication path does not follow"` beats `No LOGINACK token in response`.

## Testing

No Azure SQL instance in CI, and I have not reproduced this against a live gateway — the classification above is from reading the code, so treat the user impact as inferred rather than observed. `ParseLoginResponse`-level unit tests can pin the parse and the decision without a server; `test/cpp/test_login_error_state.cpp` already has the fixtures and helpers for exactly this shape.

oluies:
Closes #256. Implements the spec from #257. Draft: opened to get CI signal — see the risk note at the bottom.

## What changed

`TdsConnection::RunWithRoutingHops` now owns the hop loop that used to live inline in `AuthenticateWithFedAuth`; all three auth paths are thin wrappers over it.

- **D1 — the contract.** If the login response carries ROUTING, the outcome on that socket is *hop*, whatever it said about `success`. A `bool` cannot express that (it forces the driver to read `has_routing_` as a side channel off a `false` return — the arrangement that produced the bug), so an attempt reports `LoginAttemptOutcome{Success, Route, Failure}`.
- **D2 — one driver.** It also owns every state transition. `AuthenticateIntegrated` used to store `Disconnected` + close the socket at nine points inside its own body; a callback doing that mid-loop leaves the state machine dead while the next socket connects. Hop budget (5), routed-target normalization (`host\instance:port`, no SQL Browser on a hop) and per-hop resets unchanged; the hop-limit error now names the last routed target.
- **D3 — authenticator factory.** `AuthenticateIntegrated` takes `(host, port) -> IAuthenticator`. No compatibility overload: it would make replaying the gateway's ticket the easy thing. `DeriveSpn` needed no change — it already reads `info.host`/`info.port` and passes an explicit `service_principal_name` through verbatim, so both D3 requirements fall out of existing code.

**Not in the spec, found on the way:** `DoLogin7` passed `host_` as the LOGIN7 ServerName while the integrated path passed `tds_server_name_`. Once SQL auth can hop, a routed named instance would have received the instance-stripped name. Non-routed logins stay byte-identical — the driver seeds `tds_server_name_ = host_` before the first attempt.

**No change needed** in `AuthenticateWithFedAuth` after the ordering fix (the driver treats `Route` uniformly), nor at `mssql_diagnostic.cpp:144`, which inherits routing through `Authenticate()`.

## Tests

- `test/cpp/test_login_routing_hops.cpp` — **new**, 45 checks. Drives a real `TdsConnection` against a loopback fake TDS server (two listeners on `127.0.0.1:0`, `use_encrypt=false`, no DuckDB link, ~200 ms). Pins: zero-hop unchanged, single hop with LOGINACK, hop without LOGINACK, per-hop packet-id reset, hop-limit abort, integrated-auth factory re-invoked with the routed host, `Free()` before the next build, hop-failure error naming the routed host, first-attempt-null wording preserved.
- `test/cpp/test_login_error_state.cpp` — 4 parser pins added (56 checks total), including ROUTING behind a DONEINPROC run, the composition with #254.

The hop pins were confirmed to **fail** under the pre-068 behaviour before being confirmed green under the fix.

## Behaviour change

SQL auth against an endpoint answering ROUTING *and* LOGINACK now redirects instead of transacting against the gateway. Correct per [MS-TDS] — a routed session is not usable — but visible to anyone unknowingly relying on the old behaviour. Changelog entry included.

Internal API: `AuthenticateIntegrated`'s signature change affects embedders calling the TDS layer directly.

## ⚠️ The one unverified thing

The new test is wired into the **ungated** `cpp-unit-tests` job (a routing regression should fail the PR that causes it, and the `build` job is dispatch/schedule-gated). That job has no vcpkg, so the step installs simdutf via `apt-get install libsimdutf-dev` on Linux and `brew install simdutf` on macOS. **Package availability on `ubuntu-latest` was not verifiable locally** — this is the step most likely to go red here, and it fails loudly rather than silently.

## Not run

Full `make` / `make test`, the SQL Server integration suite, the `test/kerberos/` stack, and the Azure smoke job. The Windows socket code in the new test has the `#ifdef` split but has never been compiled on Windows. Tracked as open items T044–T047 in `specs/068-login-routing-unification/tasks.md`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,6 +146,46 @@ jobs:
             test/cpp/test_index_kind.cpp -o /tmp/test_index_kind
           /tmp/test_index_kind
 
+      # Spec 068: login-time routing (ENVCHANGE 20) is followed on every auth
+      # path. Drives a real TdsConnection against a loopback fake TDS server on
+      # 127.0.0.1 -- no SQL Server, no Azure, no AD. It lives in THIS job rather
+      # than the build job's "Run C++ unit tests" step on purpose: that job is
+      # dispatch/schedule-gated, and a routing regression must fail the PR that
+      # causes it. tds_connection.cpp reaches nothing outside the TDS layer, so
+      # the only externals are simdutf (via encoding/utf16.cpp) and OpenSSL
+      # (via tds_socket.hpp -> the TLS TUs; the test itself runs unencrypted).
+      - name: Install simdutf (spec 068 routing test)
+        run: |
+          if [ "${{ matrix.os.id }}" = "macos" ]; then
+            brew install simdutf
+          else
+            sudo apt-get update
+            sudo apt-get install -y libsimdutf-dev libssl-dev
+          fi
+
+      - name: Build + run login routing hop-driver test (spec 068)
+        run: |
+          set -euo pipefail
+          if [ "${{ matrix.os.id }}" = "macos" ]; then
+            SIMDUTF_PREFIX="$(brew --prefix simdutf)"
+            OPENSSL_PREFIX="$(brew --prefix openssl@3)"
+            EXTRA_INC="-I $SIMDUTF_PREFIX/include -I $OPENSSL_PREFIX/include"
+            EXTRA_LIB="-L $SIMDUTF_PREFIX/lib -L $OPENSSL_PREFIX/lib"
+          else
+            EXTRA_INC=""
+            EXTRA_LIB=""
+          fi
+          c++ -std=c++17 -pthread -Wno-deprecated-declarations \
+            -I src/include $EXTRA_INC \
+            test/cpp/test_login_routing_hops.cpp \
+            src/tds/tds_connection.cpp src/tds/tds_socket.cpp \
+            src/tds/tls/tds_tls_context.cpp src/tds/tls/tds_tls_impl.cpp \
+            src/tds/tds_packet.cpp src/tds/tds_protocol.cpp src/tds/tds_types.cpp \
+            src/tds/encoding/utf16.cpp \
+            $EXTRA_LIB -lsimdutf -lssl -lcrypto \
+            -o /tmp/test_login_routing_hops
+          /tmp/test_login_routing_hops
+
       # Spec 063 D1: who supplies a bulk-load writer's connection and how many
       # writers there may be. Every case it decides fails SILENTLY when wrong —
       # a writer given the pinned connection interleaves ROW tokens into someone
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,21 +171,25 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          SIMDUTF_VER=$(python3 -c "import json;print([o['version'] for o in json.load(open('vcpkg.json'))['overrides'] if o['name']=='simdutf'][0])")
-          # Update this hash when the vcpkg.json override moves; the mismatch
-          # message below is the reminder.
-          EXPECTED_VER=6.1.1
-          EXPECTED_SHA=28f4b565330d25801f160c97970ef06f6b861202286e9a3fef35b0d7320bbf3f
+          # shellcheck source=scripts/ci/simdutf_pin.env
+          . scripts/ci/simdutf_pin.env
+          # `next(..., '')` rather than `[0]`: vcpkg.json's own comments
+          # contemplate dropping the override once the baseline catches up, and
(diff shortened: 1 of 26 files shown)


VGSML:
Reviewed the full diff and verified locally on macOS arm64: the hop-driver test (79/79), the parser pins (56/56), a clean release build, and the SQL Server integration suite against the docker fixture — 158 cases, green. Together with the PR's own CI (Kerberos e2e, fuzzing, both unit-test legs) that covers everything except live AD+routing, which the spec already concedes doesn't exist in our test surface.

This is a faithful implementation of spec 068 that improves on it in several places:

- `LoginAttemptOutcome` instead of the spec's implied bool-plus-side-channel — the side channel *was* the original bug, so making it unrepresentable is the right call.
- The driver owning every state transition, with the per-attempt reset covering the **first** attempt — which found and pinned the latent reused-connection packet-id bug nobody asked about.
- `connect_timeout_seconds_` carried into hops (another latent find), pinned with the TEST-NET-1 test and honestly scoped in the header comment (dial only, not the handshake reads).
- The R7 ServerName fix with the strongest pin in the suite: extracting ServerName from the actual LOGIN7 bytes on both the gateway and the routed target, with the ENVCHANGE port deliberately wrong so the suffix rule is load-bearing.
- `Free()` before each hop's rebuild, pinned by `frees_before_second_build` — the kind of leak that would have survived years unnoticed.
- The simdutf amalgamation pin (version cross-checked against vcpkg.json + SHA-256, loud drift errors on all three platforms) is nicer supply-chain hygiene than most of our existing fetches.

Two cosmetic notes, neither blocking:

1. `original_sni_hostname_` in `tds_connection.hpp` is a dead member — it was dead on main too, but since this PR rewrote everything around it, dropping it here would be natural. Fine as a follow-up.
2. The `AuthenticatorFactory` doc says "must capture by value only", and the validation-path factory in `mssql_storage.cpp` captures `&info`/`&strategy_error` by reference. It's safe (strictly synchronous, same scope), but the two read as contradicting each other — worth either a softer wording in the header ("by value unless the call is strictly synchronous") or a one-line comment at the capture site saying why it's exempt.

I also checked the edge the spec didn't mention — a malformed ROUTING ENVCHANGE with an empty server name — and the parser already refuses to set `has_routing` for it, so the driver can't hop to an empty target. No action needed; noting it so the next reader doesn't re-derive it.

VGSML (dismissed):

VGSML (approved):

PR #263, opened 2026-08-11
fix: SPN from IP, SSPI coverage, a Kerberos routing hop, and error translation (#259 #260 #261 #262)

### LINKED ISSUE #259: Kerberos: SPN derived from an IP address can never match AD (no reverse DNS)
## What

`AuthStrategyFactory::DeriveSpn` (`src/tds/auth/auth_strategy_factory.cpp`) builds the default SPN by concatenating the connection-string host verbatim:

```cpp
return std::string("MSSQLSvc/") + info.host + ":" + std::to_string(info.port);
```

When a user connects by IP — `Server=10.1.2.3,1433;Trusted_Connection=yes` — that yields `MSSQLSvc/10.1.2.3:1433`. Active Directory registers SQL Server SPNs against the **FQDN**, so this can never match, and the login fails with a KDC "server not found in Kerberos database" that gives no hint the address form is the problem.

## Why now

Noticed while comparing our tests against Microsoft's own Rust TDS driver, [`microsoft/mssql-rs`](https://github.com/microsoft/mssql-rs). Its `mssql-tds/tests/test_kerberos_gssapi.rs` has a dedicated case — `test_kerberos_connection_via_ip_address` — that pins a **reverse-DNS lookup resolving the IP to an FQDN** before SPN construction. So the reference implementation treats this as a supported path; we don't.

## Suggested fix

Reverse-resolve `info.host` (`getnameinfo`) when it parses as a literal IPv4/IPv6 address, and build the SPN from the resulting FQDN. Keep the existing behaviour for names. An explicit `service_principal_name=` must still win verbatim — that part is already correct.

Worth erroring clearly rather than silently, if reverse resolution fails: "cannot derive an SPN from an IP address; pass service_principal_name=MSSQLSvc/<fqdn>:<port>" is a one-round diagnosis.

## Scope

Affects the POSIX Kerberos path and, by the same derivation, the Windows SSPI path. Not a routing issue — the SPN-over-hop work in #256 / #258 uses the same `DeriveSpn`, so this would improve both.

### LINKED ISSUE #260: Windows SSPI has no CI coverage: the tests assume a domain-joined host, but local NTLM does not need one
## What

Spec 042 Phase 4 shipped `WinSspiAuthenticator`, and there is a smoke test — `test/sql/integrated_auth/winsspi_basic.test`. It never runs in CI:

```
# description: ... Skipped unless MSSQL_WINSSPI_TEST=1 because CI runners
#              aren't domain-joined; runs manually on a real Windows host.
require-env MSSQL_WINSSPI_TEST
```

and `MSSQL_WINSSPI_TEST` is on the opt-in allowlist in `scripts/ci/check_require_env.sh`, i.e. deliberately never set. So the SSPI code path has **zero automated coverage** — the Windows CI jobs compile it and nothing exercises it.

The premise ("CI runners aren't domain-joined") is true but stronger than it needs to be.

## Why now

Microsoft's own Rust TDS driver, [`microsoft/mssql-rs`](https://github.com/microsoft/mssql-rs), gets real SSPI coverage on an undomained Windows box. `mssql-tds/tests/test_windows_sspi.rs` contains tests that need no AD at all:

- `test_sspi_localhost_select_one` — connect to `tcp:localhost,1433` with integrated auth and run `SELECT 1`; local **NTLM** satisfies SSPI without a domain
- `test_localdb_integrated_auth_connection` — `(localdb)\MSSQLLocalDB`, Windows auth over named pipes
- `test_windows_sspi_context_creation` — asserts the derived SPN is exactly `MSSQLSvc/localhost:1433` and that the context is not yet complete
- `test_windows_sspi_generate_initial_token` — asserts the initial token is non-empty

The last two need **no SQL Server at all** — they exercise `AcquireCredentialsHandle` / `InitializeSecurityContext` against the local logon session. Those are pure unit tests we could run on any `windows-latest` runner today.

## Suggested work, cheapest first

1. **No server, no domain**: a standalone test that constructs `WinSspiAuthenticator` for `localhost:1433`, asserts the derived SPN, and asserts `InitialBytes()` returns a non-empty SPNEGO blob. Runs on any Windows runner. (`mssql_winsspi_auth_test()` already does most of this as a SQL function — it needs to become an assertion, not a diagnostic.)
2. **Local SQL Server, no domain**: install SQL Server Express or LocalDB on the Windows runner and run the existing `winsspi_basic.test` against `localhost` over local NTLM, dropping the domain assumption from its `require-env`.

Step 1 alone turns "compiles" into "the SSPI handshake starts correctly", which is where the platform-specific bugs live.

## Related

Windows CI is dispatch-gated and MSVC is currently red (#165), so wiring this needs to account for the job actually running.

### LINKED ISSUE #261: test/kerberos cannot exercise a routing hop: add a gateway so SPN-over-hop is verified, not assumed
## What

Spec 068 (#256, PR #258) makes integrated auth follow a login-time ROUTING redirect, obtaining a **fresh service ticket for the routed host's SPN** rather than replaying the gateway's. That design is the spec's own #1 acknowledged risk, shipped unverified:

> No AD environment with a routing front-end exists in our test surface; the "fresh authenticator per hop" design mirrors go-mssqldb but ships unverified against real AD + MI.

Unit tests cover it with a stub authenticator (the factory is re-invoked with the routed host/port, and the previous context is freed). What no test covers is the part that can actually be wrong: whether a **real KDC** issues a usable ticket for the routed host and whether SQL Server accepts it.

## Why this is now tractable

`test/kerberos/` already stands up a KDC **we control** (MIT krb5 on ubuntu:24.04, realm `EXAMPLE.COM`) plus a domain-joined SQL Server. Adding a routing hop needs only:

1. a small **gateway** container that speaks enough TDS to answer PRELOGIN and reply to LOGIN7 with an ENVCHANGE type 20 pointing at `sql.example.com:1433` — the fake server in `test/cpp/test_login_routing_hops.cpp` is exactly this and could be reused or ported;
2. a second SPN registered in the KDC for the gateway host (`MSSQLSvc/gateway.example.com:1433`), which is one `kadmin` line since the KDC is ours;
3. a test that ATTACHes to the **gateway** and asserts the query succeeds against the routed server.

That exercises the real sequence: ticket for the gateway's SPN → ROUTING → **new** ticket for `MSSQLSvc/sql.example.com:1433` → LOGINACK. If the per-hop authenticator design is wrong, this fails; today nothing would.

## Also worth taking from mssql-rs

Comparing with [`microsoft/mssql-rs`](https://github.com/microsoft/mssql-rs):

- Their `kerberos-test/` uses a **Samba AD DC** rather than an MIT KDC, i.e. a real Active Directory with LDAP, SSSD-joined SQL Server, and generated credentials. Closer to production than our MIT KDC, and it would let us test AD-specific SPN registration behaviour. Bigger lift than the gateway above; worth considering separately.
- `test_kerberos_gssapi.rs` asserts the GSSAPI token's first byte is `0x60` (ASN.1 APPLICATION tag), confirming a properly wrapped SPNEGO blob rather than merely "non-empty". Cheap pin we could add to `mssql_kerberos_auth_test` / the krb5 unit tests.

**Not** worth copying: their auth tests skip via env var and *return success* when the infrastructure is absent. That is the false-green this repo has already been bitten by (#192, #212); `scripts/ci/check_require_env.sh` exists precisely to prevent it.

### LINKED ISSUE #262: Every login-time server error is reported as "check username and password" — including a resuming Azure serverless database
## What

`TranslateConnectionError` (`src/mssql_storage.cpp:1005`) decides an error is a credential problem by substring:

```cpp
if (lower_error.find("login failed") != string::npos || lower_error.find("authentication") != string::npos ||
    lower_error.find("18456") != string::npos) {
    return StringUtil::Format("Authentication failed for user '%s' - check username and password", user);
}
```

But the string it inspects is **our own wrapper**, not the server's message. `DoLogin7` (`src/tds/tds_connection.cpp:1038`) formats every rejected login as:

```
Authentication failed (error <N>, state <S>): <server message>
```

So the `"authentication"` arm matches unconditionally on **any** login-time error, whatever the server actually said, and the number, the state and the server's own text are all discarded before the user sees anything.

## Observed

Azure SQL error **40613** — *"Database '…' on server '…' is not currently available. Please retry the connection later"*, i.e. a paused serverless database resuming, which `test/sql/azure/azure_lazy_loading.test` explicitly warns about in its own header — surfaces as:

```
MSSQL connection validation failed: Authentication failed for user 'azure_sql_admin' - check username and password
```

Seen in CI runs [31527078032](https://github.com/hugr-lab/mssql-extension/actions/runs/31527078032) and [31531869120](https://github.com/hugr-lab/mssql-extension/actions/runs/31531869120) (both `azure_lazy_loading.test:29`), then passing in [31531723831](https://github.com/hugr-lab/mssql-extension/actions/runs/31531723831) once the database was awake. The credentials were correct the whole time. Nothing in the tree handles 40613.

That message sends the reader to rotate a secret that was never wrong. It also makes the Azure CI job look broken rather than flaky.

## It also discards the #164 work

Issue #164 added the ERROR-token **State** byte precisely because 18456 has many distinct causes — bad password vs. inaccessible initial database vs. disabled login — and the State is the only field that separates them. `DoLogin7` now carries it. On every ATTACH path (`mssql_storage.cpp:1234`, `:1245`, `:1261`, `:1280`, `:1314`) `TranslateConnectionError` throws it away again and prints "check username and password".

## Suggested fix

1. Match on the **server's** error number, not on our own prose. `LoginResponse::error_number` / `error_state` are already parsed; plumb them out rather than re-deriving intent from a formatted string.
2. Keep the friendly hint but never lose the original: append the server text and number, e.g. `Authentication failed for user 'x' (server error 18456, state 40): <server message>`.
3. Special-case the retryable ones — 40613 and 40197 deserve "the database is resuming or being moved; retry shortly", not a credential hint.

Found while verifying an unrelated Azure CI failure during spec 068 (#256 / #258); not caused by it.

oluies:
Closes #259. Closes #260. Closes #261. Closes #262.

**Stacked on #258** (base is `task/spec-068-4f420d`, not `main`) because #261's hop test only means anything with spec 068's routing code present. Retarget to `main` once #258 lands.

All four were found while verifying spec 068; none is caused by it.

## #262 — every login-time server error read as "check username and password"

`TranslateConnectionError` decided "credential problem" from the substring `"authentication"` — but the string it inspects is **our own wrapper**: `DoLogin7` formats every rejection as `Authentication failed (error N, state S): …`. The arm matched unconditionally, and the number, the state and the server's own words were discarded.

Consequences seen in this repo's CI twice today: Azure **40613** (a paused serverless database resuming, which `azure_lazy_loading.test` warns about in its own header) told the reader to rotate a secret that was never wrong. It also silently undid **#164**, which added the State byte precisely because 18456 has several distinct causes.

Classification now comes from `LoginResponse::error_number`, plumbed out via `GetLastErrorNumber()`/`GetLastErrorState()`. 40613/40197/40501/49918 are named retryable; 18456 explains what its state means; the server's text is always appended. The one surviving string arm matches the **server's** phrase (`"login failed"`), never our own word.

## #259 — an SPN built from an IP can never match

AD registers SPNs against the FQDN, so `Server=10.1.2.3,1433` produced `MSSQLSvc/10.1.2.3:1433` and a KDC failure that says nothing about the address form being the cause. IP literals (v4 and v6) are now reverse-resolved. Names pass through **untouched and unresolved** — looking them up could follow a CNAME and change an SPN that was already correct. An IP with no PTR record raises an error naming the fix rather than building something unmatchable.

## #260 — Windows SSPI had no automated coverage

Its only test assumes a domain-joined host and sits on the never-set allowlist in `check_require_env.sh`. But acquiring a Negotiate credential and producing the first token needs the **local logon session**, not a domain and not a SQL Server — so that now runs on a bare Windows runner in the MSVC job. Accepts a GSS-API-framed (`0x60`) or raw `NTLMSSP` token, since Negotiate legitimately picks NTLM on an undomained box.

## #261 — `test/kerberos` can now exercise a hop

Spec 068 D3 (a fresh ticket for the **routed** host's SPN) was verified only against a stub authenticator; nothing proved a real KDC would issue the second ticket. A `gateway` container now answers every login with `ROUTING -> sql.example.com`, and the KDC registers `MSSQLSvc/gateway.example.com:1433` so the client must obtain a ticket to reach it at all.

The assertion is on the **ccache**: two service tickets, the gateway's and the routed host's. One ticket means the hop reused the gateway's credential and the design is wrong. The final login is still expected to be rejected — SQL Server on Linux cannot map the principal without SSSD — exactly as the existing step 4 documents.

## Verified here

- 33 new checks (error translation) + 11 (SPN resolution), both in `make test-cpp`, whole suite green
- 79 routing checks, full unit suite (**5086 assertions / 158 cases**)
- **The Python gateway driven end-to-end by a real `TdsConnection`** on loopback: it routed the client to a second listener, which completed the login. The wire format — the error-prone half of #261 — is proven, not assumed
- `docker compose config` accepts the new service; `bash -n` on both scripts

## NOT verified here

- The SSPI test has never run: no Windows available. It will first execute on the next Windows dispatch
- The Kerberos hop's ticket assertions need the docker stack, which builds the extension inside Linux and takes far longer than one session. The protocol half is proven above; the Kerberos half is not

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -908,6 +908,27 @@ jobs:
       # package location; here the prefix is already built above.
       # Consequence: Windows coverage of this test is dispatch-gated, unlike
       # Linux/macOS which run it on every PR.
+      # Issue #260: the SSPI path had no automated coverage at all -- its only
+      # test assumes a domain-joined host and is therefore never run. Acquiring
+      # a Negotiate credential and producing the first SPNEGO token needs the
+      # local logon session, not a domain and not a SQL Server, so it runs here
+      # on a bare runner. secur32 is the SSPI library; the rest is the TDS layer
+      # the authenticator sits in.
+      - name: Build + run Windows SSPI context test (issue #260)
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          $prefix = "build/release/vcpkg_installed/x64-windows-static-release"
+          cl /std:c++17 /EHsc /nologo /MT /DMSSQL_ENABLE_SSPI=1 `
+            /I src/include /I "$prefix/include" `
+            test/cpp/test_winsspi_context.cpp `
+            src/tds/auth/winsspi_authenticator.cpp `
+            /Fe:test_winsspi_context.exe `
+            /link secur32.lib
+          if ($LASTEXITCODE -ne 0) { throw "compile failed" }
+          ./test_winsspi_context.exe
+          if ($LASTEXITCODE -ne 0) { throw "test failed" }
+
       - name: Build + run login routing hop-driver test (spec 068)
         shell: pwsh
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,6 +230,48 @@ jobs:
           /tmp/test_load_policy
 
   # ============================================================================
+  # Windows SSPI Unit Test - ungated, like cpp-unit-tests above.
+  #
+  # Issue #260: WinSspiAuthenticator had no automated coverage. Its only test
+  # assumes a domain-joined host and sits on the never-set allowlist in
+  # check_require_env.sh, and the Windows BUILD jobs are dispatch-gated -- so
+  # putting it there would trade a test nobody runs for a test that runs when
+  # somebody ticks a box.
+  #
+  # It belongs in its own job rather than as a windows entry in the
+  # cpp-unit-tests matrix because five of that job's steps are POSIX-only
+  # (pkg-config, readelf, shasum, `c++`) and would each need an `if:` guard.
+  #
+  # No vcpkg needed: winsspi_authenticator.cpp includes only its own header and
+  # the Windows SDK. Acquiring a Negotiate credential and producing the first
+  # token uses the local logon session -- no domain, no SQL Server, no network.
+  # ============================================================================
+  windows-sspi-test:
+    name: Windows SSPI Unit Test
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Setup Visual Studio Environment
+        uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: x64
+
+      - name: Build + run Windows SSPI context test (issue #260)
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          cl /std:c++17 /EHsc /nologo /MT /DMSSQL_ENABLE_SSPI=1 `
+            /I src/include `
+            test/cpp/test_winsspi_context.cpp `
+            src/tds/auth/winsspi_authenticator.cpp `
+            /Fe:test_winsspi_context.exe `
+            /link secur32.lib
+          if ($LASTEXITCODE -ne 0) { throw "compile failed" }
+          ./test_winsspi_context.exe
+          if ($LASTEXITCODE -ne 0) { throw "test failed" }
+
+  # ============================================================================
   # Build Job - Build against DuckDB stable on representative platforms
   # (Manual trigger only via workflow_dispatch)
   # ============================================================================
@@ -908,27 +950,6 @@ jobs:
       # package location; here the prefix is already built above.
       # Consequence: Windows coverage of this test is dispatch-gated, unlike
       # Linux/macOS which run it on every PR.
-      # Issue #260: the SSPI path had no automated coverage at all -- its only
-      # test assumes a domain-joined host and is therefore never run. Acquiring
-      # a Negotiate credential and producing the first SPNEGO token needs the
-      # local logon session, not a domain and not a SQL Server, so it runs here
-      # on a bare runner. secur32 is the SSPI library; the rest is the TDS layer
-      # the authenticator sits in.
-      - name: Build + run Windows SSPI context test (issue #260)
-        shell: pwsh
-        run: |
-          $ErrorActionPreference = "Stop"
-          $prefix = "build/release/vcpkg_installed/x64-windows-static-release"
-          cl /std:c++17 /EHsc /nologo /MT /DMSSQL_ENABLE_SSPI=1 `
-            /I src/include /I "$prefix/include" `
-            test/cpp/test_winsspi_context.cpp `
-            src/tds/auth/winsspi_authenticator.cpp `
-            /Fe:test_winsspi_context.exe `
-            /link secur32.lib
-          if ($LASTEXITCODE -ne 0) { throw "compile failed" }
-          ./test_winsspi_context.exe
-          if ($LASTEXITCODE -ne 0) { throw "test failed" }
-
       - name: Build + run login routing hop-driver test (spec 068)
         shell: pwsh
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,7 @@ docs/assets/download-metrics.svg
 website/node_modules/
 website/build/
 website/.docusaurus/
+
+# Python bytecode (e.g. test/kerberos/gateway helpers imported locally)
+__pycache__/
+*.pyc
--- a/Makefile
+++ b/Makefile
@@ -425,6 +425,8 @@ STANDALONE_TEST_SOURCES := \
     test/cpp/test_ddl_translator.cpp \
     test/cpp/test_ctas_type_mapping.cpp \
     test/cpp/test_catalog_filter.cpp \
+    test/cpp/test_connection_error_translation.cpp \
+    test/cpp/test_spn_host_resolution.cpp \
     test/cpp/test_insert_bulk_sql.cpp \
     test/cpp/test_vector_encodings.cpp \
     test/cpp/codec/test_binary_codec.cpp \
--- a/src/include/mssql_storage.hpp
+++ b/src/include/mssql_storage.hpp
@@ -240,10 +240,14 @@ struct MSSQLConnectionInfo {
 // `server_error` / `server_state` are the SERVER's ERROR-token fields (0 when
 // the failure never reached a login). When a number is present it classifies
 // the failure; otherwise a few string heuristics cover socket/TLS/DNS errors.
+// MSSQL-prefixed per CLAUDE.md's namespace rule: an exported free function in
+// the shared `duckdb` namespace needs the prefix, and this name is generic
+// enough to collide.
+//
 // Declared here rather than kept static so the classification can be unit
 // tested -- issue #262 was a misclassification that no test could have caught.
-string TranslateConnectionError(const string &error, const string &host, uint16_t port, const string &user,
-								const string &database, uint32_t server_error = 0, uint8_t server_state = 0);
+string MSSQLTranslateConnectionError(const string &error, const string &host, uint16_t port, const string &user,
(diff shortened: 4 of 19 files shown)


VGSML (dismissed):
Reviewed all four fixes and verified locally on macOS arm64: release build clean, `make test-cpp` green including the two new suites (error translation, SPN resolution), and the SQL Server integration suite against the docker fixture — 158 cases green, which matters here because the translation rewrite changes user-facing ATTACH error text and `attach_validates_credentials.test` pins a prefix of it (still matches — verified empirically, not just by reading).

**The headline first**: your "NOT verified" section is already out of date, in the good direction. The Kerberos Integration run on this PR built the gateway container and ran the new step 6 — the CI log says `step 6: OK -- the hop acquired a separate ticket for sql.example.com (spec 068 D3 verified)`. A real MIT KDC issued the second service ticket for the routed host. The one acknowledged risk spec 068 shipped with is now closed by CI evidence, not just by the loopback protocol check. That alone justifies the PR.

Per fix:

- **#262** is the right architecture, not just a patched regex: classification moved to the server's own error number (`GetLastErrorNumber()`/`GetLastErrorState()`, reset in the driver's per-attempt block so a hop can't leak a stale number), the one surviving string arm matches the *server's* phrase, and the server text is always appended so no diagnosis is lost. The 18456 message explaining state 1 as "the server logs but does not disclose" is exactly right — that masked state is what a non-Azure SQL Server usually shows a client.
- **#259**: `NI_NAMEREQD` (refusing to echo the numeric form back) is the load-bearing flag and it's there; names pass through unresolved (no CNAME risk); memoization is correctly justified by the negative-result cost on the pool-refill path; and unifying `DeriveDefaultSpn` with the diagnostic functions kills a real divergence — the diagnostics used to report an SPN the connection path wouldn't request. The throw from `DeriveDefaultSpn` is contained: both #258 factory lambdas wrap `AuthStrategyFactory::Create` in try/catch, so the no-PTR error surfaces as a clean login failure, not an exception across the TDS layer.
- **#261**: gateway.py's ROUTING ENVCHANGE is byte-correct against our parser (checked field by field), reassembles fragmented LOGIN7 (#138), and emits the DONE in the TDS 7.2+ 13-byte layout — someone learned #254's lesson. The KDC principal creation being idempotent across both the fresh-realm and existing-realm branches is the kind of detail that saves an hour of debugging a stale stack. The ccache two-tickets assertion is the correct oracle, and the Encrypt=no-propagates-across-the-hop comment in docker-compose will save the next person a false TLS chase.
- **#260**: right call putting it in an ungated job — the alternative really was a test nobody runs. Accepting either GSS-framed (0x60) or raw NTLMSSP is correct for an undomained runner, and the double-`Free()` idempotency check is a nice touch.

**The one thing to get right at merge time**: `ci.yml` triggers on `pull_request: branches: [main]`, and this PR's base is the #258 branch — so the CI workflow (C++ unit tests with the two new suites, Lint, and the new `windows-sspi-test` job) has **not run on this PR at all**. The green checks you see are the workflows without a base filter. After #258 merges and this retargets to `main`, wait for that first full CI run before merging — it's the first time the SSPI test executes anywhere, and you flagged that yourself. Everything POSIX-side I could run locally is green, so I expect no surprises except possibly the Windows leg.

Approve, conditional only on that post-retarget CI run being green.

VGSML (dismissed):

VGSML (approved):

PR #264, opened 2026-08-13
test: fix the two azure tests that require-azure skips had hidden

VGSML:
The `[azure]` group's `require azure` gate was unsatisfiable in every local and CI environment — the sqllogictest runner only loads out-of-repo extensions via `LOCAL_EXTENSION_REPO` + `--autoloading`, and nothing ever provisioned the azure extension that way. Every subtest behind the gate has been silently skipping since the files were written.

Running the whole group for real for the first time (recipe below; **168 assertions / 16 cases / 0 skips** against live Azure SQL after fixing) exposed two lines of never-executed WIP from commit b8677ef:

- **`azure_service_principal.test`**: `azuredb.mssql_scan(azuredb, ...)` — a catalog-qualified table-function lookup this extension has never registered, with a bare identifier where a VARCHAR belongs. Replaced with the documented form `mssql_scan('azuredb', ...)`, same as every green query test.
- **`azure_device_code.test`**: a `statement ok` pinning a `TENANT_ID` parameter on the azure extension's **own** `CREATE SECRET` — upstream surface, none of our code, and today's azure extension rejects the parameter. Removed rather than updated: pinning a third-party parameter list buys no coverage of this extension and breaks on upstream changes.

## Running the group locally (the recipe the skips were hiding)

```bash
# map .env, acquire AZURE_ACCESS_TOKEN via client credentials, warm the
# free-tier serverless DB (first touch gets 40613 while it resumes), then:
cp ~/.duckdb/extensions/v1.5.5/osx_arm64/azure.duckdb_extension build/release/repository/v1.5.5/osx_arm64/
HOME=<scratch> LOCAL_EXTENSION_REPO=$PWD/build/release/repository \
  ./build/release/test/unittest "[azure]" --autoloading available --force-reload
```

The isolated `HOME` matters: extensions already installed in `~/.duckdb` from another origin (community/core) make the runner's `INSTALL ... FROM <repo>` fail on the origin conflict, which surfaces as a silent skip.

Follow-up candidates (not in this PR): a `make azure-test` target encapsulating the recipe, and the same provisioning for the CI Azure job so its runs stop silently skipping the `require azure` half.


oluies:
## Code review (max effort) — 15 verified findings

Reviewed the two-file diff, then swept the surrounding lane because the PR's premise is "this group has never run". The premise is right; the fix is incomplete, and the sweep found a live product defect plus two independent holes in the guard that was written to catch exactly this class.

Every claim below was verified against the tree (and, where noted, against a live `duckdb v1.5.5` + `azure`).

---

### 1. The revived test still cannot run — `require-env` names a variable nothing sets

`test/sql/azure/azure_service_principal.test:15`

`.env.example:25-26`, `docs/TESTING.md:150-151` and `.github/workflows/ci.yml:1216-1217` all use **`AZURE_SQL_DB_HOST` / `AZURE_SQL_DB`**. The test gates on **`AZURE_SQL_HOST` / `AZURE_SQL_DATABASE`** — `git grep AZURE_SQL_HOST` finds it only inside `test/sql/azure`. `require-env` at line 15 is reached *before* `require azure` at line 25, so the file skips and exits 0 regardless of the azure extension.

`Makefile:100-109` shows this exact "nothing has ever set it" class already found and fixed twice (`MSSQL_TEST_SERVER`, `MSSQL_TEST_CONNECTION_STRING`) — the `AZURE_` case was sitting in the same file. Also strands `azure_access_token.test:15` and `azure_env_provider.test:25`.

### 2. The deleted `TENANT_ID` subtest was the canary for a live product defect

`test/sql/azure/azure_device_code.test:27`

The deletion rationale — "upstream surface, none of our code" — is factually wrong. Verified live on `duckdb v1.5.5` + `azure`:

```
CREATE SECRET (TYPE azure, PROVIDER credential_chain, CHAIN 'interactive', TENANT_ID 'common')
-> Binder Error: Unknown parameter 'tenant_id' ... provider 'credential_chain'
```

Meanwhile, in this extension's own code:

- `src/azure/azure_secret_reader.cpp:55-58` reads `tenant_id` off any azure secret
- `src/azure/azure_token.cpp:455` routes `credential_chain` + `interactive` to `AcquireInteractiveToken(info)`
- `src/azure/azure_device_code.cpp:222` passes `info.tenant_id` to `RequestDeviceCode`, defaulting to `AZURE_DEFAULT_TENANT = "common"` (`azure_token.hpp:57`)

Both ATTACH call sites (`mssql_storage.cpp:1133`, `azure_fedauth.cpp:33`) pass no override, and the plumbing that was clearly meant to supply one — `azure_tenant_id`, registered at `mssql_secret.cpp:252` with the comment *"Optional, tenant for interactive auth"* and stored at `:170` — **is never read anywhere in `src/`**.

Net effect: every org user doing interactive ATTACH is silently sent to `/common/` with no way to set a tenant. The deleted assertion was the only thing in the tree pointing at it.

### 3. Second hole in the anti-skip guard: a bare assignment satisfies it

`scripts/ci/check_require_env.sh:47`

```sh
grep -qE "^(export )?${v}[[:space:]]*[:?]?=" Makefile
```

`MSSQL_TEST_DSN_TLS` is *assigned* at `Makefile:96` but its export is deliberately commented out at `Makefile:132`. The grep matches; the variable never reaches the child process. So `tls_connection.test:17`, `tls_queries.test:16`, `tls_parallel.test:16` and `tls_multipacket.test:12` are **skipping right now** under `make integration-test`, still counting toward `MSSQL_MIN_TEST_CASES`, while the guard (ungated on every PR via `ci.yml:83`) prints `OK`.

### 4. The added comment blames the wrong gate

`test/sql/azure/azure_service_principal.test:55`

Gate order is `require mssql` (12) → `require-env AZURE_SQL_HOST` (15) → … → `require azure` (25). In CI and in any environment built from `.env.example` the runner stops at line 15. A maintainer who provisions the azure extension — the fix the comment implies — will find the file still skipping, with a committed comment asserting that was the only obstacle.

### 5. `COUNT(*) >= 0` cannot fail

`test/sql/azure/azure_service_principal.test:64`

`MSSQLPoolStatsFunction::InitGlobal` (`mssql_diagnostic.cpp:288-296`) swallows the unknown-context case into zero rows (`catch (...) { /* Not attached */ }`) — `test/sql/tds_connection/pool_stats.test:9` asserts exactly that. `COUNT(*)` over 0 rows is 0, and `0 >= 0` is true. Replace `'azuredb'` with `'typo'` and it still passes. This is one of only two assertions in the file the PR claims to have validated end-to-end. In-repo correct idiom: `tls_connection.test:63`, `SELECT connections_created >= 1 FROM mssql_pool_stats(...)`.

### 6. A late `require azure` retroactively discards four credential-free assertions

`test/sql/azure/azure_secret_validation.test:70`

Lines 8-61 are `statement error` assertions on **this extension's** `src/mssql_secret.cpp` validation — no credentials, no azure extension needed. A `require` failing mid-file aborts and marks the *whole* file skipped, so in every environment predating this PR's premise those four reported as skipped rather than passed. A regression in `MSSQLSecret` validation is currently invisible in every lane. (`azure_auth_test_function.test` has the same shape.)

### 7. Fabric has never been tested in any lane

`.github/workflows/ci.yml:1229`

Lines 1178-1180 title the job *"Azure SQL Database and Fabric tests"*; line 1229 runs `./test/unittest "[azure]"`. Catch tags derive from the parent directory, so `test/sql/fabric/fabric_types.test` is tagged `[fabric]` and is never selected. The env block (1212-1217) never sets `AZURE_WH_HOST` / `AZURE_WH_NAME` / `AZURE_WH_TOKEN`, which `fabric_types.test:21-25` requires. `make integration-test`'s path glob does reach the file, but those vars are unset there too and `AZURE_` is on `check_require_env.sh`'s `OPT_IN` list, so the guard stays silent.

### 8. The CI job cannot satisfy `require azure` either

`.github/workflows/ci.yml:1229`

The PR body names `--autoloading available` + `LOCAL_EXTENSION_REPO` as exactly what makes the gate satisfiable, then changes no CI. The job still runs plain `./test/unittest "[azure]" --force-reload`, so both edited files stay 100% skipped there while the job reports success. It is additionally `workflow_dispatch`-only and gated `github.ref != 'refs/heads/main'` (line 1187). After merge the two-line fix is unverifiable by CI and the next edit can re-break it silently.

### 9. The guard exempts `AZURE_`, so it certified this PR's rot as healthy

`scripts/ci/check_require_env.sh:34`

`OPT_IN='^(AZURE_|...)'` makes `AZURE_SQL_HOST` bypass the Makefile check entirely. The script's own header says spec 057 found this class three times in one spec and *"the fourth will look exactly like the first three."* This is the fourth, and the guard printed `OK` on it. It also scans only `require-env`, never `require <ext>`, so the azure-extension gate is outside its reach.

### 10. `make integration-test` runs DDL against a live Azure database

`test/sql/azure/azure_lazy_loading.test:30`

`Makefile:166` runs `unittest "test/sql/*"`, which recurses into `test/sql/azure/`. `azure_lazy_loading.test` gates only on `require-env AZURE_SQL_TEST_DSN` (line 11) — documented at `docs/TESTING.md:149` as the normal way to run it. Lines 57-159 then `DROP TABLE IF EXISTS` / `CREATE TABLE` / `INSERT` / `ALTER TABLE ADD` / `UPDATE` / `mssql_exec ALTER` in `dbo`. A developer who exported the documented variable gets silent create/drop against their Azure database from a target whose banner describes local Docker, with orphaned `dbo.azure_lazy_test` / `azure_lazy_test2` on a mid-file abort.

### 11. Renaming the variables is necessary but not sufficient — and the naive follow-up turns a live test red

`Makefile:118`

`-include .env` (line 84) creates *make* variables only; unexported make variables never reach the child process, and lines 119-129 export eleven `MSSQL_*` and zero `AZURE_*`. So renaming `AZURE_SQL_HOST` everywhere still leaves `make integration-test` skipping the file forever.

And adding `export AZURE_APP_ID` as-is is unsafe: `.env.example` ships `AZURE_APP_ID` / `AZURE_DIRECTORY_ID` / `AZURE_APP_SECRET` **uncommented with placeholder values** (`your-app-id`), so `azure_auth_test_function.test:47-55` would go RED with a live OAuth2 rejection on any machine that copied the example — unlike `AZURE_ACCESS_TOKEN`, which `.env.example:41` correctly leaves commented out.

### 12. The file still never tests what its description promises

`test/sql/azure/azure_service_principal.test:58`

The only real assertion is a raw `mssql_scan('azuredb', ...)`, which goes through `MSSQLQueryExecutor` and never touches the catalog (`mssql_functions.cpp:110` executes at bind time); line 64 is tautological (finding 5). So metadata load, schema/table discovery and three-part-name resolution over an Azure AD token are never exercised. A change to catalog metadata queries that breaks specifically against Azure SQL — spec 049's `sys.partitions` subquery, spec 060 native types, the `mssql_convert_varchar_max` CAST — passes here. `azure_lazy_loading.test` proves the shape is testable.

### 13. `azure_device_code.test` is the file that goes RED rather than skipping

`test/sql/azure/azure_device_code.test:7`

It is the one edited file with no `require-env` gate. If the runner records the extension as satisfiable without issuing `LOAD`, or autoload is configured but the install fails (network-restricted runner, or the `~/.duckdb` origin conflict the PR body describes), `CREATE SECRET test_azure_interactive (TYPE azure, ...)` at line 14 raises *"Secret type 'azure' not found"* under `statement ok` — a hard failure. `azure_secret_token_only.test:7` shares the shape. Anyone acting on findings 1 and 8 should expect this file to break first.

### 14. The deletion rationale contradicts what was kept

`test/sql/azure/azure_device_code.test:15`

`TYPE azure`, `PROVIDER credential_chain` and `CHAIN 'interactive'` on lines 15-19 are all duckdb-azure surface asserted by `statement ok` — verbatim the criterion given for deleting the other half. So the upstream fragility the comment claims to have removed is unchanged, while coverage of *this* extension went to zero. The retained secret is never consumed (no `mssql_azure_auth_test`, no ATTACH, no `AZURE_SECRET` reference), so "a sanity check of the precondition the device-code flow relies on" is unsupported. `# description: Test Device Code authentication error handling` is now false — `grep 'statement error'` returns 0. And "upstream surface, none of our code" is the claim that will lead someone to delete the live `tenant_id` read in `azure_secret_reader.cpp`.

### 15. Docs teach the rejected recipe, and the `# group:` model they teach is decorative

`docs/TESTING.md:932`

`docs/TESTING.md:932` lists `Interactive | provider='credential_chain', chain='interactive', tenant_id` as supported, and `test/manual/azure_device_code_test.md:32-38` gives it as copy-paste "Option B: Using specific tenant". Both now produce the Binder Error from finding 2; `AZURE.md:137-141` contradicts them by omitting tenant.

Separately, `docs/TESTING.md:400-409` and `:923` ("Group as `[azure]`") teach that the `# group:` header drives selection. It does not — all 8 top-level `test/sql/*.test` files carry `# group: [mssql]` while `[sql]` is what selects them, and no directory is named `mssql`. A maintainer who moves these files and keeps the header, as the docs instruct, silently drops them from the azure job, which has no `MSSQL_MIN_TEST_CASES` floor to notice.

---

**Cut for length**, all verified but lower severity: the ATTACH double-login plus missing 40613 warm-up (`azure_service_principal.test:49`); the bare `CREATE SECRET` losing the file's idempotent form (`azure_device_code.test:14`); the absent `make azure-test` target leaving the PR's evidence recipe uncommitted; and the `SELECT value` improvement not propagated to `azure_access_token.test:58` / `azure_env_provider.test:57`.

One earlier suspicion was **refuted**: `${VAR}` substitution is the repo-wide standard (118 uses of `${MSSQL_TESTDB_DSN}` alone), so these files do not diverge there.

oluies:
## Local integration run passed at `b71510a`

Ran the real `make integration-test` target (not a hand-rolled `unittest` call), so the Makefile changes in this branch were exercised end to end against a live SQL Server 2022 container:

```
All tests passed (5 skipped tests, 5174 assertions in 162 test cases)
integration-test: 162 cases ran (floor 150)
exit 0
```

The extension was rebuilt from scratch at this commit for the run (the CMake reconfigure invalidated the object cache, 224 targets), so these numbers come from the pushed head, not a stale artifact.

### What the run actually proves

| change | evidence |
|---|---|
| `export MSSQL_TEST_DSN_TLS` | all four `tls_*.test` files ran — `tls_connection`, `tls_queries`, `tls_parallel`, `tls_multipacket` — and none appears in the skip list. They had been skipping silently on every run while `check_require_env.sh` reported OK. |
| `~[azure] ~[fabric]` exclusion | zero files under `test/sql/azure` or `test/sql/fabric` ran. The cloud lane is out of the local target by construction, not by hoping the credentials are unset — which matters because `azure_lazy_loading.test` does `CREATE`/`ALTER`/`DROP` in `dbo` on whatever `AZURE_SQL_TEST_DSN` points at. |
| `MSSQL_MIN_TEST_CASES` tripwire | fired and passed: `162 cases ran (floor 150)`, with headroom after the exclusions removed 11 registered files. |

Every remaining skip is a documented opt-in, and nothing in that list is silent:

```
require spatial: 1
require-env MSSQL_COUNTERS: 1
require-env MSSQL_KERBEROS_TEST: 1
require-env MSSQL_NAMED_INSTANCE_HOST: 1
require-env MSSQL_WINSSPI_TEST: 1
```

For comparison, the same suite before the exclusions was 5174 assertions / 164 cases with 14 skips. The delta is exactly the two credential-free azure files moving to `make azure-test`, where they belong.

### What this does NOT cover

Three things remain unverified, and none of them is covered by the green checks on this PR:

1. **Nothing behind `require azure` has run anywhere.** No Azure credentials here, and the azure extension cannot be installed against this build at all — a shallow submodule clone stamps the version `v0.0.1`, so `INSTALL azure` gets a 404 from `extensions.duckdb.org`. That includes the new discriminating assertion for the `azure_tenant_id` override in `azure_secret_validation_with_credentials.test`.
2. **The CI provisioning fix is unproven.** The `Azure Test` job is `workflow_dispatch`-gated, so it does not run automatically. It needs a manual dispatch with the Azure secrets populated. The failure it fixes *was* reproduced locally (isolated `HOME`, no extension repository → `All tests were skipped (total skipped 20) / require mssql: 20`), but the success path has not been observed.
3. **CI is not compiling this branch.** `Build`, both Windows builds, `Smoke Test` and `Integration Test` all report `skipping` on PRs — issue #212. Current status at `b71510a` is 13 pass / 8 skipping / 0 fail, but the compile and the 5174-assertion suite above are local-only evidence, not something the checkmarks attest to.

VGSML:
@oluies — reviewed all three of your commits, verified everything locally and then live, and pushed four fix-ups on top. The verdict first: **the lane works end-to-end now.** Dispatch run [31961283249](https://github.com/hugr-lab/mssql-extension/actions/runs/31961283249) is the first green Azure job in this repo's history that actually ran something: **79 assertions against live Azure SQL, floor 60** — provisioning worked, the serverless DB was woken by the new warm-up (attempt 2), the access token was minted in-job, and the absent Fabric token produced a clean skip instead of a failure.

Your three commits held up under verification: the `azure_tenant_id` wiring is complete end to end (secret → FromSecret → ATTACH override → all four consumption sites → TokenCache key), and the discriminating AADSTS90002 assertion passed against live Azure AD here. The guard's three checks, the TLS revival (162 integration cases now, +4 files / 96 assertions — confirmed), the by-construction containment of the destructive test — all reproduced.

What the fix-ups address (each found by running, not reading):

1. **`make azure-test` died with Error 127 before any test** (a7d48d3): the glob-explainer comment added in 502e220 spliced into the `set -e; \` continuation — make strips `@` only at the start of the logical line, so the shell got the `@#` verbatim and the backticks around `*/*/` came alive as command substitution. Moved to column-0 make comments. The same splice also kept `set -e` from ever reaching the main block's shell.
2. **Quoted .env values broke the substitution** (a7d48d3): `-include .env` reads make syntax, not shell — `VAR='x'` keeps its quotes and lands in the .test SQL as `'x'`, splitting the string. The export block now strips one layer of quotes. 56 → 80 assertions on the same .env.
3. **The serverless DB answers its first login with 40613** (a7d48d3): both lanes now retry a throwaway ATTACH for up to two minutes before the suite.
4. **The archive-order link flake became deterministic on this branch** (805cab3): your tenant wiring shifted the symbol graph and `generated_extension_loader.o` started getting pulled reliably before the linker had any reason to keep `MssqlExtension`'s vtable — two red dispatches in a row. `-Wl,--start-group` on the Linux leg; ld64 already iterates, which is why Darwin never failed.
5. **A missing GitHub secret is a set-but-empty env var** (163a879): the secrets expression yields `''`, the env block exports it, and set-but-empty satisfies `require-env` — so the first live dispatch ran fabric_types with `ACCESS_TOKEN ''` and died with "Missing 'User Id'" instead of skipping. Empty AZURE_* now get unset before the suite (the job-side twin of your only-non-empty export), the `eval` in the echo loop is `${!v}` indirection now (the DSN contains `;`), AZURE_ACCESS_TOKEN is minted in-job from the service principal instead of stored (it expires hourly), and the floor moved 100 → 60 to fit the azure-only shape.

The one thing that kept your work honest through all of this is the assertion floor you insisted on: every failure above surfaced loudly precisely because an all-skip run can no longer report success. Merging.

oluies:
## Two follow-ups on the `azure-test` recipe after `a7d48d3`

`a7d48d3` fixes the continuation bug correctly — I hit the same thing independently and can confirm the mechanism: the `@#` lines spliced into the `@set -e; \` continuation reach the shell as a word `@#` (a `#` only opens a comment at the start of a word), so the backticks fire as command substitution. Reproduced as:

```
sh: build/release/: is a directory
sh: @#: command not found
```

and after moving the comment above the target, `make azure-test` runs to its intended guard for the first time:

```
azure-test: ~/.duckdb/extensions/v0.0.1/osx_arm64/azure.duckdb_extension not found.
```

Two things the current recipe still has that I'd flag. Both are Low — neither blocks the PR — and I have them implemented locally if either is wanted; I've deliberately not pushed over your work.

### 1. `azure-test: release` puts the build inside the exported-secret scope

The `MAKECMDGOALS` gate added in `502e220` narrows the `export` to invocations whose goal list contains `azure-test`. But `release` is a prerequisite, so it runs *within that same invocation* — meaning `make azure-test` on an unbuilt tree still hands `AZURE_APP_SECRET`, `AZURE_CLIENT_SECRET`, `AZURE_ACCESS_TOKEN` and the DSN password to cmake/ninja/vcpkg. That is exactly the "vcpkg — which writes build logs" case the comment above the gate says the fix removes.

What the gate actually achieved is that standalone `make release` is clean. `make azure-test` is not.

Two ways out:

- drop `release` from the prerequisites — the recipe already guards on a missing build and prints "Run 'make release' first", so the failure stays actionable — and document `make release && make azure-test`; or
- source `.env` inside the recipe (`set -a; [ -f .env ] && . ./.env; set +a`) in the same shell as the `unittest` call, so the values never enter make's exported set at all.

The first is what I'd pick: it keeps make as the `.env` parser (sourcing it as shell breaks on any unquoted value containing a space).

### 2. `ls -d $(AZURE_EXT_REPO)/*/*/ | head -1` can select a stale directory

Dropping the `v` is right. But `build/release/repository` is not cleaned between builds, so after a duckdb submodule bump it holds both the old and the new version directory, and `head -1` takes the lexicographically first. The failure is at least loud (the `$src not found` branch), but it names a version the current binary does not use — which reads as "install azure" rather than "stale repository directory".

The name is derivable from the binary, replicating `GetVersionDirectoryName()` exactly:

```make
lib_ver=$$(build/release/duckdb -noheader -list -c "SELECT library_version FROM pragma_version();"); \
src_id=$$(build/release/duckdb -noheader -list -c "SELECT source_id FROM pragma_version();"); \
case "$$lib_ver" in \
    *-dev*) ver="$$src_id" ;; \
    *)      ver="$$lib_ver" ;; \
esac; \
plat=$$(build/release/duckdb -noheader -list -c "PRAGMA platform;"); \
repo_dir="$(AZURE_EXT_REPO)/$$ver/$$plat"; \
```

Verified locally: yields `v0.0.1/osx_arm64`, matching the actual `build/release/repository/v0.0.1/osx_arm64/`. Note the `-dev` test matters — `source_id` alone would be wrong for a release build, where the directory is the tag.

### Two minor extras

- The ci.yml provisioning step uses the same `head -1`. Its `$scratch` is created fresh in that step so exactly one directory is expected — cheap insurance to assert that rather than take the first of however many turn up.
- `make help` and `docs/TESTING.md` still say plain `make azure-test`; they'd need the `make release &&` form if item 1 goes the prerequisite-dropping way.

Happy to open these as a separate PR, hand you the patch, or drop them — your call, since they touch code you've just reworked.

VGSML:
@oluies both confirmed — please open the PR, you have it implemented and they're your findings.

On 1: that one is on my review — I checked the `MAKECMDGOALS` gate against standalone `make release` and never against the prerequisite chain of `azure-test` itself, which is exactly the invocation the gate exists for. Your first option is the right one: drop the prerequisite, keep the recipe's own "Run 'make release' first" failure as the guidance, and update `make help` / `docs/TESTING.md` to the `make release && make azure-test` form. Sourcing `.env` as shell would trade a real parser (make) for one that breaks on the first unquoted space — agreed, not worth it.

On 2: agreed, and the `pragma_version()` derivation is strictly better than any glob — it asks the binary that will actually load the extension, and the `-dev` branch matches `GetVersionDirectoryName()` exactly. Take the ci.yml `head -1` assert along too; a fresh `$scratch` should contain exactly one directory, and asserting that is free.

Same lane etiquette as before: it'll get the live-dispatch treatment on the branch before merge.

PR #267, opened 2026-08-20
Spec 069: DuckDB 2.0 migration — main goes 2.0-only

VGSML:
## What

Main now targets DuckDB main (2.0-bound): submodules pinned to duckdb `d7a4366603` / extension-ci-tools `35759fd`, the whole tree migrated, `mssql_compat.hpp` retired. The 1.5.5 world lives on branch **`duckdb-v1.5.5`** for 0.2.x maintenance.

Spec: `specs/069-duckdb-v2-migration/` (inventory, decisions, plan, bench protocol + numbers).

## The clusters (A–E: what the compiler saw — ~222 errors / 35 files → 0)

- **A — `Identifier` at the API boundary.** `Catalog::GetCatalog` (34 sites), info structs via new accessors, table-function & COPY bind signatures (`vector<Identifier>`), `child_list_t`. Internal metadata/TDS/codecs/SQL text stay `string`-native.
- **B — vector const-correctness.** Writes via `GetDataMutable`/`ValidityMutable`, `ToUnifiedFormat` drops the count. The new accessors carry a real release-build check — they stay hoisted; per-value `DecodeFromTds` bodies use the `*Unsafe` forms (measured justification in bench-results.md).
- **C — filter hierarchy.** `Legacy*` renames; `TableFilterSet` via its entry API. New filter representation = phase 2, own spec.
- **D — the structural one.** Comparisons, casts and BETWEEN are `BoundFunctionExpression`s (old classes survive only as static helpers); the filter encoder dispatches on `IsCast`/`IsComparison`/`COMPARE_BETWEEN` inside the `BOUND_FUNCTION` arm. Privatized members via accessors everywhere.
- **E — scatter.** `TimePoint` collision, `dtime_t.micros`→`.value`, `WithAlias`/`WithExtensionInfo`, varargs in the `ScalarFunction` ctor, vector-header moves, `ProjectionIndex`/`TableIndex` strong types, `Vector(Value, count_t)`/`Reference(Value, count_t)`, fmt/core.h include paths for every standalone test compile.

## The clusters the compiler could NOT see (F–G: found by the suites)

- **F — `SetFallible()`**: a throwing scalar function's error becomes an INTERNAL Error without it; all 16 throwing registrations declare it (caught by `ping.test`).
- **G — 9 red integration tests, 4 roots:**
  1. **Vectors carry their own size**; the deprecated `SetCardinality` sets only the chunk count, so `IsNull`, filters and the hash join saw empty vectors (NULLs reading as valid with garbage polarity, IN passing every row, a join of two scans returning nothing). All 15 producer sites → `SetChildCardinality`, rowid-STRUCT targets → explicit `SetSize`.
  2. **Filters arrive as `ExpressionFilter` + `BoundReference(0)`**, and 2.0 does NOT re-apply TableFilters behind a `filter_pushdown` scan — a refused filter silently returned a superset. Encoder learns the BOUND_REF-through-filter-column form, and a **client-filter net** executes anything still refused (ToExpression + `SelectExpression`, with a fetch loop so a filtered-to-empty chunk does not end the scan).
  3. **`column_index_map` is no longer populated**: the binder expands the insert child to all physical columns, MOVING unnamed columns' `bound_defaults` into the projection — the null slot is the tell. Unnamed columns stay out of the column list (identity/server defaults work); the serializer addresses the full-width chunk by table ordinal.
  4. **Writer-scheduling dilution**: 2.0 splits a 110k-row scan across sink threads, putting every writer below the 102400 columnstore threshold. The threshold test pins one writer; lazy writer ramp-up recorded as a follow-up.

## Performance (interleaved same-session A/B, protocol in `bench-results.md`)

- Kernels vs 1.5.5 (501 cells): **median 1.024 / mean 1.003** after the per-value `*Unsafe` fix (was int8 1.52×, decimal 1.30× — recovered to 0.80×/1.01×). **Write-side encode kernels are 3–4× faster on 2.0.**
- E2E write vs shipped community **0.2.4** on stock DuckDB (8-cell wide-write matrix, live SQL Server): **geomean 0.985** — wall parity, gains where client CPU binds (parallel sized writers 0.90×/0.95×).

## Verification

- Local: unit suite green; **integration suite fully green — 162 cases / 5175 assertions**; live Azure: FEDAUTH token path, SQL-auth DSN + DDL, Fabric (auth + gateway routing + full types incl. DDL/write, 39 assertions) — everything reachable without the duckdb-azure extension, which does not exist for a dev SHA (404), so the azure CI lane stays off on this branch by design.
- CI: builds/smokes green on all platforms incl. Windows; the integration lane needed two of its own 2.0 fixes (the unittest now links **shared libduckdb** — the job ships the lib; and Catch exits `--list-tests` with the test COUNT, so the registered-tests guard now gates on empty output and shows stderr).

## Self-review (2 passes, verified against duckdb sources)

- Filter dispatch order safe (`IsComparison` excludes BETWEEN; DISTINCT_FROM stays unsupported); `QualifiedName` accessors return member refs (no dangling); the removed BOUND_CAST recursion arm is covered (`Child` == `GetChildren()[0]`); accessor-sed audits clean.
- Retry-loop soundness: FillChunk Resets the chunk on entry — that Reset is what un-slices dictionary buffers before a refill, and the composite-PK target pointers survive it (they alias cache-owned struct storage). The in-code comment now says exactly that.
- Known pre-existing gap (not a regression): `INSERT INTO t DEFAULT VALUES` produced an empty column list on 1.5.x too.

## Out of scope / follow-ups

- New (non-Legacy) filter representation — phase 2 spec (may unlock pushdown capability).
- Lazy writer ramp-up for parallel bulk loads (compression vs parallelism on small loads).
- `${VAR}` → `{VAR}` in .test files (2.0 runner deprecation warning, still substitutes).


oluies:
Reviewed the migration with a focus on two questions: what the test tree pins, and whether the fuzzing still covers anything. Impressive piece of work — the A–E/F–G decomposition made it reviewable. Findings below, all verified locally rather than read off the diff.

## 1. `fuzz_row_stager` does not build on this branch

`fuzz/build.sh` is the one place the cluster-E *"`fmt/core.h` include paths for every standalone test compile"* fix did not land. The Makefile got `-I duckdb/third_party/fmt/include` in 8 recipes and `run_staging_test.sh` got it too, but `fuzz/build.sh:174` still does:

```sh
INC="${INC} -I${MSSQL_DUCKDB_INC}"
```

so `TARGETS=row_stager fuzz/build.sh` dies:

```
src/codec/string_codec.cpp: fatal error: 'fmt/core.h' file not found
src/codec/uuid_codec.cpp:   fatal error: 'fmt/core.h' file not found
```

Confirmed it is only the include path: with `-I duckdb/third_party/fmt/include` added, all 19 sources in the row-stager set syntax-check clean against the pinned `d7a4366603`. One-line fix.

Nothing caught it because `row_stager` is deliberately out of the default `TARGETS` and never built in CI.

## 2. The ClusterFuzzLite gate will pass this PR having fuzzed none of it

The 5 default harnesses do build and run fine on this branch — I built them with clang 22 + ASan/UBSan/libFuzzer and ran each for 20 s against its seed corpus: no crashes, 45k–432k exec/s. So the gate is not broken.

But it is vacuous here. The default set compiles only `instance_resolver.cpp`, `tds_{token_parser,row_reader,column_metadata,types,protocol,packet}.cpp` and `encoding/utf16.cpp` — and the intersection of that list with this PR's changed files is **empty**. `cflite-pr.yml` matches on `src/tds/**` and `src/codec/**` so the job fires, and `mode: code-change` instruments only the diff, which these binaries do not contain.

`fuzz_row_stager` is the only harness that reaches the changed code (it globs all of `src/codec/*.cpp` plus staging, `bcp_row_encoder.cpp`, `type_converter.cpp`, `mssql_value_serializer.cpp` — 12 of which this PR rewrites), and per (1) it does not currently compile. Worth fixing the include and, separately, deciding whether this migration deserves a one-off manual `row_stager` campaign before merge.

## 3. The test tree still uses the API whose misuse *was* cluster G1

|  | `SetCardinality` | `SetChildCardinality` |
|---|---:|---:|
| `src/` | **0** | 15 |
| `test/` | **29** | 0 |

In 2.0 `DataChunk::SetCardinality` is `[[deprecated]]` and its own doc comment says it does **not** resize the child vectors. `test_vector_encodings.cpp`, `test_batch_builder.cpp` and `bench_materialize.cpp` still build their chunks that way — the exact shape that produced G1's garbage NULL polarity, and a path production no longer uses. Those tests structurally cannot catch a G1 regression. `test_batch_builder.cpp` is not in the diff at all. (`run_staging_test.sh` passes `-Wno-deprecated-declarations`, so the warning is invisible there too.)

## 4. None of G1–G3 got a regression test

Apart from `columnstore_batch_threshold.test`, the entire test diff (137+/104−) is clang-format churn and mechanical renames. Three silent-wrong-results bugs were found by running the suites, and none of them is now pinned by a test. Specifically:

- **G2 — the client-filter net.** New code (`ClientTableFilter`, `SelectExpression`, the fetch loop) whose failure mode is silently returning a superset. Nothing forces a *refused* filter through it. A `.test` using a shape the encoder rejects — `IS DISTINCT FROM`, which stays unsupported — asserting exact row counts would pin it, plus one case where the filter empties an entire chunk to exercise the fetch loop.
- **G3 — unnamed columns.** Behaviour changed (unnamed columns stay out of the column list); no identity / server-default test added.
- **G4 / coverage lost.** `columnstore_batch_threshold.test` now pins `mssql_copy_parallel_writers = 1`. The reasoning in the comment is sound, but the effect is that nothing asserts what a **default-settings** COPY produces against a columnstore target any more. Worth a second case at the default setting asserting whatever the diluted-writer outcome legitimately is, so the follow-up has a baseline to move.

## 5. One inconsistency in the new filter code

`src/table_scan/table_scan.cpp:459-465` — the arming loop's `else` arm silently `continue`s when a refused filter cannot be mapped to a column type, logging only at `MSSQL_SCAN_DEBUG_LOG(1)`:

```cpp
} else {
    MSSQL_SCAN_DEBUG_LOG(1, "TableScanInitGlobal: cannot map refused filter at projected idx %llu", ...);
    continue;
}
```

Twelve lines below, the sibling `catch` deliberately throws `NotImplementedException` with the comment *"instead of returning a superset"*. Same failure, opposite answers. The `else` looks defensive and may well be unreachable, but if it is ever reached it reintroduces exactly the silent wrong-results bug the net exists to prevent. It should throw like its sibling.

---

## Baseline check on `main`

For what it's worth, I ran the full docker suite on `main` (e2b8c01, v0.2.4) to have something to compare against:

- `make test` — all passed, **5182 assertions / 164 cases**, 14 skipped (env gates only)
- `make integration-test` — all passed, **5174 assertions / 162 cases**, 5 skipped

That is the same case count as this PR reports and one assertion apart, consistent with the single `+10`-line change to `columnstore_batch_threshold.test`. So "integration suite fully green" here is a genuine like-for-like result, not a reduced one — good sign for a migration of this size.

Nothing above is a blocker on the migration itself; (1) is a one-liner and (3)–(5) are about what the branch pins for the *next* change rather than whether this one is correct.

oluies (dismissed):
I am trying to fix 1 now

oluies:
Pushed finding (1) straight to this branch as 3d34633 — heads up if you have it checked out locally, you'll want to pull before your next commit.

It's the one-liner plus a guard: `MSSQL_DUCKDB_FMT_INC` derives `<duckdb>/third_party/fmt/include` from `MSSQL_DUCKDB_INC`, so the invocation documented above the block keeps working with no extra variable, and a missing `fmt/core.h` now fails with the fix in the message instead of an include error forty lines into the compile.

Verified against the pinned `d7a4366603`: `TARGETS=row_stager fuzz/build.sh` compiles all 27 objects clean, `string_codec.cpp` and `uuid_codec.cpp` included. I could not complete the link — that needs a libduckdb built at the 2.0 SHA, which I don't have locally — so the final `-lduckdb` step is still unexercised. Worth one real `row_stager` run before merge, since it's the only harness that covers the rewritten codec and staging code.

(2)–(5) from the review above are untouched and still yours to call.

oluies:
Pushed 02da1a9 — two follow-ups to your 54bedd8, both verified against a local 2.0 build (duckdb `d7a4366603`, live SQL Server). Plus one new finding at the bottom that I could not have seen without that build.

## Finding 5 has two more sites

54bedd8 fixed the arming loop in `table_scan.cpp`. `FilterEncoder::Encode` has two more of the same shape, and they're worse — both set `needs_duckdb_filter` and `continue` **without** pushing to `unhandled`, so the filter reaches neither the server nor the client net:

- `filter_encoder.cpp:167` — `projected_col_idx >= column_ids.size()`
- `filter_encoder.cpp:186` — `table_col_idx >= column_names.size()`

`needs_duckdb_filter` is assigned in two places and **read in none** — it's logged and nothing else. Harmless on 1.5.x because DuckDB re-applied TableFilters above the scan; that premise is exactly what cluster G2 says is gone. Both arms now throw `InternalException` (the type already at `table_scan.cpp:60` for this class), and I documented the field as diagnostic-only so it isn't set in place of populating `unhandled` again. Full unit suite run: neither throw fires, so they are genuinely defensive.

## G2 is reachable — `rowid IN (...)`

Your reasoning in 54bedd8 was "probed IS DISTINCT FROM and rowid ranges live, neither arms the net". Both true — I reproduced both, along with `SIMILAR TO`, `regexp_matches`, a cast, a join dynamic filter and an IN-subquery. None arm it.

But `rowid IN (...)` does. `rowid` is virtual, and the virtual-column arm of `FilterEncoder::Encode` is the one branch that pushes to `unhandled` **unconditionally** — so any TableFilter on rowid lands in the net regardless of what the combiner did upstream. It doesn't depend on spec 070 W1.

`test/sql/catalog/filter_client_net.test`, 23 assertions, every one of which returns the whole table if the net regresses. A single-row match at row 4990 of 5000 empties every earlier chunk, which is what pins the fetch loop. Measured on your tip with `MSSQL_DEBUG=1`: **nine** `client-side filter armed` lines.

Worth recording that you were right about functions: an unsupported *function* genuinely does not reach the net — DuckDB routes those to `pushdown_complex_filter` and keeps its own filter operator above the scan. That was my first draft of this test, and it passed while exercising nothing. The file says so explicitly so the next person doesn't redo it.

## New: metadata-query deadlocks, reproducible, 2.0-only

The suite is not stably green on this branch. Three full `make test` runs on 2.0:

| Run | Result | Deadlocks |
|---|---|---|
| 1 | 166 passed, **4 failed** | 7 |
| 2 | 165 passed, **5 failed** | 9 |
| 3 (on your tip + my commit) | 168 passed, **5 failed** | 5 |

Every failure is the same error, on a *metadata* query:

```
IO Error: Metadata query failed: Transaction (Process ID NN) was deadlocked
on lock resources with another process and has been chosen as the deadlock
victim. Rerun the transaction.
```

The failing set differs every run — `insert_errors`, `varchar_encoding`, `partitioned_table`, `lazy_loading`, `catalog_parsing`, `view_cast_type_mismatch`, `copy_transaction`, `preload_catalog`, `udt_type_alias` have each appeared. All of them pass in isolation, so it's contention between concurrent catalog loads, not any one test.

Same machine, same container, same TestDB, `main` at v0.2.4: **two full runs, zero deadlocks, exit 0.** So it looks 2.0-specific rather than environmental — plausibly the higher sink/scan parallelism putting more concurrent metadata sessions on the server than 1.5.x did.

Two things it costs beyond flaky CI: a deadlock mid-test leaves objects behind (one run died with `There is already an object named 'lazy_test_external'` on the *next* run), and if it reproduces in CI the integration lane's `floor 150` guard won't catch it — the count stays above the floor while individual cases fail.

I have not tried to fix this; retry-on-1205 for metadata queries is a design call, not a review nit. Happy to file it as a separate issue against spec 070 if you'd rather keep this PR to the migration.

oluies (dismissed):
Approving. Verified against a local build of this branch at 95e5365 — duckdb `d7a4366603`, macOS arm64, live SQL Server 2022 — not just read.

**The migration itself holds up.** Integration suite green on my machine: 165 cases, zero deadlocks, exit 0. That matches the `main` v0.2.4 baseline I measured earlier (162 cases / 5174 assertions), so this is a like-for-like result rather than a reduced one.

**95e5365 is the commit that justifies the whole exercise.** A real heap overflow — `ConvertDatetime` reads 8 bytes by type id while the stager had staged `column.max_length`, so a stream declaring `TDS_TYPE_DATETIME` with a 4-byte width over-reads on the constant path and strides past the buffer every row on the batch path. Pre-existing, not a 069 regression, and invisible for as long as it was precisely because `row_stager` had never compiled. `BareFixedWireWidth` is the right shape of fix: stride comes from the type id, which is the only thing the kernel actually consults, and canonicalising MONEY/SMALLMONEY to hold `stride == wire width` for every bare type is worth more than the two lines it costs. Crash input pinned as a corpus seed.

**Disclosure:** two of the commits here are mine (3d34633, 02da1a9), so this approval covers the rest of the work rather than my own.

## One thing that is still red, and why I'm approving anyway

The unit suite is not stably green. On this tip, two consecutive `make test` runs: **169 passed, 2 failed**, both times `preload_catalog` and `lazy_loading`, both times the 1205 deadlock. Your retry in 95e5365 helped materially — 5 failures down to 2, and integration went from failing to clean — but it hasn't closed it.

Two observations, offered as leads rather than conclusions:

- I could not catch the retry actually firing. `RunMetadataQuery`'s `CACHE_DEBUG` lines never appear under `make test` (the recipe doesn't propagate `MSSQL_DEBUG`), and when I ran the suite directly with debugging on, the extra logging slowed it enough that the deadlock stopped reproducing at all — 6206 cache log lines, zero deadlocks, exit 0. So the failure is timing-sensitive, and I can't tell from outside whether the surviving two are hitting the `rows_delivered == 0` guard or exhausting `MAX_ATTEMPTS`.
- If it is the guard: `preload_catalog` is by definition the bulk path that delivers many rows before it could deadlock, which is exactly the case the guard makes fatal by design. That would make those two tests the ones the current fix structurally cannot help. Worth checking before assuming three attempts is the knob.

I'm not treating this as a merge blocker: it's contention against a shared test database on one machine, both tests pass in isolation, and the integration lane — the one that gates — is clean. But it is a real behavioural difference from `main` (zero deadlocks across three runs there, same container, same database), so it should land as a spec 070 item rather than be written off as flake. Happy to file it.

One note on CI coverage while it's fresh: `floor 150` will not catch this class. The case count stays above the floor while individual cases fail, which is how a red suite can still look like a pass.

VGSML:
Second review round fully addressed — thank you, this was the review that mattered. Verdict-by-item, all verified against this tree (live SQL Server + the row_stager fuzzer built on a runner, since macOS ships no libFuzzer runtime).

**Finding 5 (your two extra sites, `02da1a9`)** — merged. Both `needs_duckdb_filter=true; continue`-without-`unhandled` paths now throw, and I made the arming-loop's sibling `else` throw too (it was the third of the same shape). A refused filter that reaches neither server nor net is now impossible by construction rather than by luck.

**G2 reachable via `rowid IN (...)`** — you were right, I was wrong to call the net unreachable. Your `filter_client_net.test` (23 assertions, the row-4990 single-match that empties earlier chunks) is exactly the pin it needed. Kept as-is.

**Metadata deadlocks (1205)** — fixed. Read-only metadata queries retry the deadlock victim (`RunMetadataQuery` + the PK path), bounded to 6 attempts / ~2.6 s, and **only while no row was delivered** — after the first row a rerun would duplicate into the callback. I could reproduce your flakiness only intermittently here (0 deadlocks in most runs vs your 5–9), but the retry closes it deterministically: **4 consecutive full `make integration-test`, 0 deadlocks, all green.** The higher-parallelism read is right — a test's CTAS DDL (Sch-M) overlaps sibling catalog loads.

**The fuzz find — you called this one before it happened.** `row_stager` was the only harness reaching the rewritten codec, and one real run found a heap overflow in 112 s. The bare temporal kernels (`ConvertDatetime` and friends) read a type-fixed length straight from `bytes.data()` with no `bytes.size()` check — the exact discipline MONEY/DATETIMEN already had and these lacked. Both the per-value entry and the batch stride are guarded now (`f713d9f`); the crashing input is pinned as a `row_stager` regression seed. **Re-run: 12.9M execs, 0 crashes** (was 112 s to crash). Not a 069 regression — the code predates it; it was invisible only because the harness had never built.

Net of the round: 6 commits from me on top of your 3. Branch is green on the full dispatch matrix + all PR workflows; the fuzz lane is manual (`fuzz-row-stager.yml`, dispatch or a `fuzz/row-stager-run*` push) because it builds libduckdb. Your call on merge from here.

PR #269, opened 2026-08-21
Spec 070 W1: single-column expression pushdown; #242 function-mapping half

### LINKED ISSUE #242: Pushed-filter function mappings diverge from DuckDB semantics: length()->LEN returns wrong rows today; dayofweek/week and '/' need an audit
Found during the spec-065 JOIN/aggregate pushdown reconnaissance (semantic divergence catalog + live probes, 2026-08-05). These are in TODAY'S filter pushdown — no new feature required to hit them.

## CONFIRMED — `length()` mapped to `LEN` loses rows

`LEN` excludes trailing spaces and counts UTF-16 code units; DuckDB `length()` counts Unicode code points including trailing spaces.

Reproduction (SQL Server 2022, extension current main):

```sql
-- server: CREATE TABLE dbo.t (s nvarchar(20)); INSERT VALUES (N'ab'), (N'ab '), (N'😀')
SELECT count(*) FROM db.dbo.t WHERE length(s) = 3;             -- pushed:  0 rows  ← WRONG
SELECT count(*) FROM (SELECT s FROM mssql_scan('db','SELECT s FROM dbo.t'))
WHERE length(s) = 3;                                            -- local:   1 row  ('ab ')
```

Two independent causes: `LEN('ab ') = 2` (trailing-space rule) and `LEN(N'😀') = 2` on non-`_SC` collations (UTF-16 units vs 1 code point).

Candidate fixes: map `length` to `LEN(x + N'.') - 1` (kills the trailing-space half; still counts UTF-16 units for astral chars), or stop pushing `length` on string arguments entirely until an exact form exists. The surrogate half has no cheap exact T-SQL form on non-`_SC` collations.

## AUDIT — `dayofweek` / `week` via `DATEPART` are `@@DATEFIRST`-dependent

The mapping table (`src/table_scan/function_mapping.cpp`) maps `dayofweek`→`DATEPART(weekday,…)` (server: 1-based AND shifted by `@@DATEFIRST`; DuckDB: fixed Sunday=0) and `week`→`DATEPART(week,…)` (non-ISO, DATEFIRST-dependent; DuckDB is ISO). A live probe at default server settings did NOT reproduce a divergence in the `dayofweek(d) = 0` shape — either the expression did not push there or it is compensated. Needs an audit of when these mappings actually fire, and either a `@@DATEFIRST`-independent emission (e.g. `(DATEPART(weekday, d) + @@DATEFIRST + 5) % 7` for Sunday=0, `DATEPART(isowk,…)` for `week`) or removal from the pushable set.

## AUDIT (latent) — division by zero: DuckDB float-Inf/NULL vs server error 8134

DuckDB `/` is float division (`10/0 = inf` under default `ieee_floating_point_ops`; integer `//` by zero yields NULL); T-SQL integer division by zero is a hard error that fails the whole query. The mapping table includes `/`. A probe (`WHERE 10 / v > 1` over a column containing 0) did NOT push in that shape — so today this may be unreachable, but nothing documents WHICH shapes reach the `/` mapping. Audit and either restrict the mapping to provably-nonzero divisors (constants) or drop it.

## Why it matters more soon

The spec-065 reconnaissance (`specs/065-dml-pushdown-recon/join-agg-pushdown.md`) plans to reuse this expression translator inside pushed GROUP BY keys, join conditions and aggregate arguments — where a wrong `length()` stops being a missing row and becomes a wrong group or a wrong sum. The catalog's rule: these three mappings must be fixed or gated out before the translator is admitted into any pushed group key or join condition.

VGSML:
First of the three spec-070 follow-ups (`docs/roadmap-v0.3.0.md` / `specs/070-duckdb-v2-followups/spec.md`). W2 (lazy writer ramp-up) and W3 (`${VAR}`→`{VAR}`) are independent and not in this PR.

## What

Registers `pushdown_expression` on the catalog scan — the 2.0 combiner now offers each single-column expression and we accept whatever the encoder can render (a **dry-run** of the same `EncodeExpression` that `pushdown_complex_filter` already uses). Accepted expressions become `EXPRESSION_FILTER`s and reach the server WHERE clause; anything the runtime encode still refuses is caught by the spec-069 client net, so a dry-run/runtime disagreement degrades to client-side filtering, **never wrong rows**. `BuildEncodeContext` factors the column-id mapping out so the two paths cannot drift.

**`year/month/day(datetime2_col) op const` now push** — `WHERE (YEAR([dt]) = 2024)`, verified live. DuckDB models our DATETIME2 as `TIMESTAMP_NS` and `year()` takes `TIMESTAMP`, so an implicit `TIMESTAMP_NS→TIMESTAMP` cast sat over the column and blocked the push; the encoder now strips a naive-timestamp precision cast **only as a date-part function argument** (never in a comparison, where `dt::DATE`/`dt::TIME` change which rows match).

Also fixes bit predicates: DuckDB BOOLEAN is SQL Server BIT, which is not a search condition — `WHERE flag` now pushes as `([flag] = 1)`, not the 4145-erroring `[flag]`.

## #242 — the function-mapping half (this PR does NOT auto-close #242)

Widening what pushes forces the audit now. Three mappings diverge from DuckDB semantics and are **removed** (not rewritten — an unmapped function falls to the client net, correct by construction):

- **`length`/`len` → `LEN`**: LEN excludes trailing spaces and counts UTF-16 code units; DuckDB `length()` counts code points including them. No exact T-SQL form on non-`_SC` collations. (The confirmed #242 bug.)
- **`/`**: SQL Server integer division on integer operands (`5/2 = 2`) vs DuckDB float (`5/2 = 2.5`).
- **`dayofweek` → `DATEPART(weekday)`** (`@@DATEFIRST`-dependent) and **`week` → `DATEPART(week)`** (non-ISO) vs DuckDB's fixed / ISO semantics.

⚠️ #242's **constant-comparison half** — SQL Server pads/collates string `=`/`IN`/`LIKE` so a pushed `[a] = N'ab'` matches `N'ab '` the client wouldn't — is **not** in this PR (that path is untouched here). Split out as **#272** with a proposed design. So #242 is left OPEN deliberately; this PR closes only the function-mapping audit.

## Tests

`test/sql/catalog/expression_pushdown.test` pins: year/month/day push and return correct rows; `dt::DATE`/`dt::TIME` comparisons stay client-side and return the right rows (the temporal-cast-scoping guard); `length(name)=3` returns the row a pushed `LEN` would have dropped (`N'ab '`); `id/2=1.5` works because `/` is not pushed; bit predicates (`flag`, `NOT flag`, `flag = true`, `flag AND year(dt)=2024`). Full integration suite green. Every existing filter-pushdown test unchanged.

## Kept

`pushdown_complex_filter` stays — it still catches multi-expression shapes the combiner does not route through `pushdown_expression`. Both feed one encoder, so the pushable set is consistent; retiring the complex-filter path is a measurement left to a follow-up (spec 070 W1 step 2).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -595,6 +595,7 @@ STANDALONE_TEST_SOURCES := \
     test/cpp/test_ddl_translator.cpp \
     test/cpp/test_ctas_type_mapping.cpp \
     test/cpp/test_catalog_filter.cpp \
+    test/cpp/test_filter_encoder.cpp \
     test/cpp/test_connection_error_translation.cpp \
     test/cpp/test_spn_host_resolution.cpp \
     test/cpp/test_insert_bulk_sql.cpp \
--- a/docs/query-execution.md
+++ b/docs/query-execution.md
@@ -34,14 +34,23 @@ Table scans are implemented as DuckDB table functions. When a query references a
 
 | Category | DuckDB Functions | SQL Server Equivalent |
 |---|---|---|
-| String | `lower`, `upper`, `length`, `trim`, `ltrim`, `rtrim` | `LOWER`, `UPPER`, `LEN`, `TRIM`, `LTRIM`, `RTRIM` |
+| String | `lower`, `upper`, `trim`, `ltrim`, `rtrim` | `LOWER`, `UPPER`, `TRIM`, `LTRIM`, `RTRIM` |
 | Date extraction | `year`, `month`, `day`, `hour`, `minute`, `second` | `YEAR`, `MONTH`, `DAY`, `DATEPART(...)` |
-| Date arithmetic | `date_diff(part, start, end)` | `DATEDIFF(part, start, end)` |
-| Date arithmetic | `date_add(date, part, amount)` | `DATEADD(part, amount, date)` (reordered) |
-| Arithmetic | `+`, `-`, `*`, `/`, `%` | Same operators |
+| Arithmetic | `+`, `-`, `*`, `%` | Same operators |
 | Pattern | prefix, suffix, contains | `LIKE 'pattern%'`, `LIKE '%pattern'`, `LIKE '%pattern%'` |
 
 **LIKE pattern escaping**: `%` → `[%]`, `_` → `[_]`, `[` → `[[]` (SQL Server bracket syntax).
+
+**Deliberately NOT mapped** (issue #242, spec 070 W1) — these are applied
+client-side by DuckDB, which is the correct answer, not a missing feature:
+
+| DuckDB Function | Why it is not pushed |
+|---|---|
+| `length` / `len` | `LEN` drops trailing spaces and counts UTF-16 code units; `length()` counts code points including them. No exact T-SQL form on non-`_SC` collations. |
+| `/` | T-SQL does INTEGER division on integer operands (`5/2 = 2`); DuckDB `/` is always floating (`5/2 = 2.5`). |
+| `date_diff`, `date_add`, `date_part` | Their date-part argument is a T-SQL *keyword*, but arrives as a VARCHAR constant that would encode as a literal (`DATEPART(N'year', x)`), which SQL Server rejects. DuckDB rewrites `date_part('year', x)` to `year(x)`, which is mapped. |
+| `dayofweek`, `week` | `@@DATEFIRST`-dependent / non-ISO week numbering. |
+| `%` on `FLOAT`/`REAL` | The operator maps, but T-SQL's modulo rejects approximate-numeric operands; float operands fall to the client filter net. |
 
 ### Projection Pushdown
 
--- a/specs/070-duckdb-v2-followups/spec.md
+++ b/specs/070-duckdb-v2-followups/spec.md
@@ -84,9 +84,35 @@ had to go with it — `length`/`len` (LEN drops trailing spaces / counts UTF-16)
 (@@DATEFIRST / non-ISO). Removed, not rewritten: an unmapped function falls to
 the spec-069 client net, correct by construction. `expression_pushdown.test`
 pins both — year pushes, `length(name)=3` returns the row LEN would have lost.
-The `pushdown_complex_filter` path is kept (it still catches multi-expression
-shapes the combiner does not route through `pushdown_expression`); retiring it
-is a measurement left to a follow-up.
+**`pushdown_complex_filter` currently SHADOWS `pushdown_expression`** (PR #269
+review — recorded here rather than fixed, because fixing it is a trade-off, not
+a bug). In `duckdb/src/optimizer/pushdown/pushdown_get.cpp` the complex-filter
+callback runs FIRST, over every pending filter, using the same
+`EncodeSearchCondition` and the same `BuildEncodeContext`; it erases everything
+it can render. `TryPushdownGenericExpression` — and therefore
+`MSSQLPushdownExpression` — only sees what is left, i.e. exactly what the
+identical encoder just refused, so the dry-run answers `false` by construction.
+What actually delivers `year(dt) = 2024` today is the temporal-cast strip in
+`EncodeFunctionExpression`, reached through `ComplexFilterPushdown`; deleting
+the `func.pushdown_expression = ...` line leaves every test passing.
+
+Making the callback reachable means restricting `ComplexFilterPushdown` to the
+shapes the combiner will not offer (it offers only expressions referencing
+exactly ONE column binding). That is not free: the combiner applies gates the
+complex-filter path does not — it skips volatile expressions, `COMPARE_IN`
+above `InClauseRewriter::IN_CLAUSE_REWRITE_THRESHOLD`, and **any** expression
+where `CanThrow() && filters.size() > 1`. A single-column expression that
+`ComplexFilterPushdown` declines under those conditions is then offered to
+nobody and runs client-side — a silent pushdown REGRESSION against what ships
+today (e.g. a cast-bearing `year(dt) = 2024` alongside a second predicate).
+Deciding between the two paths therefore needs a measurement, not a patch: the
+follow-up should either restrict `ComplexFilterPushdown` and replicate the
+combiner's gates, or drop `pushdown_expression` until it can be the only path.
+Until then the registration is a no-op kept for the API surface, and the
+encoder-level behaviour is pinned by `test/cpp/test_filter_encoder.cpp` — which
+asserts the T-SQL STRING, the thing a row-comparing sqllogictest cannot see
+(rows are identical whether the predicate ran on the server or in the spec-069
+client net).
 
 ---
 
--- a/specs/070-duckdb-v2-followups/spec.md
+++ b/specs/070-duckdb-v2-followups/spec.md
@@ -1,6 +1,8 @@
 # Spec 070 — DuckDB 2.0 follow-ups: expression pushdown, writer ramp-up, test-var syntax
 
-**Status**: DRAFT (not started; blocked on spec 069 / PR #267 merging)
+**Status**: spec 069 merged (main is 2.0). **W1 DONE** (this branch:
+`pushdown_expression` + temporal-cast passthrough + the #242 mapping removal +
+`expression_pushdown.test`). W2/W3 ready to start.
 **Follows**: spec 069, which migrated main to the 2.0 API and deliberately
 left three items out of scope. Each is an independent workstream; they share
 a spec because all three exist for the same reason — 2.0 changed the ground
@@ -8,7 +10,7 @@ under a subsystem — but they can land as separate PRs in any order.
 
 ---
 
-## W1 — Adopt the non-Legacy filter representation (and the capability it unlocks)
+## W1 — Adopt the non-Legacy filter representation (and the capability it unlocks) — DONE
 
 ### What 2.0 actually did to filters
 
@@ -70,6 +72,21 @@ Reconnaissance against duckdb `d7a4366603`:
 - A deliberately unencodable expression filter returns correct rows via the
   net (test pins both the rows and the `needs client re-filter` debug line).
 - Filter-pushdown test files extended; no existing pushdown test regresses.
+
+**Outcome (implemented).** `pushdown_expression` dry-runs the encoder and
+accepts what it can render; `year/month/day(datetime2_col) op const` now push
+(`WHERE (YEAR([dt]) = 2024)`) via a temporal→temporal cast passthrough (DuckDB
+models DATETIME2 as TIMESTAMP_NS; the implicit precision cast over the column is
+a view, so the server applies the part function natively). **#242 closed here,
+not deferred**: W1 widens what pushes, so the semantically-diverging mappings
+had to go with it — `length`/`len` (LEN drops trailing spaces / counts UTF-16),
+`/` (server integer division vs DuckDB float), and `dayofweek`/`week`
+(@@DATEFIRST / non-ISO). Removed, not rewritten: an unmapped function falls to
+the spec-069 client net, correct by construction. `expression_pushdown.test`
+pins both — year pushes, `length(name)=3` returns the row LEN would have lost.
+The `pushdown_complex_filter` path is kept (it still catches multi-expression
+shapes the combiner does not route through `pushdown_expression`); retiring it
+is a measurement left to a follow-up.
 
 ---
 
--- a/src/include/table_scan/filter_encoder.hpp
+++ b/src/include/table_scan/filter_encoder.hpp
@@ -189,8 +189,9 @@ public:
 	 * directly after WHERE). Identical to EncodeExpression except that a BOOLEAN
 	 * *value* (a bit column / constant / cast, which SQL Server rejects after
 	 * WHERE — error 4145) is coerced to `(<value> = 1)`. Use this at every
-	 * predicate position (top-level filter, AND/OR child, NOT operand); use
-	 * EncodeExpression for value operands, where a bare bit is legal.
+	 * predicate position (top-level filter, AND/OR child, NOT operand, CASE
+	 * `WHEN` operand); use EncodeExpression for value operands, where a bare bit
+	 * is legal.
 	 */
 	static ExpressionEncodeResult EncodeSearchCondition(const Expression &expr, const ExpressionEncodeContext &ctx);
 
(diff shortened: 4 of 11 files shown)


VGSML:
@oluies review please 🙏 — spec 070 W1 (single-column expression pushdown, closes #242).

Heads-up: after opening this I did a self-review and pushed a follow-up commit (`bad6e60`) fixing two real correctness bugs, so please review the tip, not just the original:
1. The temporal-cast passthrough was value-changing and silently wrong — it stripped `dt::DATE` / `dt::TIME` / `TIMESTAMP_TZ` casts in a *comparison* and pushed `[dt] = '2024-06-15'`, dropping every non-midnight row (the exact #242 class). Now scoped to date-part function args + naive-timestamp precision casts only; everything else falls to the client net.
2. Bonus fix: DuckDB BOOLEAN is SQL Server BIT, so `WHERE flag` pushed `[flag]` and errored 4145. Bare bit predicates now coerce to `([flag] = 1)`.

Both verified against a live server; `expression_pushdown.test` pins every case.

oluies (commented):
Reviewed at bad6e60, built and run against a live SQL Server (duckdb `d7a4366603`, macOS arm64). Everything below is measured, not read.

## What I verified holds

Integration suite green — **166 cases, 0 deadlocks, exit 0**, matching your claim. `expression_pushdown.test` passes (22 assertions), as do `filter_pushdown` (156), `collation_filter` (54), `rowid_filter_pushdown` (44), `filter_pushdown_hugeint` (21) and the spec-069 client-net test.

I also built a differential fixture — remote table vs a local `CREATE TABLE loc AS SELECT * FROM` copy — and compared the exact id lists for each newly-pushed shape. All agree:

`year(dt1)`, `year(dt2)`, `month`, `hour`, bare `flag`, `NOT flag`, `flag AND year(dt)=2024`, `length(a)=3`, `n % 3`, `upper(a)`, `trim(a)`, `flag = true`, `flag IS NULL`, `n+1`, `n-1`, `-n`, `n*2`.

Two things that pleased me specifically: `year(dt2)` — a date-part on the *second* datetime column, not the first — returns the right rows, so the dry-run/runtime column mapping does not drift; and the `= 1` coercion gets NULL right (`flag` excludes the NULL row, `flag IS NULL` returns only it). The cast-strip scoping is right too: `IsNaiveTimestampType` excluding DATE/TIME/TZ is the correct cut, and restricting it to date-part functions rather than comparisons is exactly the distinction that makes it safe. `all_supported` propagating out of `EncodeConjunctionExpression` means a partially-pushed AND returns `supported=false` and falls to the net rather than shipping half a predicate — I checked that specifically, since `pushdown_expression` removes DuckDB's filter operator and a partial push would have been silent wrong rows.

## Finding 1 — the `week`/`dayofweek` removal changes nothing

`GetDatePartMapping` has **zero callers**:

```
$ grep -rn 'GetDatePartMapping' src/ test/ fuzz/
src/table_scan/function_mapping.cpp:69:  static ... GetDatePartMappingTable() {
src/table_scan/function_mapping.cpp:109: bool GetDatePartMapping(...)
src/table_scan/function_mapping.cpp:111:   const auto &mappings = GetDatePartMappingTable();
src/include/table_scan/function_mapping.hpp:52: bool GetDatePartMapping(...);
```

Defined, declared, never called. The table is dead code, so editing it cannot affect what pushes.

What actually protects `week`/`dayofweek` is that DuckDB rewrites `date_part('week', col)` into a function named `week`, and `week` is not in `FunctionMapping`. Probed live, each showing the function name the encoder saw:

| predicate | function seen | pushed? |
|---|---|---|
| `date_part('week', dt) = 1` | `week` | no |
| `date_part('dayofweek', dt) = 1` | `dayofweek` | no |
| `date_part('millisecond', dt) = 1` | `millisecond` | no |
| `date_part('quarter', dt) = 1` | `quarter` | no |
| `date_part('year', dt) = 1` | — | **`(YEAR([dt]) = 1)`** |
| `date_part('second', dt) = 1` | — | **`(DATEPART(SECOND, [dt]) = 1)`** |

That was already true before this PR, so one of the three "#242 semantic fixes" is a no-op.

Two things fall out of it, both latent rather than live:

- **`millisecond` is still in that table and diverges by 1000×.** DuckDB `date_part('millisecond', TIMESTAMP '…30.500')` = **30500** (seconds×1000+ms); SQL Server `DATEPART(millisecond, …)` = **500**. Harmless while the table is dead — but it is a bigger divergence than either mapping you removed, so if the table is ever wired up it becomes a wrong-answer path that the audit passed over.
- **The `date_part` entry would emit invalid T-SQL if reached.** `{"date_part", {"date_part", "DATEPART({0}, {1})", 2}}` substitutes `{0}` with the *encoded literal*, i.e. `DATEPART(N'week', [dt])` — not a datepart keyword. Unreachable today for the same rewrite reason.

Suggestion: either delete `GetDatePartMappingTable` / `GetDatePartMapping` / the `date_part` mapping entry outright, or wire them up. As it stands the PR edits a table that cannot affect behaviour, which will mislead whoever audits this next. (Minor, same area: `IsDatePartFunction` lists `quarter`, `dayofyear`, `dayofmonth`, `week`, `dayofweek`, `isodow`, none of which are in `FunctionMapping` — harmless, but it reads as though they push.)

## Finding 2 — #242's class is not closed: comparison padding

SQL Server pads operands in string comparison; DuckDB does not. With one row holding `N'ab '` (trailing space):

| predicate | pushed as | remote | local copy |
|---|---|---|---|
| `a = 'ab'` | `([a] = N'ab')` | **[1]** | — |
| `a <= 'ab'` | pushed | **[1, 4]** | [4] |
| `a <> 'ab'` | `([a] <> N'ab')` | **[2, 4]** | [1, 2, 4] |
| `a LIKE 'ab'` | `([a] = N'ab')` | **[1]** | — |
| `a IN ('ab','zz')` | *not pushed* | — | — |

Two consequences worth stating plainly:

**The same query on the same table changes answer depending on whether the filter pushed.** Blocking pushdown with a LIMIT and nothing else:

```sql
SELECT count(*) FROM c.dbo.W1Diff WHERE a = 'ab';                        -- 1
SELECT count(*) FROM (SELECT * FROM c.dbo.W1Diff LIMIT 100) WHERE a='ab'; -- 0
```

**`=` and `IN` contradict each other** in the same session on the same data, because one pushes and the other doesn't. And `LIKE` with a wildcard-free pattern is rewritten to `=`, so it inherits padding that DuckDB's `LIKE` never has.

This is **pre-existing** — the constant-comparison path is untouched by this diff, and I confirmed the diff contains no `EncodeConstantComparison` / `LEGACY_CONSTANT` hunks. I'm raising it here for three reasons: it is the same divergence class the PR says it closes; plain string equality is a far more common predicate than `LEN`; and the stated justification for dropping `length`→`LEN` is *"LEN excludes trailing spaces"* — the identical semantics that `=`, `<=`, `<>` and `LIKE` still carry into the server.

I'm not asking you to fix it in this PR. I am asking that **"closes #242" be qualified**, or that this be split into its own issue before #242 is closed — otherwise the issue reads as resolved while the most common instance of it is still live.

## Overall

The mechanism is sound and the risky parts are the ones you got right — cast-strip scoping, partial-AND propagation, bit coercion, NULL handling, and the client net as the backstop for dry-run/runtime disagreement. Nothing here blocks merge. Finding 1 is tidy-up; finding 2 is a scope question about what #242 means.

VGSML:
Thank you — this is exactly the kind of review that earns its keep. Both findings are right; I verified each before acting.

## Finding 1 — fixed in `ad5c0b3`

You're correct that `GetDatePartMapping`/`GetDatePartMappingTable` is dead (zero callers) and the `week`/`dayofweek` edit was cosmetic. Rather than leave a table that "reads as if it matters", I removed it outright:

- Deleted `GetDatePartMappingTable` / `GetDatePartMapping` (+ the header decl). That also takes out the `millisecond` entry and its 1000× divergence you spotted — good catch; it was the biggest wrong-answer waiting in that file.
- Removed `date_diff` / `date_add` / `date_part` from the **live** `FunctionMapping` too — same defect you named for `date_part` (`DATEPART(N'year', …)`), and `date_diff`/`date_add` carry it identically. All three unreachable (the `date_part('year',…)` → `year` rewrite), none covered by a test.
- Trimmed `IsDatePartFunction` to exactly the mapped extractors (`year/month/day/hour/minute/second`), with a note to add a name only alongside its mapping.

No behaviour change — everything removed was dead or unreachable. Pushdown suite still green (expression 22, filter 174, collation 54, hugeint 21).

## Finding 2 — you're right, and it's the more important one

The comparison-padding / collation divergence is real, pre-existing, and — as you say — the same class #242 names while being a far more common predicate than `LEN`. I've split it into its own issue rather than smuggle it under #242: **#272**. It carries your repro table and a preliminary decision for discussion — keep pushdown on `=`/`IN` (they return a **superset** under SQL Server semantics) but restore correctness with a client re-check, gated by collation so `_BIN2` columns keep the fast push-and-remove path. It also flags that `collation_filter.test` currently encodes the *opposite* contract (server semantics), so the contract choice comes first. Would value your eyes on that decision.

On the "**closes #242**" wording: you're right to push back. I'll qualify it — #269 closes the function-mapping half of #242 (`length`→`LEN`, `/`, the `dayofweek`/`week` audit); the constant-comparison half is now #272. #242 shouldn't be auto-closed by this merge.

oluies:
Pushed c73bde2 addressing the roborev branch review (job 1112, 3 Medium + 1 Low). @VGSML could you review — one of the four findings is answered with an argument rather than a patch, and that's the one worth your eyes.

**Fixed**

- **CASE `WHEN` was the predicate position the bit-coercion sweep missed.** `filter_encoder.cpp:EncodeCaseExpression` encoded the `WHEN` operand with `EncodeExpression`, so a bit-valued `WHEN` emitted `CASE WHEN [flag] THEN …` — SQL Server error 4145. Because an encoded predicate is *erased* from the DuckDB plan, that fails the query outright instead of degrading to the client net. Now goes through `EncodeSearchCondition`; the header contract lists the position.
- **`%` on `FLOAT`/`REAL`.** The `/` removal audited arithmetic for divergence and left the one remaining entry with a server-side rejection: T-SQL's modulo does not accept approximate numerics, DuckDB's `%` is defined for `DOUBLE`. `EncodeFunctionExpression` now refuses float operands (the mapping stays — integers and decimals behave identically, negatives included).
- **Docs.** `docs/query-execution.md` and `website/docs/reading/queries.md` still advertised `length`/`len`, `date_diff`, `date_add`, `date_part` and `/` as pushed. `length` was removed precisely because pushing it *changes results* (#242), so the stale row was worse than untidy — both now carry an explicit "deliberately not mapped, and why" section.

**Confirmed but deliberately not patched — please sanity-check this call**

The review is right that `pushdown_expression` is unreachable: DuckDB runs `pushdown_complex_filter` **first** over every pending filter, with the same `EncodeSearchCondition` and the same `BuildEncodeContext`, and erases everything it renders — so `TryPushdownGenericExpression` only ever sees what that pass just refused, and the dry-run says `false` by construction. Deleting the registration leaves every test green.

Its suggested fix — restrict `ComplexFilterPushdown` so single-column expressions fall through — is not free, and the review didn't account for the cost. The combiner applies gates the complex-filter path does not: it skips volatile expressions, `COMPARE_IN` above `IN_CLAUSE_REWRITE_THRESHOLD`, and **any** expression where `CanThrow() && filters.size() > 1`. A single-column expression we hand over under those conditions gets offered to nobody and silently runs client-side — a pushdown regression against what ships today (e.g. a cast-bearing `year(dt) = 2024` sitting next to a second predicate). So it's a measurement, not a patch. Recorded at the call site and in `specs/070-duckdb-v2-followups/spec.md`, which no longer implies the new callback is what delivers W1 — the temporal-cast strip is.

**The half of that finding that *is* fixed: `test/cpp/test_filter_encoder.cpp`** (new, wired into `make test-cpp`).

This is why the above could hide. Every pushdown test so far asserts **rows**, and rows are identical whether the predicate ran on the server or in the spec-069 client net — that's the net's whole purpose. So the encoder could stop pushing anything and nothing would go red. These tests pin the **T-SQL string**: the `year()`-over-precision-cast strip, the casts that must *not* be stripped (`DATE`/`TIME`/TZ), bit coercion at each predicate position, the CASE case above, the #242 removals staying removed, and both modulo operand classes. Verified to fail on this branch without the fixes.

`make test-cpp`: 17/17. Full release build clean.

VGSML:
Reviewed `c73bde2` — all four calls are right, thank you.

- **CASE `WHEN`**: genuinely missed by my bit-coercion sweep — I routed the top-level filter, AND/OR children and the NOT operand through `EncodeSearchCondition` but not the `WHEN` operand, so a bit `WHEN` would have 4145'd the whole query. Good catch, and the header contract listing the predicate positions is the right place to keep it honest.
- **`%` on float**: correct — T-SQL modulo rejects approximate numerics; refusing float operands while keeping the int/decimal mapping is exactly the cut.
- **Docs**: yes, a stale "pushed" row for `length` is worse than untidy given #242 — glad you caught the website copy too.
- **`test/cpp/test_filter_encoder.cpp`**: this is the important one. You're right that every prior pushdown test asserts *rows*, which the spec-069 client net makes identical whether or not anything pushed — so the encoder could silently stop pushing and stay green. Pinning the T-SQL string is the assertion that actually guards W1. Wired into `make test-cpp`, 17/17 here too.

On the **`pushdown_expression` unreachability** — I confirm your call: keep the registration, don't act on the auto-review's suggestion. Your reasoning is exactly right: `pushdown_complex_filter` runs first with the same `EncodeSearchCondition`/`BuildEncodeContext` and erases everything it renders, so `TryPushdownGenericExpression` only ever sees refused filters and the dry-run is `false` by construction. And the suggested "make it reachable" fix would regress pushdown — the combiner's own gates (volatile, `COMPARE_IN` past the rewrite threshold, `CanThrow() && filters.size() > 1`) don't exist on the complex-filter path, so handing single-column expressions over under those conditions drops them to client-side silently. Leaving it registered + documented is correct: it costs nothing, records intent, and becomes live if/when the complex-filter path is retired (W1 step 2's measurement). The spec no longer implying the callback is what delivers W1 — the temporal-cast strip is — is the right correction.

I've qualified the PR so it no longer auto-closes #242 (function-mapping half here; constant-comparison / padding half is #272). Good to merge from my side once you approve.

oluies (approved):

PR #270, opened 2026-08-21
Spec 070 W2: lazy writer ramp-up — restore columnstore compression on default writers

VGSML:
Second of the three spec-070 follow-ups. Independent of W1 (#269) and W3.

## The bug

DuckDB 2.0 fans even a small scan across sink threads. Spec 057's one-shot writer claim then splits a sub-threshold **columnstore** load across N writers, so every rowgroup stays below SQL Server's 102400-row compression threshold — the load the user asked to compress silently landed in the delta store. Spec 069 pinned `columnstore_batch_threshold.test` to one writer to hide it; **this PR is the real fix, and dropping that pin is the acceptance test.**

## The fix

- **Per-chunk claim** replaces the one-shot claim: a thread sharing the global writer keeps re-asking, so a large load still reaches the full writer count even though every thread arrived early.
- **Volume gate** in `TryStart`: hold extra writers until the load has produced one `flush_rows` batch — enough to land a compressible rowgroup before a second writer dilutes it.
- Applied symmetrically to COPY and CTAS. (The CTAS shared-BCP path must advance the gate counter too, or a load that starts entirely on the shared writer never ramps — that was a real bug caught in bench.)

## Two gate designs measured and rejected (recorded in the spec)

1. `rows_sunk >= active * flush_rows` (the spec's first sketch): **1.3–1.5× slower at threads=4** — kept ~active·flush_rows rows serialized on the shared writer, far past the compression-return point.
2. A single unconditional warm-up batch (`rows_sunk >= flush_rows`): still **1.2×** — serializing even 102400 rows on a large load costs ~30% by arithmetic (100k/rate warm-up vs 25k/rate parallel), not the "~0.1–0.2 s noise" the cost model assumed.

The serialization penalty is inherent, so it is paid **only where it buys compression**: the gate is active for a **COLUMNSTORE** target and off for heap/rowstore, which fan out immediately as before. `total_rows_expected` is never populated (no size test available), but a heap load has no compression to protect, so gating on target shape is both sufficient and free on the common path.

## Results

- `columnstore_batch_threshold.test` green **at default writers** (pin removed).
- 1M heap CTAS reaches 4/4 writers with ~38 ms ramp overhead.
- Wide-write A/B (heap fixture, 1M rows, threads 1 & 4): **geomean 0.961 vs main, 0.989 vs community 0.2.4**, threads=4 within noise (0.86–1.07× vs main).
- Integration suite green — 165 cases / 5226 assertions, 0 deadlocks.


VGSML:
@oluies review please 🙏 — spec 070 W2 (lazy writer ramp-up).

Follow-up commit `e348c64` on top of the original addresses a self-review finding: the per-chunk claim re-blocked a 30 s `Acquire()` every chunk when the pool was exhausted (it only stopped asking on the *cap*, not on acquisition failure). `TryStart` now returns a 3-way `Claim` (`Started`/`GateClosed`/`Unavailable`) so a genuine acquire failure is terminal — pre-W2 permanent fallback restored. Also rewrote the stale `TryStart` doc (it described the rejected gate design). Ramp still opens 2/4→3/4→4/4 as rows cross flush_rows; copy/ctas/dml green.

oluies (commented):
Built and measured this one rather than read it — it's the concurrency change of the three, and the CI that runs on it excludes the integration lane.

## Both halves verified

| check | result |
|---|---|
| `columnstore_batch_threshold.test` at **default** writers (your acceptance test) | pass, 17 assertions |
| Integration suite | **165 cases, 0 deadlocks, exit 0** |
| 1M-row heap COPY, `MSSQL_COUNTERS=1` | `writers=8/8` |
| 110k columnstore COPY | `writers=5/8` |

The last two are the interesting pair. On the 110k columnstore load the rowgroups came out:

```
COMPRESSED   1 group   102400 rows
OPEN         1 group     7600 rows
```

That is the gate doing exactly what it claims: everything stays on the first writer until it has produced one full `flush_rows` batch — which compresses — and the remaining 7600 rows then fan out. Meanwhile the 1M heap load reaches the full 8/8, so the ramp costs nothing where there's no compression to protect. On a machine with `threads=18`, both numbers are what the design predicts.

Dropping the spec-069 one-writer pin also closes the coverage loss I raised as finding 4 on #267, and the `CsDefaults` case you delete really is redundant now — `COUNT(*) = 110000` on `CsThreshold` already proves nothing was lost between writers, which was all that assertion added.

## One gap worth naming

**The fan-out half has no regression test.** `parallel_writers.test` and `ctas_parallel_writers.test` set `mssql_copy_parallel_writers = 4`, but everything they assert — `count(*)`, `count(DISTINCT id)`, `sum(id)`, the `EXCEPT ALL` — holds at *any* writer count. So if the gate regressed to "never opens", the whole suite stays green and only throughput collapses.

`parallel_writers.test` already documents this ("What is NOT asserted here… `connections_created > 1`… is useless, because the count is 2 even on a single thread"), so the gap predates you. What W2 changes is its weight: writer count used to be a derived constant, and is now behind a volume gate with per-chunk re-claiming — a genuinely new way for it to silently become 1. The evidence for that half in this PR is a bench number, not a test.

It also became testable in the process: both the COPY and CTAS close paths already emit `writers=used/limit` under `MSSQL_COUNTERS=1`, which is precisely what I used above, and `make counters-test` already exists to run the suite with them on. A single large-load case asserting `writers > 1` would close it. Not a merge blocker — the behaviour is right today and I measured it — but worth pinning while the mechanism is fresh, rather than discovering it from a throughput report later.

## Note on CI

Seven jobs report `skipping` on this PR — Build, Integration Test, Smoke Test, Windows MSVC/MinGW, Azure. So the integration lane did not run here; the 165-case/0-deadlock result above is from my machine, not from CI. Fine for this change, but worth knowing that a concurrency PR merges on a green that never exercised it.

Happy to approve once you've decided whether the `writers > 1` pin belongs in this PR or a follow-up.

VGSML:
Thank you — the review is exactly right, and I added the pin (`c999e45`) rather than defer it: the mechanism is fresh, and "writer count silently becomes 1" is precisely the failure this PR makes newly reachable.

`test/sql/copy/parallel_writers_fanout.test`. Since `writers=used/limit` is stderr-only and not assertable from sqllogictest, it reads the count indirectly: on a **fresh pool** loading a **pre-created** columnstore table (no DDL of its own), each parallel writer acquires its own pooled connection, so `mssql_pool_stats.connections_created` equals the number of writers that ran. A 600k-row load (well past `flush_rows`) opens the gate and fans out; the assertion is `connections_created > 1`.

I verified it's a real guard, not a tautology: the **identical load at `mssql_copy_parallel_writers = 1`** gives `connections_created = 1`, so a gate stuck closed turns the query red. It asserts only that fan-out *happened*, not how far — absolute writer count is machine- and speed-dependent (you measured 5/8 on threads=18; I get 2/4 on threads=8), so pinning an exact number would flake. Copy suite green (867 assertions, 28 cases).

On the CI note — you're right that the integration lane `skipping` means neither your 165-case run nor this new test executes in CI here; it's a local guard (gated by `require-env MSSQL_TESTDB_DSN`), same footing as the rest of the integration suite. Worth a separate look at wiring the integration lane onto these PRs, but out of scope for W2.

Good to merge from my side whenever you're happy with the pin.

oluies:
Rebased onto `main` after #269 merged, and pushed a9f59c3 addressing the roborev branch review (job 1115: 2 Medium + 1 Low). @VGSML review please — the first one is a real regression reachable from a documented setting.

**Medium — the warm-up gate was keyed on the user's number, not the server's.** `bulk_load_session.cpp` held extra writers until the shared writer had sunk one `flush_rows` batch. But the thing that gate exists to protect — SQL Server writing a bulk batch straight into a COMPRESSED rowgroup — triggers at a *fixed server constant* of 102400 rows. `mssql_copy_flush_rows` is user-settable and validated only as non-negative, so the shape this branch's own spec text says was measured 1.3–1.5× slower and rejected re-entered through the setting: at `flush_rows = 1000000` the first **million** rows serialized onto one writer. The mirror case was waste rather than slowness — below 102400 no batch can compress at all, yet the gate still held writers back for nothing.

New named `MSSQL_COLUMNSTORE_ROWGROUP_ROWS` sits next to `MSSQL_DEFAULT_COPY_FLUSH_ROWS` (same number today, which is exactly why they got conflated), the gate waits on the server's threshold, and it stays open entirely when `flush_rows` is below it.

**This one is pinned by a test that bites.** `parallel_writers_fanout.test` gains a 600000-row columnstore load at `flush_rows 1000000` asserting `connections_created > 1`. Reverting just the gate keying makes it fail with `connections_created = 1` — the load collapsing onto a single writer, verified locally, not argued.

**Medium — docs.** The branch replaced a documented invariant and no doc moved with it. `DATAMODEL.md` still said a writer is "claimed by a **thread**, on its first chunk", which per-chunk re-claiming makes false — the mermaid edge label and the spec-057 row said it too. The Writers section now carries the three claim outcomes (`Started` / `GateClosed` / `Unavailable`), why the gate exists, that it is columnstore-only, and that its threshold is the server's. `CLAUDE.md`'s `mssql_copy_parallel_writers` row read as an unconditional ceiling, so setting it to 8 and loading 100k rows into a columnstore target silently yielded one writer with nothing explaining why.

**Low — dead null-check.** `MSSQLCatalog::GetConnectionPool()` returns a reference, so `pool != nullptr` could never be false in either sink. Dropped. The header comment claimed the pointer is "Null when claiming is disabled (transaction-pinned, limit of one, or init failed)" — none of the three is true; `may_claim` is the single switch and a failed catalog lookup throws. Corrected, with a note not to guard on the pointer.

**Verification** (local SQL Server on :14330, since none of this runs in CI):

- Full integration suite — `test/sql/*` minus `[azure]`/`[fabric]`: **172 cases, 5053 assertions, all pass**, 5 skipped for optional env.
- That includes the 8 files gated on `MSSQL_TEST_SERVER` / `MSSQL_TEST_CONNECTION_STRING`, run with both actually set.
- `make test-cpp` 17/17; release build clean.

One suite-mechanics note while I was in there: passing several filters to `unittest` **concatenates** them into one pattern and reports "No tests ran" with **exit 0** — the same false-green shape as #212. Worth a guard if that ever gets scripted.

oluies (approved):
LGTM

PR #271, opened 2026-08-21
test: sweep deprecated sqllogictest substitution forms to 2.0 syntax (spec 070 W3)

VGSML:
Last of the three spec-070 follow-ups. Mechanical, no C++ change — done last on purpose because it touches every `.test` file and would sit noisily inside any other PR's diff.

## What

The DuckDB 2.0 sqllogictest runner deprecates two substitution forms and prints a `Replacing deprecated ...` warning for **every occurrence, once per run**, while still substituting.

- **`${VAR}` → `{VAR}`** across **159 `.test` files** (334 occurrences). `require-env` gating is unchanged: only a variable the same file declares substitutes; an undeclared name still arrives literally.
- **`__TEST_DIR__` → `{TEST_DIR}`** in `copy/vector_encodings_bcp.test` (2 occurrences). The W3 assertion is *"zero `Replacing deprecated` lines in the suite log"*, not *"zero `${`"* — the runner deprecates `__TEST_DIR__` with the identical warning line, so it had to go for the assertion to hold.
- **`docs/TESTING.md` + `test/README.md`** examples converted, with a short authoring note naming the brace-only form canonical and restating the require-env gating rule.
- **`test/TLS_TESTING.md` left untouched** — its `${VAR}` are real shell variables in `bash`/`curl` examples, not runner substitutions.

## Verify

Full integration suite green with **zero** `Replacing deprecated` lines in the log (grep the suite log — that absence *is* the assertion; a numeric floor on it would rot):

```
All tests passed (5 skipped tests, 5232 assertions in 165 test cases)
--- deprecation lines: 0
```

Scope check: `git diff --name-only` touches only `.test` files, `docs/TESTING.md`, `test/README.md`, and the spec.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #273, opened 2026-08-22
docs: bring the v0.3.0 roadmap up to date with what actually shipped

oluies:
@VGSML — the roadmap is where we split the work, so a stale row costs something. Two went stale when #269 merged.

**State flips**

- The **#242 fix** row still read *"open bug, repro in #242"*. It shipped in #269 alongside 070 W1.
- The **070** row still read *"spec drafted (069 branch)"*. W1 is merged (#269), W3 landed ahead of the spec (#271), W2 is open for your review (#270).

**One that needed more than a state flip**

The "settled during recon" note said the mappings *"need an audit before any of them may enter group keys or join conditions."* That audit happened — and it was resolved by **removal, not by finding exact T-SQL forms**. `length`/`len`, `/` and `date_diff`/`date_add`/`date_part` are gone from the mapping table; `dayofweek`/`week` were already out.

That distinction matters for the join/agg workstream, which lists `#242 fixed` as a dependency. Read as a bare "fixed", the next reader concludes `length()` may now enter a group key. It may not — it falls to the client net, which is the correct answer and is also the answer for join conditions. Anything wanting it server-side needs an exact form first, and there isn't one on non-`_SC` collations. The note now says so, including the one function the audit **kept**: `%`, gated to exact-numeric operands because T-SQL's modulo rejects float/real.

**Order of battle** marks what's done, and calls out the thing I think is the real headline: **step 0, the cardinality callback, is still not started** — size S, depends on nothing, and gates 066 → 067 → MERGE, i.e. most of the release. Everything else in flight is downstream of it or independent.

**Not touched on purpose**: `specs/070-duckdb-v2-followups/spec.md`, whose status line still says "W2/W3 ready to start". #270 already edits that file and this would conflict with it — worth a one-line fix inside #270 instead.

Owners are unchanged; the file says to claim by editing, so swap freely.

### DIFF
--- a/docs/roadmap-v0.3.0.md
+++ b/docs/roadmap-v0.3.0.md
@@ -32,7 +32,7 @@ cut by then.
 | — | MERGE pushdown (T-SQL MERGE from DuckDB MERGE INTO; semantics mapped in 065 research) | recon only | 067 | M | VGSML |
 | 061 | Collation-aware ORDER BY pushdown — makes the spec-039 ORDER BY/TOP pushdown default-safe (today experimental, opt-in `mssql_order_pushdown`); the remaining half of #58 / discussion #59 | spec on main (spun off 060) | nothing | M | oluies |
 | — | JOIN / aggregation pushdown (`join-agg-pushdown.md` on the recon branch): reduction-vs-relocation ladder, materialize-then-decide; the community ask in discussion #75 | recon only | 066; #242 fixed | L | pair — design review together, then split |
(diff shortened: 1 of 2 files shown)


oluies:
@VGSML — #270 merged while I was writing this, so I rebased and updated: **spec 070 is now complete** (W1 #269, W2 #270, W3 #271 all in). The row and the order-of-battle line say that rather than "W2 in review".

Which makes the remaining picture short: everything left in v0.3.0 is either **step 0** or downstream of it. The cardinality callback is size S, depends on nothing, is unclaimed in practice, and gates 066 → 067 → MERGE. It's the whole critical path now.

VGSML (dismissed):

VGSML (approved):

PR #274, opened 2026-08-23
feat: report scan cardinality to the optimizer (roadmap v0.3.0 step 0)

oluies:
@VGSML — roadmap "step 0". Closes the critical-path item that gates 066 → 067 → MERGE.

## The bug, reproduced before fixing

Every MSSQL scan planned as `~1 row`. Two real tables, 200000 rows and 50 rows:

```
╭─ Hash Join ───┴──────────────╮
│ Conditions: id = id          │
│ ~1 row                       │
╰───────────────┬──────────────╯
╭─ Mssql Catalog Scan ─────────╮  ╭─ Mssql Catalog Scan ─────────╮
│ ~1 row                       │  │ ~1 row                       │
```

Indistinguishable — so join order and build-side choice around any MSSQL scan were decided arbitrarily. After:

```
╭─ Mssql Catalog Scan ─────────╮  ╭─ Mssql Catalog Scan ─────────╮
│ ~200,000 rows                │  │ ~50 rows                     │
```

## Why it was invisible, and why the work already existed

`MSSQLTableEntry::GetStorageInfo` has computed the real count since spec 023 — statistics cache, DMV fallback — and its comment even says *"Table-level cardinality is provided via GetStorageInfo"*. That is true for a **table** scan. An MSSQL scan is a **table function**, and the optimizer asks the function's `cardinality` callback instead of reading `TableStorageInfo`. The number existed; it just never reached the planner. `TableFunction::cardinality` was unset.

## Design points worth your review

**Cheap and non-throwing, unlike `GetStorageInfo`.** This runs inside the optimizer, once per scan per plan, so it must not acquire a connection — a query must not pay a network round trip to be *planned*. Statistics cache first (`TryGetCachedRowCount`, refreshed by `PreloadRowCount` after a load), then the count the catalog already loaded with the metadata, then nothing. Any throw degrades to "no estimate", never a failed plan.

**0 means unknown, not empty.** A VIEW has no `sys.partitions` rows and reports 0 while returning millions — claiming 0 would make the planner pick it as a build side, far worse than claiming nothing. So a count of 0 yields no estimate. The cost: a genuinely empty table also plans as unknown, losing an estimate nobody had before this PR. I think that trade is right, but it's the judgment call in here.

**`mssql_enable_statistics` was dead.** `LoadStatisticsEnabled` existed with **no callers** — the setting was documented in CLAUDE.md and did nothing. This callback is the one optimizer-facing consumer of those row counts, so it now gates this, and is the escape hatch if an estimate ever plans worse than no estimate did.

## Testing

`test/sql/catalog/scan_cardinality.test` asserts the **plan**, not rows — rows are identical at any estimate, which is exactly why this sat broken without a single test going red. Verified it fails with the registration reverted.

- Full integration suite: **168 cases / 5291 assertions**, all pass (local SQL Server).
- `make test-cpp`: 17/17. Release build clean.

Docs moved with it per the DATAMODEL.md rule: the statistics-layer section explains why the count travels through `cardinality` rather than `TableStorageInfo`, and the CLAUDE.md row says what the setting now does.

## Follow-up not done here

The estimate is the table's **full** row count; pushed-down filters don't reduce it. DuckDB applies its own selectivity on top, so this is not wrong, but a filter-aware estimate would be better and is a separate change.


oluies:
Pushed b2c5c68 — filter-aware estimates, the follow-up this PR listed as not-done. It turned into a finding, so it's worth reading before the code.

## A pushed filter reduced the estimate by nothing

`WHERE id < 100` over 200000 rows planned as **200000** — and not just in a single-table query where the join-order optimizer doesn't run. Inside a join, where it does, still 200000.

## Why — and DuckDB is not missing the feature

`RelationStatisticsHelper::ExtractGetStats` already walks `get.table_filters` and applies either a statistics-derived selectivity or `DEFAULT_SELECTIVITY` (0.2). It had nothing to walk.

`ComplexFilterPushdown` encodes each predicate into `complex_filter_where_clause` — **a string** — and erases the expression from the plan. So `get.table_filters.HasFilters()` is false and the estimator applies nothing at all. Confirmed with `MSSQL_DEBUG=1` against a live server:

```
ComplexFilterPushdown: processing 1 expression(s)
  filter[0]: encoded -> ([id] < 100)
ComplexFilterPushdown: 1 expressions handled, 0 remaining for DuckDB
```

(`PushdownExpression` never logs — that's job 1112's finding 1 again, from a different direction.)

## What landed

The scan applies DuckDB's own constant rather than inventing one — same 0.2, same reason. 200000 → 40000 filtered; unfiltered untouched. `top_n` is additionally reported as a hard `max_cardinality`, since TOP N is a bound the server cannot exceed rather than a guess.

**It is labelled a stopgap in the code and in DATAMODEL.md.** If single-column predicates ever reach `pushdown_expression` and survive as EXPRESSION_FILTERs in `table_filters`, DuckDB applies selectivity itself and this must be deleted or it double-counts.

## The part I'd like your read on

This is the **second** cost of `pushdown_complex_filter` shadowing `pushdown_expression`. The first was a dead callback (#269). The second is that **the planner cannot see a single predicate on an MSSQL scan** — every filter we push becomes a string it cannot reason about, so no selectivity, no always-false pruning, no filter-aware join ordering.

In #269 I argued against restricting `ComplexFilterPushdown` because the combiner applies gates the complex-filter path doesn't (volatile, oversized IN, `CanThrow() && filters.size() > 1`), so deferring could silently lose pushdown. That risk is still real — but it was weighed against *no* benefit at the time. Now there's a concrete one on the other side. Spec 070's W1 outcome already asks for a measurement; this is the argument for actually doing it.

## Testing

`scan_cardinality.test` extended: filtered estimate drops, unfiltered doesn't, rows unchanged either way — including the join case, where DuckDB propagates the predicate transitively to the other side and that filter pushes too (both sides reduce, result still 50).

Full integration suite **168 cases / 5297 assertions** green; `make test-cpp` 17/17.

VGSML (approved):

PR #275, opened 2026-08-23
docs: step 0 done + the W1 measurement, with VGSML's 061-gates-W1 decision folded in

oluies:
Docs-only. Updated after @VGSML's decision — the question this PR was titled around is **answered**; folded in here as asked, since #277 deliberately left `docs/roadmap-v0.3.0.md` alone.

## What's recorded

**The edge:** `061 → W1 restriction → planner-visible filters → DML-collapse + join/agg`.

I proposed that edge as a SELECT-correctness story. VGSML's reason is stronger and is the one now written down: **a server-side `UPDATE`/`DELETE` has nothing downstream to re-check it.** A predicate matching one row too many doesn't return an extra row — it *modifies data DuckDB would never have touched*. SELECT survives a loose predicate because a client pass fixes it; DML doesn't survive it at all.

Also carried, because both are easy to lose:

- **Rowid is not automatically safe.** It works today only because a PK is unique *under the server's own comparison*. Move any rowid comparison to binary semantics client-side and that breaks quietly.
- **My #276 conclusion is corrected, not left standing.** "Blocked" was too strong: the range rewrite returns a **superset** of DuckDB's answer, so the restriction was *missing the re-check*, not wrong.

**The 061 row is retitled** — collation-**faithful**, not ORDER-BY-only, and it now supplies the exactness DML needs (`native AND … COLLATE …_BIN2` plus a trailing-space sentinel, added beside the native predicate so the Index Seek survives). It stops being an independent item.

**The W1 row** moves from "blocked" to "measured, gated on 061".

## One thing deliberately left open

#277 § 4.1 takes SELECT exactness from the client net, which applies DuckDB's **binary** comparison — so `WHERE name = 'abc'` would stop matching `'ABC'` on a CI collation. That contradicts the first line of this file's own do-not-re-litigate list (*"NATIVE server semantics by default"*) and the user-facing promise in `website/docs/reading/queries.md:61-63`.

Superset-and-re-check and native-by-default are **mutually exclusive** for SELECT: the native answer is 2, the range gives 3, the re-check gives 1. Raised on #277; recorded here as **unanswered** rather than resolved in either direction, with a note not to design relocation-vs-reduction on top of it until it is.

## Also in this PR

- Step 0 marked done (#274); **066 is the head of the critical path**.
- The W1 measurement's result, including that my #269 objection (combiner gates would lose pushdown) **did not reproduce**.
- A workstream row for the W1 restriction, so #274's stopgap has something scheduled to remove it — three roborev rounds on this branch flagged that it had none.
- The step 0 row corrected: the TOP N "hard bound" I shipped is really an estimate clamp, since nothing consumes `max_cardinality`.


oluies:
Ran the measurement this PR recommends. **Answer: no, not as-is** — and the dependency edge points the other way from what I wrote here.

I restricted `ComplexFilterPushdown` to defer single-column predicates to `pushdown_expression`, so they survive as EXPRESSION_FILTERs in `get.table_filters`. Branch `spec/070-w1-measure-shadowing`.

**Most of it worked.** No pushdown lost — every shape the combiner gates still reached the server (`IN` at the 6-value threshold, `year(dt)=2024` beside a second predicate, `LIKE`). The planner could finally see predicates: `EXPLAIN` showed `Filters: id < 100` and applied real selectivity instead of the blind spot this PR describes. Full suite green, 168 cases / 5297 assertions.

**My #269 objection did not reproduce.** I argued the combiner's gates (`volatile`, oversized `IN`, `CanThrow() && filters.size() > 1`) would silently lose pushdown. They didn't. That reasoning was wrong.

**What actually killed it** is unrelated to any of that. DuckDB rewrites a prefix predicate into a range — `prefix(col,'n')` → `col >= 'n' AND col < 'o'` — which is exact for its own binary comparison and *not* for SQL Server, where the collation decides ordering. On `SQL_Latin1_General_CP1_CI_AS`, SQL Server's **default**, `'ñu'` sorts between `'n'` and `'o'`: the server's `LIKE N'n%'` returns 2 rows, the range returns 3. `main` returns 2; the restricted build returns 3. Silent wrong rows — and the entire suite passed in that state, because nothing covered it.

#276 adds that missing coverage (test only, no behaviour change), asserted against the server's own answer so it holds on any collation.

**Consequence for this document.** The restriction has a prerequisite: the encoder must refuse or correctly translate range-rewritten prefix filters before those predicates may become TableFilters. That is spec **061** — already on the roadmap, listed as independent. So:

> **061 gates the W1 restriction → which gates planner-visible filters → which is what join/agg pushdown actually needs.**

061 is no longer "independent, any time". If you're happy with that reading I'll amend this PR's section and the order-of-battle to record it, rather than leaving the text recommending a measurement that now has an answer.

VGSML:
## Decision: **yes — 061 gates the W1 restriction. But the driving reason is DML, not SELECT.**

Worked it through with measurements; the spec update is **#277**.

### Why the answer is yes

The v0.3.0 DML-collapse workstream wants `UPDATE`/`DELETE` to run server-side without the rowid round trip. That removes the one thing that makes a loose predicate survivable: **nothing downstream can re-check a server-side DML**. A predicate matching one row too many does not return an extra row — it *modifies data DuckDB would never have touched*. So for DML the pushed filter must be **exactly** equivalent, and today, on the default `_CI_AS` collation, a string predicate is not. That is 061's job, and it is why the edge exists.

Worth stating plainly because "061 gates W1" read as a SELECT-performance story badly undersells it: **rowid is not automatically safe either.** It is safe today only because a PK is unique *under the server's own comparison*, so `pk = value` can't hit two rows. Lean on that without noticing and it breaks the moment any rowid comparison happens client-side with binary semantics.

### But your #276 finding is narrower than it looks — it's a *superset*

Measured on live SQL Server 2025:

| | rows |
|---|---|
| server's `LIKE N'n%'` | 2 |
| rewritten range `>= 'n' AND < 'o'` | **3** (adds `ñu`) |
| DuckDB's own exact answer | **1** (`nero`) |

The range is a **superset** of DuckDB's answer, not a different answer. So under a "push the native predicate, re-check exactly in DuckDB" rule it is not a bug — the client net drops `Nero` and `ñu` and the result is DuckDB's.

**Which means your restriction was not wrong; it was missing the re-check.** DuckDB 2.0 does not re-check a `TableFilter` behind a `filter_pushdown` scan (the contract in `CLAUDE.md`), so once your branch turned those predicates into `EXPRESSION_FILTER`s, nothing re-applied them. The spec-069 client net already does exactly this job for *refused* filters; it needs arming for "pushed, but not authoritative". That is a small, well-understood change — not all of 061.

### The split that makes it cheap (§ 4 of #277)

- **SELECT** — push only the **native** predicate. It stays SARGable (Index Seek), the plan shape and cardinality estimate are untouched, and DuckDB does the exact pass. Cost: transferring the superset delta, a handful of rows on a selective predicate.
- **DML** — emit **both** (`native AND … COLLATE Latin1_General_BIN2`, plus a sentinel on equality for trailing spaces, which no collation fixes). Measured: the pair keeps the **Index Seek** and adds a residual filter; `COLLATE` *alone* degrades to an Index Scan, which is why it must be added, never substituted. The duplicate-predicate estimate skew is near-free here — a single-table DML has no join order to lose.
- **`<>` / `NOT LIKE`** — negation inverts the inclusion, so the server returns a **subset** and no re-check recovers it. These never collapse: client-side for SELECT, rowid for DML.

### For the roadmap text

Record the edge as `061 → W1 restriction → planner-visible filters → DML-collapse + join/agg`, and retitle the 061 row — it is no longer ORDER-BY-only. I deliberately left `docs/roadmap-v0.3.0.md` untouched in #277 since this PR is already editing it; fold it in here however reads best.

Also: you were right in #276 that your #269 objection didn't reproduce, and right to record it. Appreciated — that's the kind of correction that saves the next person a week.

oluies:
Folded your decision in — `d1ec279`, and retitled the PR since the question it was named after is answered.

- **Edge recorded** as `061 → W1 restriction → planner-visible filters → DML-collapse + join/agg`, with your DML reasoning as the stated *why*, not mine. "Nothing downstream can re-check a server-side DML — it modifies data DuckDB would never have touched" is the sentence that makes it land; my SELECT-performance framing genuinely did undersell it.
- **061 row retitled** to collation-**faithful**, no longer ORDER-BY-only, including the `native AND … COLLATE …_BIN2` + trailing-space sentinel shape and *why it is added beside rather than substituted* (the seek). Marked as a prerequisite rather than independent.
- **Rowid-is-not-automatically-safe** carried into the section — that one would have been lost otherwise, and it is the kind of thing someone leans on without noticing.
- **My #276 "blocked" conclusion corrected in place** rather than left standing, with the superset framing.
- W1 row: "blocked" → "measured, gated on 061".

The § 4.1 semantics question is recorded as **unanswered**, with a note not to design relocation-vs-reduction on top of it until it is settled — detail on #277. If you'd rather this PR not carry an open question at all, I can drop that paragraph and let #277 own it entirely; I left it in because the recon list it contradicts lives in *this* file.

Ready to merge whenever — still needs an approving review, and I can't approve my own.

oluies:
Pushed `f98aa58` (rebased onto `main`, so it now includes #276's test that this text cites). Four roborev reviews on this branch; all four addressed. One of them says I got a material claim wrong, and it was right — flagging that first because it partly underpinned the decision recorded here.

## The "costs nothing in plan shape" claim was wrong

I measured that `LIKE N'n1%'` and `>= N'n1' AND < N'n2'` both produce an Index Seek, and concluded refusing the rewrite was free. That measurement is real but answers a different question: **both forms would be pushed.** Under the W1 restriction the LIKE form isn't on the table at all.

Verified live: `ComplexFilterPushdown` claims `prefix()` **today** and emits `[name] LIKE N'n%'`, so `TryPushdownPrefixFilter` never sees it. Stop claiming it there — which is the entire point of W1 — and the combiner rewrites to a range first. So "refuse the range" means **nothing is pushed**, the client net applies the predicate under DuckDB's binary comparison, and the Index Seek is exactly what's lost.

**And refusal is not semantics-neutral either.** Measured on a `_CI_AS` column: the same predicate returns **2 rows pushed, 1 client-side**. Refusing breaks the native contract from the other side, just as arming the net would.

**The decision survives — native stays the default — but it means mechanism (a):** `ComplexFilterPushdown` keeps claiming string-pattern expressions (today's behaviour; seek and semantics kept), and the trade is **planner visibility, not plan shape**. String comparisons cannot become planner-visible without 061. The section says that now instead of the opposite.

## The refuted objection was still live in the code

Three of the four reviews raised this independently, and it's the one with real downstream cost. `table_scan.cpp`'s reachability note and spec 070's W1 outcome **both still argued** that restricting `ComplexFilterPushdown` would lose pushdown for the combiner-gated shapes, and both still called the measurement pending — while the roadmap table asserted the opposite as tracked fact. Whoever picked up that unassigned row would have read the spec and re-derived the refuted version. Both now record that it did not reproduce, and name the real blocker.

## Also

- `DATAMODEL.md` still carried the two claims the roadmap had just corrected — the unscoped *"cannot see any predicate"* and `top_n` as a *"hard `max_cardinality`"* — so the new pointer sent readers straight to the uncorrected copy. Fixed there and in `table_scan.cpp`; `max_cardinality` is now described as reported-but-not-consumed, with the reason.
- That pointer named a heading that doesn't exist. It's Layer 4 (Cache & registries).
- The 061 row said *"spec on main; widened in #277"* — #277 is **open**, and the spec on main is still ORDER-BY-only.
- Blast radius understated: `NativeDuckDBType` returns a plain type for **any** `max_length <= 0`, so `text`/`ntext`/`varchar(max)` on pre-existing user tables are exposed too — not just what CTAS creates.
- *"the recon list below"* — it's above.
- The superseded "missing the re-check" paragraph was left standing two lines above the decision that voids it.

Build clean; suite 168 cases / 5297 assertions; `make test-cpp` 17/17.

VGSML (approved):

Issue #272, opened 2026-08-21
Pushed string comparison adopts SQL Server collation/padding semantics: same query returns different rows depending on whether the filter pushed


VGSML:
Found by @oluies while reviewing #269 (spec 070 W1). **Pre-existing** — the constant-comparison encoding path is untouched by that PR — and distinct from #242 (which is about pushed-filter *function mappings*), so tracking it separately.

## Symptom

SQL Server compares strings with **collation** (case/accent-insensitive by default) and **trailing-space insensitivity**; DuckDB's `=` is exact. When a string comparison is pushed, the filter is **removed** from the DuckDB plan and the server's looser semantics win. When the same predicate can't push (e.g. behind a `LIMIT`), DuckDB evaluates it exactly. So the answer depends on whether the filter pushed.

With one row holding `N'ab '` (trailing space):

| predicate | pushed as | remote (pushed) | local copy (exact) |
|---|---|---|---|
| `a = 'ab'` | `([a] = N'ab')` | `[1]` | — |
| `a <= 'ab'` | pushed | `[1, 4]` | `[4]` |
| `a <> 'ab'` | `([a] <> N'ab')` | `[2, 4]` | `[1, 2, 4]` |
| `a LIKE 'ab'` | `([a] = N'ab')` | `[1]` | — |
| `a IN ('ab','zz')` | *not pushed* | — | — |

```sql
SELECT count(*) FROM t WHERE a = 'ab';                        -- 1  (pushed, server padding)
SELECT count(*) FROM (SELECT * FROM t LIMIT 100) WHERE a='ab'; -- 0  (client, exact)
```

`=` and `IN` contradict each other on the same data because one pushes and the other doesn't; `LIKE` with a wildcard-free pattern is rewritten to `=` and inherits the padding.

## Relationship to #242

Same divergence *class* (pushed T-SQL ≠ DuckDB semantics) but a different code path — #242 is the function-mapping audit (`length`→`LEN`, `dayofweek`/`week`, `/`), which #269 addresses. This one is the constant/`LIKE`/`IN` comparison path. #242 should **not** be considered to cover it; keep them separate.

## Preliminary decision (for discussion — this warrants its own spec)

Keep pushdown on `=` (it's the most common predicate and drives index seeks), and restore correctness with a client re-check, gated by collation:

[...]

PR #276, opened 2026-08-23
test: pin that a pushed LIKE agrees with the server's LIKE, not with a range rewrite

oluies:
@VGSML — this came out of running the measurement spec 070's W1 outcome asks for. **The measurement's answer is "no, not as-is", and this test is why.**

## What I measured

I restricted `ComplexFilterPushdown` to defer single-column predicates to `pushdown_expression`, so they survive as EXPRESSION_FILTERs in `get.table_filters`. That is the change #269 discussed and #275 recommends measuring.

**Most of it went well.** Against a live server:

- No pushdown lost — every shape DuckDB's combiner gates still reached the server (`IN` at the 6-value threshold, `year(dt)=2024` alongside a second predicate, `LIKE`).
- The planner could finally *see* predicates: `EXPLAIN` showed `Filters: id < 100` on the scan and applied real selectivity, instead of the string-shaped blind spot described in #275.
- Full integration suite: **168 cases, 5297 assertions, green.**

My stated objection in #269 — that the combiner's gates (`volatile`, oversized `IN`, `CanThrow() && filters.size() > 1`) would silently lose pushdown — **did not reproduce.** That reasoning was wrong.

## What killed it

DuckDB rewrites a prefix predicate into a range: `prefix(col,'n')` → `col >= 'n' AND col < 'o'`. Exact for DuckDB's binary string comparison. **Not** exact for SQL Server, where the collation decides the ordering.

On `SQL_Latin1_General_CP1_CI_AS` — SQL Server's **default** collation — `'ñu'` sorts between `'n'` and `'o'`:

| query | rows |
|---|---|
| server's own `LIKE N'n%'` | **2** (`nero`, `Nero`) |
| the range `>= N'n' AND < N'o'` | **3** (adds `ñu`) |
| `main` today | **2** ✅ |
| with the restriction | **3** ❌ |

Silent wrong rows, on the default collation. And **the entire 168-case suite passed in that state** — nothing covered it.

## So this PR is just the test

No behaviour change. It pins the invariant on `main`, where the LIKE is encoded by `ComplexFilterPushdown` and reaches the server *as a LIKE*. Verified it fails (line 51) against the restricted build.

It asserts against **the server's own answer**, not a fixed count, so it holds under any collation — BIN2 answers 1, CI_AS answers 2, and either way the pushed query must agree with the unpushed one.

## What this means for the roadmap

The restriction is not dead, but it now has a **prerequisite**: the encoder must refuse — or correctly translate — DuckDB's range-rewritten prefix filters before those predicates may become TableFilters. That is spec **061** (collation-aware pushdown) territory, which is already on the roadmap and independent.

So the dependency edge is the other way around from what #275 assumes: **061 gates the W1 restriction, which gates planner-visible filters, which is what join/agg pushdown actually wants.** I'd suggest recording that. The measurement branch is `spec/070-w1-measure-shadowing` if you want to reproduce any of it.

### DIFF
--- a/test/sql/catalog/like_pushdown_collation.test
+++ b/test/sql/catalog/like_pushdown_collation.test
@@ -1,0 +1,81 @@
+# name: test/sql/catalog/like_pushdown_collation.test
+# description: A pushed LIKE must agree with the SERVER's own LIKE, not with a range rewrite
+# group: [mssql]
+
+require mssql
+
+require-env MSSQL_TESTDB_DSN
+
+# =============================================================================
+# WHY THIS EXISTS
+#
+# DuckDB rewrites a prefix predicate into a RANGE: `prefix(col, 'n')` becomes
+# `col >= 'n' AND col < 'o'` (FilterCombiner::TryPushdownPrefixFilter). That is
+# exact for DuckDB's own binary string comparison. It is NOT exact for SQL
+# Server, whose collation decides the ordering:
+#
+#   on SQL_Latin1_General_CP1_CI_AS — SQL Server's DEFAULT collation —
+#   'ñu' sorts BETWEEN 'n' and 'o', so the range MATCHES it
+#   while `LIKE N'n%'` does NOT.
+#
+# Measured: 4 rows ('nero', 'Nero', 'ñu', 'oslo'), `LIKE 'n%'` -> 2 on the
+# server, the range -> 3. A scan that pushed the range would silently return an
+# extra row.
+#
+# Today the LIKE is encoded by ComplexFilterPushdown and reaches the server AS a
+# LIKE, so this passes. It is pinned because any change that lets DuckDB's
+# TableFilter machinery see these predicates — which is desirable for other
+# reasons, it is how the planner gets selectivity — makes the range rewrite
+# reachable and this WRONG. The full suite passed in exactly that state; nothing
+# else covers it.
+#
+# The assertion is written against the SERVER'S OWN ANSWER rather than a fixed
+# count, so it holds under any collation: a BIN2 database answers 1, a CI_AS
+# database answers 2, and either way the pushed query must agree.
+# =============================================================================
+
+statement ok
+ATTACH '{MSSQL_TESTDB_DSN}' AS lc (TYPE mssql);
+
+statement ok
+SET mssql_exec_invalidate_cache = true;
+
+statement ok
+CREATE OR REPLACE TABLE lc_src AS SELECT * FROM (VALUES ('nero'), ('Nero'), ('ñu'), ('oslo')) v(name);
+
+statement ok
+COPY lc_src TO 'lc.dbo.LikeCollation' (FORMAT bcp, CREATE_TABLE true, replace true);
+
+# The invariant: pushing LIKE must not change which rows come back, whatever the
+# server's collation says. Both sides ask the same server the same question.
+query I
+SELECT (SELECT count(*) FROM lc.dbo.LikeCollation WHERE name LIKE 'n%')
+     = (SELECT c FROM mssql_scan('lc', 'SELECT COUNT(*) AS c FROM dbo.LikeCollation WHERE name LIKE N''n%'''));
+----
+true
+
+# Same for a suffix and an infix, which take the same encoder path.
+query I
+SELECT (SELECT count(*) FROM lc.dbo.LikeCollation WHERE name LIKE '%o')
+     = (SELECT c FROM mssql_scan('lc', 'SELECT COUNT(*) AS c FROM dbo.LikeCollation WHERE name LIKE N''%o'''));
+----
+true
+
+query I
+SELECT (SELECT count(*) FROM lc.dbo.LikeCollation WHERE name LIKE '%er%')
+     = (SELECT c FROM mssql_scan('lc', 'SELECT COUNT(*) AS c FROM dbo.LikeCollation WHERE name LIKE N''%er%'''));
+----
+true
+
+# And the accented row is the one that separates the two forms: it must NOT be
+# returned by a prefix match, on any collation where LIKE excludes it.
+query I
+SELECT count(*) FROM lc.dbo.LikeCollation WHERE name LIKE 'n%' AND name = 'ñu';
+----
+0
+
+statement ok
+SELECT mssql_exec('lc', 'DROP TABLE dbo.LikeCollation');
+
+statement ok
+DETACH lc;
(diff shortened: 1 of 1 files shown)


oluies:
Pushed f9f95fa — roborev job 1131 reviewed this branch and found the test could not fail, three separate ways. All three were right, so the file has changed substantially.

**Tautological suffix/infix cases.** Both sides were the same server evaluating the same LIKE, and the patterns I picked could not separate the two semantics even in the regression state: `'%o'` matches `nero`/`Nero`/`oslo` under binary *and* CI_AS (3 = 3), `'%er%'` matches `nero`/`Nero` under both (2 = 2). Now upper-case, and measured: server `'%ER%'` = **2** vs DuckDB binary **0**; server `'%O'` = **3** vs binary **0**. They now separate "the server answered" from "the client answered".

**The whole file went vacuous under a binary collation.** `_BIN2` sorts `'ñu'` *after* `'o'`, so `LIKE` and the range agree and the rewrite this exists to catch would sail through. `mssql_ctas_text_type` defaults to NVARCHAR, so the column inherits the *database* default — whichever server the suite points at silently decides whether this file tests anything. My header said it "holds under any collation", which is true and was the wrong claim: holding is not detecting. There is now a guard assertion that the server's own LIKE count differs from its range count (2 vs 3 here), so a server where this cannot discriminate fails loudly instead of reassuring.

**The `AND name = 'ñu'` case is gone, not repaired.** It asserted a hard-coded `0` while the file's own rule is server-relative; it is *wrong* on an accent-insensitive collation, where the server's LIKE does match `'ñu'`; and in the regression it targeted, DuckDB can fold `name >= 'n' AND name < 'o' AND name = 'ñu'` to empty under its own ordering without consulting the server — so it would read 0 in both states. It only looked like coverage. The prefix assertion carries the real signal.

Re-verified: still fails on the restricted build (line 69, the prefix invariant), passes on `main`. Full suite **169 cases / 5307 assertions** green.

oluies:
Pushed d531b61 — roborev job 1133 caught that the guard I added in the last round measured the wrong pair. It was right, and following the fix turned up something worth knowing.

**The guard is gone.** It compared the server's `LIKE` against the server's *range*, when what the assertions need is that the server's answer differs from **DuckDB's binary** answer. Wrong in both directions: false green on `Latin1_General_CS_AS` (upper-case patterns match nothing on either side — all three assertions vacuous while the guard says the environment is fine) and false red on `Latin1_General_CI_AI` (prefix discriminates perfectly, but LIKE and range agree). It also only probed the accent-ordering the *prefix* case needs, never the case-insensitivity the suffix/infix cases need.

Taken the review's other suggestion instead: **pin the collation rather than probe it** — explicit `CREATE TABLE ... COLLATE Latin1_General_CI_AS` (the `varchar_encoding.test` pattern) and `COPY` into it. Nothing left to probe. The pin itself is asserted against `sys.columns`, so a server that substituted its own collation fails loudly instead of quietly.

**Pinning the collation was not sufficient, and that's the interesting part.** The first pinned version used `NVARCHAR(100)` and **passed on the restricted build** — the regression went undetected.

Spec 060 reports a length-carrying string column as `MSSQL_NVARCHAR(n)`, so DuckDB reaches the VARCHAR overload through an implicit no-op cast. That makes the column ref a `BOUND_FUNCTION`, and `FilterCombiner::TryPushdownPrefixFilter` requires a bare `BOUND_COLUMN_REF` — so **those columns are accidentally immune to the rewrite**. Measured on the restricted build:

| column | pushed SQL |
|---|---|
| `NVARCHAR(100)` → `mssql_nvarchar(100)` | `[name] LIKE N'n%'` ✅ |
| `NVARCHAR(MAX)` → plain `varchar` | `[name] >= N'n' AND [name] < N'o'` ❌ |

The fixture is now `NVARCHAR(MAX)` — the shape at risk, and the shape COPY and CTAS create by default. That also narrows the blast radius of the underlying hazard for whenever #275's dependency question gets settled: it's the untyped-string columns that are exposed, not every string column.

Re-verified both directions: fails on the restricted build, passes on `main`. Full suite **169 cases / 5310 assertions**.

VGSML:
Good test — and the finding behind it turned out to be the useful kind: it's real, and it's *narrower* than "the restriction is dead".

I reproduced your case on a live server and measured what DuckDB itself would answer:

| | rows |
|---|---|
| server's own `LIKE N'n%'` | 2 (`nero`, `Nero`) |
| the rewritten range `>= N'n' AND < N'o'` | **3** (adds `ñu`) |
| **DuckDB's own exact answer** | **1** (`nero`) |

The range is a **superset** of DuckDB's answer, not a contradictory one. That reframes it: with the pushed predicate still re-checked in DuckDB, the client net drops `Nero` and `ñu` and the answer is DuckDB's — correct, and the seek still happened server-side.

So the restriction you measured isn't blocked; it was **missing the re-check**. DuckDB 2.0 does not re-apply a `TableFilter` behind a `filter_pushdown` scan (the contract we recorded in `CLAUDE.md`), so the moment your branch turned single-column predicates into `EXPRESSION_FILTER`s, nothing was left to apply them exactly. The spec-069 net already does this for *refused* filters — it needs arming for "pushed, but not authoritative". Much smaller than all of 061.

Where 061 genuinely does gate things is **DML**: a server-side `UPDATE`/`DELETE` has no client pass, so one row too many *modifies data DuckDB would never have touched*. There the exact predicate must be in the T-SQL. Measured for that path: `COLLATE Latin1_General_BIN2` reproduces DuckDB's semantics including your range case (3 → 1), trailing spaces need a sentinel because no collation fixes them, and the forced predicate has to be **added** beside the native one — `COLLATE` alone turns the Index Seek into an Index Scan, while the pair keeps the seek plus a residual filter. All of that is written up in **#277**.

On this PR: **approve from me.** It pins a real invariant on `main`, it asserts against the server's own answer rather than a fixed count (so it survives any collation), and it's the coverage whose absence let a 168-case suite pass in a wrong state. Worth keeping even after the client-net change lands — at that point it becomes the test that proves the re-check is actually running.

One suggestion, optional: a companion case asserting the *pushed* result equals the *un-pushed* one (same query behind a `LIMIT` to block pushdown) would catch the push/no-push disagreement class directly — that's issue #272's shape, and it's the assertion I'd want guarding the client-net change.

VGSML (approved):
Approving: pins a real invariant on main, asserts against the server's own answer rather than a fixed count, and covers the gap that let a 168-case suite pass in a wrong state. See the comment for how this interacts with #277.

PR #277, opened 2026-08-24
spec 061: widen to collation-faithful pushdown — answers #275's question

VGSML:
Spec-only. This is @VGSML's answer to the decision #275 asks for, worked out with the measurements to back it.

**Short answer to #275: yes, 061 gates the W1 restriction — but the reason is DML, not SELECT.** And the gate is narrower than it first looked, because @oluies's #276 finding turns out to be a *superset*, which the client net already knows how to handle.

## What changed in the spec

061 was "push ORDER BY only when the collation already agrees" — correct but passive: on a default `_CI_AS` install nothing qualifies. It now covers the one question asked in three places — *when may a string comparison run on SQL Server, given the server compares by collation and DuckDB compares by bytes?*

### Measured (live SQL Server 2025, § 3)

| predicate | server default | `COLLATE …BIN2` | DuckDB |
|---|---|---|---|
| `= 'Nero'` | 2 | **1** ✅ | 1 |
| `LIKE 'n%'` | 2 | **1** ✅ | 1 |
| **range `>= 'n' AND < 'o'`** (#276) | **3** (`ñu`) | **1** ✅ | 1 |
| `'ab' = 'ab '` | true | **true — not fixed** ❌ | false |
| `(a+'~') = ('ab'+'~') COLLATE …BIN2` | — | **1** ✅ | 1 |

So `COLLATE …_BIN2` fixes case/accents/ordering/LIKE **and the rewritten range #276 found**. Trailing spaces are not a collation property at all — a sentinel fixes those (that's #272's case).

### The SARGability catch, and the way around it (§ 3.2)

| emitted | plan |
|---|---|
| `a = 'nero'` (today) | Index Seek |
| `a = 'nero' COLLATE …BIN2` **alone** | **Index Scan** ⛔ |
| `a = 'nero' AND a = 'nero' COLLATE …BIN2` | **Index Seek** + residual ✅ |
| `a LIKE 'n%' AND a LIKE 'n%' COLLATE …BIN2` | **Index Seek** ✅ |
| `a = 'ab' AND (a+'~') = ('ab'+'~') COLLATE …BIN2` | **Index Seek** ✅ |

Forced collation must be **added**, never substituted.

## The rule differs by who can re-check (§ 4)

- **SELECT** — push only the **native** predicate (SARGable, superset); DuckDB applies the exact one via the spec-069 client net. No forced `COLLATE`, so seek, plan shape and cardinality estimate are all untouched. **This is what unblocks the W1 restriction**: under this rule #276's 3-vs-2 is the expected superset, and the missing piece was the re-check (DuckDB 2.0 does not re-check a `TableFilter` behind a `filter_pushdown` scan), not the restriction.
- **DML-collapse** — nothing can re-check a server-side `UPDATE`/`DELETE`; one row too many *modifies data DuckDB would not touch*. So both predicates go server-side. The duplication's usual cost (a skewed estimate) is near-free on a single-table DML, which is why the expensive form is confined here.
- **ORDER BY** — unchanged: `TOP k` under the wrong order returns the wrong *set*, and no re-check recovers it.
- **`<>` / `NOT LIKE`** — superset property inverts, so these never collapse: client-side for SELECT, rowid for DML.

## Recorded costs, not hidden

- It **flips `collation_filter.test`'s contract** (server semantics → DuckDB's). A decision to take deliberately and rewrite the test, not a test to refit.
- The prefix superset property is **argued, not proven** for collations with expansions/ignorable characters — gated per family, unproven defaults to "don't push".
- `NVARCHAR` `_BIN2` compares UTF-16 code units vs DuckDB's UTF-8 — agree across the BMP, disagree above U+FFFF.
- The sentinel changes expression type/length (`CONVERT_IMPLICIT` in the plan).

## Not touched on purpose

`docs/roadmap-v0.3.0.md` — #275 is already editing it and this would conflict. The 061 row wants retitling (it is no longer ORDER-BY-only) and the edge `061 → W1 restriction → DML-collapse` recording; worth folding into #275 instead. Issue #272 is subsumed by § 4.1 + the sentinel.

### DIFF
--- a/specs/061-collation-aware-order-pushdown/spec.md
+++ b/specs/061-collation-aware-order-pushdown/spec.md
@@ -272,6 +272,15 @@ the `_BIN2` case § 2 already allows for free. So:
 - **Mixed keys** → the rule is per key; a forced key and a native key can appear
   in one `ORDER BY` since each is independent.
 
+**The transformation must stay local to a key, and that is a hard constraint,
+not a preference.** An `ORDER BY` carries several keys, so anything emitted has
+to compose by concatenation: `ORDER BY a COLLATE …, b COLLATE …, c` is still one
+clause and still one plan. A `COLLATE` suffix has that property; a rewrite of the
+statement's *shape* does not. It is the structural reason the two-branch NULL
+form of D3a is deferred rather than merely descoped — partitioning on the
+null-ness of n keys needs 2^n branches, so the trick that is free for one key is
+unusable for three.
+
 Unchanged: the existing NULL-ordering check still gates, ties remain
 tie-broken by the server (already true for non-string keys today, so not new
 here), and the above-BMP caveat of § 5's D1 applies to the forced key exactly as
@@ -374,8 +383,16 @@ reports. Nothing new has to be fetched.
     its own collation. This *changes results* (that is the point: it is the
     SSMS-like ordering, and it makes ORDER BY agree with the native filter
     semantics of § 4.1), and it buys the ordered-index path on any collation, no
-    sort at all. It is the knob for someone who wants the server's linguistic
-    order deliberately.
+    sort at all.
+
+    **Under this flag every gate drops away, including the NULL one.** The user
+    has asked for the server's ordering, and where the server puts NULLs *is*
+    the server's ordering — so declining a nullable `ASC` key here would be
+    refusing to deliver the thing that was requested. Collation is likewise
+    irrelevant: nothing is being matched against DuckDB. So the flag means
+    **push every key**, no collation test, no `IsNullOrderCompatible` call, and
+    the ordered index is used throughout. It is both the fastest mode and the
+    simplest one to implement — a straight emit.
 
   The two compose in one direction only: the native flag chooses **how** a key
   is pushed, never **whether**. With `mssql_order_pushdown = false` nothing is
@@ -497,6 +514,10 @@ O4. `SET mssql_order_pushdown_native_collation = true` on a `_CI_AS` key emits n
     "on" row (`_x | Ähre | Apple | apple`), which is a different *set* under
     `LIMIT`. The divergence is the feature; it is pinned so it cannot regress
     into the faithful path unnoticed.
+O4a. Under the same flag a **nullable `ASC`** key pushes too — the gate that
+    declines it by default does not run here — and returns the server's NULL
+    placement (NULLs first). Asserted because it is the one place where the flag
+    widens *what* is pushed rather than only *how*.
 O5. NULL handling is unchanged (D3a): a `NOT NULL` string key of any collation
     now pushes, a nullable key sorted plain `ASC` is still declined, and a
     nullable `DESC` key still pushes. Asserted as behaviour, so that widening the
--- a/specs/061-collation-aware-order-pushdown/spec.md
+++ b/specs/061-collation-aware-order-pushdown/spec.md
@@ -382,13 +382,60 @@ reports. Nothing new has to be fetched.
   pushed no matter what the native flag says, and that must be asserted — an
   opt-in that silently re-enables a disabled pushdown is the worst of both.
 
-- **D3a NULL ordering is untouched, and is now the only blanket decline.** The
-  existing check (§ 1) refuses a nullable key because SQL Server sorts NULLs
-  first in ASC and DuckDB sorts them last. Worth noting the same forcing trick
-  would fix it — `ORDER BY CASE WHEN k IS NULL THEN 1 ELSE 0 END, k COLLATE …` —
-  at the cost of the ordered-index path, exactly as the collation force does.
-  Deliberately **not** in this spec's scope: it is a second, independent
-  widening, and bundling it would make the ORDER BY change impossible to bisect.
+- **D3a NULL ordering: split it server-side, do not put it in the sort key.**
+  The existing check (§ 1) declines a nullable key outright because SQL Server
+  sorts NULLs first in ASC and DuckDB sorts them last. An earlier draft of this
+  decision said the fix — moving the NULL test into the ORDER BY as
+  `CASE WHEN k IS NULL THEN 1 ELSE 0 END` — would cost the ordered-index path.
+  **Measured, that is wrong**, and the working form is a `UNION ALL` of two
+  bounded `TOP`s rather than a compound sort key:
+
+  ```sql
+  SELECT TOP k … FROM (
+      SELECT TOP k …, 0 AS g FROM t WHERE key IS NOT NULL ORDER BY key
+      UNION ALL
+      SELECT TOP k …, 1 AS g FROM t WHERE key IS NULL
+  ) x ORDER BY g, key
+  ```
+
+  On a nullable `_BIN2` key over 200 000 rows the plan contains **no Sort
+  operator at all** —
+
+  ```text
+  Top(10)
+    └─ Merge Join(Concatenation)
+         ├─ Top(10) → Index Seek(key IsNotNull) ORDERED FORWARD
+         └─ Top(10) → Index Seek(key = NULL)    ORDERED FORWARD
+  ```
+
+  — because both branches arrive already ordered and the server concatenates
+  them by merge. **~0 ms, against ~6–7 ms for the `CASE`-in-ORDER-BY form**,
+  which degrades to `Sort(TOP k)` over an unordered Index Scan. `IS NOT NULL`
+  becomes a *seek predicate*; the `CASE` expression cannot, which is the whole
+  difference.
+
+  Two consequences:
+
+  - **A nullable key stops being a blanket decline.** For a `_BIN2` or
+    non-string key it now pushes on the free path — no sort — where today it
+    does not push at all.
+  - **For a forced-collation key the NULL split is free anyway**, because that
+    query is already paying a `Sort(TOP k)` for the collation: measured, the
+    non-NULL branch keeps its `Index Seek(IsNotNull) ORDERED FORWARD` feeding
+    that same bounded sort.
+
+  The naive shapes both fail and are recorded so they are not retried: putting
+  the `CASE` in the sort key loses the seek (above), and wrapping an
+  index-ordered subquery in an outer sort — `SELECT … FROM (SELECT TOP 100
+  PERCENT … ORDER BY key) x ORDER BY CASE …` — is **flattened by the optimizer
+  into exactly the same plan as the plain `CASE` sort**, measured identical.
+  T-SQL does not preserve a subquery's order without a bound, so the ordering
+  has to be re-expressed as something the optimizer can turn into a seek, which
+  is what the two-branch form does.
+
+  Still deliberately a **separate change from the collation work**: it is a
+  second, independent widening, and landing it in the same commit would make an
+  ORDER BY regression impossible to bisect.
 - **D4 Tests that assert the ORDER, not the row count.** The regression this
   guards is a reordering, so a test that counts rows cannot see it. The shape is
   the § 1 table: the same seven rows under a `_CI_AS` column and under a `_BIN2`
@@ -458,8 +505,12 @@ O4. `SET mssql_order_pushdown_native_collation = true` on a `_CI_AS` key emits n
     "on" row (`_x | Ähre | Apple | apple`), which is a different *set* under
     `LIMIT`. The divergence is the feature; it is pinned so it cannot regress
     into the faithful path unnoticed.
-O5. A nullable key is still declined (D3a), under every combination of the two
-    settings.
+O5. A nullable key pushes via the two-branch form (D3a) and returns DuckDB's
+    NULLS-LAST order; on a `_BIN2` key the plan is asserted to contain **no Sort
+    operator**, since the whole point is that the seek survives. Both naive
+    shapes are pinned as rejected: the `CASE`-in-sort-key form and the
+    `TOP 100 PERCENT` subquery form produce a sort, so the assertion is on the
+    plan, not the rows — the rows are identical either way.
 O6. A mixed ORDER BY (one `_BIN2` key, one `_CI_AS` key) pushes both — each key is
     judged independently, the `_BIN2` one bare and the other forced.
 O7. On Fabric, the default pushes with no `COLLATE` clause — the warehouse
--- a/specs/061-collation-aware-order-pushdown/spec.md
+++ b/specs/061-collation-aware-order-pushdown/spec.md
@@ -414,6 +414,29 @@ reports. Nothing new has to be fetched.
   becomes a *seek predicate*; the `CASE` expression cannot, which is the whole
   difference.
 
+  **The metadata for this already exists and is already read.**
+  `MSSQLColumnInfo::is_nullable` is populated by all three catalog query shapes
+  (`mssql_metadata_cache.cpp`) and the gate consults it today —
+  `mssql_optimizer.cpp:306` feeds it to `IsNullOrderCompatible`. So the two-branch
+  form needs no new metadata and no extra round trip; it needs the gate to emit a
+  different query instead of giving up.
+
+  **And the gate is narrower than "nullable declines".** Reading it against
+  DuckDB's default (`default_null_order = NULLS_LAST`, verified, for ASC *and*
+  DESC):
+
+  | key | DuckDB asks | SQL Server's default | today |
+  |---|---|---|---|
+  | `NOT NULL` | — | — | **pushes** (the check returns early) |
+  | nullable, `DESC` | NULLS LAST | DESC → NULLs last | **pushes** — they agree |
+  | nullable, `ASC` | NULLS LAST | ASC → NULLs first | **declined** |
+  | nullable, `ASC NULLS FIRST` (explicit) | NULLS FIRST | NULLs first | **pushes** |
+
+  So the only shape that loses is **a nullable key sorted plain ascending** —
+  which is `ORDER BY col`, the most common ORDER BY there is. That is what the
+  two-branch form buys back, and why it is worth more than the collation work it
+  is bundled beside.
+
   Two consequences:
 
   - **A nullable key stops being a blanket decline.** For a `_BIN2` or
--- a/specs/061-collation-aware-order-pushdown/spec.md
+++ b/specs/061-collation-aware-order-pushdown/spec.md
@@ -217,14 +217,65 @@ Forms whose superset property is unproven (§ 3.3) do not collapse: they fall
 back to the existing rowid round trip, which targets rows by identity rather
 than by predicate and is correct under any collation.
 
-### 4.3 ORDER BY — § 2's passive rule, plus an option
+### 4.3 ORDER BY and TOP — the widening the same trick buys
 
-Sorting has no superset: `TOP k` under the wrong order returns the wrong *set*,
-and a client re-sort of truncated rows cannot recover it. So § 2 stands — push
-when the key is already `_BIN2` or non-string. `ORDER BY a COLLATE …_BIN2` is
-available and correct, but it forces a server-side sort (the index order no
-longer applies), so it is worth it only when `TOP k` avoids transferring the
-table. Offered as an opt-in, not a default.
+Sorting has no superset to fall back on: `TOP k` under the wrong order returns
+the wrong **set**, and no client re-sort of already-truncated rows recovers it.
+§ 1 measures exactly that — four rows, two disjoint answers.
+
+Note this is **not** in tension with § 4.1's native-by-default. Both apply the
+same principle — *preserve what ships* — to surfaces where what ships differs.
+Filters push today, so their shipped behaviour is the server's semantics. ORDER
+BY does **not** push today (`mssql_order_pushdown` has been `false` since spec
+039), so its shipped behaviour is **DuckDB's** order. Preserving each therefore
+points in opposite directions, and that is a consequence of history, not an
+inconsistency.
+
+**§ 2's rule (push only a `_BIN2` or non-string key) is correct but empty on a
+default install.** The active form of § 3 fixes that, and here it is genuinely
+cheap. Measured, 200 000-row `nvarchar(100) COLLATE SQL_Latin1_General_CP1_CI_AS`
+with an index on the key:
+
+| pushed | plan | 5-run elapsed |
+|---|---|---|
+| `TOP 10 … ORDER BY s` (native) | `Top` → **Index Scan ORDERED FORWARD** — no sort at all | ~0–1 ms |
+| `TOP 10 … ORDER BY s COLLATE …BIN2` | `Top` → **`Sort(TOP 10)`** → Compute Scalar → Index Scan | ~6–7 ms |
+
+Two things follow.
+
+**It is correct.** On § 1's own seven values the collation-forced order is
+DuckDB's, element for element:
+
+```text
+DuckDB           ORDER BY s LIMIT 4 :  Apple | BANANA | Zebra | _x
+server, native CI_AS              :  _x    | Ähre   | apple | Apple
+server, COLLATE …_BIN2            :  Apple | BANANA | Zebra | _x   ← identical
+```
+
+Including `_x` sorting *after* the uppercase letters, which is where the
+linguistic collation departs most sharply.
+
+**It is affordable.** The forced order costs the ordered-index fast path, but
+SQL Server answers with a **bounded `Sort(TOP 10)`** — a priority queue over the
+scan, not a full sort — so the cost is one pass, ~6 ms per 200 k rows. Set that
+against the alternative for a non-`_BIN2` column, which today is *no pushdown at
+all*: transferring the whole column to DuckDB to find ten rows.
+
+And there is no regression to trade away. The ordered-index path exists only
+when the column's own collation already matches the requested order — precisely
+the `_BIN2` case § 2 already allows for free. So:
+
+- **`_BIN2` or non-string key** → push as today, ordered index scan, no sort.
+- **Any other string key** → push `ORDER BY key COLLATE Latin1_General_BIN2`
+  with the `TOP`, taking the bounded sort. Widens TopN pushdown from "nothing on
+  a default install" to every string column.
+- **Mixed keys** → the rule is per key; a forced key and a native key can appear
+  in one `ORDER BY` since each is independent.
+
+Unchanged: the existing NULL-ordering check still gates, ties remain
+tie-broken by the server (already true for non-string keys today, so not new
+here), and the above-BMP caveat of § 5's D1 applies to the forced key exactly as
+it does to a naturally-`_BIN2` one.
 
 ### 4.4 Summary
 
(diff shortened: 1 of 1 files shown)


oluies:
The DML half is convincing and the superset framing is a real improvement on where I left it in #276 — "missing the re-check", not "blocked", is the better description.

**But § 4.1 changes a documented, shipped guarantee, and the spec doesn't say so.**

Under "SELECT — native predicate only, exactness from the DuckDB client net", the client pass applies DuckDB's binary semantics. So the result of every pushed string predicate becomes **DuckDB's answer, not the server's**. Measured just now on a `Latin1_General_CI_AS` column (`nero`, `Nero`, `ñu`, `oslo`):

| query | today | under § 4.1 |
|---|---|---|
| `WHERE name = 'nero'` | **2** (`nero`, `Nero`) | **1** (`nero`) |
| `WHERE name LIKE 'n%'` | **2** | **1** |

That is not a bug in your rule — it is the rule working as written. The problem is what it lands on:

- `website/docs/reading/queries.md:61-63` promises the opposite, to users, today: *"Pushed string predicates follow the **server's comparison semantics**: on a case-insensitive collation, `WHERE name = 'abc'` matches `'ABC'` — exactly what the same query returns in SSMS."*
- The roadmap's **"Settled during recon — do NOT re-litigate"** list opens with: *"Semantic contract: **NATIVE server semantics by default**, strict semantics via annotation."*

So § 4.1 silently inverts the default from native to strict, for every existing user on a CI collation — the common case, since `_CI_AS` is SQL Server's default. `SELECT * FROM t WHERE name = 'abc'` quietly stops matching `'ABC'`.

**Worth noting the re-check cannot rescue the native answer either.** For the range case the correct native-semantics answer is 2, but pushing the range gives 3 and re-checking in DuckDB gives 1. Nothing in the superset-plus-re-check shape produces 2. The only way to keep 2 is what `main` does today: push the **exact** predicate (`LIKE N'n%'`) and refuse DuckDB's range rewrite. So this isn't a detail to patch — superset-and-re-check and native-semantics-by-default are mutually exclusive for SELECT.

Which makes it a genuine fork, and I don't think it's mine to pick:

1. **Adopt strict-by-default** — § 4.1 as written. Then it is a deliberate breaking change and needs saying so: the recon line rewritten rather than silently overridden, `queries.md` updated, a release note, and probably an opt-out for people relying on SSMS-matching behaviour.
2. **Keep native-by-default** — then SELECT must push the *exact* predicate and refuse the inexact rewrite, which is closer to today's behaviour and makes the client net unnecessary for these forms. Costs the SARGability that D7 is protecting.

§ 4.1's "Nothing is forced on the SELECT path" reads as a performance decision; it is also a semantics decision, and that half is currently unstated. Happy to take either branch — but the contract line should move explicitly rather than by implication.

(#276 is merged, so the LIKE-vs-range invariant is pinned on `main` and will go red for whichever of these lands without updating it — which I think is the right kind of tripwire to have.)

oluies:
**Decision (oluies): keep NATIVE server semantics as the default — SELECT pushes the exact predicate and refuses DuckDB's inexact rewrites.** The recon line and `queries.md` stand as written; § 4.1 needs changing rather than the contract.

**And a correction to my own framing.** When I put the two options I said this branch "costs the SARGability that D7 is protecting". That was wrong, and it made the chosen option look more expensive than it is. Measured on a 20 000-row indexed `NVARCHAR(100) COLLATE Latin1_General_CI_AS` column, `SHOWPLAN_ALL`:

| predicate sent | access |
|---|---|
| `name LIKE N'n1%'` | **Index Seek** |
| `name >= N'n1' AND name < N'n2'` | **Index Seek** |

SQL Server converts a prefix `LIKE` into a seek range itself. So refusing the rewrite and pushing the native `LIKE` keeps the seek, the plan shape *and* the semantics — for the prefix case there is no trade at all. D7's concern ("nothing is forced on the SELECT path") is satisfied by the exact predicate too.

### What this means for § 4.1

- Push the **exact** predicate (`= N'abc'`, `LIKE N'n%'`, `IN (…)`). Native semantics preserved, seek preserved.
- **Refuse** DuckDB's prefix→range rewrite rather than pushing it and repairing downstream. That is the "encoder must refuse or correctly translate" half — and it is much less than the full `_BIN2` machinery.
- **Do not arm the client net for "pushed but not authoritative" on SELECT.** That was the neat part of your proposal and it is precisely what breaks native: the net applies DuckDB's binary comparison, so it returns 1 where the contract requires 2. The net stays as it is — for *refused* filters only.
- `<>` / `NOT LIKE` are unaffected: still client-side, since the server returns a subset and nothing recovers it.

### One consequence I'd like you to confirm rather than assume

If native semantics are the default *everywhere*, then a server-side `UPDATE`/`DELETE` whose native predicate matches `'ABC'` for `name = 'abc'` is doing what the contract promises — the same rows SSMS would touch. It is not "modifying data DuckDB would never have touched" in a contract-violating sense; it is the documented behaviour, just applied to a destructive statement.

If that reading is right, the `native AND … COLLATE …_BIN2` pair plus trailing-space sentinel is needed only under the **strict annotation**, not on the default DML path — which narrows § 4.2 considerably. If it is wrong, then DML is a deliberate exception to native-by-default and § 4.2 should say so in as many words, because "the default differs between SELECT and DML" is exactly the kind of thing that gets discovered by someone's deleted rows.

Your call — the measurement work in § 3.2 stands either way, it is the scope it applies to that moves.

I'll leave #277 to you and keep the roadmap in sync. #276 is on `main` and pins the LIKE-vs-range invariant, so whichever way § 4.1 is rewritten, that test will hold the line.

oluies:
**Correction to my previous comment — the measurement in it does not support what I claimed.** Please read this before acting on it; #275 has merged with the corrected version (`e123e18`).

I wrote that keeping native semantics "costs nothing in plan shape", citing that `LIKE N'n1%'` and `>= N'n1' AND < N'n2'` both produce an Index Seek. The measurement is real; the inference was wrong. **Both those forms are ones that would be pushed.** Under the W1 restriction the `LIKE` form is not on the table at all:

- Verified live — `ComplexFilterPushdown` claims `prefix()` **today** and emits `[name] LIKE N'n%'`, so `TryPushdownPrefixFilter` never sees it. That is why `main` gets both the seek and the server's semantics.
- Stop claiming it there — the entire point of W1 — and the combiner rewrites to a range *before* offering it. So "refuse the range" pushes **nothing**: the client net applies the predicate under DuckDB's binary comparison, and the seek is what's lost.
- **Refusal is therefore not semantics-neutral either.** Measured on `_CI_AS`: the same predicate returns **2 rows pushed, 1 client-side**. It breaks the native contract from the other direction, exactly as arming the net would.

### What that means for § 4.1

The decision stands — native stays the default — but it implies **mechanism (a)**, and the trade is planner **visibility**, not plan shape:

- **String-pattern expressions must stay claimed by `ComplexFilterPushdown`** until 061 supplies a collation-faithful form. That keeps the exact `LIKE`, the seek, and the semantics. They stay planner-invisible meanwhile — the accepted cost.
- **Non-string predicates need no gate at all** and can be deferred to `pushdown_expression` today.

So W1 is not one blanket change, and § 4.1 shouldn't describe it as one. `specs/070-duckdb-v2-followups/spec.md` and `src/table_scan/table_scan.cpp` now carry that split on `main`.

### Two things still open, both yours

1. **Is DML a deliberate exception to native-by-default?** If native applies everywhere, a server-side `DELETE` matching `'ABC'` for `name = 'abc'` is the documented behaviour, not a violation — and the `_BIN2` pair is needed only under the strict annotation, which narrows § 4.2 considerably. If DML is meant to be strict, § 4.2 should say so outright: a default that differs between SELECT and DML is the kind of thing found through someone's deleted rows.

2. **That answer can withdraw one of the two reasons for `061 → W1`.** The edge currently rests on DML exactness. Answer "native everywhere" and that lapses; only the mechanism-(a) planner-visibility reason survives. Three surfaces encode the edge now (roadmap, spec 070, `table_scan.cpp`), and the merged roadmap says they must be revisited together if the answer comes back that way.

One smaller correction while I'm here: "the default `_CI_AS`" overstates it — `SQL_Latin1_General_CP1_CI_AS` is the US-English install default, and the hazard holds for **any** non-binary ordering. `like_pushdown_collation.test` pins its own collation precisely because the ambient one can't be relied on.

oluies:
**Answer to question 1 (oluies): yes — DML is a deliberate exception.** SELECT native, DML strict. § 4.2 stands as you wrote it: the `native AND … COLLATE …_BIN2` pair plus the trailing-space sentinel is required on the server-side DML path, not merely under the strict annotation.

That also settles question 2 in the direction that *keeps* the edge: the DML-exactness reason for `061 → W1` does **not** lapse, so `061` stays on the critical path with both reasons intact. No revisiting needed of the roadmap row, the order-of-battle edge, or the notes in `specs/070-duckdb-v2-followups/spec.md` and `src/table_scan/table_scan.cpp`.

**The one thing this now requires is loud documentation**, because it is exactly the shape I flagged as dangerous: a default that differs between read and write.

- `SELECT ... WHERE name = 'abc'` matches `'ABC'` on a CI collation — server semantics, as `website/docs/reading/queries.md:61-63` promises today.
- `DELETE ... WHERE name = 'abc'` will **not** delete `'ABC'` — strict semantics.

So the same predicate, on the same column, selects four rows and deletes one. That is defensible (a destructive statement should be the conservative one), but it will surprise anyone who reads the existing contract line — the roadmap's *"NATIVE server semantics by default"* is now only true of reads, and nothing user-facing mentions DML semantics at all.

I'll open a follow-up that qualifies the recon line and adds the DML statement to the docs, so the asymmetry is written down where someone about to run a `DELETE` would see it. Shout if you'd rather it live in #277 instead.

oluies:
Reviewed. The DML analysis (§ 4.2), the `<>`/`NOT LIKE` asymmetry (§ 3.3), and the measurement that `COLLATE` alone degrades the seek while the *pair* keeps it are all convincing, and D5/D6/D7 are the right decomposition. Two things follow — one a correction of mine, one blocking.

## First, a correction I owe you

I said earlier that § 4.1 "silently inverts the default from native to strict" and that the semantics half was "currently unstated". **That was wrong.** Risks says it plainly:

> *"This changes a documented contract… Under § 4.1 the answer becomes DuckDB's (`Apple` only). That is the point — it is what makes pushed and un-pushed agree — but it must be a decision recorded and the test rewritten, never a test quietly fitted to new behaviour."*

That is exactly the right framing, and F1 pins the invariant that motivates it (#272's push/no-push disagreement). I had read § 4.1 without § 7. Apologies for the mischaracterisation.

## The blocking item: the decision went the other way

@oluies decided **native-by-default stands** — SELECT pushes the *exact* predicate and refuses DuckDB's inexact rewrites, rather than pushing a superset and re-checking. Recorded on #275 and now merged to `main` (`3579f63`), including the order-of-battle edge and the W1 row.

So § 4.1 as written proposes the branch that was not taken. The two are mutually exclusive: the native answer is 2 rows, the range gives 3, the re-check gives 1. § 4.2 is unaffected — DML strict is *also* decided, and your "nothing can re-check a server-side DML" reasoning is what carried it.

Worth noting the consequence you'd feel first: under native-by-default, `SET mssql_statistics_cache…`-style opt-outs aside, **`SELECT … WHERE name = 'abc'` keeps matching `'ABC'`** and `collation_filter.test` stands as written. The cost is the one you identify — pushed and un-pushed still disagree, so #272 stays open rather than being closed by this spec.

## A technical finding: § 4.1's W1 claim is necessary but not sufficient

> *"This is also what unblocks the W1 restriction… The missing piece was the re-check, not the restriction."*

The re-check does not restore two pushdown losses the measurement found *after* #276 — both now recorded on `main` (roborev job 1130, verified against source):

1. **`rowid` loses pushdown entirely.** A rowid predicate is single-column so it defers, but `FilterEncoder::Encode` refuses every virtual column (`table_col_idx >= VIRTUAL_COL_START`) while `ComplexFilterPushdown` reached `EncodeColumnRef`, which rewrites rowid to the scalar PK. `WHERE rowid > 100` goes from `WHERE [id] > 100` to a **full table scan**. Rows stay correct — so the client net "fixes" it — but the wire cost is unbounded and no re-check can recover it.
2. **OR-chains, non-dense `IN` and temporal-cast** push through `CreateOptionalExpressionFilter`, i.e. an `optional_filter` scalar function the encoder has no mapping for. It refuses, so the scan pushes **nothing at all** — again correct rows, again a full stream.

Neither blocks W1; both are shapes the deferral must **exclude**. But § 4.1 currently reads as though the re-check is the whole missing piece, and an implementer taking that at face value would ship the two regressions above as "expected superset behaviour". Worth one sentence.

## Suggestion

Rewrite § 4.1 for native-by-default (push the exact predicate; refuse the inexact rewrite; do **not** arm the net for pushed filters), keep § 4.2–4.4 as they are, and add the exclusion list above to whatever § 4.1 becomes. If you'd rather argue for strict-by-default instead, that's a legitimate case — it closes #272, which native-by-default does not — but it needs to go back to @oluies as a contract change with the docs and a release note, not through this spec.

Happy to make either edit if you'd prefer I do it.

VGSML:
Rewritten per your suggestion — `8215e04`. Native-by-default accepted; no argument from me, the shipped contract in `queries.md` is the stronger claim and your arithmetic (2 / 3 / 1) settles it. Thanks also for the self-corrections along the way — the `LIKE`-is-claimed-today observation is what makes the mechanism-(a) split obvious in hindsight.

## § 4.1, rewritten

Push the **exact** predicate; **refuse** the prefix→range rewrite; **do not** arm the client net for pushed filters. W1 recorded as the split rather than a blanket change — string-pattern expressions stay claimed by `ComplexFilterPushdown` until this spec supplies a collation-faithful form, non-string predicates defer today.

The mutual exclusivity is now in the spec as the table, because it is the clearest single statement of why this isn't patchable.

**Your exclusion list is in, and I verified the rowid one against source rather than taking it:** `EncodeColumnRef` tests `COLUMN_IDENTIFIER_ROW_ID` at line 884, *before* the virtual-column gate at 897; `FilterEncoder::Encode` refuses everything `>= VIRTUAL_COL_START` into `unhandled` at 262. So the two paths disagree by construction and deferring a rowid predicate turns `WHERE [id] > 100` into a full scan. Confirmed, listed.

Consequences stated rather than implied: **#272 is not closed** by this design (pushed and un-pushed `=` still disagree — that's the price of the contract, and the issue tracks it), and the read/write asymmetry is called out in § 4.4 with a pointer to your #286.

## New § 4.5 — since we were in the spec anyway

@VGSML asked whether the pushdown surface could be widened while we're here. It can, and the finding is a little pointed: **#242 removed four mappings as having no exact T-SQL form. Three of them do.** Measured DuckDB-vs-server on the same inputs:

| DuckDB | exact T-SQL | evidence |
|---|---|---|
| `/` | `({0} * 1.0 / NULLIF({1}, 0))` | `5/2`→2.5, `7/2`→3.5 both sides. `NULLIF` also defeats something the removal never named: raw `5*1.0/0` **aborts the query** on the server while DuckDB yields `inf`. NULL and `inf` behave identically in predicate position — verified. |
| `week` | `DATEPART(ISO_WEEK, {0})` | 2024-01-04→1, 2024-12-30→1, 2021-01-01→53 on both. DuckDB's `week()` **is** the ISO week; the original mapping just used the wrong datepart. |
| `dayofweek` | `((DATEDIFF(day,'19000107',{0}) % 7) + 7) % 7` | Sun→0, Wed→3, and 1899-05-10→3 *before* the anchor. Anchored on a known Sunday ⇒ **`@@DATEFIRST`-independent**, which is exactly what disqualified `DATEPART(weekday, …)`. The `+7 %7` is load-bearing: without it T-SQL's sign-preserving `%` answers −4. |
| `length` | `LEN({0} + N'~') - 1` | `'ab '`→**3** both sides (naive `LEN` says 2) — the § 3.1 sentinel again. **But `LEN` counts UTF-16 code units**: `'a😀b'` is 3 in DuckDB, 4 on the server. BMP-only, so it ships gated or not at all — that one's a review call, not a measurement. |

Also verified exact and currently unmapped: `abs`, `floor`, `ceiling`, `round` (agrees on half-away-from-zero, on negatives, and in the 2-arg form), `substring` (**including** the `start = 0` edge — both answer `ab`).

SARGability is unchanged by any of it: a function wrapping a column defeats a seek whether it's `LEN(x + N'~') - 1` or the `YEAR(x)` we already push. The gain is not pushing the whole column.

§ 4.5 is deliberately **independent** of § 4.1–4.4 — it needs no collation machinery and could ship separately, or as its own PR if you'd rather keep this spec purely about collation. It's parked here because #242 is where the removals are recorded and the next reader of that issue should find the exact forms beside them.

F4 added to acceptance to pin each restored form on the value that broke the original, including the `b = 0` row that must not error.

VGSML:
Second pass — `92c229c`. @VGSML's calls, both measured before writing.

## § 4.5 narrowed to the simple forms

**Taken:** `abs` → `ABS`, `floor` → `FLOOR`, `ceil` → `CEILING`, `round` → `ROUND`, `substring` → `SUBSTRING`. Direct 1:1, verified on the cases that usually separate implementations — `round` half-away-from-zero, negatives and the 2-arg form; `ceil`/`floor` of a negative; `substring`'s `start = 0` edge (both answer `ab`).

**Left to DuckDB:** `/`, `week`, `dayofweek`, `length`. The reasoning is the one you'd apply yourself: each needs a *compound* rewrite, and a compound expression over a column defeats the index exactly as the naive mapping did, while the queries using them are narrow. So DuckDB computes them over the rows the other predicates already cut down. Their exact forms stay in the spec as a table — the point is that the next reader of #242 shouldn't re-derive them, not that we ship them.

`length` is rejected on correctness as well: `LEN` counts UTF-16 code units, so `'a😀b'` is 3 in DuckDB and **4** on the server. Exact only below the BMP — not a guarantee worth shipping.

## § 4.3 — TopN turned out to be where the real widening is

This was the thinnest section; it now carries something. § 2's rule (push only a `_BIN2` or non-string key) is correct but **empty on a default install** — nothing qualifies. Forcing the key's collation fixes that, and measured on 200 000 rows (`nvarchar(100) CI_AS`, indexed):

| pushed | plan | 5-run elapsed |
|---|---|---|
| `TOP 10 … ORDER BY s` | `Top` → **Index Scan ORDERED FORWARD**, no sort | ~0–1 ms |
| `TOP 10 … ORDER BY s COLLATE …BIN2` | `Top` → **`Sort(TOP 10)`** → Index Scan | ~6–7 ms |

The forced form gets a **bounded** Top-N sort — a priority queue over one pass, not a full sort.

And it is correct, on § 1's own seven values — the set that motivated this whole spec:

```
DuckDB                  : Apple | BANANA | Zebra | _x
server, native CI_AS    : _x    | Ähre   | apple | Apple
server, COLLATE …_BIN2  : Apple | BANANA | Zebra | _x   ← identical to DuckDB
```

including `_x` sorting *after* the uppercase letters, which is where the linguistic collation departs most sharply.

**There is no regression to trade away**, which is what makes this easy: the ordered-index fast path exists only when the column's own collation already matches the requested order — precisely the `_BIN2` case § 2 already takes for free. So `_BIN2`/non-string keys keep the free path, and every *other* string key goes from **no pushdown at all** to a ~6 ms bounded sort instead of transferring the column to find ten rows.

## One thing worth your eye

§ 4.3 now records why it isn't in tension with § 4.1's native-by-default, since that looks contradictory at a glance: both preserve *what ships*, and what ships differs. Filters push today, so their shipped behaviour is the server's semantics. ORDER BY does **not** push (`mssql_order_pushdown` `false` since spec 039), so its shipped behaviour is DuckDB's order. Preserving each points opposite ways — history, not inconsistency. If you read that differently, § 4.3 is the place to push back.

F4 rewritten for the five taken forms, plus a guard that the four rejected ones stay unmapped (so § 4.5's decision can't be undone by someone reading only the exact-forms table). F5 added for the TopN order.

VGSML:
Settings design in — `aed6937`.

The logic @VGSML pointed at is the clean part: `mssql_order_pushdown` shipped disabled for exactly one reason — **it changed results** — and § 4.3's forcing removes that reason. Once the pushed order *is* DuckDB's, element for element, the only remaining question is where the sort runs, which is a performance choice. So the default can flip.

## Two settings

| setting | type | default | meaning |
|---|---|---|---|
| `mssql_order_pushdown` | BOOLEAN *(unchanged type)* | **`false` → `true`** | Push `ORDER BY`/`TOP` **preserving DuckDB's order**: bare key for `_BIN2`/non-string (ordered index scan, free), `COLLATE Latin1_General_BIN2` for any other string key (bounded `Sort(TOP k)`). |
| `mssql_order_pushdown_native_collation` | BOOLEAN *(new)* | `false` | Push **without** forcing — the server orders by its own collation. **Changes results by design**: the SSMS-like ordering, and it makes ORDER BY agree with § 4.1's native filter semantics. Buys the ordered-index path on *any* collation, no sort at all. |

Kept BOOLEAN rather than the three-valued `off`/`safe`/`always` the first draft sketched, for two reasons: an existing `SET mssql_order_pushdown = false` keeps working unchanged, and the three-way version made the user choose between correct and fast — which the measurement says they no longer have to.

**They compose in one direction only.** The native flag chooses *how* a key is pushed, never *whether*. With `mssql_order_pushdown = false` nothing pushes regardless of the native flag — asserted in O3, because an opt-in that silently re-enables a disabled pushdown is the worst of both worlds.

## Acceptance rewritten around the flip

The old O1 said *"with the default setting, generated SQL is byte-identical to today's"* — that is now precisely wrong, so it is replaced by "the default pushes, and returns DuckDB's order asserted element by element on § 1's seven values".

Two that are worth calling out as deliberate:

- **O2 asserts the emitted SQL, not just the rows**, for `_BIN2` and non-string keys: there must be **no** `COLLATE` clause. A regression that forces the collation everywhere would still return correct rows and sail through O1 — it would only show up as a lost index seek. Cost regressions need a cost assertion.
- **O4 pins the native flag's divergence** as § 1's "on" row (`_x | Ähre | Apple | apple`) — a different *set* under `LIMIT`. The divergence is the feature, so it is pinned rather than tolerated, and cannot drift back into the faithful path unnoticed.

## One thing left deliberately undone

**D3a**: with collation handled, the NULL-ordering check becomes the *only* blanket decline — a nullable key is still refused because SQL Server sorts NULLs first in ASC and DuckDB last. The same forcing trick would fix it (`ORDER BY CASE WHEN k IS NULL THEN 1 ELSE 0 END, k COLLATE …`), at the same cost — the ordered-index path. I left it out on purpose: it is a second, independent widening, and bundling it would make the ORDER BY change impossible to bisect if something regresses. Recorded so it isn't lost.

Risks updated too: flipping the default is a behaviour change *even though results don't move* — the sort relocates to SQL Server, so plans, memory profile and where a slow query surfaces all change, and a loaded server absorbs work DuckDB used to do. Results being identical makes it safe, not invisible; it wants a release note.

VGSML:
@VGSML's question — *"what about an index-ordered subquery with the NULL test in an outer sort, right on the SQL Server side?"* — turned out to be worth measuring: **the shape is right, the expression is not, and the working form makes nullable keys free rather than expensive.** So D3a's claim that the NULL fix costs the ordered-index path was wrong. Corrected in `f0bc73d`.

## What fails

| form | plan | 200 k rows |
|---|---|---|
| `ORDER BY CASE WHEN k IS NULL THEN 1 ELSE 0 END, k` | `Top` → **`Sort(TOP 10)`** → Compute Scalar → Index **Scan** | ~6–7 ms |
| `SELECT … FROM (SELECT TOP 100 PERCENT … ORDER BY k) x ORDER BY CASE …` | **flattened by the optimizer into exactly the plan above** — measured identical | ~6–7 ms |

The subquery buys nothing because T-SQL does not preserve a subquery's order without a bound, and the optimizer simply removes it. A `CASE` expression also cannot become a seek predicate — that is the whole difference.

## What works — two bounded `TOP`s, one statement, all server-side

```sql
SELECT TOP k … FROM (
    SELECT TOP k …, 0 AS g FROM t WHERE key IS NOT NULL ORDER BY key
    UNION ALL
    SELECT TOP k …, 1 AS g FROM t WHERE key IS NULL
) x ORDER BY g, key
```

On a nullable `_BIN2` key the plan has **no Sort operator at all**:

```text
Top(10)
  └─ Merge Join(Concatenation)
       ├─ Top(10) → Index Seek(key IsNotNull) ORDERED FORWARD
       └─ Top(10) → Index Seek(key = NULL)    ORDERED FORWARD
```

**~0 ms against ~6–7 ms.** `IS NOT NULL` becomes a *seek predicate*, both branches arrive already ordered, and the server concatenates them by merge. The outer sort has at most 2k rows to consider and disappears entirely on this plan.

## Two consequences

- **A nullable key stops being a blanket decline.** On a `_BIN2` or non-string key it now pushes **on the free path** — no sort — where today it does not push at all. That is a bigger widening than the collation one: `mssql_order_pushdown` currently refuses every nullable key regardless of collation.
- **For a forced-collation key the split is free anyway**, because that query already pays a `Sort(TOP k)` for the collation — measured, the non-NULL branch keeps its `Index Seek(IsNotNull) ORDERED FORWARD` feeding that same bounded sort.

## On testing this

O5 rewritten to assert **the plan**, not the rows: a nullable key must push, and on a `_BIN2` key the plan must contain no `Sort`. The rows are identical under all three shapes, so only a plan assertion catches a regression back to the naive form — same reasoning as O2 for the stray `COLLATE`. Both rejected shapes are pinned in the spec so they are not re-derived and retried.

Still kept as a **separate change** from the collation work — independent widening, and landing both in one commit would make an ORDER BY regression impossible to bisect.

VGSML:
Scoped down per @VGSML — `56c38cc`. **TopN for non-nullable keys; the nullable-ASC case is deferred, not built.**

So `IsNullOrderCompatible` (`mssql_optimizer.cpp:265`) is **not touched** by this spec at all. The collation work simply applies to whatever that gate already passes, which makes the change considerably smaller to review.

## What that gate actually passes

Answering the nullability question directly: yes, we have it — `MSSQLColumnInfo::is_nullable`, populated by all three catalog query shapes and already read at `mssql_optimizer.cpp:306`. Reading the gate against DuckDB's verified default (`default_null_order = NULLS_LAST`, for ASC *and* DESC):

| key | DuckDB asks | SQL Server's default | gate |
|---|---|---|---|
| `NOT NULL` | — | — | **passes** (returns early) |
| nullable, `DESC` | NULLS LAST | DESC → NULLs last | **passes** — they already agree |
| nullable, `ASC NULLS FIRST` (explicit) | NULLS FIRST | NULLs first | **passes** |
| nullable, `ASC` | NULLS LAST | ASC → NULLs first | **declined** |

So this spec widens TopN to **`NOT NULL` string keys of any collation** — today only `_BIN2` qualifies — plus the nullable shapes that already pass. Nullable plain-`ASC` keeps being declined exactly as today: no regression, no new behaviour, nothing new to get wrong.

## The nullable-ASC measurement is kept, not deleted

It's recorded in D3a so nobody re-derives it. Short version, for whoever picks it up: `ORDER BY col` on a nullable column is the single shape that loses — and it's the most common ORDER BY there is — but it's recoverable **server-side** with two bounded branches rather than a compound sort key:

```sql
SELECT TOP k … FROM (
    SELECT TOP k …, 0 AS g FROM t WHERE key IS NOT NULL ORDER BY key
    UNION ALL
    SELECT TOP k …, 1 AS g FROM t WHERE key IS NULL
) x ORDER BY g, key
```

On a nullable `_BIN2` key over 200 000 rows that plan has **no Sort operator at all** — `Top → Merge Join(Concatenation) → two Index Seeks (`IsNotNull` / `= NULL`) ORDERED FORWARD`, **~0 ms** — because `IS NOT NULL` becomes a seek predicate and both branches arrive already ordered.

Both obvious alternatives are written down as dead ends, since they measure ~6–7 ms and look plausible: `CASE WHEN k IS NULL …` in the sort key degrades to `Sort(TOP k)` over an unordered scan, and an index-ordered `TOP 100 PERCENT` subquery under an outer sort gets **flattened by the optimizer into that same plan** (T-SQL doesn't preserve a subquery's order without a bound).

And the follow-up would need no new metadata and no extra round trip — `is_nullable` is already there and already consulted.

O5 now asserts the NULL behaviour is **unchanged**, specifically so that widening the collation rule can't quietly widen the NULL rule alongside it.

VGSML:
Both points in — `0b2b025`. The first one supplied an argument I was missing, and the second closes the design.

## Keep the emit per-key — and that's structural, not stylistic

@VGSML: *an ORDER BY can have several keys, so don't complicate the statement.* Right, and it's worth stating as a constraint rather than a preference: anything we emit has to **compose by concatenation**. `ORDER BY a COLLATE …, b COLLATE …, c` is still one clause and one plan — a `COLLATE` suffix has that property, a rewrite of the statement's *shape* does not.

That is the real reason the two-branch NULL form is deferred, better than "smaller to review": **partitioning on the null-ness of n keys needs 2ⁿ branches.** Free for one key, unusable for three. I'd been treating that as a scope call; it's actually a structural limit on the technique, and the spec now says so.

## Under the native flag, every gate drops — including the NULL one

This is the part that tidies the whole design. With `mssql_order_pushdown_native_collation = true` the user has asked for **the server's ordering** — and where the server puts NULLs *is* the server's ordering. Declining a nullable `ASC` key there would be refusing to deliver the thing that was requested. Collation is equally beside the point, since nothing is being matched against DuckDB.

So the flag means simply: **push every key.** No collation test, no `IsNullOrderCompatible` call, ordered index throughout. It ends up being both the fastest mode and the simplest to implement — a straight emit with no analysis at all.

The two modes now read cleanly against each other:

| | what is pushed | gates | plan |
|---|---|---|---|
| **default** (faithful) | keys the existing NULL gate passes; `COLLATE …_BIN2` on non-`_BIN2` string keys | collation per key, NULL gate unchanged | free path for `_BIN2`/non-string, bounded `Sort(TOP k)` otherwise |
| **native flag** | **every key**, bare | none | ordered index throughout, no sort |

O4a added to pin it: under the flag a nullable `ASC` key pushes and returns the server's NULL placement. Worth its own criterion because it's the one place the flag widens *what* is pushed rather than only *how* — and therefore the one place a careless implementation could leave the gate running and silently under-push.

PR #280, opened 2026-08-25
ci: run build + integration on pull requests, and export the TLS DSN in the CI lane

### LINKED ISSUE #278: TLS integration tests skip in CI: the Makefile export was never mirrored into scripts/ci/integration_test.sh, and check_require_env.sh only inspects the Makefile
## Summary

The four TLS integration files have **never run in CI**. `MSSQL_TEST_DSN_TLS` is exported by the `Makefile` but not by `scripts/ci/integration_test.sh`, which is what the CI lane actually runs — and `scripts/ci/check_require_env.sh`, the guard written to catch exactly this class, only inspects the `Makefile`, so it reports OK.

Affected: `tls_connection.test`, `tls_queries.test`, `tls_parallel.test`, `tls_multipacket.test`.

## Evidence

CI run [32644956651](https://github.com/hugr-lab/mssql-extension/actions/runs/32644956651), `Integration Test (linux_amd64)` — a **passing** job:

```
All tests passed (9 skipped tests, 5174 assertions in 163 test cases)
Skipped tests for the following reasons:
  require-env MSSQL_TEST_DSN_TLS: 4
```

The same suite run locally with the variable set reports **167** cases. The four-case difference is entirely this.

## Why the guard misses it

`Makefile:136` exports the variable, added specifically to fix this after the PR #264 review — the comment there recounts the history:

> `MSSQL_TEST_DSN_TLS` was assigned above but deliberately NOT exported, so the four TLS files … skipped on every run — and `scripts/ci/check_require_env.sh` accepted the bare assignment as proof they did not.

But the CI lane does not go through that Makefile target. `.github/workflows/ci.yml:879` runs `scripts/ci/integration_test.sh`, which exports ten variables at `:31-46` — `MSSQL_TEST_HOST`, `MSSQL_TEST_PORT`, `MSSQL_TEST_USER`, `MSSQL_TEST_PASS`, `MSSQL_TEST_DB`, `MSSQL_TEST_DSN`, `MSSQL_TEST_URI`, `MSSQL_TESTDB_DSN`, `MSSQL_TESTDB_URI`, `MSSQL_TEST_SERVER`, `MSSQL_TEST_CONNECTION_STRING` — and not this one.

And `check_require_env.sh:87` greps for the export in **`Makefile` only**:

```bash
if grep -qE "^export[[:space:]]+${v}([[:space:]]|=|:=|\?=|$)" Makefile; then
```

So the fix landed in one of the two environments that run the suite, and the guard is blind to the other. It currently prints `OK — 30 variables exported or documented-opt-in`.

This is the fourth instance of the class that script's own header enumerates (`MSSQL_TEST_SERVER`, `MSSQL_TEST_CONNECTION_STRING`, the tag-filter one, then `MSSQL_TEST_DSN_TLS` itself) — arriving through a path the guard does not look down.

## Suggested fix

1. Export `MSSQL_TEST_DSN_TLS` in `scripts/ci/integration_test.sh` alongside the others. SQL Server 2022 has TLS on by default with a self-signed certificate — the server the CI service container starts — so there is nothing to opt into, which is the same reasoning as the Makefile comment.
2. Widen `check_require_env.sh` to require the export in **both** `Makefile` and `scripts/ci/integration_test.sh`. A variable exported in one and not the other is precisely the shape that produced this.
3. Consider asserting the case count in the CI lane, not just the exit status — 163 vs 167 is the only signal this gap ever produced, and nothing was watching it.

## Impact

TLS is in the connection path. `tls_multipacket` in particular covers fragmentation behaviour that the packet-size work (spec 055) touched. None of it has CI coverage; it runs only on a developer machine that happens to export the variable.

### LINKED ISSUE #279: Pull requests still never compile the extension or run the integration suite (residual of #212 gap 1)
## Summary

#212 was closed, but only half of gap 1 was fixed. The `Build`, `Smoke Test` and `Integration Test` jobs now run on **merge to `main`** — they still do not run on a **pull request**. So a C++ change is compiled and integration-tested only *after* it lands: detection after the fact rather than prevention.

## Evidence

`.github/workflows/ci.yml` triggers on `pull_request` (`:5-6`), but every heavy job gates it out:

```yaml
# :287  Build            # :715  Smoke Test
if: ${{ (github.event_name == 'workflow_dispatch' && inputs.run_build)
        || github.event_name == 'schedule'
        || github.event_name == 'push' }}

# :776  Integration Test — same, plus && inputs.run_integration_tests
```

`push` is `branches: [main]` (`:3-4`), so on a `pull_request` event all three skip. Confirmed on this week's PRs: the only C++ compiled at PR time is the standalone unit-test job, which builds the libduckdb-free TDS files and never touches `table_scan.cpp`, the codec sources, or the copy path.

The build-on-merge addition is genuinely worth having — the comment at `:281-286` explains it well — but it changes *when you find out*, not *whether a broken PR can be approved*.

## Why it still bites

Concretely, from this week: #269, #270 and #274 all changed `src/table_scan/` or `src/copy/`, and none of them was compiled by CI before merge. Each needed a local `make release` plus a local SQL Server to verify. The integration suite that covers those paths (167 cases) ran only because it was dispatched by hand.

`scripts/ci/check_require_env.sh:46-49` already names this as blocking a related fix:

> It deliberately does NOT cover the standalone C++ tests (`make test-cpp`), which went stale the same way inside the same spec: they link against the built extension archive, and no CI job builds the extension on a pull request at all (issue #212). That gate is blocked behind it.

## A second, smaller false-green in the same family

Passing several filters to `unittest` **concatenates** them into one pattern rather than unioning them:

```
$ build/release/test/unittest test/sql/a.test test/sql/b.test
Filters: test/sql/a.test test/sql/b.test
No test cases matched 'test/sql/a.testtest/sql/b.test'
No tests ran
$ echo $?
0
```

Exit 0. That is #212 gap 2's shape — a selection that matches nothing reporting success — surviving in the runner itself. `make integration-test` guards it with a `MSSQL_MIN_TEST_CASES` floor, but `scripts/ci/integration_test.sh` builds its own selection with `find`, so the floor does not cover the CI lane. Any script that loops filters is one refactor away from silently testing nothing.

## Suggested fix

1. Run `Build` on `pull_request` — at minimum `linux_amd64`, which is the platform the integration lane needs anyway. Cost is a compile per PR; the alternative is what #212 documented three times in one week.
2. Run `Integration Test` on `pull_request` too, or state explicitly that post-merge detection is the accepted policy and say so in `CLAUDE.md`, so nobody reads a green PR as "the integration suite passed".
3. Assert a case-count floor in `scripts/ci/integration_test.sh` the way the Makefile target does.

## Relationship to #212

Not a duplicate: #212's gap 2 (the glob selecting zero files) is fixed, and gap 1 is half-fixed. This tracks the residual — PRs still do not compile — plus the runner-level concatenation false-green found while verifying the CI lane this week.

oluies:
Closes #278, closes #279. @VGSML

**This PR is its own test.** If the #279 change works, this PR should be the first one in the repo's history that CI compiles and integration-tests before merge — and the integration job should report **167** cases rather than 163, because of the #278 change. Both numbers are checkable on the checks tab below.

## #278 — the four TLS files have never run in CI

`MSSQL_TEST_DSN_TLS` *is* exported by the Makefile (fixed after the PR #264 review). But CI doesn't go through that target — it runs `scripts/ci/integration_test.sh`, which exported ten variables and not that one. The job reported PASSED at **163 cases against 167 locally**, and the `MSSQL_MIN_TEST_CASES` floor of 150 cannot see a four-case difference.

The reason it stayed hidden is the more useful half. `check_require_env.sh` — written precisely to catch this class, and whose own header names `MSSQL_TEST_DSN_TLS` as the bug that motivated its export check — greps **`Makefile` only**. So the fix landed in one of the two lanes that run this suite, and the guard was blind to the other.

It now requires the export in **both**. Verified in both directions:

```
# before the fix, with the widened guard:
check_require_env: MSSQL_TEST_DSN_TLS is not exported by scripts/ci/integration_test.sh
    — these files SKIP silently there:
    tls_connection.test tls_queries.test tls_parallel.test tls_multipacket.test

# after:
check_require_env: OK — 30 variables exported or documented-opt-in
```

## #279 — pull requests still never compiled the extension

#212 gap 1 was addressed by building on merge to `main`, which converts prevention into post-hoc detection. Concretely: **#269, #270 and #274 all changed `src/table_scan/` or `src/copy/`, and none was compiled by CI before it landed.** Each needed a local `make release` and a local SQL Server to verify.

`Build` and `Integration Test` now run on `pull_request`.

**On a PR the build is Linux only** — it's the platform the integration lane consumes, and one compile per PR is the price of the check. macOS stays on merge/schedule/dispatch, where a platform-specific break is still caught before a release rather than at it. Docs-only PRs remain free via `paths-ignore`.

### One implementation note worth reading

That restriction is a `fromJSON` matrix, not a condition on the job's `if`, because **GitHub does not expose the `matrix` context to `jobs.<id>.if`**. My first version used `matrix.platform.id == 'linux_amd64'` there — it evaluates to empty, so the job would have silently never run on PRs. That failure is indistinguishable from the coverage gap being fixed, which is a bad way for a CI change to be wrong. The comment in the file says so.

## Not changed

`scripts/ci/integration_test.sh` already carries a case-count floor (#212 gap 2), so the third suggestion on #279 was already implemented — I've corrected that on the issue. Its limitation stands: it catches "almost nothing", not "quietly four fewer". The guard widening is what covers that class.

yamllint clean; both `fromJSON` arrays verified to parse.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,32 +278,40 @@ jobs:
   build:
     name: Build (${{ matrix.platform.id }})
     runs-on: ${{ matrix.platform.runner }}
-    # Also on a merge to `main` (issue #212 gap 1). A pull request never compiled
-    # the extension, so a C++ regression could reach main behind green checks —
-    # three times in one week when that issue was filed. Building on the merge
-    # catches it on main rather than at the next release, and costs nothing on
-    # the PRs themselves: the `push` trigger is `branches: [main]` and the
-    # workflow's `paths-ignore` keeps docs-only merges free.
-    if: ${{ (github.event_name == 'workflow_dispatch' && inputs.run_build) || github.event_name == 'schedule' || github.event_name == 'push' }}
+    # On a merge to `main` (issue #212 gap 1) AND on the pull request itself
+    # (issue #279). Building only on the merge turned a green PR into "nobody has
+    # compiled this yet" — the regression is still found, just after it has
+    # landed. #269, #270 and #274 all changed src/table_scan/ or src/copy/ and
+    # none was compiled by CI before merge.
+    #
+    # On a pull request this is LINUX ONLY, deliberately: it is the platform the
+    # integration lane needs anyway, and one compile per PR is the price of the
+    # check. macOS and Windows stay on merge/schedule/dispatch, where a
+    # platform-specific break is still caught before a release rather than at it.
+    # Docs-only PRs are already free via the workflow's `paths-ignore`.
+    if: >-
+      ${{ (github.event_name == 'workflow_dispatch' && inputs.run_build)
+          || github.event_name == 'schedule'
+          || github.event_name == 'push'
+          || github.event_name == 'pull_request' }}
     strategy:
       fail-fast: false
       matrix:
-        platform:
-          # Linux x86_64 - with Docker support for integration tests
-          - id: linux_amd64
-            os: linux
-            arch: amd64
-            runner: ubuntu-22.04
-            docker_available: true
-            vcpkg_triplet: x64-linux
-
-          # macOS ARM64 - load-only testing
-          - id: osx_arm64
-            os: osx
-            arch: arm64
-            runner: macos-14
-            docker_available: false
-            vcpkg_triplet: arm64-osx
+        # Linux x86_64 (Docker, and what the integration lane consumes) plus
+        # macOS ARM64 (load-only). A PULL REQUEST builds Linux only (issue
+        # #279): it is the platform integration needs, and one compile per PR is
+        # the price of the check — macOS stays on merge / schedule / dispatch,
+        # where a platform-specific break is still caught before a release.
+        #
+        # Written as fromJSON rather than a condition on the job's `if`, because
+        # GitHub does NOT expose the `matrix` context to `jobs.<id>.if`. A
+        # `matrix.platform.id == '…'` test there evaluates to empty and the job
+        # silently never runs — the failure would look exactly like the coverage
+        # gap this is fixing.
+        platform: >-
+          ${{ github.event_name == 'pull_request'
+              && fromJSON('[{"id":"linux_amd64","os":"linux","arch":"amd64","runner":"ubuntu-22.04","docker_available":true,"vcpkg_triplet":"x64-linux"}]')
+              || fromJSON('[{"id":"linux_amd64","os":"linux","arch":"amd64","runner":"ubuntu-22.04","docker_available":true,"vcpkg_triplet":"x64-linux"},{"id":"osx_arm64","os":"osx","arch":"arm64","runner":"macos-14","docker_available":false,"vcpkg_triplet":"arm64-osx"}]') }}
 
     env:
       # Send vcpkg's binary cache to a directory actions/cache persists (see
@@ -769,11 +777,16 @@ jobs:
     name: Integration Test (linux_amd64)
     runs-on: ubuntu-22.04
     needs: build
-    # On a merge to `main` as well: the suite went from weekly to per-merge, which
-    # is what makes a SQL regression visible while the change is still fresh.
-    # Only worth having because the selection is fixed — this job used to report
-    # PASSED having run zero files (#212 gap 2).
-    if: ${{ (github.event_name == 'workflow_dispatch' && inputs.run_build && inputs.run_integration_tests) || github.event_name == 'schedule' || github.event_name == 'push' }}
+    # On a merge to `main`, and on the pull request itself (issue #279): the suite
+    # went from weekly to per-merge to per-PR, which is what makes a SQL
+    # regression visible while the change is still a proposal rather than
+    # history. Only worth having because the selection is fixed and floored —
+    # this job used to report PASSED having run zero files (#212 gap 2).
+    if: >-
+      ${{ (github.event_name == 'workflow_dispatch' && inputs.run_build && inputs.run_integration_tests)
+          || github.event_name == 'schedule'
+          || github.event_name == 'push'
+          || github.event_name == 'pull_request' }}
 
     services:
       sqlserver:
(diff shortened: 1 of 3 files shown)


oluies:
**Both fixes verified in CI, on this PR.** [Run 32892193378](https://github.com/hugr-lab/mssql-extension/actions/runs/32892193378) — conclusion `success`.

### #279 — a pull request now compiles the extension

```
Build (linux_amd64): success
Integration Test (linux_amd64): success
```

No `Build (osx_arm64)` job was spawned, so the `fromJSON` matrix restricts to Linux on a PR exactly as intended. This is the first pull request in the repo's history that CI has compiled — everything before it was approved on the strength of unit tests that never touch `table_scan.cpp` or the codec sources.

### #278 — the TLS files run

```
All tests passed (5 skipped tests, 5310 assertions in 169 test cases)

Skipped tests for the following reasons:
  require spatial: 1
  require-env MSSQL_COUNTERS: 1
  require-env MSSQL_KERBEROS_TEST: 1
  require-env MSSQL_NAMED_INSTANCE_HOST: 1
  require-env MSSQL_WINSSPI_TEST: 1
```

`require-env MSSQL_TEST_DSN_TLS: 4` is **gone** from the skip list. Case count went **163 → 169**: the four TLS files, plus the two test files added since the earlier measurement (`scan_cardinality`, `like_pushdown_collation`).

The number worth noting is that CI's 169/5310 is now **identical to a local `make integration-test` run**. Those two lanes agreeing is the actual fix — the divergence between them is what hid this for as long as it did, and what `check_require_env.sh` now enforces.

### Note on the remaining 5 skips

All five are genuine opt-ins that need a developer to supply something (a spatial extension, a Kerberos KDC, a named instance, a Windows host, or the counters flag), and all five are in the guard's `OPT_IN` allowlist. Nothing in that list is silently skipping for want of a variable CI could have set.

VGSML (approved):

PR #282, opened 2026-08-26
docs: "no pushdown lost" was too broad — rowid and relaxation-only shapes do lose, and must be excluded from any W1 deferral

oluies:
@VGSML — roborev job 1130 reviewed the measurement branch and found **two High-severity pushdown losses my measurement never tested**. Both verified against the source. This corrects a claim that is currently on `main` in four places.

## What still stands

The narrow claim is fine: the #269 gates objection (volatile, oversized `IN`, `CanThrow() && filters.size() > 1`) **did not** reproduce — those shapes were tested and each reached the server.

## What does not

The generalisation that grew out of it — **"no pushdown lost"** — is false.

**1. Relaxation-only shapes.** `GenerateTableScanFilters` returns `PUSHED_DOWN_PARTIALLY` for `LIKE`, OR-chains, non-dense `IN` and temporal-cast filters, and `pushdown_get.cpp` then *skips* `TryPushdownGenericExpression` for anything not `NO_PUSHDOWN`. The server gets the combiner's **relaxation** — prefix bounds, an optional filter, margin-adjusted temporal bounds — while the exact predicate stays above the scan.

I recorded `LIKE` as "reached the server" on the strength of seeing `[name] >= N'n' AND [name] < N'o'` in the debug log. **That is the relaxation, not the predicate.** I read a weaker predicate as a successful push, then built a conclusion on it. `WHERE name LIKE 'ab%cd'` over 200k rows streams the entire `ab` prefix range to the client.

**2. `rowid` loses pushdown entirely.** A rowid predicate is single-column, so it defers — but `FilterEncoder::Encode` refuses every virtual column (`table_col_idx >= VIRTUAL_COL_START`) and routes it to the client net, whereas `ComplexFilterPushdown` reached `EncodeColumnRef`, which rewrites rowid to the scalar PK and pushes real T-SQL:

| | pushed |
|---|---|
| today | `WHERE [id] > 100` |
| under the deferral | *(nothing)* — full table scan |

Rows stay correct; the wire cost does not. My measurement covered `id`, `dt` and `name` — never `rowid`.

## Why this is still not "blocked"

Neither is a reason the restriction can't be done. Both are shapes the deferral must **exclude** — which is a narrower and more actionable statement than either "no pushdown lost" or "blocked on 061". Corrected in all four surfaces that carried the overstatement: the roadmap section, its W1 row, spec 070's W1 outcome, and the `table_scan.cpp` reachability note.

Build clean; suite **169 cases / 5310 assertions**; `make test-cpp` 17/17.

---

**Process note worth acting on.** This sat unread because `roborev fix --list` defaults to the **current branch**. From `main` it reports "No open jobs found" while **nine reviews across seven branches are open — 33 findings, including these two High**. Enumerating `roborev list --branch <b> --open` per branch is the only way to see them. Every merged feature branch in this session has unread findings against code now on `main`.

### DIFF
--- a/docs/roadmap-v0.3.0.md
+++ b/docs/roadmap-v0.3.0.md
@@ -31,7 +31,7 @@ cut by then.
 | 067 | DML staging: UPDATE/DELETE via #temp bulk load + set-based JOIN; match-key ladder makes rowid/PK optional; closes #140 | spec on recon branch | 062 (bulk-load path), 066 | L | VGSML |
 | — | MERGE pushdown (T-SQL MERGE from DuckDB MERGE INTO; semantics mapped in 065 research) | recon only | 067 | M | VGSML |
 | 061 | **Collation-faithful pushdown** (widened in #277 from ORDER BY only). Makes the spec-039 ORDER BY/TOP pushdown default-safe (today experimental, opt-in `mssql_order_pushdown`) — the remaining half of #58 / discussion #59 — AND supplies the exactness that server-side DML requires: `native AND … COLLATE …_BIN2` plus a trailing-space sentinel, added beside the native predicate so the Index Seek survives. **Now a prerequisite, not an independent item** | spec on main (ORDER BY only); widening proposed in #277 — **OPEN**, mechanism lives on that PR branch | nothing | M | oluies |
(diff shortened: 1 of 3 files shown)


oluies:
@VGSML — ready for review, and **only you can unblock it**: it's authored by `oluies`, GitHub won't take a self-approval, and `oluies` has `maintain` not `admin`, so `--admin` is rejected too. One approving review from you is the only path.

**Docs-only.** It retracts a claim of mine that is currently on `main`.

I wrote that the W1 measurement showed **"no pushdown lost."** The narrow part is true — the #269 gates objection (volatile, oversized `IN`, `CanThrow() && filters.size() > 1`) did not reproduce. The generalisation is false, and roborev job 1130 caught two classes that do lose, neither of which I had tested:

- **Partially-pushed shapes.** `GenerateTableScanFilters` returns `PUSHED_DOWN_PARTIALLY` for `LIKE`-prefix, OR-chains, non-dense `IN` and temporal-cast, and `pushdown_get.cpp` then skips `TryPushdownGenericExpression`. Only `LIKE`-prefix pushes bounds the encoder can render; the other three push an `optional_filter` scalar function it has **no mapping** for, so the scan pushes *nothing* and streams the table.
- **`rowid` loses pushdown entirely.** It's single-column so it defers, but `FilterEncoder::Encode` refuses virtual columns while `ComplexFilterPushdown` reached `EncodeColumnRef`, which rewrites rowid to the scalar PK. `WHERE rowid > 100` goes from `WHERE [id] > 100` to a full scan.

Neither blocks the restriction — both are shapes the deferral must **exclude**, which is more actionable than either "no pushdown lost" or "blocked".

CI green: Build, Integration (5m39s), both C++ unit lanes, Windows SSPI, Lint, yamllint.

VGSML (approved):

PR #283, opened 2026-08-26
roborev backlog: nine reviews that were invisible from main — two real defects, one contradictory criterion

oluies:
@VGSML — **stacked on #282**; merge that first. Clears the nine open roborev reviews (33 findings) that `roborev fix --list` could not see, because it defaults to the *current branch* and these live on merged feature branches.

## Two real defects in merged code

**The statistics cache was invalidated by nothing.** `InvalidateAll` / `InvalidateTable` / `InvalidateSchema` / `SetCacheTTL` had **no callers anywhere** — `RefreshCache`, `InvalidateMetadataCache` and `InvalidateTableEntry` cleared only `metadata_cache_`. Harmless while row counts fed `GetStorageInfo` (a number the planner ignores for a table function), but since the spec-070 cardinality callback they *are* what the optimizer plans on: a stale count survived an explicit `mssql_invalidate_cache()` and drove join order for the full TTL. All three paths now clear it, `RefreshCache` applies the documented `mssql_statistics_cache_ttl_seconds` (which reached the provider through nothing), and the callback reads the **catalog** count first — the one a refresh actually updates.

**A search condition was accepted in value position** (in the previous commit of this PR): `CASE … THEN id > 5` and `(id > 1) IS NULL` produced T-SQL the server rejects, and since the expression is erased from the plan the query *failed* rather than degrading to the client net. Plus the modulo gate couldn't see `money` (it arrives as `DECIMAL`), and `flush_rows = 0` opened the columnstore warm-up gate backwards — 0 means *unbounded*, the case the gate matters most for.

## One criterion that two surfaces defined differently

The roadmap said **"string comparisons"** must stay claimed; spec 070 and `table_scan.cpp` said **"string-pattern expressions"**. That difference *is* the type test an implementer has to write — `name = 'abc'` is deferrable under one and gated under the other.

Unified on the broader term, with the reason stated, because there are **two** hazards and only one was ever given:

1. `TryPushdownPrefixFilter` rewrites `prefix()` into a range under binary ordering — **patterns only**;
2. once *any* string comparison is a visible filter, the combiner prunes contradictions and derives transitive predicates over its constants under **binary** semantics, on predicates this path used to erase before it saw them. `name = 'abc' AND name = 'ABC'` is contradictory under binary and satisfiable on `_CI_AS`.

The second is why the broader term is right: `name = 'abc'` encodes identically either way — making it **visible** is what isn't free.

## Claims that were wrong

- *"`PreloadRowCount` refreshes the cache after a bulk load"* — it doesn't; its only callers are metadata loads.
- *"the same 16 behind the `text`→16 CAST truncation"* — asserted a **fixed** bug (#197) as live; `GetNVarcharLength` special-cases `text`/`ntext` to MAX, and the 16 is *why*.
- `ci.yml` said Windows stays on "merge/schedule/dispatch" — it is **dispatch-only**.

## And one I reintroduced in #280

The `linux_amd64` matrix entry was written out **twice** inside a single expression, so an edit to one copy would make PRs build a different configuration than merges — *green PR, broken main*, the class #279 exists to close, one level up. A `build-matrix` job now owns the list once.

Also: `dynamic_cast` removed from the planning path (the codebase avoids RTTI across the extension boundary and says so twice); `check_require_env`'s best diagnostic was suppressed in exactly the case that motivated the script; `MSSQL_MIN_TEST_CASES` raised 150 → 165 (the #278 commit diagnosed that 150 can't see four files vanish, then left it at 150); and the two tests gained the assertions their own comments said mattered most — the TOP-N clamp, VIEW-reports-0, and the column's reported *type* alongside its collation.

## Verification — incomplete, and here is exactly how

Build clean, `make test-cpp` **17/17**, `make test-load-policy` green including seven new boundary assertions.

**The integration suite could not run.** The docker daemon is unreachable — `/var/run/docker.sock` is a symlink to a socket that doesn't exist — so the new `.test` assertions are **unexecuted**. Once #280 is on `main` this PR's own CI will run them; until then that half is unverified and I'd rather say so than imply otherwise.


oluies:
Container is back and the suite has now run — **169 cases / 5318 assertions, green**. `make test-cpp` 17/17, `make test-load-policy` green.

**It found one failure, and it was one of the assertions I'd added blind** — which is the argument for running them rather than reasoning about them.

The TOP-N `EXPLAIN` assertion rested on job 1124's premise: that `MSSQLOptimizer` folds the `LogicalTopN` *into* the scan and removes the node, leaving the get claiming the full table count for an operator emitting at most N. **Measured, that does not reproduce.** `ORDER BY id LIMIT 10` over 200 000 rows keeps a `Top N` operator *above* the scan, and the scan still reports `~200,000` — with `mssql_order_pushdown = true` as well as the default `false`:

```
╭─ Top N ────────────────╮
│ Top: 10   ~10 rows     │
╰───────────┬────────────╯
╭─ Mssql Catalog Scan ───╮
│ ~200,000 rows          │
╰────────────────────────╯
```

So the clamp in `MSSQLCatalogScanCardinality` is correct but **nothing reaches it**, and the assertion was failing for the wrong reason. Removed rather than repaired, with the reason recorded in both the test file and the code comment — which is exactly what job 1127 proposed for this outcome: *"if that estimate proves not to be observable, say so in the code comment instead of adding a test."*

The **VIEW-reports-0** assertion added alongside it passes and stays — that one pins the rule with the largest blast radius (a view reports 0 rows from `sys.partitions` while returning millions, so it must yield *no* estimate rather than `~0`).

`0a0ae49` pushed. This PR is now fully verified end-to-end, integration included.

oluies:
Another roborev round on this stack (jobs 1190/1191/1193/1197). **Two more real code defects**, both of them mine from earlier in this PR.

**I fixed four value positions and left three.** `THEN`/`ELSE` and `IS [NOT] NULL` were guarded individually — comparison operands, BETWEEN's input/bounds and mapped-function arguments were not. `WHERE flag = (id > 5)` encodes to `([flag] = ([id] > 5))`, which SQL Server rejects, and since `ComplexFilterPushdown` erases the accepted expression the query **fails** rather than falling to the client net. That is exactly the failure the previous commit claimed to have closed ("all four now refuse"). Now one `EncodeValueExpression` routed through **every** value position, so the next one can't be added without the check.

**The statistics invalidation I added missed the path that matters.** Wiring `RefreshCache` / `InvalidateMetadataCache` / `InvalidateTableEntry` left `InvalidateSchema` with **zero callers** — and the schema path is what COPY/bulk load, CTAS, CREATE/DROP TABLE and `mssql_invalidate_cache(ctx, schema)` all take. So a 1M-row `COPY` into a previously-empty table *still* left the pre-load count driving plans (and `duckdb_tables()`) for the full TTL. The "invalidated by nothing" defect, still open on its commonest route.

Also: `mssql_statistics_cache_ttl_seconds` now applies at the point of use, so `SET … = 0` works without a manual `mssql_refresh_cache()`.

**And two claims of mine that were wrong:**

- The sentence this stack exists to retract — *"No pushdown was lost"* — was still standing, unscoped, **nine lines above its own retraction**. Rescoped to the gates actually measured.
- The retraction itself understated the loss for three of its four shapes. Only `LIKE`-prefix pushes comparison bounds the encoder can render — and that's a string predicate, already excluded. OR-chains, non-dense `IN` and temporal-cast push an `optional_filter` scalar function the encoder has **no mapping** for, so the scan pushes **nothing** and streams the table. Those three are the genuinely new exclusion. My headline `LIKE 'ab%cd'` example demonstrated nothing — an interior wildcard is refused today anyway.
- My "measured" explanation for the unreachable TOP-N clamp was a property of the **fixture**, not the code: `CardScanBig` is COPY-created so every column is nullable, DuckDB's default NULLS LAST fails `IsNullOrderCompatible`, and `TryPushTopN` bailed for a reason unrelated to the clamp. Against a `NOT NULL` column the TopN *would* fold. The conclusion survives; the reason didn't — the real one is stage order, which the pre-existing comment already gave.

Suite **169 cases / 5318 assertions** green; `make test-cpp` 17/17; `make test-load-policy` green.

oluies:
@VGSML — ready for review, **stacked on #282** (merge that first). Same blocker: authored by `oluies`, so no self-approval, and `maintain` can't `--admin`. Your review is the only way this lands.

CI is now fully green — `Build (linux_amd64)` 35m26s, `Integration Test` 6m46s, concurrency stress 33m, all unit lanes.

## What it is

The nine roborev reviews that were **invisible from `main`** — `roborev fix --list` is branch-scoped, so from `main` it reported "No open jobs" while 33 findings sat open on merged feature branches. Plus the rounds that followed.

**Real defects fixed in already-merged code:**

- **Invalid T-SQL from value position.** `WHERE flag = (id > 5)`, `CASE … THEN id > 5`, `(id > 1) IS NULL`, BETWEEN bounds, function args — each emitted T-SQL SQL Server rejects, and since `ComplexFilterPushdown` erases the expression the query **fails** rather than degrading to the client net. Now one `EncodeValueExpression` routed through every value position.
- **Modulo couldn't see `money`.** It excluded FLOAT/DOUBLE, but `money`/`smallmoney` arrive as `DECIMAL(19,4)` and T-SQL rejects them for `%` with the same 8117. Now a whitelist of exact integers.
- **`flush_rows = 0` opened the columnstore warm-up gate backwards.** 0 means *unbounded* — the case the gate matters most for — but compares numerically below the threshold. Extracted as pure `MSSQLWarmupGateRows`, boundaries pinned.
- **The statistics cache was invalidated by nothing.** `InvalidateAll`/`InvalidateTable`/`InvalidateSchema`/`SetCacheTTL` had zero callers. Since the spec-070 cardinality callback those counts are what the optimizer plans on, so a stale count survived `mssql_invalidate_cache()` and drove join order for the whole TTL. Wired into every path — including the **schema** path, which is what COPY/CTAS/DDL actually take.
- **A cross-session TTL leak** (introduced while fixing the above, then found): the provider is one object per catalog, the setting is per session. Every read now takes its TTL as an argument; the mutable knob is deleted so the type can't express the misuse.

**And two CI gaps this branch exposed by fixing them:**

- `make test-cpp` had **never run on Linux** — GNU ld's single-pass archive resolution breaks the link, which `ci.yml`'s hand-written step already documented and worked around while the Makefile didn't. Now `--start-group`, and CI runs the same `STANDALONE_TEST_SOURCES` list both lanes share.
- The failure output said only *"see build/test/<n>.log"* — a file that lives on the runner and is never uploaded. It now tails the log.

## Worth knowing before you read it

This took **eight roborev rounds**, and every round found a real defect in the previous round's fix — twice of a kind that fails queries outright. The pattern was consistent: each fix was correct where I looked and incomplete one step beyond (planner path but not refresh path; four value positions but not seven; invalidation wired to three callers but not the one COPY uses). Worth a sceptical eye on the seams rather than the centre.

VGSML (dismissed):

oluies:
@VGSML — thank you for the approval, and sorry: **it was dismissed by my own push.** You approved at 08:46; I pushed `c3d184d` a few minutes later and branch protection invalidated the approval as stale. Entirely my sequencing error — I had a review round in flight and requested review anyway.

**#282 is merged** (`3579f63`). This one needs a re-approval on `c3d184d`, and **I will not push again** until it lands, so it can't happen twice.

The one round between your approval and now is small, and closes the CI thread you'd otherwise inherit:

- **Deleted the duplicated `test_vector_encodings` block from `ci.yml`.** It's in `STANDALONE_TEST_SOURCES`, so once `make test-cpp-run` was wired into the build job it was being statically linked and run **twice in the same job**, from two copies of the link line that had already drifted (`-lgssapi_krb5` in one, `--start-group` in the other). That drift is what cost three CI rounds here. The Makefile's is now the only static link line — which was the point of the change that accidentally created the duplicate.
- **Scoped the `from_catalog_metadata` TTL exemption**, which I had made unconditional. `PreloadRowCount` is the statistics cache's only populator on the reachable paths, so *every* entry was exempt: the TTL governed nothing, `GetRowCount`'s DMV re-read became unreachable for any table the catalog had loaded, and `mssql_statistics_cache_ttl_seconds` documented a setting with no effect. Now opt-in per read, and only `GetStorageInfo` — the listing path, where ageing a count out costs a connection plus a DMV query per table — asks for it.
- **That exemption had also made its own test vacuous.** `IsCacheValid` returned on `from_catalog_metadata` *before* reading `ttl_seconds`, so the test guarding the caller-supplied-TTL invariant passed even with that invariant reverted — the regression it exists to catch, already shipped twice (jobs 1203, 1216). Verified fixed by simulating the revert, which now fails on the right assertion.

CI green on `c3d184d`: `Build (linux_amd64)`, `Integration Test`, concurrency stress, all unit lanes, yamllint.

VGSML (dismissed):

VGSML (dismissed):

VGSML (dismissed):

VGSML (approved):

VGSML (approved):

PR #284, opened 2026-08-29
roborev 1259: remove the twelve remaining CI duplicates, pin the other half of the TTL exemption

oluies:
@VGSML — **stacked on #283**, base is `fix/roborev-backlog`, so it merges after that one. Deliberately a separate PR: #283 is waiting on your re-approval and pushing to it again would dismiss that a second time.

Three Low findings from roborev job 1259. None is a correctness defect; two are the same "fixed where I looked" pattern.

## I removed one of thirteen duplicates and claimed I'd removed them all

The previous round deleted the hand-rolled `test_vector_encodings` block and left a comment saying *"one link line now: the Makefile's"*. That wasn't true of the job: `test_datetime_codec`, the `ddl_translator`/`ctas_type_mapping`/`catalog_filter` loop and the eight-family codec loop are **all** in `STANDALONE_TEST_SOURCES`, so `make test-cpp-run` at the end of the same job built and ran each of them a second time — from a second compile line (`-L build/release/src -lduckdb` rather than the static archive set) free to drift from the Makefile's.

That drift is not hypothetical: it is exactly how the `--start-group` bug survived. One copy carried the workaround, the other didn't, and only the Makefile's ran on Linux. All twelve are gone now. What remains is only what `STANDALONE_TEST_SOURCES` doesn't cover — the TDS-only tests that link no libduckdb, and the staging tests.

## The exemption predicate was half-pinned

`PreloadRowCount` is the only cache populator the test could reach, so every entry was catalog-sourced — and weakening `exempt_catalog_sourced && stats.from_catalog_metadata` to `exempt_catalog_sourced` alone left the whole suite **green**.

That weakening is a live hazard: `GetStorageInfo` passes `true` unconditionally, and `GetRowCount` writes DMV entries back with `from_catalog_metadata = false`. The blanket form would make `SHOW ALL TABLES` serve a DMV-sourced count that never ages out, at any TTL, for the life of the catalog.

Added `SeedRowCountForTesting` so a DMV-shaped entry can be created without a connection, and asserted such an entry is refused at TTL 0 **even when the exemption is requested**. Verified by applying the weakening — the suite now fails.

## The setting documented one meaning, the code had two

`CLAUDE.md` and `docs/architecture.md` said "Statistics cache TTL". It now says which path honours it, which exempts catalog-loaded counts and why (a connection plus a `sys.dm_db_partition_stats` round trip **per table** on a listing), and that DMV-sourced counts age normally everywhere. `DATAMODEL.md` gains the ownership invariant: session setting, per-catalog provider, passed at the read, never written back — and that the exemption is opt-in *per read* rather than a property of the entry, since making it unconditional exempted every entry the cache holds in practice.

Suite **169 cases / 5318 assertions** green; `make test-cpp-run` 18/18; `test-load-policy` green; yamllint clean.


VGSML (approved):

VGSML:
@oluies heads-up — **this one never actually reached `main`.** The stacked setup ate it.

Its base was `fix/roborev-backlog`, the head branch of #283. I squash-merged #283 into `main` at 21:47:29, and this merged at 21:48:21 — into `fix/roborev-backlog`, which by then was already a dead end. GitHub says `MERGED` and it is, technically: merged into its stated base, from which nothing flows any more. `a607c78d` is not an ancestor of `main`.

Nothing lost — both branches are still on the remote. But `main` is missing the whole of this PR: the `ci.yml` dedup (the 12 duplicate test builds are still there), `SeedRowCountForTesting` + the job-1259 assertion, and the doc updates.

Worth noting the squash merge makes this easy to miss: `git log main..fix/roborev-backlog` lists #283's commits as "missing" even though their content is in, so the commit view looks like noise. It only shows up by content — `SeedRowCountForTesting` simply isn't in `main`.

**Re-landed as #287.** Cherry-pick of `54692a9` onto `main`, your authorship preserved.

Why a cherry-pick and not a PR straight from `fix/roborev-backlog`: that branch is 5 commits behind, so the diff **reverts** docs merged since — including all 76 lines of your `website/docs/writing/dml.md` from #286. One conflict in `docs/architecture.md`, resolved keeping `main`'s rows (they carry #286's newer "Registered but READ BY NOTHING" findings) and taking your rewritten `mssql_statistics_cache_ttl_seconds` row. Net ±1 line there.

PR #292, opened 2026-08-31
docs+ci: make the PR gate actually fail on a broken link, and split the github-actions group by update-type

oluies:
Two findings from roborev job 1300 on `63aea12` (the #288 follow-up). #288 was already merged, so this lands on a fresh branch off `main`.

## MEDIUM — `docs-build.yml` could not catch the defect class it was written for

`docs-build.yml` was added so an npm PR carries evidence the site still builds, and the same commit fixed a dangling `#collation` anchor. But `onBrokenAnchors` was unset, so Docusaurus defaults it to **`warn`** — that exact breakage produced a warning and `npx docusaurus build` **still exited 0**. Which is why it reached `main` from #286 in the first place. `onBrokenMarkdownLinks` was at `'warn'` too, so a renamed target behind the newly-introduced relative links would also have sailed through.

Both are now `'throw'`. `onBrokenMarkdownLinks` moves to `markdown.hooks`, where Docusaurus 3.10 wants it — the top-level key is deprecated and removed in v4, so the migration also clears a warning printed on every build.

**Verified in both directions**, since a gate that has never been seen to fail is not a verified gate:

- No pre-existing anchor or markdown-link breakage anywhere in `docs/` **or** `versioned_docs/` — the tightening needs no cleanup pass first, which was the risk roborev flagged.
- Re-breaking the `#collation` anchor now fails:

  ```
  Error: Docusaurus found broken anchors!
  - Broken anchor on source page path = /mssql-extension/next/reading/queries/
  EXIT=1
  ```

  Before this commit the identical edit exited **0**.

⚠️ This also tightens the **production Pages deploy**, which builds the same config: a future broken anchor fails the deploy instead of shipping. That is deliberate — a dangling link in published docs is worth a red deploy — but it is a real behaviour change to `pages.yml`'s outcome, so flagging it rather than burying it.

## LOW — the version-escape fix reached one site out of many

`63aea12` fixed `queries.md` but left the same defect elsewhere. An absolute link like `/writing/table-options/` resolves into the **released** docs, so a page published under `/next/` sends the reader to stale content. It only avoids being a hard break while the released snapshot happens to still have that page.

Roborev named three sites; I swept the live tree instead and found **12**:

| file | links |
|---|---|
| `docs/index.md` | 2 |
| `docs/reference/settings.md` | 4 |
| `docs/reference/troubleshooting.md` | 1 |
| `docs/writing/ctas.md` | 2 |
| `docs/writing/dml.md` | 1 |
| `docs/writing/transactions.md` | 2 |

All now version-relative markdown links (`../writing/table-options.md`). **Three carried anchors** — `#order-by-pushdown-experimental` and two in `transactions.md` — which is the case that breaks silently, and the case the first finding is about.

Every rewrite was resolved against a target file that exists on disk; the script reported no unresolved targets.

## Verified

- `npx docusaurus build` green with both options at `'throw'`.
- Deprecation warning gone.
- Deliberate-break test above, restored and rebuilt green.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,17 +28,26 @@
 version: 2
 
 updates:
-  # GitHub Actions. Workflows pin major-version tags (actions/checkout@v7), so
-  # every update here is in practice a major bump; one grouped PR per month.
+  # GitHub Actions. Split by update-type for the same reason as npm below, and
+  # the two pinning styles in use here are why: most workflows pin a major tag
+  # (actions/checkout@v7) where the only possible bump is a major, but
+  # codeql.yml pins a full patch version (github/codeql-action@v4.37.3) that
+  # moves on patch releases. So this ecosystem really does have both streams,
+  # and grouping all of it would ride a major in on the same approval as a
+  # patch -- PR #291 did exactly that, landing setup-node v6 -> v7 alongside a
+  # codeql-action patch. Majors are left ungrouped and get their own PR.
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: monthly
     open-pull-requests-limit: 5
     groups:
-      github-actions:
+      github-actions-minor-patch:
         patterns:
           - "*"
+        update-types:
+          - minor
+          - patch
     cooldown:
       default-days: 7
     labels:
--- a/website/docs/index.md
+++ b/website/docs/index.md
@@ -26,7 +26,7 @@ ATTACH 'Server=localhost;Database=AdventureWorks;User Id=sa;Password=...' AS mss
 SELECT * FROM mssql.dbo.SalesOrderHeader LIMIT 10;
 ```
 
-→ **[Getting Started](/getting-started/)** · **[Settings Reference](/reference/settings/)** · **[Download metrics](https://hugr-lab.github.io/mssql-extension/metrics/)**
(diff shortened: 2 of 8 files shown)


oluies:
Folded in a third change (`da4ddea`), and rebased onto `2dd3b15`.

## Also — the justification for grouping github-actions whole was wrong

#288 collapsed the entire `github-actions` ecosystem into one group on this reasoning, written into the config:

> Workflows pin major-version tags (actions/checkout@v7), so every update here is in practice a major bump; one grouped PR per month.

That is true of most workflows but not all. `codeql.yml` pins a **full patch version** — `github/codeql-action@v4.37.3` — and moves on patch releases. So the ecosystem has both streams, and the premise the group rested on does not hold.

**#291 showed the consequence the same day.** It carried `setup-node` **v6 → v7** — a major, touching both `docs-build.yml` and `pages.yml` — in one PR under one approval, alongside a `codeql-action` v4.37.3 → v4.37.8 patch:

```
- uses: actions/setup-node@v6      →  @v7      (major, docs-build.yml + pages.yml)
- uses: github/codeql-action@v4.37.3 → @v4.37.8 (patch, codeql.yml ×3)
```

It was fine on the merits. The problem is that the group is what removed the choice — the same defect roborev job 1300 flagged in the `website-major` group, which I fixed on the npm side and left standing here.

Now split the same way: `github-actions-minor-patch` groups minor/patch, majors fall through to individual PRs.

VGSML (approved):

PR #294, opened 2026-09-01
docs+ci: finish the link sweep the machine should have done, and stop Dependabot re-raising an uninstallable bump

oluies:
Roborev jobs 1316 and 1318 on the merged #292 work, plus the recurrence of the closed #289 as #293.

## The #292 sweep missed 17 links — and #292 armed a gate in front of them

#292 said it swept the docs tree and rewrote 12 links. It rewrote **12 of 29**. The script carried a hand-written list of section prefixes — `writing|reading|reference|getting-started|guides|intro|troubleshooting` — and the tree also has `connection/` and `performance/`. Everything under those was invisible both to the sweep and to the grep that "verified" it.

Measured on `origin/main` before this branch:

```
16  ](/connection/
 1  ](/performance/
```

Three of the 17 are anchored, which is the part that matters: **#292 also set `onBrokenAnchors: 'throw'`.** An absolute anchor link in `docs/` is validated against the *released snapshot's* copy of the page, so renaming a heading in `docs/connection/azure.md` would fail the build — and the Pages deploy — while pointing at a file that looks fine. Tightening the gate was right. Leaving live rounds behind it was not.

Swept again with **no allowlist**: a link is internal if the path it names resolves to a file inside that version's root.

| root | rewritten | unresolved |
|---|---|---|
| `website/docs` | 17 | 0 |
| `versioned_docs/version-0.2.3` | 29 | 0 |
| `versioned_docs/version-0.2.4` | 29 | 0 |
| **total** | **75** | **0** |

The frozen snapshots had it worse than `docs/` — a reader on 0.2.3 following "COPY TO" silently landed on the 0.2.4 page (roborev's third finding, pre-existing).

## A check, because hand-sweeping has now failed twice

`scripts/ci/check_docs_links.py`, wired into `docs-build.yml` **ahead of** `npm ci` so a link error reports in seconds rather than after a 90-second install.

Docusaurus cannot catch this class itself: an absolute link resolves to a page that genuinely exists in the released version, so `onBrokenLinks` sees nothing wrong. That is precisely why two hand sweeps and a green build all missed it.

Verified in both directions:

```
$ python3 scripts/ci/check_docs_links.py
checked 3 docs root(s): no absolute internal links        # EXIT=0

$ # put one link back
  website/docs/reference/functions.md:188
      ](/connection/azure/)
      use ](../connection/azure.md)                        # EXIT=1
```

## Dependabot — job 1318, and why #289 came back

**`ignore` for `@mermaid-js/layout-elk >=0.2.0`.** Not a judgement about the release: `@docusaurus/theme-mermaid@3.10.2` declares `peerOptional: "^0.1.9"` (confirmed in its own `package.json`), so 0.2.x cannot install — `npm ci` fails `ERESOLVE`. **Closing the PR does not hold:** #289 was closed by hand, and Dependabot re-raised it as #293 **67 seconds** after the next push to `main`, because it re-evaluates on every push. The entry carries its own removal condition.

The three low findings, all correct:

- The file header still described the shape as *"group each ecosystem into one PR"* — neither ecosystem does that any more. It now states the per-stream rule it actually implements.
- The `github-actions` comment quoted `codeql-action@v4.37.3`; #291 had already bumped it to `v4.37.8`. The claim was right, the citation stale.
- `open-pull-requests-limit` 5 → **10** on both. With majors ungrouped, ~20 distinct actions can each want a PR, and 5 would silently *withhold* updates rather than defer them.

## Note

Once this merges, #293 should close itself on Dependabot's next run — an open PR matching a new `ignore` rule gets closed automatically. I'll confirm rather than assume.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,15 @@
 # Dependabot configuration.
 #
-# Shape follows GitHub's "tame Dependabot" guidance: group each ecosystem into
-# one PR, keep the cadence slow, and let security updates ignore both. Security
-# fixes are raised as soon as an advisory lands -- `groups`, `schedule` and
-# `cooldown` shape VERSION updates only -- so slowing the version stream costs
-# nothing in vulnerability response time.
+# Shape follows GitHub's "tame Dependabot" guidance -- group your updates, slow
+# the cadence, keep security fast -- with one deliberate departure: the grouping
+# is per-STREAM, not per-ecosystem. Both ecosystems group only minor/patch and
+# leave majors ungrouped, so a major arrives in a PR that can be merged,
+# reverted or bisected on its own instead of riding in under a routine
+# approval. Cadence is monthly, with a 7-day cooldown.
+#
+# Security fixes are raised as soon as an advisory lands -- `groups`, `schedule`
+# and `cooldown` shape VERSION updates only -- so slowing the version stream
+# costs nothing in vulnerability response time.
 #
 # Ecosystems deliberately NOT covered, recorded so the omissions read as
 # decisions rather than oversights:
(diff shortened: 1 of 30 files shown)


VGSML (approved):
Approved — merge it whenever suits you.

The `#289 → closed by hand → #293 sixty-seven seconds after the next push` detail is the part that settles it, and I didn't have it when I wrote my comment on #293. It turns "close the PR" from a reasonable option into a non-option: Dependabot re-evaluates on every push, so anything not written into the config doesn't hold. Same lesson as #284 landing in a dead branch — if the rule isn't encoded, it isn't a rule.

Your `versions: [">=0.2.0"]` is also the better scope than the `semver-minor` ignore I suggested. It pins the actual incompatibility boundary rather than a bump *class*, so a future 0.1.x patch still flows and only the uninstallable range is held back. And the comment carries the condition for its own removal, which is what stops an ignore entry outliving its reason.

Checked before approving:
- All 11 checks pass, including **Build docs site** — the one failing on #293 — and Integration Test.
- The ignore entry is valid Dependabot syntax and scoped to the one package.
- `check_docs_links.py` shouldn't false-positive: the pattern requires a leading `/` while excluding `//` (protocol-relative), so external `http(s)://` links can't match, and it only flags absolute links whose target resolves to a real `.md` page — assets and non-page paths fall out via the `target is None` skip.

The link gate itself is the right shape for the same reason as the ignore rule: Docusaurus genuinely cannot catch this class, since an absolute link resolves to a page that really exists in the released version, so `onBrokenLinks` sees nothing wrong. Two hand sweeps that both came back is exactly the evidence for making it a check rather than a habit. Putting it *before* the ~90 s npm install is a nice touch.

PR #301, opened 2026-09-06
ci: retry the DuckDB submodule clone, and stop a wedged runner burning six hours

oluies:
`main` went red on `1f1be81` (the #300 merge). It is **flake, not a defect** — #300 changed comments and one markdown file and cannot break a build — but it exposed two things worth fixing.

## What actually happened

`Build (osx_arm64)` failed with **no log and no failing step**, which is why `gh run view --log-failed` returns nothing useful. The API has it:

```json
{ "name": "Checkout DuckDB submodule", "status": "in_progress", "conclusion": null }
  started_at 17:54:17Z    completed_at 18:39:17Z
```

The job entered the submodule checkout, stayed there, and died 45 minutes later — a wedged or lost runner.

## Two things made it worse than it needed to be

**1. No timeout anywhere.** `ci.yml` sets `timeout-minutes` on no job, so GitHub's default applies: **six hours**. Measured normal durations from the last six runs on `main`:

| job | typical |
|---|---|
| `Build (osx_arm64)` | 16–22 min |
| `Build (linux_amd64)` | 23–32 min |

So a wedge can sit for an order of magnitude longer than the work takes. Now `timeout-minutes: 60` on the build matrix (≈2× the slowest observed) and `90` on the two Windows builds.

**2. No retry on the clone that hung.** `git submodule update --init --recursive duckdb` got exactly one attempt. Now three, with backoff.

## It is deliberately still a full clone

`--depth 1` is the obvious reflex here and it is **wrong**. Our own Makefile records why:

> A shallow clone stamps v0.0.1 and hides this; a full one does not.

An extension stamped `v0.0.1` will not `LOAD` into a versioned CLI, so shallow trades a slow clone for a broken artifact. The flakiness is absorbed by retrying instead, and the reason now sits at the call site rather than waiting to be rediscovered.

## What this does not change

A PR still builds **Linux only**. That is existing, deliberate policy, documented at `build-matrix`:

> A pull request builds Linux only… macOS stays on merge / schedule / dispatch, where a platform-specific break is still caught before a release rather than at it.

That is why this surfaced on `main` rather than on #300 — a cost decision, not an oversight. Changing it belongs with the CI-budget conversation in #298, not here.

## Verified

- `yamllint` clean (the ruleset CI uses) and **`actionlint` clean**.
- The retry block parses under `bash -n`.
- A re-run of the failed job is in flight to confirm the flake diagnosis; I will report the result rather than assume it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,16 +133,27 @@ jobs:
       - name: Issue -> test index is well-formed
         run: python3 scripts/ci/issue_links.py
 
-      # scripts/ci/ now holds five shell scripts on the CI critical path, one
-      # of which does `rm -rf` against a path it derives. Their clean state was
-      # established by hand once and nothing kept it that way -- an unquoted
-      # expansion reintroduced later would be found by the next incident, not
-      # by the PR. ~25 s in a 10-minute job.
-      - name: shellcheck the CI scripts
+      # Seven shell scripts across three directories sit on the CI critical
+      # path, one of which does `rm -rf` against a path it derives. Their clean
+      # state was established by hand once and nothing kept it that way -- an
+      # unquoted expansion reintroduced later would be found by the next
+      # incident, not by the PR. ~25 s in a 10-minute job.
+      - name: shellcheck the workflow-invoked shell scripts
         run: |
-          sudo apt-get update
-          sudo apt-get install -y shellcheck
-          shellcheck -S warning scripts/ci/*.sh
+          # shellcheck ships on the ubuntu-latest image, so this normally
+          # installs nothing. The install is conditional rather than
+          # unconditional because the step above already runs `apt-get update`
+          # in this job, and a second index refresh is exactly the command
+          # whose Signed-By conflict broke this repo's CI once before.
+          command -v shellcheck >/dev/null 2>&1 || {
+            sudo apt-get update
+            sudo apt-get install -y shellcheck
+          }
+          shellcheck --version | head -2
+          # Every shell script a workflow invokes, not just scripts/ci/:
+          # fuzz/build.sh runs in fuzz-row-stager.yml and
+          # run_staging_test.sh in this file.
+          shellcheck -S warning scripts/ci/*.sh fuzz/build.sh test/cpp/codec/run_staging_test.sh
 
   # ============================================================================
   # C++ Unit Tests Job - lightweight, no SQL Server / DuckDB build required.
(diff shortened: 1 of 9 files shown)


VGSML (dismissed):

VGSML (dismissed):
Approving.

The diagnosis is the valuable part: a step that **hangs** produces no log and no failing step, so `--log-failed` returns nothing and the job looks broken for no reason. Reaching for the API and finding `"status": "in_progress"` with 45 minutes between `started_at` and `completed_at` is what turned that into something actionable.

Both fixes are right, and the reasoning inside `checkout_duckdb.sh` is the part I checked hardest:

- **A plain retry loop is useless against a hang** — it only runs when the command *returns*. Naming that explicitly, then bounding the attempt, is the correct order of thinking.
- **The perl wrapper is correct, and the comment explains why it is not the obvious form.** `fork` + `setpgrp` + `kill("KILL", -$p)` kills the process *group*; `exec` + `alarm` would signal only the top-level `git submodule update` and orphan `git clone` / `git-remote-https` per nested submodule — which would then still be writing into `duckdb/` while the next attempt's `reset_state` does `rm -rf` on it. That is a real bug avoided, not a stylistic preference.
- **`GIT_HTTP_LOW_SPEED_*` demoted to a secondary signal** rather than the guard, because it bounds the HTTP transfer only and not connect, delta resolution or worktree writes. The failure never identified which of those it was, so guarding one would have been a guess.
- **Not shallow, on purpose**, with the `v0.0.1` consequence recorded at the call site. Right call: an unloadable artifact is worse than a slow clone.
- **`reset_state` removing both `--git-dir/modules/duckdb` and `--git-common-dir/modules/duckdb`**, with the empty-string guard so a failed `rev-parse` cannot turn into `rm -rf "/modules/duckdb"`. Worth the paranoia given the path is derived and relative.
- **90 rather than 60 on `build`**, sized on the cold-ccache case because the cache key carries the submodule SHA. On a repo tracking a duckdb branch that is the routine case, not the rare one — sizing on the warm number would make every bump PR time out looking exactly like the wedge this is for. (Immediately relevant: I have a submodule bump open in #306.)

**One note on scope, not a change request.** The description is about timeouts and the retry, but the diff also carries five shell fixes (`fuzz/build.sh`, `check_require_env.sh`, the two `trap` quotings, the shellcheck disable) and a new **shellcheck step in the `lint` job**. That is a reasonable thing to add alongside a PR that introduces a script doing `rm -rf` on a derived path — it just is not what the title says, and a reviewer skimming the header would not expect a new CI gate.

**Merge-order note:** #303 inserts its step at the same anchor in `lint` (right after `Every require-env variable is actually set`), so whichever lands second will conflict there. Trivial to resolve.

VGSML (approved):

PR #303, opened 2026-09-07
test: make the issue → regression-test link a declaration a machine can read

oluies:
Item **4** from #298 — "regression tests keyed to issue numbers", after nlohmann/json's `unit-regression*.cpp`.

## What I proposed in #298 was wrong for this repo

I suggested a new `test/sql/regression/issue_NNN.test` directory, on the assumption that regression tests were missing. Looking at the tree first is what showed otherwise:

```
$ grep -rloiE "issue #?[0-9]{2,4}" test/sql test/cpp | wc -l
55
```

**38 issues are already referenced across 55 test files** — including #181, #189 and #224. The linkage exists; it is prose. Nothing can answer *"is #181 still covered?"* without a grep, and nothing notices when the test that was a guard gets renamed or deleted.

A `regression/` directory would also have been actively harmful: it splits feature-organised suites that share setup, for no gain the header field does not already give.

## Why a declaration and not a grep

This is the whole design, and there is a live example of it in the tree.

`test/sql/query/legacy_lob_types.test` is a test for **#197**. Of #224 its prose says:

> its code-page decode without the cast is issue #224's territory … **deliberately not pinned here**

A grep-built index reports #224 as **covered** when the file says the opposite. That is worse than having no index — it answers "still covered?" confidently and wrongly. A declaration cannot be produced by accident; a mention can.

## The convention

```sql
# name: test/sql/copy/copy_varchar_length.test
# description: BCP COPY into an existing VARCHAR(n) must accept n characters
# group: [copy]
# issue: 181
```

```cpp
// test/cpp/test_load_policy.cpp
// issue: 189
```

Queried by `scripts/ci/issue_links.py`:

| | |
|---|---|
| (no args) | validate + summary — **runs in CI**, in the `lint` job, no build needed |
| `--index` | every issue → its tests |
| `--issue 181` | which tests guard it |
| `--backlog` | names an issue in prose, declares none |

Current state: **28 issues guarded by 39 declared tests**, 18 files in the backlog.

## Back-fill, and what was deliberately left alone

Back-filled onto the **36 files that already named an issue in their header** — header placement is declared intent. The heuristic validates itself: `legacy_lob_types.test` lands as a guard for #197, with #224 correctly left undeclared.

The **18 body-prose-only** files stay in `--backlog`. Whether such a mention is really a guard is a judgement about what the test pins, so the tool surfaces the list and leaves the call to a person. The backlog *is* the tracked remainder rather than something quietly forgotten.

Also normalises `copy_nvarchar_length_validation.test` — the one file of 185 whose `# group:` sat below a blank line and therefore outside its own header. All 185 now carry `name`/`description`/`group` uniformly.

## Design details worth the review

- **The header block is narrow on purpose**: comment lines at the top, ending at the first blank line. If it ran through every leading comment, a declaration could sit forty lines down inside a preamble — easy to miss when reading, easy to add by accident when editing.
- **A malformed declaration is an error**, not an absent one. `# issue: nonsense` would otherwise fall through to "declares nothing", land silently in the backlog, and read as if nobody had tried — exactly what a typo produces.

## Verified

- **All 185 tests still register** with sqllogictest after the header edits (`--list-tests` count unchanged), and an edited test still parses and runs.
- Validator exits **1** on a malformed field and on an out-of-range number; exits **0** clean.
- `--issue 181` resolves to its test; `--issue 224` correctly reports **no guard** — which is the honest answer, and the one a grep would have got wrong.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,15 @@ jobs:
       - name: Every require-env variable is actually set
         run: scripts/ci/check_require_env.sh
 
+      # The issue -> regression-test index. Validates the `# issue:` /
+      # `// issue:` header declarations and prints the count; the backlog of
+      # files that name an issue in prose without declaring one is reported,
+      # not enforced, because deciding whether a mention is a guard needs a
+      # human. See docs/TESTING.md "Linking a test to the issue it guards".
+      # No build needed -- it reads files.
+      - name: Issue -> test index is well-formed
+        run: python3 scripts/ci/issue_links.py
+
   # ============================================================================
   # C++ Unit Tests Job - lightweight, no SQL Server / DuckDB build required.
   # Runs on every push/PR (unlike the heavy build job, which is manual).
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -416,7 +416,10 @@ C++ tests use the same field in their banner comment, under the path line:
 
 Several issues on one test is `# issue: 140, 141`. The field must sit in the
 **header block** -- the run of comment lines at the very top, ending at the
-first blank line.
+first blank line. That block is often longer than it sounds: many `.test` files
+continue straight into a prose preamble with a bare `#` instead of a blank
+line, so the rule is about a simple, predictable boundary rather than a short
+one. Put the field directly under `# group:`, where the back-fill puts it.
 
 Query the index with `scripts/ci/issue_links.py`:
 
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -395,6 +395,50 @@ statement ok
 DETACH testdb;
 ```
 
+### Linking a test to the issue it guards
+
+A test that exists to keep a fixed bug fixed declares the issue in its header,
+under `# group:`:
+
+```sql
+# name: test/sql/copy/copy_varchar_length.test
+# description: BCP COPY into an existing VARCHAR(n) must accept n characters
+# group: [copy]
+# issue: 181
+```
+
+C++ tests use the same field in their banner comment, under the path line:
+
+```cpp
+// test/cpp/test_load_policy.cpp
+// issue: 189
+```
+
+Several issues on one test is `# issue: 140, 141`. The field must sit in the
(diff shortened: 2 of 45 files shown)


VGSML (dismissed):

VGSML (approved):
Approving. I fetched the branch and ran the tool rather than reading it only:

```
$ python3 scripts/ci/issue_links.py
30 issue(s) guarded by 45 declared test(s)
16 file(s) mention an issue without declaring one (see --backlog; not an error)
$ python3 scripts/ci/issue_links.py --issue 224 ; echo "exit=$?"
no test declares itself a guard for #224
exit=3
```

(The counts have moved past the 28/39 in the description — later commits on the branch.)

The design argument is the right one and `legacy_lob_types.test` proves it: it declares `# issue: 197` and its prose says of #224 *"deliberately not pinned here"*. A grep-built index would report #224 covered; this reports it uncovered, which is the true answer. Validating the **whole field** rather than scanning it for digits is the part that makes the declaration unfakeable — `# issue: 181 (see spec 057)` indexing both 181 and 57 is exactly the accident the field exists to prevent.

Also right: distinct exit 3 for "no guard" versus 1 for "index invalid", printing `bad` in every mode, and `missing_roots()` — a CI gate that passes because it indexed nothing is the classic version of this failing silently, and we have already been bitten by that shape (`make integration-test` matching zero files and exiting 0).

**One observation, follow-up not blocker.** `--backlog` lists files that declare **nothing**, so a file that guards one issue while mentioning another never appears — and that is your own showcase example: `legacy_lob_types.test` declares 197, mentions 224, and is absent from the backlog. The index is right either way, but the more useful question is issue-keyed rather than file-keyed: *which issues are named somewhere in the tree and guarded nowhere?* That would surface #224 as a known-but-unguarded issue instead of leaving it visible only to whoever thinks to run `--issue 224`. Small change to `collect()` — worth a follow-up once #305 removes #224 from that list anyway.

**Merge-order note:** this and #301 both insert a step into `lint` right after `Every require-env variable is actually set`. Same anchor, so the second one in will conflict — trivial to resolve, just don't be surprised.

PR #305, opened 2026-09-07
fix(#224): refuse to put code-page bytes into a DuckDB VARCHAR

oluies:
**Draft** — see *Not yet verified* at the bottom. The code is complete; I could not compile it or run the repro in this session.

`mssql_scan()` put raw CHAR/VARCHAR bytes into a DuckDB VARCHAR with no transcoding and no validation. DuckDB VARCHAR is UTF-8 by contract, so a legacy-collation column produced an invalid string that *looked* like data — `upper()` on a scanned `'naïve'` returned `"NA"`.

The catalog path was never affected: `BuildColumnExpression` wraps non-Unicode string columns in a server-side `CAST` to `NVARCHAR`, so UTF-16 arrives. Only a raw scan reaches the decoder.

## Follows the issue's decision, but triggers differently — on purpose

#224 decides **fail loudly, do not transcode**. This does that. It does *not* key the failure off the collation's UTF-8 flag as the issue's implementation note suggests. It validates the **bytes**, for two reasons:

1. **It cannot emit an invalid string whatever the protocol flag says.** The flag version depends on getting a bit position right inside a 5-byte `COLLATION` field. Wrong in one direction it breaks every UTF-8 column; wrong in the other it reintroduces the bug silently. I could not verify that bit against a live server here, and I would rather the correctness not depend on it.
2. **ASCII-only data in any code page keeps working** — which is most legacy data in practice. Classifying by collation refuses those rows, turning the fix into a regression for anyone whose columns happen to be ASCII.

So a UTF-8 column passes *because it is valid UTF-8*, and the collation is read only to make the error name the culprit:

```
Column "v" is a non-Unicode CHAR/VARCHAR whose bytes are not valid UTF-8
(collation 0x00D00409, sort id 52). DuckDB VARCHAR must be UTF-8, and this
extension does not transcode legacy code pages. Either CAST it in the query --
CAST("v" AS NVARCHAR(4000)) -- or read the table through the attached catalog,
which casts server-side.
```

The **ASCII fast path runs first**. `IsAsciiRun` already exists in this file with a measured rationale (~3.1 ns for a `validate_utf8` call versus a fraction of that), so the common case never reaches simdutf.

## The fifth collation byte

The parser advanced past it under the comment *"collation is 5 bytes but we only store 4"*. That byte is the **SortId** — what identifies the code page for the `SQL_*` collations — so without it the error could not say which collation it had met. Now stored.

## Verified by reading

The staged batch decoder `DecodeChunkImpl` is **UTF-16 only** — it works in `PayloadSize()/2` units through `convert_utf16le_to_utf8` — so CHAR/VARCHAR necessarily land in `DecodeFromTds`, which is where the guard is. That was the one structural risk: a guard on the wrong path would have looked right and done nothing.

## Not yet verified

- **Not compiled.** No build was run.
- **Repro not exercised.** No SQL Server available in this session (port 14330 closed, no container).

The test added alongside encodes the issue's own reproduction, plus the two cases that matter beyond the error itself: an **ASCII row still streaming** through a raw scan, and the **`CAST` workaround the error recommends actually working**.

Happy to run `make` and the integration test against a live server on request — that is what takes this out of draft.

## Note

The test declares `# issue: 224` per the convention in #303. That declaration is inert until #303 merges, at which point `--issue 224` stops reporting "no guard" — which is currently the honest answer, and the one a grep would have got wrong.

### DIFF
--- a/DATAMODEL.md
+++ b/DATAMODEL.md
@@ -474,6 +474,40 @@ sequenceDiagram
 
 ## Layer 5 — Codec (spec 045)
 
+**Invariant: single-byte text is validated as UTF-8 before it is published (issue #224).**
+
+DuckDB `VARCHAR` is UTF-8 by contract. SQL Server's `CHAR` / `VARCHAR` / `TEXT`
+carry bytes in the column's own code page, and the TDS type token does not say
+which — with `UTF8SUPPORT` negotiated, a UTF-8-collated column and a CP1252 one
+arrive as the *same* type. So the bytes are checked, not classified by
+collation, and a column that is not valid UTF-8 raises rather than publishing a
+string every downstream function would mangle.
+
+Three decoders publish these bytes, and the guard has to sit on all of them —
+each was found only after the previous fix was believed complete:
+
+| path | decoder | when |
+|---|---|---|
+| staged batch | `binary::DecodeChunkFromStaging` | the ordinary scan |
+| constant | `binary::DecodeFromTds` via `TryEmitConstant` → `DecodeFirstValue` | a chunk whose values are uniform and non-NULL |
+| per value | `string::DecodeFromTds` | `TypeConverter::ConvertValue`, INSERT…RETURNING |
+
+Single-byte text reaches the **binary** kernel because its bytes are copied
+verbatim, exactly like `VARBINARY` (`P2StageBinary` / `PlpStageBinary` /
+`LobStageBinary` → `FinalizeKernel::Binary`); only the destination vector
+differs. `IMAGE` shares the staging arm but lands in a `BLOB`, where arbitrary
+bytes are the point, and is excluded.
+
+The catalog path is normally exempt because `BuildColumnExpression` casts
+non-Unicode string columns to `NVARCHAR` server-side — **except** a declared
+`VARCHAR(MAX)` when `mssql_convert_varchar_max` is off, which
+`NeedsNVarcharConversion` deliberately opts out. That configuration reaches the
+decoder through the catalog and now errors where it previously returned a
+corrupt string.
+
+Shared check and message: `src/include/codec/utf8_guard.hpp`.
+
+
 ```mermaid
 classDiagram
     class TypeFamily {
--- a/src/codec/binary_codec.cpp
+++ b/src/codec/binary_codec.cpp
@@ -29,6 +29,7 @@
 
 #include "codec/binary_codec.hpp"
 
+#include "codec/utf8_guard.hpp"
 #include "codec/vector_format.hpp"
 #include "copy/target_resolver.hpp"
 #include "duckdb/common/exception.hpp"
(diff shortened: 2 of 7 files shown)


oluies:
Out of draft. Built and tested against a live SQL Server — and the test found that **my first fix was on a path `mssql_scan()` never takes**.

## What the test caught that reading did not

I had reasoned that `string::DecodeChunkImpl` is UTF-16 only, therefore CHAR/VARCHAR must land in `string::DecodeFromTds`, therefore the guard there covers it. Running it disproved that: the query still failed, but with DuckDB's own error, not mine —

```
Invalid Input Error: Invalid unicode (byte sequence mismatch) detected in value construction
```

— which means the guard never fired. The real route is through the **binary** kernel, because single-byte CHAR/VARCHAR is staged exactly like VARBINARY:

```
BIGVARCHAR → column_ops: P2StageBinary / PlpStageBinary
           → FinalizeKernel::Binary
           → binary::DecodeChunkFromStaging
           → VARCHAR vector
```

Only the *destination* differs, and that difference is the bug. `column_ops.cpp` states the assumption that makes it look safe:

> collation handling lives in the scan's SELECT list (spec 026), not here

True of the catalog path, whose `CAST` to `NVARCHAR` happens server-side. False of a raw `mssql_scan()`, which has no such CAST. That comment is precisely what #224 breaks.

The guard now sits in `binary::DecodeChunkFromStaging`, gated on `col.type_id` being a char type so **genuine VARBINARY is untouched** — its bytes are meant to be arbitrary. One `simdutf::validate_utf8` over the whole staged payload in the good case (concatenated valid UTF-8 is valid UTF-8, so one call clears every row); the per-row loop runs only on failure, only to name the offending row.

## Verified, against `mssql-dev` on 14330

| | |
|---|---|
| **compiles** | ✅ — it never had been when this was opened |
| **new test** | **22/22 assertions** |
| **full integration suite** | **170 cases, 5340 assertions, all passed** |
| `make test-cpp-run` | 18/18, incl. `test_binary_codec`, `test_string_codec` |
| clang-format-14 | clean |

The integration run is the one that matters here — this touches a kernel shared with VARBINARY and every string read, so the regression risk was never in the test I wrote. It includes `utf8_support.test`, the #225 case this must not break.

The test asserts all four arms: the catalog path still returns `café`/`naïve`; an ASCII row still streams through a raw scan; the non-UTF-8 rows fail with the named error instead of publishing a corrupt string; and the `CAST` the error recommends actually works.

Fixture bytes were confirmed on the server before relying on them — `'caf'+CHAR(233)` stores as `0x636166E9` and `'na'+CHAR(239)+'ve'` as `0x6E61EF7665`: a lone `E9`, and an `EF` followed by `0x76`, which is not a valid continuation byte.

VGSML:
Thanks for this — the analysis in it is right, and two of its findings are keepers. But we've decided **not to take the runtime guard**. Reasoning below, plus the measurement you couldn't run and the one piece of evidence you said you were missing.

## What we're deciding

`mssql_scan()` takes T-SQL the caller wrote. Someone in that position can add a `CAST`. Paying for a check on the scan's hot path so that a hand-written query fails loudly instead of quietly isn't a trade we want to make — so #224 becomes documented behaviour, not enforced behaviour. Docs PR is #307; it replaces a line in `troubleshooting.md` that claimed a raw scan "may fail with UTF-8 validation error", which it never did.

## The measurement

I built a microbenchmark shaped exactly like `DecodeChunkFromStaging` — `IsAsciiRun` over the whole payload, then `validate_utf8` plus the per-value continuation-byte test — same simdutf, arm64:

| value size | ASCII | non-ASCII UTF-8 |
|---|---|---|
| 10 B | 0.59 ns/row | 0.92 ns/row |
| 40 B | 1.24 ns/row | 3.02 ns/row |
| 200 B | 6.49 ns/row | **14.15 ns/row** |

Really ~0.03 ns/byte on ASCII and ~0.07 ns/byte on non-ASCII — it scales with payload, not rows. Against the ~80 ns/value read floor that's 1.5–4% on typical widths.

**Where that lands is the problem.** A non-UTF8 `varchar` never reaches this kernel on the catalog path — `NeedsNVarcharConversion` casts it to `NVARCHAR` server-side, so it arrives as 0xE7 and takes the UTF-16 kernel. Only two things reach the binary kernel through the catalog: a column with a **UTF-8 collation** (no cast needed), and the declared `VARCHAR(MAX)` + `mssql_convert_varchar_max = false` corner you documented. So the guard bills the configuration we optimised hardest for — UTF-8 collations with `UTF8SUPPORT`, the 85.8 MB → 43.9 MB and 0.44 s → 0.25 s from #225 — and bills it for confirming what the collation already stated.

## The bit you couldn't verify

> *I could not verify that bit against a live server here, and I would rather the correctness not depend on it.*

Verified now. SQL Server 2025, a table with four `varchar` columns of different collations, dumping the five raw COLLATION bytes out of COLMETADATA:

| column | collation | word | LCID | flags | fUTF8 |
|---|---|---|---|---|---|
| `v_utf8` | `Latin1_General_100_CI_AS_SC_UTF8` (cp 65001) | `0x24D00409` | 0x0409 | 0x4D | **set** |
| `v_bin2u8` | `Latin1_General_100_BIN2_UTF8` (cp 65001) | `0x26000409` | 0x0409 | 0x60 | **set** |
| `v_cp1252` | `Latin1_General_CI_AS` (cp 1252) | `0x00D00409` | 0x0409 | 0x0D | clear |
| `v_cp1251` | `Cyrillic_General_CI_AS` (cp 1251) | `0x00D00419` | 0x0419 | 0x0D | clear |

Everything decodes without remainder: LCID 0x409 = en-US and 0x419 = ru-RU; `CI` sets `fIgnoreCase`, `AS` leaves `fIgnoreAccent` clear, `BIN2` sets `fBinary2`. The test on the 32-bit word this repo already stores is **`col.collation & 0x04000000`** (flags occupy bits 20–27, `fUTF8` is 0x40 within them).

Recording it because it settles the question for whoever meets it next — and it degrades correctly on old servers, where the bit is never set because UTF-8 collations don't exist before SQL Server 2019.

Had we gone the enforcement route, that bit is what it should have keyed on: zero cost on every UTF-8 column, byte validation only behind a cleared bit. Your reason for not classifying by collation — that it would refuse ASCII-only legacy data, which is most legacy data — stays correct and is exactly why the byte check would still have belonged *behind* the bit rather than instead of it.

## Two things worth keeping

1. **The fifth collation byte.** `offset += 5; // collation is 5 bytes but we only store 4` discards the SortId, which is what names the code page for the `SQL_*` collations. Worth landing on its own — it costs nothing and it's the difference between an error that names the collation and one that shrugs.
2. **The `DATAMODEL.md` table of which decoder publishes what** (staged batch / constant / per value), and the observation that `TryEmitConstant` → `DecodeFirstValue` bypasses the batch kernel entirely. That's a genuine structural fact about the codec that isn't written down anywhere, and it's the sort of thing that makes the *next* change to that layer wrong if it isn't. Would you split it out?

## Proposal

Close this PR in favour of #307 (docs), and open a small one carrying the SortId parse plus the `DATAMODEL.md` decoder table. #224 gets closed as documented-by-design with a link to the entry. If you'd rather keep this open while you look at the numbers, that's fine too — but the enforcement half isn't going in.

For the record: the structural work here was sound. You found all three publication paths, correctly identified that `binary::DecodeChunkFromStaging` is where single-byte text lands (I re-derived it through `ResolveKernel` — `P2StageBinary` / `PlpStageBinary` / `LobStageBinary` → `FinalizeKernel::Binary`, and `FinalizeKernel::Text` can't take a VARCHAR target because of the spec-060 "two VARCHARs never diverge" guard), and the packed-payload reasoning — validity of the concatenation plus "no value starts on a continuation byte" implies per-value validity — is correct, including catching the `C3` / `A9` join across a boundary. None of that was wasted; it's why we could reason about the cost precisely enough to decide against it.

VGSML:
#320 lands the half of this that costs nothing, and proposes closing this one.

Concretely, what came across:

- **`collation_sort_id`** — your 5th byte, both parse arms. I found the argument for it that neither of us had: `SQL_Latin1_General_CP1_CI_AS` (CP1252) and `SQL_Latin1_General_CP1251_CI_AS` (CP1251) both report **LCID 0x0409** and differ only as SortId 52 vs 106. The first four bytes genuinely cannot tell those apart, so this is completing a parse, not adding a field.
- **`DATAMODEL.md`** — your three-decoder table, rewritten to describe the behaviour that exists rather than the guard.
- **The fUTF8 bit you said you could not verify** — verified on a live SQL Server 2025 by dumping COLMETADATA for known collations. It is `0x04000000` in the word we already store; the evidence table is in the header. `Latin1_General_100_BIN2_UTF8` → `0x26000409` set, `Cyrillic_General_CI_AS` → `0x00D00419` clear, each decoding without remainder.

What did not come across is the per-value guard, and #307 is the reasoning. One thing from measuring it is worth having on the record here, because it is the part that decided it and it is not obvious:

**A non-UTF-8 `varchar` never reaches the binary kernel on the catalog path at all.** `NeedsNVarcharConversion` casts it to `NVARCHAR` server-side, so it arrives `0xE7` and takes the UTF-16 kernel. The only two things that reach `binary::DecodeChunkFromStaging` through the catalog are a **UTF-8-collated** column and a declared `VARCHAR(MAX)` with `mssql_convert_varchar_max` off. So the guard would have billed almost exactly the configuration #225 optimised hardest for, and almost none of the columns it was protecting. Cost was ~0.03 ns/byte ASCII and ~0.07 non-ASCII — scaling with **payload**, not rows: 1.24 ns/row at 40 bytes, 6.49 at 200.

Your underlying point — that the caller is handed something invalid and told nothing — was right, and that half is fixed rather than dropped. It is said once per query over COLMETADATA, where it costs nothing:

```
mssql: column "cp1251" has a non-UTF-8 collation (LCID 0x0419, SortId 0); its bytes are
returned verbatim and may not be valid UTF-8. CAST it to NVARCHAR in the query to convert.
```

Looking for somewhere to put that turned up a second hole: `SurfaceWarnings` walked `info_messages_` and did nothing, on the grounds that DuckDB had no warning API. That was true when it was written and is not now, so every `PRINT`, every `RAISERROR` below severity 11 and every procedure progress notice the server has ever sent us was being dropped on the floor. Both go to `duckdb_logs` at `WARNING` now.

Closing is your call rather than mine — say the word, or close it, whichever you prefer.

PR #304, opened 2026-09-07
ci: run the checks on stacked PRs too — closes #285

oluies:
Closes #285.

`ci.yml` triggered on `pull_request: branches: [main]`, so a PR whose base was another branch ran **none** of it. Stacked PRs are a normal way to work here, and they arrived unverified.

## The evidence

Two PRs open the same day, same repo, differing only in base branch:

| PR | base | checks |
|---|---|---|
| #283 | `main` | **14** — Build, Integration, C++ Unit Tests ×2, Lint, Windows SSPI, smoke… |
| #284 | `fix/roborev-backlog` | **4** |

The four were `Concurrency stress`, `Scan PR prose for prompt injection`, `Secret scan (TruffleHog)` and `yamllint` — every one from a workflow with its *own* triggers. No Build. No Integration. No C++ Unit Tests. No Lint.

**#284 makes the point sharply**: its entire change was to `ci.yml` itself, deleting twelve hand-rolled test blocks on the argument that `make test-cpp-run` already covers them. That argument is exactly what a `Build` job verifies — and the PR's own checks could not. It had to be dispatched by hand ([run 33250064017](https://github.com/hugr-lab/mssql-extension/actions/runs/33250064017)) to get coverage, and that result attaches to a manual run, not to the PR.

The failure was quiet in the familiar way: four checks ran and passed, so the page looked green. Only counting them against another PR revealed the ten that never ran.

## Why option 1, not option 2

The issue offered listing the stacking bases (`branches: [main, 'fix/**', 'spec/**']`). Dropping the filter entirely is better: a prefix list needs maintaining and will be wrong for whatever prefix someone picks next.

It also makes `ci.yml` **consistent with the rest of the repo**. Checked every workflow — `ci.yml` was the *only* one filtering `pull_request` by base branch. `cflite-pr`, `concurrency-tests`, `docs-build`, `kerberos` and `named-instance` all scope by `paths:` instead, which is precisely why they *did* run on #284.

## Verified before changing anything

- **No job conditions needed changing.** `build` and `integration-test` already fire on `github.event_name == 'pull_request'` irrespective of base — added by #279.
- **Concurrency is safe.** The group keys on `github.ref`, which for a PR is `refs/pull/N/merge`, so a stacked PR's run cannot cancel another PR's or `main`'s.
- `smoke-test` deliberately still excludes `pull_request` (dispatch/schedule/push only) — unchanged, and why it still shows `skipping`.
- yamllint and actionlint clean.

## Cost

[...]

PR #306, opened 2026-09-08
chore: track duckdb's v2.0 pre-release branch (v2.0-cyanoptera)

VGSML:
DuckDB cut **`v2.0-cyanoptera`** on 2026-09-03 as the stabilisation branch for v2.0 — the API track this repo already targets (spec 069). The submodule follows it now instead of `main`.

That is not a cosmetic choice. The two branches have diverged: **613 commits live on cyanoptera that `main` does not have**, going back to 2026-07-21, and `main` carries 13 that cyanoptera does not. Tracking `main` means tracking the branch where 2.0 is *not* being stabilised.

Pin moves `d7a4366603` → `070ce1f695` (**1679 commits**). The old pin is a strict ancestor, so this is a fast-forward within the same lineage, not a switch.

### 1. `ColumnList` ownership moved out of `TableCatalogEntry`

duckdb [`888cd17bee`](https://github.com/duckdb/duckdb/commit/888cd17bee) — *"moving ColumnList ownership out of TableCatalogEntry and into DuckTableEntry"*. `GetColumns()` is now `= 0`, and the base constructor no longer copies `info.columns` out of the `CreateTableInfo` it is handed:

```cpp
TableCatalogEntry::TableCatalogEntry(Catalog &catalog, SchemaCatalogEntry &schema, CreateTableInfo &info)
    : StandardEntry(...), constraints(std::move(info.constraints)) { ... }   // no columns
```

`MSSQLTableEntry` was therefore abstract and the build stopped on `error: allocating an object of abstract class type`. It owns its own `ColumnList` now.

`MakeTableInfo` also **stops filling `info.columns`** — nothing reads them there any more, so populating the list would build every table's columns twice and leave a copy nobody looks at.

### 2. cpp-httplib now takes its roots from the macOS keychain

The bundled httplib enables `CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN` by itself for Apple + clang + OpenSSL, so `azure_http.cpp` pulls in `SecTrustSettingsCopyCertificates` and the CoreFoundation accessors. The extension failed to link:

```
Undefined symbols for architecture arm64:
  "_SecTrustSettingsCopyCertificates", referenced from:
      duckdb_httplib_openssl::tls::load_system_certs(void*) in azure_http.cpp.o
```

Both frameworks are linked on `APPLE`. Taking the keychain roots is the right side of that choice rather than setting `CPPHTTPLIB_DISABLE_MACOSX_AUTOMATIC_ROOT_CERTIFICATES`: **vcpkg's static OpenSSL ships no CA bundle**, so disabling the feature leaves Azure AD token acquisition verifying against an empty trust store.

### Also

Two `EntryLookupInfo(CatalogType, Identifier)` call sites move to the `QualifiedName` overload — duckdb deprecated the old one in the same window, and it was the only warning the bump produced. `QualifiedName(info.GetQualifiedName().Name())` rather than the whole qualified name, so the DROP SCHEMA lookup keeps exactly today's semantics.

`CLAUDE.md` and the `ci.yml` comment that both said the submodule tracks `branch = main` are updated.

## Verified

- **Clean build, zero warnings** (`GEN=ninja make`, macOS arm64).
- **SQL suite green against SQL Server 2025**: 175 passed, 8 skipped.
  The two failures on a 12-worker run were `Transaction (Process ID N) was deadlocked on lock resources` on *metadata* queries — parallel workers against a freshly-initialised database, not this change. Both pass run on their own (`preload_catalog.test`, `partitioned_table.test`).
- `clang-format-14` clean; `yamllint` clean under the ruleset CI uses.

## One thing this does NOT fix, and it is worth knowing

The `azure-test` job installs the azure extension from `extensions.duckdb.org/<source_id>/…`, which exists only for commits DuckDB's own extension CI published. **It 404s for this pin — and it 404s for the current pin too**:

```
070ce1f695: 404
d7a4366603: 404
```

So that lane was already broken before this PR and is not a regression from it. It is manual-dispatch only, so nothing here goes red. Worth its own issue: any pin that tracks a development branch will keep hitting this, and the job's own error text already predicts it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1270,8 +1270,9 @@ jobs:
           # ExtensionHelper::GetVersionDirectoryName(), which returns the
           # version tag only for a RELEASE build; for anything carrying `-dev`
           # it returns DuckDB::SourceID() -- a 10-char commit hash, which never
-          # starts with `v`. The duckdb submodule is pinned to `branch = main`
-          # and this workflow passes neither OVERRIDE_GIT_DESCRIBE nor
+          # starts with `v`. The duckdb submodule is pinned to a SHA on
+          # `branch = v2.0-cyanoptera` and this workflow passes neither
+          # OVERRIDE_GIT_DESCRIBE nor
           # DUCKDB_EXPLICIT_VERSION, so `git describe` lands off-tag,
           # DUCKDB_VERSION becomes vX.Y.Z-devN and the hash is what gets used.
           # A `v*` glob matches nothing here and the step dies before running a
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -485,6 +485,24 @@ jobs:
         run: |
           mkdir -p build/release
           cd build/release
+          # -DENABLE_UNITTEST_CPP_TESTS=FALSE drops DuckDB's OWN C++ unit tests
+          # (its test/ subdirs api, appender, arrow, catalog, common, logging,
+          # memoryleak, secrets, optimizer, serialize, sql, ossfuzz, mbedtls,
+          # encryption, persistence). Not an optimisation dressed up as a fix:
+          # `make` has ALWAYS passed it -- extension-ci-tools has it in
+          # BUILD_FLAGS -- so every cmake line in these workflows was configuring
+          # a different build from the one developers run, and compiling upstream
+          # test code that nothing in this repo executes. Measured on this tree:
+          # 1937 -> 1706 build edges, with the `unittest` and test/sqlite targets
+          # identical either way. They sit OUTSIDE the gate in DuckDB's
+          # test/CMakeLists.txt, which is why the sqllogictest driver the
+          # integration job runs is unaffected.
+          #
+          # It stopped being merely wasteful on the v2.0-cyanoptera bump: a
+          # compile error in duckdb/test/common/test_http_transport_manager.cpp
+          # -- returning unique_ptr<MockHTTPParams> where unique_ptr<HTTPParams>
+          # is declared, which GCC rejects -- failed Build (linux_amd64) on a PR
+          # that touched none of it.
           cmake -G "Ninja" \
             -DCMAKE_BUILD_TYPE=Release \
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
@@ -492,6 +510,7 @@ jobs:
             -DCMAKE_TOOLCHAIN_FILE="$(pwd)/../../vcpkg/scripts/buildsystems/vcpkg.cmake" \
             -DVCPKG_MANIFEST_DIR="$(pwd)/../.." \
             -DDUCKDB_EXTENSION_CONFIGS="$(pwd)/../../extension_config.cmake" \
+            -DENABLE_UNITTEST_CPP_TESTS=FALSE \
             ../../duckdb
           cmake --build . --config Release
 
@@ -948,6 +967,7 @@ jobs:
             -DVCPKG_HOST_TRIPLET=x64-windows-static-release `
             -DVCPKG_MANIFEST_DIR="$(Get-Location)/../.." `
             -DDUCKDB_EXTENSION_CONFIGS="$(Get-Location)/../../extension_config.cmake" `
+            -DENABLE_UNITTEST_CPP_TESTS=FALSE `
             ../../duckdb
           cmake --build . --config Release
           Pop-Location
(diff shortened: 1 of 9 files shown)


VGSML:
`Build (linux_amd64)` failed, and the cause turned out to be more interesting than the pin.

## The failure

Not our code — [duckdb/test/common/test_http_transport_manager.cpp:294](https://github.com/duckdb/duckdb/blob/v2.0-cyanoptera/test/common/test_http_transport_manager.cpp):

```
error: could not convert 'result' from
  'unique_ptr<duckdb::MockHTTPParams, ...>' to 'unique_ptr<duckdb::HTTPParams, ...>'
  294 |    return result;
```

Introduced upstream by `c3dbf06693` ("refactor transport manager to have separate client pool"), which changed `make_uniq<HTTPParams>` to `make_uniq<MockHTTPParams>` and left the `return result;` below it — landing on the branch via [#25358](https://github.com/duckdb/duckdb/pull/25358) on 2026-09-07, the day before this pin was taken. GCC rejects it.

## Why we were compiling DuckDB's test suite at all

`make` has **always** passed `-DENABLE_UNITTEST_CPP_TESTS=FALSE` — extension-ci-tools puts it in `BUILD_FLAGS`. The workflows don't use the Makefile; they call `cmake` on `../../duckdb` directly, and **none of the seven invocations passed it**.

So CI has been configuring a *different build from the one anybody runs locally*, compiling DuckDB's `api` / `appender` / `arrow` / `catalog` / `common` / `logging` / `memoryleak` / `secrets` / `optimizer` / `serialize` / `sql` / `ossfuzz` / `mbedtls` / `encryption` / `persistence` suites — none of which this repo executes. "Run C++ unit tests" compiles our own standalone tests; the integration job uses `unittest` purely as the sqllogictest driver for `test/sql`.

That is also why my "clean build" claim in the description didn't catch it: `make` on a developer machine has **no `test_common` target to fail in**. I should have said "clean build of what `make` builds", and now the two agree so the distinction stops mattering.

## The fix, measured

Flag added to all seven cmake invocations (`ci.yml` ×3, `codeql.yml`, `release.yml` ×3). Configured this tree both ways:

| | build edges | `test_common` targets | `unittest` + `test/sqlite` |
|---|---|---|---|
| without flag | 1937 | 4 | 15 |
| with flag | **1706** | **0** | **15** |

`unittest` and `test/sqlite` sit outside the gate in DuckDB's `test/CMakeLists.txt`, so the driver the integration job depends on is untouched.

`release.yml` is in deliberately rather than left as a follow-up: it has the same three unguarded lines, so the next release off this pin would have failed identically, in the worst possible place to find out.

## Pin stays at the tip

The alternative was moving back 31 commits to `de6a4203c1` (the last commit before #25358 landed). That dodges this one bug and fixes nothing: CI would still compile upstream test code and still be hostage to the next such commit. The tip is also the better-validated choice — it is what the local suite ran against.

Re-verified at the tip after the change: build clean, no warnings; SQL suite **176 passed / 8 skipped**. The single failure is the known 12-worker artifact — `Transaction (Process ID N) was deadlocked on lock resources` on a *metadata* query, `partitioned_table.test`, which passes on its own (75 assertions).

Worth reporting `return result;` → `return std::move(result);` upstream; say the word and I'll open it.

oluies (approved):
Approved.

The reasoning in the description is what carries this: the pin move is a fast-forward within one lineage rather than a branch switch, and both breakages are real API changes with the upstream commit named, not churn worked around.

Checked what I could without re-fetching duckdb:

- `MERGEABLE` against `a700177`, so no textual conflict with the three CI PRs that landed today (#301, #303, #304) despite both touching all three workflow files.
- `Build (linux_amd64)` 27m38s and `Integration Test (linux_amd64)` 5m35s both pass — for a 1679-commit submodule move that is the evidence that matters, more than any reading of the diff.
- The `ColumnList` change is the kind that fails loudly (`allocating an object of abstract class type`), so a green build is strong evidence it is complete.

Two notes, neither blocking:

**CI freshness.** These checks ran before `a700177` landed at 08:41, so they were green against an older base. Nothing in this PR adds shell scripts or `.test` files, so the two gates added today — `shellcheck scripts/ci/*.sh fuzz/build.sh test/cpp/codec/run_staging_test.sh` (#301) and `issue_links.py` (#303) — have no new input to reject. Worth watching the post-merge run on `main` rather than assuming.

**`MakeTableInfo` no longer filling `info.columns`** is the change I would keep an eye on. The argument is sound — nothing reads them there now, and populating would build every table's columns twice — but it is the one edit whose failure mode is silent rather than a compile error, and it is not covered by a test that would notice a future reader of that field.

PR #307, opened 2026-09-08
docs(#224): mssql_scan hands back code-page bytes, and the docs promised an error it never raised

VGSML:
Closes the documentation half of #224. **No behaviour change.**

## The bug in the docs

`website/docs/reference/troubleshooting.md` told readers:

```sql
-- Without CAST: may fail with UTF-8 validation error for extended ASCII
FROM mssql_scan('db', 'SELECT name FROM dbo.customers');
```

It does not fail. It returns the column's **code-page bytes** into a DuckDB `VARCHAR`, which for anything outside ASCII is not valid UTF-8 — so the value is stored, displayed, and then mangled by whatever touches it:

```sql
SELECT upper(name) FROM mssql_scan('db', 'SELECT name FROM dbo.t');
-- 'naïve' -> 'NA'
```

CP1252's `ï` is the single byte `0xEF`, which UTF-8 reads as a three-byte lead, so `upper()` eats the two bytes after it.

A documented error that is never raised is worse than no documentation at all: a reader concludes the case is handled and stops looking.

## What the entry now says

The real behaviour, the fix, and **precisely where it cannot bite** — which turns out to be most places:

- **Catalog scans** rewrite every non-Unicode string column to `CAST(col AS NVARCHAR(...))` server-side, so three-part names — and the `COPY` / `CREATE TABLE AS` that read through them — are safe. The one exception is the one the setting already documents: a declared `VARCHAR(MAX)` under `mssql_convert_varchar_max = false`.
- **`NCHAR` / `NVARCHAR` / `NTEXT` / `XML`** are UTF-16 on the wire and always decoded.
- **UTF-8 collations** (`..._UTF8`, SQL Server 2019+) are already UTF-8. Worth saying out loud that the *default* collation of a SQL Server installation is **not** one of these — `SQL_Latin1_General_CP1_CI_AS` is CP1252 — so a `varchar` column is in a legacy code page unless somebody chose otherwise.
- **ASCII-only data** is valid UTF-8 in every code page.

The recommended cast is `NVARCHAR(MAX)`, not `NVARCHAR(n)`: SQL Server does not raise on a narrowing character `CAST`, so a fixed width truncates silently — the same trap the `VARCHAR(n > 4000)` note directly above it already records.

## Why documentation rather than a check

Stated in the entry itself, so the decision does not have to be rediscovered: validating every value on read would sit on the scan's hot path and be paid mostly by UTF-8 columns that cannot be affected, while the case it protects is a query the user wrote and can fix with one `CAST`. See the discussion on #305.

## Verified

- `scripts/ci/check_docs_links.py` clean.
- Anchor `#raw-mssql_scan-does-not-transcode-code-pages` is referenced from `reading/types.md` and from the Known Issues bullet; both are relative links, which is what the checker enforces.

### DIFF
--- a/website/docs/reading/types.md
+++ b/website/docs/reading/types.md
@@ -30,6 +30,30 @@ sidebar_position: 3
 | `NCHAR(n)`        | `VARCHAR`      | UTF-16LE decoded             |
 | `NVARCHAR(n)`     | `VARCHAR`      | UTF-16LE decoded             |
 
+**Non-Unicode `CHAR` / `VARCHAR` carry a code page, and only the catalog casts it away.**
+`CHAR`, `VARCHAR` and `TEXT` hold bytes in the *column's* code page —
+`Latin1_General_CI_AS` is CP1252, `Cyrillic_General_CI_AS` is CP1251 — while
+DuckDB `VARCHAR` is UTF-8 by contract. A **catalog scan** rewrites those columns
+to `CAST(col AS NVARCHAR(...))` server-side, so three-part-name queries (and the
+`COPY` / `CREATE TABLE AS` that read through them) always get UTF-8. A raw
+`mssql_scan()` has no such rewrite: the bytes arrive unchanged, and for anything
+outside ASCII the result is **not valid UTF-8** — it is not rejected, it is
+stored and then mangled by whatever touches it (`upper('naïve')` → `'NA'`).
+
+Add the cast yourself in the T-SQL:
+
+```sql
+SELECT * FROM mssql_scan('db',
+    'SELECT id, CAST(name AS NVARCHAR(MAX)) AS name FROM dbo.customers');
+```
+
+`NVARCHAR(MAX)` rather than a fixed width: SQL Server does not raise on a
+narrowing character `CAST`, so `NVARCHAR(50)` would silently truncate. Columns
+with a UTF-8 collation (`..._UTF8`, SQL Server 2019+) need nothing — they are
+already UTF-8 — but note that the default collation of a SQL Server
+installation is *not* one of those. See
+[Troubleshooting](../reference/troubleshooting.md#raw-mssql_scan-does-not-transcode-code-pages).
+
 ### Binary Types
 
 | SQL Server Type   | DuckDB Type    | Notes                        |
--- a/website/docs/reference/troubleshooting.md
+++ b/website/docs/reference/troubleshooting.md
@@ -184,11 +184,34 @@ SELECT id, name FROM db.dbo.customers;
   legacy code page unless somebody chose otherwise.
 - **ASCII-only data**, in any code page: every byte is already valid UTF-8.
 
-Tracked as [issue #224](https://github.com/hugr-lab/mssql-extension/issues/224).
-Validating every value on read was considered and rejected: the check would sit
-on the scan's hot path, and it would be paid mostly by UTF-8 columns that cannot
-be affected. An explicit `CAST` in a query you wrote yourself is the cheaper
-contract.
+### Why this is not validated for you
+
+Tracked as [issue #224](https://github.com/hugr-lab/mssql-extension/issues/224),
+and deliberately answered with documentation rather than a runtime check.
+
+The obvious fix is to validate every value as it is decoded and raise on the
+first one that is not UTF-8. It was considered and rejected, because of where
+that check would have to live.
+
+**The string codec is built to avoid exactly that call.** Its ASCII fast path
+exists for one reason, recorded next to it in `codec/string_codec.hpp`: measured
+on 12-byte values, *a simdutf call costs ~3.1 ns of pure overhead* against *~3 ns
+of actual byte work* — the call is as expensive as the transcoding it performs.
+Adding validation to the read path puts one back, per value, and as a **second
+pass over bytes the decoder has already copied**. That is a real cost on every
+scan, paid in full by the UTF-8 and `NVARCHAR` columns that cannot be affected by
+this in the first place.
+
+**And it would be paid on the wrong path.** The catalog scan — three-part names,
+and the `COPY` / `CREATE TABLE AS` built on them — is the path that carries
+volume, and it is already safe here: it casts non-Unicode string columns
+server-side, so the bytes it decodes are Unicode by construction. `mssql_scan()`
+is the escape hatch for hand-written T-SQL. Slowing the main path to guard the
+escape hatch is the wrong trade.
+
+**What the escape hatch gets instead is this page.** Someone writing raw T-SQL is
+already choosing the exact columns and can add `CAST(col AS NVARCHAR(MAX))`; what
+they were missing was any statement that they had to. That is what changed.
 
 
 ### Unicode Transcoding (simdutf)
--- a/website/docs/reference/troubleshooting.md
+++ b/website/docs/reference/troubleshooting.md
@@ -123,15 +123,72 @@ SET mssql_convert_varchar_max = false;
 
 **Notes:**
 
-1. **Catalog queries only**: This conversion applies only to catalog-based queries (three-part naming like `db.schema.table`). When using `mssql_scan()` with raw SQL, you must manually add CAST expressions:
+1. **Catalog queries only**: This conversion applies only to catalog-based queries (three-part naming like `db.schema.table`). When using `mssql_scan()` with raw SQL, you must add the CAST yourself — see [Raw `mssql_scan()` does not transcode code pages](#raw-mssql_scan-does-not-transcode-code-pages) below for what happens if you do not.
 
 ```sql
--- Without CAST: may fail with UTF-8 validation error for extended ASCII
+-- No CAST: the raw code-page bytes are returned as-is. For a non-UTF8
+-- collation that is NOT valid UTF-8, and string functions will mangle it.
 FROM mssql_scan('db', 'SELECT name FROM dbo.customers');
 
--- With CAST: properly handles extended ASCII characters
-FROM mssql_scan('db', 'SELECT CAST(name AS NVARCHAR(100)) AS name FROM dbo.customers');
+-- With CAST: SQL Server transcodes to UTF-16 and the extension decodes it.
+FROM mssql_scan('db', 'SELECT CAST(name AS NVARCHAR(MAX)) AS name FROM dbo.customers');
 ```
+
+### Raw `mssql_scan()` does not transcode code pages
+
+**Symptom.** A string read through `mssql_scan()` looks right in the output but
+behaves as if it were truncated. The classic shape is a string function
+returning less than it was given:
+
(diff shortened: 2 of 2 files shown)


oluies (commented):
The writing here is good — the CP1252 `0xEF` explanation of why `upper('naïve')` returns `'NA'` is the clearest statement of this bug anywhere in the repo, and `NVARCHAR(MAX)` rather than a fixed width is exactly right for the reason given.

**I have not approved it, because it contradicts #305 on substance rather than on files.**

#305 (open, mine, 11 checks green, blocked only on review) makes `mssql_scan()` **raise** on non-UTF-8 single-byte text instead of publishing it. This PR documents the opposite as settled:

> It is not rejected — it is stored, displayed, and then mangled by whatever touches it.

and explicitly rejects the approach #305 takes:

> Validating every value on read was considered and rejected: the check would sit on the scan's hot path, and it would be paid mostly by UTF-8 columns that cannot be affected.

They do not conflict on files — this is `website/docs/`, #305 is `src/`, `test/`, `DATAMODEL.md` — so both will merge cleanly and the result will be documentation describing behaviour the shipped code does not have. That is precisely the fault this PR's own title names, inverted.

## On the performance argument specifically

It is answerable, and #305 answers it — that objection was raised in review there (roborev 1420) against an earlier version that did validate per value, and the current form does not:

- **all-ASCII payload** → one `IsAsciiRun` pass, no `validate_utf8` call at all. This is the common case.
- **non-ASCII** → **one** `simdutf::validate_utf8` per *chunk*, plus one byte test per value checking no value begins on a continuation byte. Whole-payload validity alone is unsound because staged values are packed with no separator, so `0xC3` ending one value and `0xA9` starting the next validate as `é` across the boundary; the boundary test closes exactly that.
- The per-row loop runs only on failure, to name the offending row.

So a UTF-8 column pays one call per chunk and O(count) byte tests — not a call per value. Whether that is still too much is a fair question, but it is a different question from the one the paragraph answers.

## What I think should happen

Not my call, but the options as I see them:

1. **#305 merges first, then this is amended** — change the "not rejected" paragraph to describe the error and keep everything else, which is most of the page and all of it good.
2. **This merges as-is and #305 is closed** — a legitimate choice if the hot-path cost is judged unacceptable. It would want issue #224 reclassified from a bug to documented behaviour, and #305's test deleted rather than left orange.

Either way the two should be decided together. What must not happen is both merging in the current order, which ships docs promising the old behaviour against code with the new one.

Happy to do the amendment in (1) against your text if that is the direction — I would keep this page essentially intact and rewrite one paragraph.

VGSML:
You are right about the important thing, and right that my paragraph argued badly. Taking those in order, then a third option neither of us listed.

## The process point is yours, unreservedly

Two PRs that describe opposite behaviour must be decided together, and merging both in this order ships documentation promising the old behaviour against code with the new one — the exact fault this PR's title names, inverted. That is not a nuance, it is the whole point of the change. Nothing should merge here until #305 is decided.

## My performance paragraph was wrong for your design

It argued against a per-value `validate_utf8`, and #305 does not do that. One `IsAsciiRun` on the payload, one `validate_utf8` per **chunk** when that fails, plus one byte test per value for the continuation-byte boundary. I was answering a question you had already stopped asking, and the paragraph as written is not a fair description of your PR.

## But it is not free, and here is the measurement

I built a microbenchmark shaped exactly like your kernel — `IsAsciiRun` over the whole payload, then `validate_utf8` plus the per-value boundary test — same simdutf, arm64:

| value size | ASCII (fast path) | non-ASCII UTF-8 |
|---|---|---|
| 10 B | 0.59 ns/row | 0.92 ns/row |
| 40 B | 1.24 ns/row | 3.02 ns/row |
| 200 B | 6.49 ns/row | **14.15 ns/row** |

Really **~0.03 ns/byte on ASCII and ~0.07 ns/byte on non-ASCII** — it scales with payload, not with rows, because `IsAsciiRun` is a **second full pass over bytes the decode has already copied**. Against the ~80 ns/value read floor that is 1.5–4% at typical widths and ~18% on a 200-byte column.

## Where that cost lands is the real objection

Not on legacy columns. Trace which single-byte text actually reaches `binary::DecodeChunkFromStaging` **through the catalog**: `NeedsNVarcharConversion` casts non-Unicode `varchar` to `NVARCHAR` server-side, so it arrives as `0xE7` and takes the UTF-16 kernel — no guard, no cost. Exactly two things reach the binary kernel that way:

1. a column with a **UTF-8 collation** (no cast needed), and
2. a declared `VARCHAR(MAX)` with `mssql_convert_varchar_max = false`.

So the guard bills the configuration we optimised hardest for — UTF-8 collations with `UTF8SUPPORT`, the 85.8 MB → 43.9 MB and 0.44 s → 0.25 s from #225 — and bills it to confirm what the collation already states. `mssql_scan()` is the escape hatch; the catalog scan is the path that carries volume. That asymmetry, not the constant, is the objection.

## The third option: gate on the collation's fUTF8 bit

You wrote in #305 that you could not verify that bit against a live server. **It is verified now.** SQL Server 2025, four `varchar` columns of different collations, dumping the five raw COLLATION bytes out of COLMETADATA:

| collation | word | LCID | flags | fUTF8 |
|---|---|---|---|---|
| `Latin1_General_100_CI_AS_SC_UTF8` | `0x24D00409` | 0x0409 | 0x4D | **set** |
| `Latin1_General_100_BIN2_UTF8` | `0x26000409` | 0x0409 | 0x60 | **set** |
| `Latin1_General_CI_AS` (cp1252) | `0x00D00409` | 0x0409 | 0x0D | clear |
| `Cyrillic_General_CI_AS` (cp1251) | `0x00D00419` | 0x0419 | 0x0D | clear |

Everything decodes without remainder — LCID 0x409 = en-US, 0x419 = ru-RU; `CI` sets `fIgnoreCase`, `AS` leaves `fIgnoreAccent` clear, `BIN2` sets `fBinary2`. The test on the 32-bit word we already store is **`col.collation & 0x04000000`** (flags occupy bits 20–27, `fUTF8` is 0x40 within them). It degrades correctly on old servers: UTF-8 collations do not exist before SQL Server 2019, so the bit is never set there and the byte check runs.

With that gate first:

- **fUTF8 set** → skip. One `AND` per column per chunk. The catalog path and every modern UTF-8 column pay **nothing**.
- **fUTF8 clear** → genuinely a legacy code page; run your ASCII probe and validation, at 0.03 ns/byte, on the only columns that can be wrong.

And your reason for *not* classifying by collation stays correct and is exactly why the byte check belongs **behind** the bit rather than instead of it: ASCII-only data in a legacy code page must keep working, and only the bytes can tell you that.

That would make #305 cost approximately nothing where I objected, and it needs the fifth collation byte you already parse in that PR.

## Where the decision stands

The call is @VGSML's, and as of now it is your option 2 — documented behaviour, #224 reclassified, this page carrying the reasoning. I have rewritten the offending paragraph (`1b8b2e2`) so it argues from the measured second-pass cost and the path asymmetry rather than from a per-value call that #305 does not make.

If the fUTF8 gate changes that judgement, I would rather it did — enforcement that costs nothing is better than documentation, and your structural work in #305 (all three publication paths, the packed-payload boundary reasoning, the fifth collation byte) is what would carry it. Say the word and I will hold this PR until #305 is resolved either way.

VGSML:
> Whether that is still too much is a fair question, but it is a different question from the one the paragraph answers.

Agreed, so I measured the question you actually asked — #305 against `main`, same DuckDB submodule pin, on the shipped path rather than a microbenchmark.

**First, a correction to my own number.** I estimated up to ~14 ns/row on a 200-byte column from an isolated kernel benchmark. That was too high: in the shipped path the payload is already in L1 after the decode's `memcpy`, so the second pass is cheaper than in isolation. The real figure is a third of that.

## Method

`test/bench/bench_live_server.sh`, which exists to separate **client CPU** from wall clock — the guard is pure client work and wall clock drowns it in server I/O. Two statically-linked builds differing only by #305's diff (I merged current `main` into the PR branch first, so the delta is the guard and not its 11-commit-stale base). 500K rows, `threads=1`, median of 7 reps, 20 query iterations per process, `startup` control subtracted. Two alternating rounds so machine drift lands on both.

The harness has the control built in: `str*` steps are `NVARCHAR`, which takes the **UTF-16 kernel** and cannot reach the guard at all. If those move with the treatment, the measurement is worthless.

## Result

| step | `main` | #305 | delta | | same-binary spread |
|---|---|---|---|---|---|
| `read_str16u` — NVARCHAR(16) | 41.2 | 42.1 | +0.9 | +2.3% | 0.8–1.1 → **noise** |
| `read_str200` — NVARCHAR(200) | 292.4 | 294.0 | +1.6 | +0.5% | 0.6–2.8 → **noise** |
| **`read_vstr16u`** — VARCHAR(32) UTF-8 | 34.4 | **36.9** | **+2.5** | **+7.3%** | 0.8 → **3× noise** |
| **`read_vstr200`** — VARCHAR(200) UTF-8 | 159.3 | **165.0** | **+5.7** | **+3.6%** | 1.2–1.4 → **4× noise** |

*(client CPU ns per value, median of 7, two rounds averaged; Apple M4 Max, arm64)*

The controls stayed inside their own run-to-run spread. The guarded steps did not, and the direction was consistent in **both** rounds — 36.5 and 37.3 against 34.8 and 34.0, no overlap. So the effect is real at **+7.3% of client CPU on a 32-byte UTF-8 varchar column**, not the 18% I implied and not zero.

## What the same run says about which columns pay

| | 16-char column | 200-char column |
|---|---|---|
| `NVARCHAR` | 45.8 | 293.8 |
| `VARCHAR` + UTF-8 collation | **33.6** | **163.0** |

A UTF-8 varchar is **~1.4–1.8× cheaper** on client CPU than the nvarchar carrying the same text — which is #225's whole point, visible here in CPU rather than in wire bytes.

And that is where the 7.3% lands. Trace what reaches the binary kernel *through the catalog*: `NeedsNVarcharConversion` casts non-Unicode `varchar` to `NVARCHAR` server-side, so it arrives `0xE7` and never sees the guard. Exactly two things get there — a **UTF-8-collated** column, and `VARCHAR(MAX)` with `mssql_convert_varchar_max = false`.

So the tax falls on the fastest read path we have, to check columns whose collation already guarantees the answer. That is the objection; the constant only sets its size.

## Which is why the fUTF8 gate matters more than this table

With `col.collation & 0x04000000` tested first (verified values in my previous comment), the 7.3% goes to one `AND` per column per chunk for every UTF-8 column, and your byte validation runs where the bytes genuinely can be wrong. Your reason for not classifying by collation stays intact and is exactly why the check belongs **behind** the bit rather than instead of it.

I would rather ship that than ship this page. Raw data in the session if you want it; happy to re-run on Linux/x86 before anyone treats a single arm64 number as decisive — simdutf picks a different kernel there.

oluies (approved):

PR #308, opened 2026-09-08
fix(catalog): the 1205 retry gave up exactly where the deadlock actually lands

VGSML:
Found while investigating why `make test` at 12 workers loses 1–2 files per run to
`Transaction (Process ID N) was deadlocked on lock resources`. CI never saw it: `integration_test.sh` ends in a single `"$UNITTEST_BIN" -f /tmp/mssql_selected.txt`, so the files run **serially** and one test's `DROP TABLE` never overlaps another's catalog load.

## The guard was right about the hazard and wrong about the case

`RunMetadataQuery` already retried a deadlock victim — six attempts, backoff — but only while `rows_delivered == 0`:

```cpp
if (result.error_number == 1205 && rows_delivered == 0 && attempt < MAX_ATTEMPTS) {
```

The reasoning was sound: rerunning a partially-consumed stream appends the same rows twice, which is the *"Column with name x already exists!"* class. The problem is that it made the retry useless exactly where 1205 lands. The bulk catalog query streams **one row per column of every table in a schema** — by the time two sessions collide it has almost always delivered rows. The retry protected the queries that rarely deadlock and gave up on the one that does.

Callers now pass a `reset` that undoes their partial accumulation, and the rerun becomes legal mid-stream.

## Every call site answers the question

Including the ones that are already idempotent — *"is this restartable?"* is something each site should be made to state, not inherit:

| site | reset |
|---|---|
| `GetTableMetadata` | re-empty the slot the erase/emplace created |
| `LoadAllTableMetadata` | re-clear the map it cleared, reset the group-by cursor |
| `BulkLoadAll` | cursor + this schema's share of the totals **only** |
| `EnsureSchemasLoaded` | nothing — `emplace()` on an existing key is a no-op |
| `EnsureTablesLoaded` / `LoadSchemas` / `LoadTables` | clear entries written from a catalog mid-change |
| `LoadColumns` | `columns.clear()` — `push_back` is the one that is outright wrong to repeat |
| `PrimaryKeyInfo::Discover` | `columns.clear()` — a composite PK aborted after its first column would list it twice, and the rowid STRUCT built from that is **wrong**, not missing |

`BulkLoadAll` is the one worth reading twice: its reset deliberately does **not** `tables.clear()`. That pass *merges* into a map which may hold entries this query never covers — a table excluded by `table_filter`, say — so clearing would drop them. It does not need to: emptying `current_table` makes the rerun's first row take the new-table branch, which clears that table's columns before refilling them.

## What this is not

**It is not a fix for the deadlock.** Reading the graphs, the cycle is between our multi-table catalog join and a concurrent `DROP TABLE`, and `DROP` acquires in a fixed order our plan is not guaranteed to follow. Across **8 captured reports** in one test session:

- ours was the **victim every time** (`DuckDB MSSQL Extension` vs `DROP TABLE [dbo].[ctas_*]`, `insert_multirow`, `ExecCacheTest`, …);
- **`sys.sysrowsets` — the `sys.partitions` leg — was one of the two resources in 7 of the 8**;
- both inversions occur: we hold `sysrowsets`-S and want `Sch-S` on the object while DROP holds `Sch-M` and wants `sysrowsets`-X (3 reports); and we hold `sysschobjs`-S and want `sysrowsets`-S while DROP holds the reverse (4 reports).

Two dead ends, checked rather than assumed:

- **Isolation level is not a lever.** Catalog reads ignore the session isolation level — wrapping the query in `REPEATABLE READ` + `BEGIN TRAN` and inspecting `sys.dm_tran_locks` afterwards leaves only a `DATABASE` S lock; the key locks were taken and released inside the scan regardless. (`READ UNCOMMITTED` would also reintroduce the duplicate-row hazard this very guard exists for.)
- **`sys.dm_db_partition_stats` instead of `sys.partitions` does not remove the cycle.** Plan references to `sysrowsets` fall 30 → 10, but `sysidxstats` appears (12). It reshuffles the surface rather than removing it.

## Measured

`make test` at 12 workers: **1–2 failures per run → four consecutive runs of 177 passed / 0 failed**. Build clean, `clang-format-14` clean.

## Follow-ups this suggests

Not in this PR:

1. **Cut the exposure window.** 186 metadata rows take ~3 ms server-side but ~50 ms streamed to a *local* `sqlcmd` — the catalog locks are held for as long as the **client** takes to consume, and our client is slower than sqlcmd. Materialising into a `#temp` server-side and streaming from tempdb would bound it by server speed instead.
2. **`SET DEADLOCK_PRIORITY LOW`** on metadata connections. We are already the usual victim (a read-only transaction is the cheapest to roll back), but "usually" is not "always", and the bad outcome is a *user's* DDL aborting because our catalog refresh happened to collide.

### DIFF
--- a/src/catalog/mssql_metadata_cache.cpp
+++ b/src/catalog/mssql_metadata_cache.cpp
@@ -194,7 +194,7 @@ static bool IsTTLExpired(const std::chrono::steady_clock::time_point &last_refre
 using MetadataRowCallback = std::function<void(const vector<string> &values)>;
 
 static void RunMetadataQuery(tds::TdsConnection &connection, const string &sql, MetadataRowCallback callback,
-							 int timeout_ms) {
+							 int timeout_ms, const std::function<void()> &reset) {
 	// Log the query being executed (truncated for readability)
 	CACHE_DEBUG(1, "RunMetadataQuery: timeout=%dms, sql=%.120s%s", timeout_ms, sql.c_str(),
 				sql.size() > 120 ? "..." : "");
@@ -203,10 +203,21 @@ static void RunMetadataQuery(tds::TdsConnection &connection, const string &sql, 
 	// reads and the server's own message says "Rerun the transaction". DuckDB
 	// 2.0's higher scan/sink parallelism overlaps a test's DDL (Sch-M) with
 	// catalog loads on sibling pooled connections often enough to make 1205 a
-	// per-run event (PR #267 review). Retry ONLY when the victim query
-	// delivered no rows yet — after the first row the callback has consumed
-	// state that a rerun would duplicate ("Column already exists" class), so a
-	// mid-stream deadlock stays fatal.
+	// per-run event (PR #267 review).
+	//
+	// The rerun used to be allowed ONLY before the first row, because after it
+	// the callback holds state a second pass would duplicate ("Column with name
+	// x already exists!"). That guard made the retry useless exactly where 1205
+	// actually lands: the bulk catalog query streams one row per COLUMN of every
+	// table, so by the time two sessions collide it has almost always delivered
+	// rows. `reset` closes that — the caller undoes its partial accumulation and
+	// the query is re-run from the top.
+	//
+	// Note what this is and is not. It does not stop the deadlock; the cycle is
+	// between our multi-table catalog join (holding a key in sys.sysschobjs,
+	// wanting sys.sysrowsets) and a concurrent DROP TABLE holding the reverse,
+	// and no client-side isolation setting reaches it — catalog reads ignore the
+	// session isolation level. This only makes losing the race recoverable.
 	constexpr int MAX_ATTEMPTS = 6;
 	auto start = std::chrono::steady_clock::now();
 	for (int attempt = 1;; attempt++) {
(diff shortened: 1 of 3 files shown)


VGSML:
@oluies — context for reviewing this, because the ground moved after I opened it.

**This PR is the backstop, not the fix.** The structural fix is spec 071, now up as **#309**. I'd propose taking this one **as-is** anyway, and the reasoning is below rather than "trust me".

## What the measurement found after this PR was written

Against a synthetic **200,000-table / 1,200,000-column** catalog on SQL Server 2025:

`sys.partitions` reads `sysrowsets`, which is clustered on `rowsetid` — `object_id` is derived and has **no index**, so filtering by it buys nothing:

| lookup by `object_id` | base table | logical reads |
|---|---|---|
| `sys.partitions` | `sysrowsets` | **3200** (2 scans) |
| `sys.dm_db_partition_stats` | `sysrowsets` | **1600** (1 scan) |
| `sys.indexes` | `sysidxstats` | **6** (seek) |

And the thing forcing that `GROUP BY` is `COUNT(*) AS partition_count` — a field `ParseTableShape` writes and **nothing reads**, not even copied into `MSSQLTableEntry`. Spec 049 added it for a write path that never picked it up.

So the single most contended resource in the catalog queries is there to populate a dead field. Taking it out: single-table metadata **18.1 ms → 0.7 ms** net, `sysrowsets` reads **3200 → 0**, and on the deadlock axis (3 reps each, 150 metadata queries against 3 DDL sessions):

| DDL churn | with the aggregate | without |
|---|---|---|
| CREATE/DROP INDEX | 127, 141, 93 | **0, 0, 0** |
| CREATE/DROP TABLE | 63, 73, 46 | **32, 24, 24** |

Which lines up with the 8 captured cycles in this PR's description: `sysrowsets` was one of the two resources in **7 of 8**.

## Why this PR still stands on its own

1. **It does not go to zero.** Table DDL stays at 24–32 per 150 after the aggregate leaves, because the `sysschobjs` ↔ `syscolpars` / `syssingleobjrefs` class remains — that was the 8th cycle. Those still need a correct retry, and today the retry gives up on them for the same reason it gave up on the others: rows have already been delivered.
2. **It fixes a contract, not a query.** The `rows_delivered == 0` guard is wrong independently of what the SQL looks like — a partially-consumed stream must either be undoable or fatal, and "fatal" was chosen without the callers ever being asked whether they could undo. Every call site now answers that question, and the answers stay valid however 071 reshapes the queries.
3. **071 makes one of these resets simpler, not unnecessary.** W3 replaces the streaming group-by with client-side hash grouping keyed on `object_id`, which retires the `current_table` cursor — the fiddliest reset here, the `BulkLoadAll` one. The mechanism is still needed; that particular unwind gets easier.

No conflict between the two: this touches the retry plumbing, 071 touches the SQL and the load paths. Merge order does not matter.

## Measured effect of this PR alone

`make test` at 12 workers, unchanged queries: **1–2 failures per run → four consecutive runs of 177 passed / 0 failed**. Deadlocks still occur; they are now absorbed.

Happy to hold this until 071 lands if you'd rather review them together — but they are independent, and the failing-test noise this removes is in the way of everything else meanwhile.

oluies (approved):

Issue #86, opened 2026-03-25
SHOW ALL TABLES fails after MSSQL attach


sashapodgoreanu:
## Description
After attaching a SQL Server database using the mssql extension, basic queries work correctly, but `SHOW ALL TABLES` fails.

## Environment
- OS: Windows
- DuckDB version: 1.5
- SQL Server: 2022
- Extension: mssql (hugr-lab)

## Steps to Reproduce
1. Load the extension:
   LOAD mssql;

2. Attach a SQL Server database (using either SECRET or connection string):
   ATTACH 'connection_string' AS mssql_db (TYPE MSSQL);

3. Run a query:
   SELECT * FROM mssql_db.some_table;
   -- Works correctly

4. Run:
   SHOW ALL TABLES;

## Expected Behavior
`SHOW ALL TABLES` should list tables from all attached databases, including MSSQL.

## Actual Behavior
The command hangs with no error.

## Notes
- Direct SELECT queries on MSSQL tables work fine.
- The issue seems specific to metadata listing.

VGSML:
Thanks for reporting this. Could you provide the error?
What the version of the sql server and deployment azure/selfhosted? How many tables/schemas

sashapodgoreanu:
Hello,

We are using a self-hosted SQL Server 2022 instance.
Our environment is quite large, with approximately 10,000 schemas and between 100 and 10,000 tables per schema.

I can confirm that SHOW ALL TABLES does not fail. However, the query executed by the extension in this phase (to read metadata from the attached database) is extremely slow in our scenario.

The main issue is not SHOW ALL TABLES itself, but what happens afterward. When we attempt to attach another database, DuckDB triggers a scan that results in reading metadata from SQL Server.
In our case, this metadata scan is very expensive and causes the operation to hang or take an excessive amount of time.


We created a Test DB, created 10000 Schemas with 0 tables, and the show all tables is very slow! 30 minutes and still running. 

I enabled debug and i see the queries scrolling in the terminal.

<img width="916" height="472" alt="Image" src="https://github.com/user-attachments/assets/536eb859-2fce-49d9-a727-81457aeefd30" />

sashapodgoreanu:
Can "lazy loading" of metadatata of a schema/table be an option as a new feature? 
We can use  schema_filter to filter some schemas, but 10K tables we need them all. 

Thanks.

VGSML:
[...]

PR #309, opened 2026-09-08
spec 071: what catalog metadata actually costs

VGSML:
Spec only — no behaviour change here. Proposes the fix for [#86](https://github.com/hugr-lab/mssql-extension/issues/86) and the defect found alongside it. Implementation of W1 follows in its own PR.

> **This spec has been rewritten once already.** Measurement killed most of the first draft — including two of its own proposals and one claim about our code. §5 keeps every dead end with its numbers, because that is half the value here.

Everything was measured against a synthetic **200,000-table / 1,200,000-column** catalog on SQL Server 2025. The reproduction recipe is §0.

## The row count never needed `sys.partitions`

Every metadata query carries an aggregate over `sys.partitions`, which reads `sysrowsets` — clustered on `rowsetid`, with `object_id` derived and unindexed. So filtering by `object_id` buys nothing. Five sources measured for **one** object's row count:

|---|---|---|---|
| `sys.sysindexes` | `sysidxstats` | 6 | 1 ms |
| `sys.dm_db_stats_properties` | `sysidxstats` | ~17 | 1 ms |

`OBJECTPROPERTYEX` answers out of object metadata, agrees with `sys.partitions` **exactly** (5000/5000, 150000/150000, 0 for empty, NULL for a VIEW — the same 0 the catalog already carried), needs no statistics, and is not deprecated. **Every** row count in the database: **84 ms** against **2400 ms**.

The shape half splits the same way: `sys.indexes` seeks (6 reads); `sys.partitions` cannot. What forced them to share a subquery is `COUNT(*) AS partition_count` — a field `ParseTableShape` writes and **nothing reads**, not even copied into `MSSQLTableEntry`.

**Result: nothing is given up.** Counts stay, the planner's cardinality stays, only the dead field and the scan go. Single-table **18.1 ms → ~1 ms**, `sysrowsets` reads **3200 → 0**.

## Issue #86 is that same aggregate plus a harder problem

An **empty** schema, database holding 200K objects:

| | CPU |
|---|---|
| current | **2708 ms** |
| with the seekable sources | **484 ms** |

484 ms to return zero rows, because `sys.objects` filters metadata visibility per object in the *database*. **And there is no per-schema seek to buy** — `sys.objects`, `sys.tables` and `INFORMATION_SCHEMA.TABLES` all plan as a Clustered Index Scan on `sysschobjs.clst`, with a literal `nsid = 1` as much as with `SCHEMA_ID()`, because the nonclustered `(nsid, name)` index does not cover `type`/`nsclass`/`pclass`.

So the catalog pass is a floor **per statement**, and the only lever is one statement instead of N (W2). 10,000 empty schemas: ~7.5 h today, ~80 min after W1, **~2 s** as one query.

## Withdrawn from the first draft

- **"Stop fetching the row count on the single-table path."** `scan_cardinality.test` catches it: the planner reads the catalog's copy first, so removing it plans every direct query at `~1 row`. Making the count cheap was the better answer.
- **"Feed catalog counts to `PreloadRowCount`."** Already implemented — `MSSQLTableSet::Scan` calls it for every table (`mssql_table_set.cpp:179`). The draft claimed `mssql_preload_catalog.cpp` was the only caller; that came from a **truncated grep**, and the conclusion drawn from it (that `SHOW ALL TABLES` runs a DMV query per table) was wrong.

## Dead ends, with numbers

Isolation-level tricks (**catalog reads ignore the session isolation level**); the DMV swap; correlating the aggregate; splitting it into a second statement (−42% at 200 tables, a wash at 200K); server-side `#temp` materialisation (**+89%**, and its first measurement showed a 96% *win* because `SELECT INTO` was failing on Msg 1038 and I was timing a query that did nothing); the sargable schema predicate; **the bulk statistics load** (2400 ms at 200K, and restricting it to one schema barely helps at 2105 ms — the cost is the scan, verified by discarding rows server-side); `sys.sysindexes` (cheap but deprecated); `sys.dm_db_stats_properties` (cheap but **NULL for a table with no statistics** — exactly a table CTAS or COPY has just created).

Two of this investigation's own numbers are **retracted** for not surviving repetition: an `ALTER TABLE` deadlock rate (contaminated by churn sessions left running from the previous DDL kind) and a `TRUNCATE` comparison (4–80 per 150 on identical conditions).

## Relationship to #308

Complementary. #308 makes losing a catalog deadlock recoverable; W1 removes `sysrowsets`, which was one of the two resources in **7 of 8** captured cycles. Merge order does not matter.

### DIFF
--- a/specs/071-catalog-metadata-cost/spec.md
+++ b/specs/071-catalog-metadata-cost/spec.md
@@ -1,0 +1,268 @@
+# Spec 071 — What catalog metadata actually costs
+
+**Status**: PROPOSED.
+**Closes**: [#86](https://github.com/hugr-lab/mssql-extension/issues/86) (`SHOW ALL TABLES` unusable on a database with many schemas).
+**Relates to**: PR #308 (the 1205 deadlock-victim retry — the reset callback that
+makes losing a deadlock recoverable). This spec attacks the same queries from the
+other side: W1 removes the resource that 7 of 8 captured deadlock cycles were
+fighting over, so the retry has far less to absorb.
+
+Four defects in the metadata layer, found by measuring rather than reading. They
+are independent, they land in ascending order of size, and three of them are the
+same root cause seen from different paths: **the `sys.partitions` aggregate that
+every catalog query carries, for a column nothing reads.**
+
+---
+
+## 0. How everything below was measured
+
+Reproduce before trusting. SQL Server **2025** (17.0.4075.5) in Docker — the 2022
+image crash-loops on Apple Silicon, see `docs/TESTING.md` — with a synthetic
+catalog in `TestDB`:
+
+```sql
+-- 200,000 tables x 6 columns, one schema `big`, plus an empty schema `empty1`
+CREATE TABLE big.tN (id INT, c1 VARCHAR(50), c2 INT, c3 DECIMAL(18,4),
+                     c4 DATETIME2(3), c5 NVARCHAR(100));
+```
+
+Resulting size: **200,000 tables / 1,200,000 columns / 200,230 rows in
+`sys.partitions`**. Creation ran at ~1 ms per table and did **not** degrade with
+catalog size (0.8 ms/table at 2K, 0.97 ms/table at 97K) — the write side of the
+catalog scales fine, which is why everything below is about reads.
+
+- **Wall clock**: `sqlcmd -i q.sql -o /dev/null` with `GO N` inside one session,
+  so connection setup is paid once. A `SELECT 1` script measured the same way is
+  subtracted as the baseline (~20.7 ms/rep here — it is `docker exec` plus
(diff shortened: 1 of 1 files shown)


VGSML:
Folded into **#312**, which now carries the spec and its implementation in one PR.

Separating them was the mistake I have been arguing against on #307: this PR was `MERGEABLE` on its own, so it could have merged first and left `specs/071-catalog-metadata-cost/spec.md` describing an `OBJECTPROPERTYEX` row-count source that no code in `main` used. The spec was rewritten twice under measurement — dropping two of its own workstreams, then replacing the whole row-count source — so by the time the implementation existed it described that implementation rather than proposing it. It belongs in the same change.

All three commits (`e58e43f`, `1540eb3`, `a08ceaf`) are cherry-picked into #312 unchanged.

PR #310, opened 2026-09-08
fix(catalog): spec 071 — what catalog metadata actually costs

VGSML:
Implements **spec 071** (#309) in full — W1 and W2 — plus a comment correction folded in from #311. **Stacked on #308**, because all three rewrite the same parse loops; #304 means CI runs on stacked PRs now.

Closes [#86](https://github.com/hugr-lab/mssql-extension/issues/86).

Three commits, each standing on its own measurement.

---

## W1 — the row count never needed `sys.partitions`

Every metadata query fetched `approx_rows`, `index_type` and `partition_count` from one aggregate over `sys.partitions`. That view reads `sysrowsets`, clustered on `rowsetid` — `object_id` is derived and unindexed, so a lookup by it is a full scan in **any** form. For **one** object's row count, on a 200,000-table catalog:

|---|---|---|---|

`OBJECTPROPERTYEX` agrees with `sys.partitions` **exactly** — 5000/5000, 150000/150000, 0 for empty, NULL for a VIEW, which maps to the same 0 the catalog already carried. It needs no statistics (unlike `sys.dm_db_stats_properties`, which returns NULL for a 5000-row heap with none — exactly a table CTAS or COPY has just made) and is not deprecated (unlike `sys.sysindexes`). Every row count in the database: **84 ms** against **2400 ms**.

What forced the two halves to share a subquery was `COUNT(*) AS partition_count` — a field `ParseTableShape` wrote and **nothing read**. Deleted; `is_partitioned` replaces it, which is what every comment already claimed it meant.

**Nothing is given up.** Counts stay, the planner's cardinality stays. Dropping the count was tried first and reverted — `scan_cardinality.test` catches it.

## W2 — one query for every schema

Listing **one** schema costs a full pass over `sys.objects` whatever the predicate says: metadata visibility is filtered per object in the *database*. An **empty** schema measured **484 ms** of server CPU at 200K objects. And there is no seek to buy — `sys.objects`, `sys.tables` and `INFORMATION_SCHEMA.TABLES` all plan as a **Clustered Index Scan on `sysschobjs.clst`**, with a literal `nsid` as much as with `SCHEMA_ID()`, because the nonclustered `(nsid, name)` index does not cover `type`/`nsclass`/`pclass`.

DuckDB drives that **once per schema**, so the shape was O(schemas × objects) — the reporter's 10,000 *empty* schemas extrapolate to hours. One query for all of them: **1917 ms**. Break-even is under five schemas.

What stood in the way was not size but the `ORDER BY`, and `BulkLoadAll` said so itself:

> *"ORDER BY … on millions of rows exceeds the memory grant and causes non-linear performance degradation"*

The sort existed only so a **streaming** parse could see table boundaries. Group on `object_id` in a hash map and the sort goes, and with it the loop. `object_id` rather than the name, because two schemas may hold tables of the same name.

Both callers move: the listing path and `mssql_preload_catalog()` with no schema named — preload was built the same per-schema way, which is why recommending it on #86 lowered the constant and never the exponent.

**Ordinary queries stay lazy.** `SELECT * FROM db.sch.tbl` goes to `GetTableMetadata` — one table, one seek — and never comes here. That split is why loading every schema on a *listing* costs nobody anything they were not already paying.

## Folded in from #311 — the Fabric flag never disabled BCP

Comment and log only. The source said *"Fabric endpoints don't support BCP/INSERT BULK, need fallback to INSERT"*. `is_fabric_endpoint` has exactly one consumer — `RequiresSingleByteText()`, because Fabric has no `nvarchar` — and `src/copy/` never tests it, so the fallback did not exist. The premise is also out of date: `bcp` is supported on Fabric as a preview feature, and `INSERT BULK` is what `bcp` speaks.

## Results at 200K tables / 1.2M columns

| | before | after |
|---|---|---|
| single-table metadata | 18.1 ms, `sysrowsets` 3200 reads | **~1 ms, 0 reads** |
| an **empty** schema's discovery | 2708 ms CPU | **484 ms** |
| every row count in the DB | 2400 ms | **84 ms** |
| 10,000 empty schemas (#86) | ~7.5 h | **~2 s** |

Also relevant to #308: `sysrowsets` was one of the two resources in **7 of 8** captured deadlock cycles. Removing it took index-DDL contention from 127/141/93 to **0/0/0** per 150 metadata queries.

## Verified

- **SQL suite: 177 passed, 8 skipped** against SQL Server 2025 — including `scan_cardinality.test`.
- Every rewritten template executed against a live server before building, to check the SQL independently of the C++.
- Row counts compared per object against `sys.partitions` on tables with rows, an empty table and a view.
- `OBJECTPROPERTYEX` platform support confirmed from the documentation for Azure Synapse and all three Fabric surfaces, and against Fabric's own unsupported list — see the comment below.

### DIFF
--- a/src/catalog/mssql_metadata_cache.cpp
+++ b/src/catalog/mssql_metadata_cache.cpp
@@ -1,4 +1,5 @@
 #include "catalog/mssql_metadata_cache.hpp"
+#include <algorithm>
 #include <chrono>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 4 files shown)


VGSML:
**The Fabric/Synapse question in the PR description is resolved** — from the documentation, not by assumption.

## `OBJECTPROPERTYEX` covers every target we support

Its *Applies to* list:

SQL Server · Azure SQL Database · Azure SQL Managed Instance · **Azure Synapse Analytics** · Analytics Platform System (PDW) · **SQL analytics endpoint in Microsoft Fabric** · **Warehouse in Microsoft Fabric** · **SQL database in Microsoft Fabric**

The page's own monikers confirm it, including the preview surface: `fabric`, `fabric-sqldb`, `azure-sqldw-latest`, `aps-pdw-2016`. The `Cardinality` property is marked *"Applies to: SQL Server 2012 (11.x) and later"* — an **engine-version** bound, not a platform one.

Checked the negative side too. Fabric's [T-SQL surface area](https://learn.microsoft.com/en-us/fabric/data-warehouse/tsql-surface-area) page enumerates what is **not** supported, and it does not list `OBJECTPROPERTYEX`, `OBJECTPROPERTY`, `sys.partitions`, `sys.indexes` or `sys.partition_schemes`.

**Still not verified by execution** on Fabric or Synapse — no access here — so the reason it is safe regardless is worth stating: `ISNULL(..., 0)` yields 0, which `MSSQLCatalogScanCardinality` already treats as *unknown* and answers with no estimate. That is the same path a VIEW takes today.

The docs actually supply a second and likelier route to that path: *"Returns `NULL` on error or if a caller does not have permission to view the object"*, with metadata-visibility restrictions applied to the result set. So a **low-privilege user** degrades to "no estimate" rather than to an error — which is the behaviour `sys.partitions` gave them too.

## One correction worth having in writing

`mssql_enable_statistics` is **not** a kill switch for this, despite the name:

- it is `SetScope::GLOBAL` — not per attached catalog;
- it is read only by `MSSQLCatalogScanCardinality` (`LoadStatisticsEnabled`), so it gates whether an estimate **reaches the planner**, not whether the metadata query asks for one. The SQL goes to the server either way.

If a per-catalog escape ever proves necessary, the seam that already exists is `MSSQLCatalog::IsFabricEndpoint()` (`src/catalog/mssql_catalog.cpp:927`) — the same one that already disables BCP for Fabric.

Recorded in spec 071 §4 (#309, commit `a08ceaf`), so the next person does not re-open the question.

Sources:
- [OBJECTPROPERTYEX (Transact-SQL)](https://learn.microsoft.com/en-us/sql/t-sql/functions/objectpropertyex-transact-sql)
- [T-SQL surface area in Fabric Data Warehouse](https://learn.microsoft.com/en-us/fabric/data-warehouse/tsql-surface-area)

VGSML:
Superseded by **#312** — same branch, same three commits, based on `main`.

This PR was auto-closed by GitHub when #308 merged and its branch `fix/metadata-deadlock-reset`, which was this PR's base, was deleted. A closed PR cannot have its base changed nor be reopened once the base branch is gone, so retargeting was not possible.

**Nothing was lost.** #308 landed as a merge commit rather than a squash, so `65efaaa` is a genuine ancestor of `main` and this branch's diff against `main` is exactly the three commits it always was — W1, the Fabric comment fix, W2. Verified by content: `grep` for `OBJECTPROPERTYEX`, `LoadAllSchemasMetadata` and `is_partitioned` in `origin/main` returns nothing, and for `MetadataResetCallback` returns one hit.

PR #311, opened 2026-09-08
docs(fabric): the flag never disabled BCP, and Fabric supports bcp anyway

VGSML:
**Comment and log-message only. No behaviour change** — the behaviour was already right, which is the point.

## What the source claimed

`src/mssql_storage.cpp`, at the ATTACH-time endpoint check:

```cpp
// Fabric endpoints don't support BCP/INSERT BULK, need fallback to INSERT
connection_info->is_fabric_endpoint = connection_info->IsFabricEndpoint();
if (connection_info->is_fabric_endpoint) {
    MSSQL_STORAGE_DEBUG_LOG(1, "Fabric endpoint detected: %s (BCP disabled, using INSERT fallback)", ...);
}
```

and the header field: `// True if targeting Microsoft Fabric (no BCP/INSERT BULK support)`.

## None of it is true

`is_fabric_endpoint` has **exactly one consumer** — `MSSQLCatalog::RequiresSingleByteText()`, which exists because Fabric Warehouse has no `nvarchar`. Through it the flag reaches the CTAS planner (`mssql_ctas_planner.cpp:54`), COPY's target types (`copy_function.cpp:269`) and the collation choice (`mssql_catalog.cpp:945,989`). That is the whole list.

Grep `src/copy/` for it and **there is no hit**. The bulk-load path has never tested it, so the "fallback to INSERT" the comment advertised does not exist and never did.

## And the premise is out of date

`bcp` **is** supported on Fabric, as a preview feature. From Microsoft's [T-SQL surface area in Fabric Data Warehouse](https://learn.microsoft.com/en-us/fabric/data-warehouse/tsql-surface-area), under Limitations:

> `BULK LOAD`, though `bcp` is supported as a preview feature.

`INSERT BULK` is the protocol `bcp` itself speaks — it is what this extension implements. And our own troubleshooting page already implied the load runs there: it records that a `#temp` table cannot be a BCP **target** on Fabric, which is a limit only a working load can reach.

## Why it is worth a commit

Same defect class as the `#224` documentation fix (#307): a documented behaviour that does not exist, wrong in the direction that costs most. A reader deciding whether to trust `COPY ... TO` against Fabric would have concluded from the source that it silently degrades to row-by-row INSERT, and either avoided it or gone looking for a switch that is not there.

Found while checking `OBJECTPROPERTYEX`'s platform support for spec 071 (#309).

## Verified

- SQL suite **177 passed, 8 skipped** — unchanged, as expected for a comment change.

### DIFF
--- a/src/include/mssql_storage.hpp
+++ b/src/include/mssql_storage.hpp
@@ -147,7 +147,12 @@ struct MSSQLConnectionInfo {
 	//===----------------------------------------------------------------------===//
 	// Endpoint Type Flags (T040-T041: cached at ATTACH time for performance)
 	//===----------------------------------------------------------------------===//
-	bool is_fabric_endpoint = false;  // True if targeting Microsoft Fabric (no BCP/INSERT BULK support)
+	// True if targeting Microsoft Fabric. Read ONLY by
+	// MSSQLCatalog::RequiresSingleByteText(): Fabric has no `nvarchar`, so CTAS,
+	// COPY and the collation choice pick single-byte text there. It does not gate
+	// the bulk-load path — `bcp`, whose INSERT BULK protocol this extension
+	// speaks, is supported on Fabric as a preview feature.
+	bool is_fabric_endpoint = false;
 
 	// Issue #225: did the server grant the LOGIN7 UTF8SUPPORT feature? Observed by
 	// the ATTACH-time validation login, which happens on every non-lazy attach, so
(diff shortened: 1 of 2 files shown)


VGSML:
Folded into #310 as commit `164d445` — spec 071's implementation touches the same file, and keeping this separate would have meant two PRs racing for the same lines. The change and its reasoning are unchanged.

PR #312, opened 2026-09-08
fix(catalog): spec 071 — what catalog metadata actually costs

### LINKED ISSUE #86: SHOW ALL TABLES fails after MSSQL attach

VGSML:
Replaces **#310**, which GitHub auto-closed when #308 merged and its branch — #310's base — was deleted. Same branch, same three commits, now based on `main`. Nothing was lost: #308 landed as a merge commit, so its work is on `main` and this branch's diff against `main` is exactly the three commits below.

Implements **spec 071** (#309) in full — W1 and W2 — plus the comment correction folded in from #311. Closes [#86](https://github.com/hugr-lab/mssql-extension/issues/86).

---

## W1 — the row count never needed `sys.partitions`

Every metadata query fetched `approx_rows`, `index_type` and `partition_count` from one aggregate over `sys.partitions`. That view reads `sysrowsets`, clustered on `rowsetid` — `object_id` is derived and unindexed, so a lookup by it is a full scan in **any** form. For **one** object's row count, on a 200,000-table catalog:

|---|---|---|---|

`OBJECTPROPERTYEX` agrees with `sys.partitions` **exactly** — 5000/5000, 150000/150000, 0 for empty, NULL for a VIEW, which maps to the same 0 the catalog already carried. It needs no statistics (unlike `sys.dm_db_stats_properties`, which returns NULL for a 5000-row heap with none — exactly a table CTAS or COPY has just made) and is not deprecated (unlike `sys.sysindexes`). Every row count in the database: **84 ms** against **2400 ms**.

What forced the two halves to share a subquery was `COUNT(*) AS partition_count` — a field `ParseTableShape` wrote and **nothing read**. Deleted; `is_partitioned` replaces it, which is what every comment already claimed it meant.

**Nothing is given up.** Counts stay, the planner's cardinality stays. Dropping the count was tried first and reverted — `scan_cardinality.test` catches it.

## W2 — one query for every schema

Listing **one** schema costs a full pass over `sys.objects` whatever the predicate says: metadata visibility is filtered per object in the *database*. An **empty** schema measured **484 ms** of server CPU at 200K objects. And there is no seek to buy — `sys.objects`, `sys.tables` and `INFORMATION_SCHEMA.TABLES` all plan as a **Clustered Index Scan on `sysschobjs.clst`**, with a literal `nsid` as much as with `SCHEMA_ID()`, because the nonclustered `(nsid, name)` index does not cover `type`/`nsclass`/`pclass`.

DuckDB drives that **once per schema**, so the shape was O(schemas × objects) — the reporter's 10,000 *empty* schemas extrapolate to hours. One query for all of them: **1917 ms**. Break-even is under five schemas.

What stood in the way was not size but the `ORDER BY`, and `BulkLoadAll` said so itself:

> *"ORDER BY … on millions of rows exceeds the memory grant and causes non-linear performance degradation"*

The sort existed only so a **streaming** parse could see table boundaries. Group on `object_id` in a hash map and the sort goes, and with it the loop. `object_id` rather than the name, because two schemas may hold tables of the same name.

Both callers move: the listing path and `mssql_preload_catalog()` with no schema named — preload was built the same per-schema way, which is why recommending it on #86 lowered the constant and never the exponent.

**Ordinary queries stay lazy.** `SELECT * FROM db.sch.tbl` goes to `GetTableMetadata` — one table, one seek — and never comes here.

## Folded in from #311 — the Fabric flag never disabled BCP

Comment and log only. The source said *"Fabric endpoints don't support BCP/INSERT BULK, need fallback to INSERT"*. `is_fabric_endpoint` has exactly one consumer — `RequiresSingleByteText()`, because Fabric has no `nvarchar` — and `src/copy/` never tests it, so the fallback did not exist. The premise is also out of date: `bcp` is supported on Fabric as a preview feature, and `INSERT BULK` is what `bcp` speaks.

## Results at 200K tables / 1.2M columns

| | before | after |
|---|---|---|
| single-table metadata | 18.1 ms, `sysrowsets` 3200 reads | **~1 ms, 0 reads** |
| an **empty** schema's discovery | 2708 ms CPU | **484 ms** |
| every row count in the DB | 2400 ms | **84 ms** |
| 10,000 empty schemas (#86) | ~7.5 h | **~2 s** |

Also relevant to #308: `sysrowsets` was one of the two resources in **7 of 8** captured deadlock cycles. Removing it took index-DDL contention from 127/141/93 to **0/0/0** per 150 metadata queries.

## Verified

- **SQL suite: 177 passed, 8 skipped** against SQL Server 2025 — including `scan_cardinality.test`.
- Every rewritten template executed against a live server before building, to check the SQL independently of the C++.
- Row counts compared per object against `sys.partitions` on tables with rows, an empty table and a view.
- `OBJECTPROPERTYEX` platform support confirmed from the documentation for Azure Synapse and all three Fabric surfaces, and against Fabric's own unsupported list.

### DIFF
--- a/specs/071-catalog-metadata-cost/spec.md
+++ b/specs/071-catalog-metadata-cost/spec.md
@@ -1,0 +1,268 @@
+# Spec 071 — What catalog metadata actually costs
+
+**Status**: PROPOSED.
+**Closes**: [#86](https://github.com/hugr-lab/mssql-extension/issues/86) (`SHOW ALL TABLES` unusable on a database with many schemas).
+**Relates to**: PR #308 (the 1205 deadlock-victim retry — the reset callback that
+makes losing a deadlock recoverable). This spec attacks the same queries from the
+other side: W1 removes the resource that 7 of 8 captured deadlock cycles were
+fighting over, so the retry has far less to absorb.
+
+Four defects in the metadata layer, found by measuring rather than reading. They
+are independent, they land in ascending order of size, and three of them are the
+same root cause seen from different paths: **the `sys.partitions` aggregate that
+every catalog query carries, for a column nothing reads.**
+
+---
+
+## 0. How everything below was measured
+
+Reproduce before trusting. SQL Server **2025** (17.0.4075.5) in Docker — the 2022
+image crash-loops on Apple Silicon, see `docs/TESTING.md` — with a synthetic
+catalog in `TestDB`:
+
+```sql
+-- 200,000 tables x 6 columns, one schema `big`, plus an empty schema `empty1`
+CREATE TABLE big.tN (id INT, c1 VARCHAR(50), c2 INT, c3 DECIMAL(18,4),
+                     c4 DATETIME2(3), c5 NVARCHAR(100));
+```
+
+Resulting size: **200,000 tables / 1,200,000 columns / 200,230 rows in
+`sys.partitions`**. Creation ran at ~1 ms per table and did **not** degrade with
+catalog size (0.8 ms/table at 2K, 0.97 ms/table at 97K) — the write side of the
+catalog scales fine, which is why everything below is about reads.
+
+- **Wall clock**: `sqlcmd -i q.sql -o /dev/null` with `GO N` inside one session,
+  so connection setup is paid once. A `SELECT 1` script measured the same way is
+  subtracted as the baseline (~20.7 ms/rep here — it is `docker exec` plus
+  sqlcmd's own formatting, and it is large relative to the signal, which is why
+  the per-invocation form of this measurement is useless and was discarded).
+- **Server cost**: `SET STATISTICS TIME ON` (CPU ms) and `SET STATISTICS IO ON`
+  (logical reads per base table). These are the numbers that carry the argument;
+  wall clock is only corroboration.
+- Variants were run **interleaved** and repeated, never one-after-another —
+  three passes, and any effect that did not survive all three is reported as
+  noise below rather than as a result.
+
+---
+
+## 1. The findings
+
+### F1 — `sys.partitions` cannot be seeked, in any form
+
+All three metadata queries fetch `approx_rows` / `index_type` /
+`partition_count` from one uncorrelated subquery:
+
+```sql
+LEFT JOIN (SELECT p.object_id, SUM(p.[rows]) AS [rows],
+                  MAX(ISNULL(i.type, 0)) AS index_type, COUNT(*) AS partition_count
+           FROM sys.partitions p
+           LEFT JOIN sys.indexes i ON i.object_id = p.object_id AND i.index_id = p.index_id
+           WHERE p.index_id IN (0, 1)
+           GROUP BY p.object_id) p ON p.object_id = o.object_id
+```
+
+`sys.partitions` reads `sys.sysrowsets`, which is clustered on `rowsetid`;
+`object_id` is derived and has no index. Filtering by it changes nothing:
+
+| lookup by `object_id` | base table | logical reads |
+|---|---|---|
+| `sys.partitions` | `sysrowsets` | **3200** (2 scans) |
+| `sys.dm_db_partition_stats` | `sysrowsets` | **1600** (1 scan) |
+| `sys.indexes` | `sysidxstats` | **6** (seek) |
+
+So the row count is O(catalog) however it is asked for, and the index *shape* is
+O(1). They do not belong in the same subquery.
+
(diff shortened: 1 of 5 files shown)


VGSML:
**Rebased onto the new `main` (duckdb `v2.0-cyanoptera`), and spec 071 is now part of this PR rather than a separate one.**

## The spec moved here from #309

Keeping them apart was a mistake, and precisely the one I have been arguing about on #307: a document and the code it describes must land together, or the document can merge first and describe behaviour `main` does not have. #309 was `MERGEABLE` the whole time — it could have gone in alone, leaving `specs/071-catalog-metadata-cost/spec.md` describing an `OBJECTPROPERTYEX` source that no code used.

That risk was not hypothetical for this spec in particular, because it was **rewritten twice under measurement** — first dropping two of its own workstreams, then replacing the entire row-count source. By the time the implementation existed, the spec described it rather than proposing it.

Six commits now: three implementation, three spec.

## Re-verified against the pin that is actually in `main`

The earlier "177 passed" was measured against duckdb `d7a4366603`, which #306 has since replaced. That build no longer exists, so the claim needed redoing rather than repeating — #306 made `TableCatalogEntry::GetColumns()` pure virtual, and while its fix (`MSSQLTableEntry` owning its own `ColumnList`) is orthogonal to the metadata-cache changes here, "orthogonal" is a prediction until it is run.

Against `main` at `656666f`, duckdb `070ce1f695`:

- build clean, no warnings
- **SQL suite: 177 passed, 8 skipped** — including `scan_cardinality.test`, the test that caught and reverted the first attempt at this change

No conflicts with #306: it touches `mssql_table_entry.{cpp,hpp}` and `CMakeLists.txt`, this touches `mssql_metadata_cache.{cpp,hpp}`, `mssql_storage.{cpp,hpp}` and `specs/`.

VGSML:
@oluies — green now (all checks pass, including `Integration Test (linux_amd64)`), so this is ready when you have time.

706 lines is a lot to read straight through, so here is where I would spend the attention. **Three places where I think the risk is real:**

1. **The reset in `LoadAllSchemasMetadata`.** It rolls back `tables` only for the schemas *this pass touched*, not all of them — the reasoning being that a schema the query never reached legitimately still holds what it had. That interacts with your 1205 retry from #308, and if I got it wrong the symptom only appears under a deadlock, which is exactly where tests do not look.

2. **A schema the query returned no rows for is marked `LOADED` and empty.** Otherwise the next `Scan` asks again forever. But it means "filtered out by `schema_filter`" and "genuinely empty" land in the same state, so the filter has to be applied before that point. It is — worth a second pair of eyes on the ordering.

3. **Sorting each table's columns by `column_id` after the hash grouping.** Rows now arrive unordered and every consumer indexes columns by position. Forget that sort and you get silently transposed column types — and on a small test table the arrival order can coincide with the right one, so no existing test would catch it.

**What you can skip:**

- `OBJECTPROPERTYEX` values — compared per object against `sys.partitions` on a table with rows, an empty table and a view; they agree exactly, including NULL→0 for the view, which is the value the catalog already carried.
- Platform support — checked against the documentation (Azure Synapse plus all three Fabric surfaces) *and* against Fabric's own list of what is unsupported. The failure mode if it is ever wrong is benign: `ISNULL(…, 0)` lands on the "unknown" path `MSSQLCatalogScanCardinality` already takes for a VIEW.

The spec (`specs/071-catalog-metadata-cost/spec.md`) is in this PR rather than its own, deliberately — §5 records nine dead ends with their numbers, including two of my own proposals that measurement killed and two of my own figures that did not survive repetition.

oluies (approved):

Issue #239, opened 2026-08-04
Materialise a scan of our own catalog before the sink, instead of documenting it as a limitation


VGSML:
## What we document as a limitation, `duckdb-postgres` solves

`README.md` tells users that reading and writing the same attached catalog inside
one explicit transaction does not work: the transaction's connection is pinned,
and a pinned connection cannot stream a result set and receive a bulk load at the
same time.

`duckdb-postgres` hits the identical constraint — one connection per transaction,
no parallel sink — and removes it in the planner instead of documenting it. In
`PostgresCatalog::PlanCreateTableAs`:

```cpp
auto &inner_plan = AddCastToPostgresTypes(context, planner, plan);
MaterializePostgresScans(inner_plan);
```

plus, on the scan side, `bind_data.max_threads = 1; bind_data.can_use_main_thread = true;`.

The scan of their own catalog is **materialised before the sink starts**, so the
single connection is never streaming and loading at once.

## Why this is worth doing

- It is the same shape of fix for the same shape of problem, from the reference
  implementation for this class of extension.
- It converts a documented limitation into working SQL. The workaround we suggest
  today is for the user to restructure the statement.
- It is planner work, isolated from the write path.

## Why it is not part of spec 063

063 merges the COPY and CTAS bulk-load machinery — already the riskiest refactor
on the write path. Adding planner changes on top would make it unreviewable: a
reviewer could not separate what moved from what changed.

## What to investigate

1. Where DuckDB lets an extension rewrite its own scans in the plan for a sink —
   whether the `PlanCreateTableAs` / `PlanInsert` hooks give us the same reach.
2. Whether the constraint is really the pinned connection alone, or also the
   catalog-level locks the bulk load takes on the target.
3. What it costs: materialising means the full source is buffered before a single
   row is sent, so the pipelining we get today disappears for that statement.
   Postgres pays this unconditionally; we would only pay it when a scan of the
   same catalog is in the plan.
4. Whether the same rewrite covers COPY, not just CTAS.

Reference: `README.md` — "Reading and writing the same catalog in one transaction";
spec 057 (#234) — where the pinned-connection conflict was root-caused.

VGSML:
Fixed in spec 075 (https://github.com/hugr-lab/mssql-extension/pull/341) — no longer a documented limitation.

[...]

Issue #129, opened 2026-05-28
Support SQL Server as a DuckLake metadata backend


sashapodgoreanu:
## Request

I would like to use this DuckDB `mssql` extension together with DuckLake, using SQL Server as the DuckLake metadata backend.

I understand that SQL Server is not currently listed as an officially supported DuckLake metadata backend. Still, if this extension could work reliably in that scenario, it could make the extension more useful for DuckLake users who already run SQL Server infrastructure.

This issue is a compatibility/roadmap question for the `mssql` extension, with a reproducible example of the current failure.

## Environment

- OS: Windows
- DuckDB: `1.5.2`
- DuckLake: `1.0`
- DuckDB extensions used: `ducklake`, `httpfs`, `mssql`
- Metadata backend tested: SQL Server
- Object storage tested: MinIO/S3

## Attached Repro

I am attaching this files for repro:

[01-init.sql](https://github.com/user-attachments/files/28339322/01-init.sql)
[docker-compose.yml](https://github.com/user-attachments/files/28339304/docker-compose.yml)
[duckdb_repro.sql](https://github.com/user-attachments/files/28339354/duckdb_repro.sql)

The repro starts SQL Server on `127.0.0.1:32813` and MinIO on `127.0.0.1:32814`.

It creates:

- SQL Server database: `ducklake_metastore`
- SQL Server schema: `main`
- MinIO bucket: `ducklake-metastore`

## Steps To Reproduce

Start the attached containers:

```powershell
docker compose up -d
```

Then run this DuckDB script:

```sql
LOAD ducklake;
LOAD httpfs;
LOAD mssql;

CREATE OR REPLACE SECRET minio_s3 (
    TYPE s3,
    KEY_ID 'minioadmin',
    SECRET 'minioadmin',
    REGION 'us-east-1',
    ENDPOINT '127.0.0.1:32814',
    URL_STYLE 'path',
    USE_SSL false
);

CREATE OR REPLACE SECRET mssql_backend (
    TYPE mssql,
    host '127.0.0.1',
    port 32813,
    database 'ducklake_metastore',
    user 'sa',
    password 'YourStrong!Passw0rd',
    use_encrypt false
);

CREATE OR REPLACE SECRET ducklake_mssql (
    TYPE ducklake,
    METADATA_PATH '',
    DATA_PATH 's3://ducklake-metastore/sql-server-metastore/ducklake/',
    METADATA_SCHEMA 'main',
[...]

PR #313, opened 2026-09-08
release: v0.2.5 — the catalog's default schema, and two scans on one pinned connection

VGSML:
Targets **`duckdb-v1.5.5`** (the released v1.5.5 / v0.2.x maintenance line), not `main` — `main` has moved to the duckdb 2.0 pre-release, and the consumer that needs this ships on released duckdb.

Closes [#129](https://github.com/hugr-lab/mssql-extension/issues/129) and [#239](https://github.com/hugr-lab/mssql-extension/issues/239).

Both fixes are things a **generic catalog consumer** hits immediately. Found while building a DuckLake catalog manager over an attached MSSQL database, but neither is specific to that: they are about the extension behaving like a catalog rather than like a query function.

---

## R1 — the default schema is `dbo`, not `main` (#129)

`MSSQLCatalog` did not override `Catalog::GetDefaultSchema()`, so the base answer `main` came back — a schema no SQL Server database has. Anything resolving an unqualified name through the catalog's default failed:

```
Schema 'main' not found in MSSQL database
```

#129 shows a user working around it by CREATEing a schema literally called `main` on the server.

A constant, not a per-login `SCHEMA_NAME()` lookup: that would put a round trip on ATTACH and needs an answer on the lazy-validation path, where no connection exists yet. A login with a different default still addresses its tables by qualified name; only the unqualified default is wrong for it, which is today's behaviour minus the failure.

## R2/R3 — two scans of one catalog on the pinned connection (#239)

A transaction pins **one** TDS connection per catalog and routes every read there. A scan holds that connection from `InitGlobal` until its last row is drained, and DuckDB does not promise to drain one source before initializing the next.

In an explicit transaction, a plan with more than one scan of the same catalog now drains each into a buffer-managed `ColumnDataCollection` at `InitGlobal` and releases the connection before returning. **Autocommit is untouched** — there each scan takes its own pooled connection, which is why the same query always worked outside a transaction. The mechanism mirrors duckdb-postgres, which has the same one-connection constraint, and matches spec 066 D1.

### Two findings that changed the plan, both from running it

**1. The reproducing shape is not the obvious one.** A correlated subquery in the SELECT list —

```sql
SELECT p.id, (SELECT list(c.v) FROM child c WHERE c.pid = p.id) FROM parent p
```

— does **not** reproduce it. DuckDB decorrelates that into a hash join whose sources are drained in sequence, and it passes on the broken build. What reproduces it is a correlated subquery in a **predicate**, which plans as `LEFT_DELIM_JOIN` + `DELIM_SCAN` and initializes both sides before draining either:

```sql
SELECT p.id FROM parent p WHERE p.id > (SELECT count(*) FROM child c WHERE c.pid = p.id)
```

The regression test uses that shape and was verified to fail on the pre-fix binary with the exact documented errors. A test built on the first shape would have passed on a broken build.

**2. Materialization alone fixes `threads = 1` and *not* `threads > 1`.** There the two `InitGlobal`s run **concurrently**, so the drain has not happened when the second batch is sent, and the failure is a torn stream rather than `not in Idle state`. The race is between the initializations, not inside them. So a materialized scan also holds a per-catalog mutex across batch-and-drain. Verified at threads 1, 4 and 8.

| | before | after |
|---|---|---|
| `threads = 1` | `Cannot execute: connection not in Idle state (current: Executing)` | 4997 rows |
| `threads = 4` | `Connection closed while waiting for COLMETADATA` | 4997 rows |
| `threads = 8` | (same) | 4997 rows |
| autocommit | passed | passed, and takes no new path |

## Verified

- **Suite: 166 test cases, 5225 assertions, 0 failures** against SQL Server 2025 (164/5182 before; the two new files are the delta).
- **Both new tests confirmed to fail on the pre-fix binary** — `not in Idle state` for #239, a failed `current_schema()` assertion for #129. A regression test that passes before the fix is decoration.
- `clangd --check` on every touched file: 0 compile diagnostics.

## Note for the forward-port to `main`

duckdb 2.0 changed the signature:

```cpp
// 1.5.5   virtual string GetDefaultSchema() const;
// 2.0     virtual optional<Identifier> GetDefaultSchema() const;
```

R1 as written here **will not compile** on `main`, and the port is not a cast: 2.0 gives `nullopt` the meaning *"this catalog has no default schema, do not probe an implicit one"* and documents an empty `Identifier` as *"unspecified"*, so the port must deliberately return `Identifier("dbo")`.

R2/R3 should port more cleanly, but `LogicalGet`, `TableFunctionInitInput` and `MaxThreads()` all need re-checking on 2.0 — `TableFunctionInitInput::projected_types()` does not exist on 1.5.5 and the types are derived by hand here, which may be avoidable there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.5] - 2026-09-08
+
+Two fixes that a generic catalog consumer hits immediately. Found while building
+a DuckLake catalog manager on top of an attached MSSQL database, but neither is
+specific to that: both are about the extension behaving like a catalog rather
+than like a query function.
+
+### Fixed
+
+- **The catalog's default schema is `dbo`, not DuckDB's `main`**
+  ([#129](https://github.com/hugr-lab/mssql-extension/issues/129)).
+  `MSSQLCatalog` did not override `Catalog::GetDefaultSchema()`, so the base
+  answer `main` came back — a schema no SQL Server database has. Anything that
+  resolves an unqualified name through the catalog's default failed with
+  `Schema 'main' not found in MSSQL database`; the workaround in the wild was to
+  CREATE a schema literally called `main` on the server. `SELECT
+  current_schema()` now answers `dbo` after `USE <catalog>`, and unqualified
+  names resolve.
+
+- **Two scans of one catalog inside an explicit transaction no longer break the
+  pinned connection**
+  ([#239](https://github.com/hugr-lab/mssql-extension/issues/239)).
+  A transaction pins ONE TDS connection per catalog and routes every read on
+  that catalog to it, while a scan holds that connection from init until its
+  last row is drained. DuckDB does not promise to drain one source before
+  initializing the next — a correlated subquery in a predicate plans as
+  `LEFT_DELIM_JOIN` and initializes both — so the second scan met a busy
+  connection:
+
+  - `threads = 1` → `Cannot execute: connection not in Idle state (current: Executing)`
+  - `threads > 1` → `Connection closed while waiting for COLMETADATA`, a torn stream
+
+  In an explicit transaction, a plan holding more than one scan of the same
+  catalog now drains each of them into a buffer-managed collection at init and
+  releases the connection before returning, and holds a per-catalog lock across
+  batch-and-drain so concurrent initializations serialize. **Autocommit is
+  unchanged**: there every scan takes its own pooled connection, which is why
+  the same query always worked outside a transaction.
+
+  The mechanism mirrors duckdb-postgres, which has the same one-connection
+  constraint. Memory is bounded by the buffer manager, and only the plans that
+  would otherwise fail are affected.
+
 ## [0.2.4] - 2026-08-17
 
 ### Fixed
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ set(EXTENSION_NAME ${TARGET_NAME}_extension)
 set(LOADABLE_EXTENSION_NAME ${TARGET_NAME}_loadable_extension)
 
 # Extension version - update this when creating a new release tag
-set(MSSQL_EXTENSION_VERSION "0.2.4")
+set(MSSQL_EXTENSION_VERSION "0.2.5")
 
 # Project configuration
 project(${TARGET_NAME})
--- a/src/catalog/mssql_catalog.cpp
+++ b/src/catalog/mssql_catalog.cpp
@@ -292,6 +292,11 @@ string MSSQLCatalog::GetCatalogType() {
 	return "mssql";
 }
 
+string MSSQLCatalog::GetDefaultSchema() const {
+	// See the header for why this is a constant and not SCHEMA_NAME().
+	return "dbo";
+}
+
 //===----------------------------------------------------------------------===//
 // Schema Operations
 //===----------------------------------------------------------------------===//
--- a/src/include/catalog/mssql_catalog.hpp
+++ b/src/include/catalog/mssql_catalog.hpp
@@ -70,6 +70,43 @@ public:
 
 	string GetCatalogType() override;
 
+	//! SQL Server's default schema, not DuckDB's `main`.
+	//!
+	//! The base Catalog answers `main`, which no SQL Server database has, so
+	//! anything that resolves an unqualified name through the catalog's default
+	//! fails with "Schema 'main' not found in MSSQL database". ducklake hits this
+	//! on ATTACH — DuckLakeTransaction::GetDefaultSchemaName() asks for it before
+	//! any table exists — and issue #129 shows a user working around it by
+	//! CREATEing a schema literally called `main` on the server.
+	//!
+	//! `dbo` is the constant answer rather than a per-login lookup: it is the
+	//! default for every login that has not been given another, and resolving
+	//! SCHEMA_NAME() would put a round trip on ATTACH and need an answer on the
+	//! lazy-validation path, where no connection has been made yet. A login whose
+	//! default schema is not `dbo` still addresses its tables by qualified name;
+	//! only the unqualified default is wrong for it, which is the pre-existing
+	//! behaviour minus the crash.
+	string GetDefaultSchema() const override;
+
+	//! Serializes materialized catalog scans against each other (issue #239).
+	//!
+	//! R2 drains a scan inside InitGlobal so the pinned connection is Idle again
+	//! before the next scan starts — which is enough at threads = 1, where DuckDB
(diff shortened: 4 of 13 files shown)


oluies (approved):

oluies (commented):
Looked hard at the two things you flagged. The shape of the fix is right — materialize to release, lock to order, autocommit untouched — and the two findings in the description are the valuable part of the PR: that the correlated-subquery-in-SELECT shape decorrelates into a hash join and *passes* on the broken build, and that materialization alone fixes `threads = 1` and not `threads > 1` because the race is between the initializations. Both are the kind of thing only running it tells you.

Two things on the gate, one of which I think is a real hole, and a view on the mutex.

## The optimizer gate misses `mssql_scan()`

`CollectCatalogScans` filters on `get.function.name == "mssql_catalog_scan"`. But `mssql_scan()` is a separate table function (`mssql_functions.cpp:472`) whose bind data also carries a `context_name`, and it acquires its connection through the same `ConnectionProvider::GetConnection` — which, in an explicit transaction, returns **the same pinned connection**:

```cpp
// mssql_connection_provider.cpp:139-146
// In an explicit DuckDB transaction (BEGIN was issued) - use pinned connection
auto pinned = txn->GetPinnedConnection();
if (pinned) { return pinned; }
```

So these shapes still hit #239, and the gate does not see them:

```sql
BEGIN;
-- one catalog scan + one raw scan: count is 1, no materialization
SELECT p.id FROM db.dbo.parent p
 WHERE p.id > (SELECT count(*) FROM mssql_scan('db','SELECT pid FROM dbo.child') c WHERE c.pid = p.id);

-- two raw scans: count is 0
SELECT * FROM mssql_scan('db','...') a JOIN mssql_scan('db','...') b USING (id);
```

The `< 2` test is per *counted* scan, not per *consumer of the pinned connection*, and those two sets differ. Whether that is in scope for a release cut is your call — the catalog path is what DuckLake exercises — but it is worth being explicit in the CHANGELOG that the fix covers catalog scans rather than "two scans of one catalog", because the issue title reads as the latter.

## `IsAutoCommit()` is read at optimize time, used at execute time

Worth confirming rather than assuming: if a prepared statement is planned in autocommit and later executed inside a transaction, does 1.5.5 re-run the optimizer extension? If the physical plan is cached, the gate was evaluated under the wrong transaction state and the flag is stale in the unsafe direction — `false` when it needed to be `true`. `PREPARE p AS SELECT ...; BEGIN; EXECUTE p; EXECUTE p;` would be the probe. If it does re-plan, this is a non-issue and worth a sentence in the comment saying so.

## The mutex: right mechanism, one level too coarse

Short answer: it works and it is safe, but the catalog is not the object that owns the contended resource.

The resource is **one pinned TDS connection per transaction**, and `MSSQLTransaction` owns it (`GetPinnedConnection()`). Two independent DuckDB connections in concurrent transactions against the same attached catalog have **two different pinned connections** and cannot conflict — but they share `MSSQLCatalog`, so they serialize on its mutex anyway. That is correctness-preserving and costs concurrency that nothing requires. Scoping to `MSSQLTransaction` matches the resource exactly.

That relocation also subsumes the first finding. A lock on the transaction (or on the pinned connection itself) is held by *whatever* uses that connection, so `mssql_scan()`, and DML on the same pinned connection, get ordered without the optimizer having to enumerate them. The gate would still be needed to decide *materialization* — draining is what makes the connection Idle again, and no lock achieves that — but ordering would stop depending on a plan-shape census that has to stay exhaustive.

One smaller note: the lock is held across the batch, the full network drain and `ColumnDataCollection::Append`, which can spill through the buffer manager. That is a long critical section holding a mutex across unbounded I/O. Fine as written, since there is exactly one mutex and no ordering to get wrong, but it is worth a comment saying so — the next person to add a second lock anywhere near this path needs to know.

## Things I checked and liked

- `static_cast` with the job-1124 reference rather than `dynamic_cast`, and the reason stated (the function-name test is the discriminator). That is the right call and the right place to justify it.
- The drain reusing `FillChunk` rather than a second row reader, so a materialized scan and a streamed one cannot diverge in projection, rowid or warnings. That is the sort of duplication that would have rotted.
- Marking **all** scans rather than all-but-the-first, with the reason (which source DuckDB initializes first is scheduling, not a plan property).
- `GetDefaultSchema()` as a constant with the tradeoff stated, and the forward-port note about `optional<Identifier>` meaning "do not probe" rather than "unspecified" — that is exactly the sort of thing that would be lost otherwise.

VGSML:
Ran all three. One is a real hole, one is a non-issue, and the third changes what the right fix looks like.

## 1. The `mssql_scan()` gap — confirmed, both shapes

Reproduced exactly as you wrote them, on the merged code:

```
A: catalog scan + mssql_scan  →  IO Error: ... connection not in Idle state (current: Executing)
B: two mssql_scan             →  IO Error: ... connection not in Idle state (current: Executing)
```

You are right that the `< 2` test counts the wrong set. It counts scans the optimizer can *see* as catalog scans, and the set that matters is *consumers of the pinned connection*, which also includes `mssql_scan()` and DML.

## 2. `IsAutoCommit()` at optimize time — not a problem, and now verified

```sql
PREPARE p AS SELECT ... ;   -- planned in autocommit
BEGIN; EXECUTE p; EXECUTE p; COMMIT;
```

Both executions return 4997. duckdb re-plans, so the flag is never evaluated under one transaction state and used under another. Worth the sentence you suggested; I will add it rather than leave the next reader to re-derive it.

## 3. The mutex — you are right that the catalog is the wrong owner, and there is a trap in the fix

Agreed on the scoping: the resource is one pinned connection per **transaction**, and two DuckDB connections in concurrent transactions against the same catalog have two of them and serialize on my mutex for nothing.

But moving the lock alone does **not** subsume finding 1, and the way it fails is worse than today. A lock only helps a holder that *releases within `InitGlobal`* — which is exactly what materialization does. If `mssql_scan()` took a transaction-scoped lock without being materialized, it would take the lock, stream lazily, and hold it until drained; the second scan would block in `InitGlobal`, so the pipeline could never drain the first. **That is a hang instead of an error message.**

So the complete fix is not "relocate the mutex" but "gate and materialize every consumer of the pinned connection, then scope the lock to the transaction" — and materializing `mssql_scan()` means giving it the same drain, since it has its own init/execute path.

## What I propose for the release

Ship v0.2.5 as it stands and be **honest in the CHANGELOG about what it covers** — your point that "two scans of one catalog" reads as the general claim is correct, and the general claim is not what shipped. It covers catalog scans, which is the whole of what DuckLake exercises, and that is the release this was cut for.

The three-part follow-up — `mssql_scan()` in the gate, its own materialization, and the lock moved to `MSSQLTransaction` — lands as its own change on both lines, with the hang above written down so nobody does half of it.

Filing it now with your two repro shapes and that ordering constraint, and amending the CHANGELOG line before I tag. Thanks — the `mssql_scan()` one I would not have found, because I only ever exercised the path DuckLake uses.

PR #314, opened 2026-09-08
fix(catalog): the v0.2.5 fixes on the duckdb 2.0 line, plus mssql_scan on the pinned connection (#129, #239, #316)

### LINKED ISSUE #129: Support SQL Server as a DuckLake metadata backend

VGSML:
Started as the forward-port of **#313** (v0.2.5, on `duckdb-v1.5.5`) to the 2.0 line. It has since picked up a fix that exists only here, so it is no longer a pure port: no version bump, but it does carry a `[Unreleased]` CHANGELOG entry now.

Closes [#129](https://github.com/hugr-lab/mssql-extension/issues/129), [#239](https://github.com/hugr-lab/mssql-extension/issues/239) and [#316](https://github.com/hugr-lab/mssql-extension/issues/316) on `main`. Read #313 for the reasoning behind the first two; this describes what 2.0 made different, and #316 in full.

## What the port is, and what it is not

**`GetDefaultSchema` changed its signature *and its meaning*:**

```cpp
// 1.5.5   virtual string GetDefaultSchema() const;
// 2.0     virtual optional<Identifier> GetDefaultSchema() const;
```

2.0 gives the three possible answers distinct meanings, documented in `catalog.hpp`:

| answer | means |
|---|---|
| a value | probe this schema for unqualified names |
| `nullopt` | this catalog **has** no default schema — never probe an implicit one |
| empty `Identifier` | "unspecified" elsewhere in the catalog |

We mean the first, so this returns `Identifier("dbo")` and deliberately neither empty form. The header now says that, because someone reading `return Identifier("dbo")` on a 2.0 header cannot otherwise tell that `nullopt` was considered and rejected — and getting it wrong would silently stop unqualified names resolving at all, which is the very bug being fixed.

**`Catalog::GetCatalog` takes an `Identifier` on 2.0**, so the new call site wraps at the API boundary and keeps `std::string` inside — the rule CLAUDE.md states and the rest of the file already follows.

Everything else — the optimizer gate, the materialization drain, the per-catalog mutex — ports unchanged.

## #316 — `mssql_scan()` was never reachable by the #239 gate

The v0.2.5 fix counts scans on the optimized plan, and the optimizer knows one function name: `mssql_catalog_scan`. @oluies flagged that `mssql_scan()` takes the same pinned connection and is not covered. It turns out it cannot be covered there at all.

`MSSQLScanBind` **executes the query during binding** — that is where the result's column types come from, out of the batch's own COLMETADATA — and hands the live stream forward, holding the connection until execution drains it. So inside a transaction the second scan of that catalog fails in its own `Bind`, before any `InitGlobal` exists to materialize and long before the optimizer runs. Both shapes reproduced against a live server on this build:

```
catalog scan + mssql_scan   IO Error: Connection closed unexpectedly
two mssql_scan              IO Error: Cannot execute: connection not in Idle state (current: Executing)
```

Bind now drains into a collection and closes the stream whenever the session is not in autocommit. Closing is the operative half — draining alone leaves the connection `Executing`.

**The trigger is "in a transaction", not "more than one scan."** Bind cannot see the rest of the plan; by the time a count would be available the connection is already held. So a lone `mssql_scan()` in a transaction is buffered too. The defence for that is narrow and worth stating rather than glossing: inside a transaction such a scan could not have coexisted with anything else on that catalog anyway, so this turns a failing case into a slower one rather than taking a working case away.

The collection is a `shared_ptr` on the bind data, not a `unique_ptr` on the global state: `FunctionData::Copy` has to hand out the same rows, and one bind data can initialize more than one global state.

## Also carries #318

@oluies found that the materialized collection could not spill — built from `Allocator::Get(context)`, which is duckdb's `IN_MEMORY_ALLOCATOR` overload, unaccounted against the buffer manager, while the docs promised the opposite. Confirmed unchanged on the `070ce1f6` pin, so it is not a 1.5.5 artefact. Both call sites here use the `ClientContext` overload.

His three doc findings are applied too: the page no longer claims *"only queries that would otherwise fail are affected"* (the gate flags every scan of a catalog once two appear, so a plain hash join over two of its tables buffers both sides — deliberate, but not what was written), it now says what `mssql_scan()` does in a transaction, and the example alias is `sqlserver` like the rest of the page.

## Verified

- **Full SQL suite: 159 cases, 4925 assertions, 0 failures** against a live SQL Server on the `v2.0-cyanoptera` pin.
- The #239 regressions pass here as on 1.5.5, including the `threads = 4` case that needs the mutex rather than the drain.
- `transaction_mssql_scan_materialization.test` is new: both of the shapes above, rows-not-just-counts, read-your-writes through a drained scan, and autocommit still streaming. Verified **failing** on the pre-fix binary before it was written the other way round.

## Merge order

No longer coupled: #313 is merged and `v0.2.5` has been retagged over #318 and #319, so this branch has nothing left to wait for.

## Not in this PR

Scoping the materialize lock to `MSSQLTransaction` rather than the catalog (@oluies' third point). Agreed, but it changes locking behaviour and wants its own test for two transactions on one catalog not blocking each other.

### DIFF
--- a/src/catalog/mssql_catalog.cpp
+++ b/src/catalog/mssql_catalog.cpp
@@ -292,6 +292,12 @@ string MSSQLCatalog::GetCatalogType() {
 	return "mssql";
 }
 
+optional<Identifier> MSSQLCatalog::GetDefaultSchema() const {
+	// See the header: a value (not nullopt, not an empty Identifier) is what says
+	// "probe dbo for unqualified names".
+	return Identifier("dbo");
+}
+
 //===----------------------------------------------------------------------===//
 // Schema Operations
 //===----------------------------------------------------------------------===//
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Forward-port of the v0.2.5 fixes to the duckdb 2.0 line** (#314). The
+  released v0.2.5 sits on the `duckdb-v1.5.5` maintenance branch; these are the
+  same two fixes against duckdb `main`.
+  - **An MSSQL catalog's default schema is `dbo`** (#129). `Catalog::GetDefaultSchema()`
+    answered `main` from the base class, which no SQL Server database has, so
+    anything resolving an unqualified name through the catalog default failed
+    with `Schema 'main' not found` — ducklake hits it on ATTACH, before any
+    table exists. On the 2.0 API the signature is `optional<Identifier>`, where
+    `nullopt` means "this catalog has no default, do not try"; we return
+    `Identifier("dbo")`.
+  - **Two scans of one catalog inside a transaction** (#239). A transaction pins
+    one connection and routes every read of that catalog to it, but a scan holds
+    that connection from `InitGlobal` until its last row, and DuckDB does not
+    promise to drain one source before starting the next. Such plans are now
+    materialized into a buffer-managed collection as each scan starts, freeing
+    the connection immediately. Autocommit is untouched.
+
+- **`mssql_scan()` inside a transaction took the pinned connection and kept it**
+  (#316). The #239 gate runs on the optimized plan and sees only
+  `mssql_catalog_scan`. `mssql_scan()` is unreachable from there by construction:
+  it executes its query during **binding** — that is how it reads the result's
+  column types — so a second scan of the catalog failed inside its own Bind,
+  before any `InitGlobal` and long before the gate. Reported shapes were
+  `Connection closed unexpectedly` (catalog scan + `mssql_scan`) and
+  `Cannot execute: connection not in Idle state` (two `mssql_scan`). It now
+  drains at Bind whenever the session is not in autocommit — the trigger is "in
+  a transaction", not "more than one scan", because Bind cannot see the rest of
+  the plan. Autocommit still streams.
+
+- **The materialized scan could not spill** (#318, thanks @oluies). It was built
+  from `Allocator::Get(context)`, and that overload is duckdb's
+  `IN_MEMORY_ALLOCATOR` one: unaccounted against the buffer manager and unable to
+  spill, so a large materialized scan grew in process memory until it OOMed while
+  the docs promised the opposite. Built from the `ClientContext` overload, which
+  defaults to `BUFFER_MANAGER_ALLOCATOR`.
+
 ## [0.2.4] - 2026-08-17
 
 ### Fixed
--- a/src/include/catalog/mssql_catalog.hpp
+++ b/src/include/catalog/mssql_catalog.hpp
@@ -70,6 +70,51 @@ public:
 
 	string GetCatalogType() override;
 
+	//! SQL Server's default schema, not DuckDB's `main`.
+	//!
+	//! The base Catalog answers `main`, which no SQL Server database has, so
+	//! anything that resolves an unqualified name through the catalog's default
+	//! fails with "Schema 'main' not found in MSSQL database". ducklake hits this
+	//! on ATTACH — DuckLakeTransaction::GetDefaultSchemaName() asks for it before
+	//! any table exists — and issue #129 shows a user working around it by
+	//! CREATEing a schema literally called `main` on the server.
+	//!
+	//! `dbo` is the constant answer rather than a per-login lookup: it is the
+	//! default for every login that has not been given another, and resolving
+	//! SCHEMA_NAME() would put a round trip on ATTACH and need an answer on the
+	//! lazy-validation path, where no connection has been made yet. A login whose
+	//! default schema is not `dbo` still addresses its tables by qualified name;
+	//! only the unqualified default is wrong for it, which is the pre-existing
+	//! behaviour minus the crash.
+	//!
+	//! On the duckdb 2.0 line the return type is optional<Identifier>, and the
+	//! three answers are distinct: a value means "probe this schema for
+	//! unqualified names", nullopt means "this catalog HAS no default, never
+	//! probe", and an empty Identifier means "unspecified" elsewhere in the
+	//! catalog. We mean the first, so this returns Identifier("dbo") and neither
+	//! of the empty forms. (The v0.2.x line on duckdb 1.5.5 returns a plain
+	//! string; that is the only difference between the two ports.)
+	optional<Identifier> GetDefaultSchema() const override;
+
+	//! Serializes materialized catalog scans against each other (issue #239).
+	//!
+	//! R2 drains a scan inside InitGlobal so the pinned connection is Idle again
+	//! before the next scan starts — which is enough at threads = 1, where DuckDB
+	//! initializes the two sources in sequence. With threads > 1 the two
+	//! InitGlobals run CONCURRENTLY and the drain has not happened yet when the
+	//! second one sends its batch: the failure becomes a torn stream
+	//! ("Connection closed while waiting for COLMETADATA") instead of the clean
+	//! "not in Idle state". Materialization alone cannot fix that, because the
+	//! race is between the initializations, not inside them.
+	//!
+	//! So a materialized scan holds this from before its batch until after its
+	//! drain. It is contended only by scans that are already serialized by
+	//! construction — they share one pinned connection and cannot run in parallel
+	//! anyway — so it costs nothing that was not already sequential.
+	std::mutex &MaterializeMutex() {
+		return materialize_mutex_;
+	}
+
 	optional_ptr<SchemaCatalogEntry> LookupSchema(CatalogTransaction transaction, const EntryLookupInfo &schema_lookup,
 												  OnEntryNotFound if_not_found) override;
 
@@ -280,6 +325,11 @@ protected:
 	void DropSchema(ClientContext &context, DropInfo &info) override;
 
 private:
+	//! See MaterializeMutex(). Not the transaction's connection_mutex_: that one
+	//! guards the pinned-connection member accessors and is taken and released
+	//! inside them, where this must span a whole batch-and-drain.
+	std::mutex materialize_mutex_;
+
 	// Issue #225: -1 = not observed yet, 0 = declined, 1 = granted.
 	std::atomic<int8_t> utf8_support_acked_{-1};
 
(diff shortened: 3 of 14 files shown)


oluies:
Ran a roborev review against `511b85f7` (job 1475). **Verdict: fail**, four findings. The HIGH is not confined to this PR, which is why I am posting rather than waiting for the draft to be finished.

## HIGH — the memory guarantee does not exist, and it is already on `v0.2.5`

`transactions.md` promises buffer-manager-bounded spilling; #313's description says the same (*"Memory is bounded by the buffer manager — ColumnDataCollection spills"*). The code does not do that. Verified in the duckdb source rather than inferred:

```cpp
// duckdb/src/common/types/column/column_data_allocator.cpp:14
ColumnDataAllocator::ColumnDataAllocator(Allocator &allocator)
    : type(ColumnDataAllocatorType::IN_MEMORY_ALLOCATOR) { … }

// :18 — only this one is buffer-managed
ColumnDataAllocator::ColumnDataAllocator(BufferManager &buffer_manager, …)
    : type(ColumnDataAllocatorType::BUFFER_MANAGER_ALLOCATOR) { … }
```

The drain constructs with `Allocator::Get(context)`, so the collection allocates straight from the allocator: not accounted against the buffer manager, and **it cannot spill**. A large in-transaction multi-scan grows in process memory until it OOMs, having told the user in writing that it will not.

**The same line is on the released line, not just here:**

```
origin/duckdb-v1.5.5:src/table_scan/table_scan.cpp:614
pr-314:src/table_scan/table_scan.cpp:686
    result->materialized = make_uniq<ColumnDataCollection>(Allocator::Get(context), chunk_types);
```

`v0.2.5` is tagged. The fix is one argument — `make_uniq<ColumnDataCollection>(context, chunk_types)` — and belongs on `duckdb-v1.5.5` first so this PR inherits it. Happy to raise that.

This also compounds the third finding below: the gate materializes **both sides** of a plain hash join over two tables in one catalog inside `BEGIN…COMMIT` — a query that worked before and never needed materializing — which is exactly the shape most likely to be large.

## MEDIUM — the documented scope is still wider than the gate

`transactions.md:35-41` and `:81-83` describe the fix as covering two scans of one catalog, but the optimizer collects only `mssql_catalog_scan`. #315 corrected the CHANGELOG for this; this page did not get the same treatment, and it documents `mssql_scan()` batches immediately below the new section, so a reader will reasonably assume coverage. One sentence saying raw `mssql_scan()` reads are not materialized and still cannot overlap on a pinned connection would close it.

## LOW ×2

- *"Only queries that would otherwise fail are affected"* overstates the gate's precision — it flags every scan whenever a plan holds two or more of the same catalog, making no attempt to decide whether they would have overlapped. Restating it as what the code does is both simpler and true.
- The new example uses alias `mssql` where the rest of the page uses `sqlserver`.

Full output under roborev job 1475 if you want the line-level detail.

VGSML:
All four addressed, and the HIGH changed the release plan rather than just this PR. Pushed as `930013c` + `e90db87`.

## HIGH — fixed, and it does not ship

Right on every point, including that it was already on the released line. #318 approved and merged; I confirmed the same constructor on the 2.0 pin `070ce1f6` before applying it here, so this is not a 1.5.5-only artefact:

```cpp
ColumnDataAllocator::ColumnDataAllocator(Allocator &allocator) : type(ColumnDataAllocatorType::IN_MEMORY_ALLOCATOR) {
```

**The `v0.2.5` tag has been moved rather than released-and-patched.** Nothing was published — the community-extensions PR was never opened — so the window was clean:

- run on the old commit cancelled, `v0.2.5` retagged `139ae0c` → `d35b617`, new run `34264795965` building
- `#319` corrects the two claims in the `[0.2.5]` entry that this changes
- the community descriptor pins `ref: "v0.2.5"` by name, so it picks the moved tag up with no edit

That also settles the open question on #318: nobody had compiled it on the 1.5.5 toolchain, and the retagged run is that compile — six platforms plus the linux_amd64 integration job.

## MEDIUM — true when you wrote it, false in the other direction now

You asked for *"one sentence saying raw `mssql_scan()` reads are not materialized"*. I went the other way and made them materialized (#316), because the two shapes you named are real and I reproduced both against a live server on this build:

```
catalog scan + mssql_scan   IO Error: Connection closed unexpectedly
two mssql_scan              IO Error: Cannot execute: connection not in Idle state (current: Executing)
```

`mssql_scan()` is not reachable from the optimizer gate by construction, not by oversight: `MSSQLScanBind` executes the query at **bind** time — that is where the column types come from — and holds the connection until execution drains it. So the second one fails inside its own Bind, before any `InitGlobal` exists to materialize and long before the optimizer runs. Materializing at `InitGlobal`, the way the catalog scan does it, is already too late.

So Bind drains and closes whenever the session is not in autocommit. **The trigger is "in a transaction", not "more than one scan"** — Bind cannot see the rest of the plan, and by the time a count would be available the connection is held. A lone `mssql_scan()` in a transaction is therefore buffered too. That is a real cost, and the honest defence is narrow: inside a transaction such a scan could not have coexisted with anything else on that catalog anyway, so it turns a failing case into a slower one rather than taking a working case away.

`test/sql/transaction/transaction_mssql_scan_materialization.test` covers both of your shapes, the rows-not-just-counts check, read-your-writes through the drained scan, and autocommit still streaming. Verified failing on the pre-fix binary first. Full suite after: **159 cases, 4925 assertions**.

## LOW ×2 — both taken

- *"Only queries that would otherwise fail are affected"* is gone from the page and from the CHANGELOG (#319). Replaced with what the gate does, quoting your framing: it flags every scan of a catalog once two appear, so a plain hash join over two of that catalog's tables buffers both sides. I kept the reason next to it — the plan carries no drain order, and guessing the other way costs a hung connection rather than a slow query — because otherwise the breadth reads as sloppiness rather than a choice.
- Alias unified to `sqlserver` across the page. It was wrong in the pre-existing example too, not only the new one.

## Left alone deliberately

Your point 3 about scoping the lock to `MSSQLTransaction` instead of the catalog — agreed, and not in this PR. It is a behaviour change to the locking, it wants its own test for two transactions on one catalog not blocking each other, and I would rather not put it under a tag that is about to move.

src/catalog/mssql_catalog.cpp
```diff
@@ -292,6 +292,12 @@ string MSSQLCatalog::GetCatalogType() {
 	return "mssql";
 }
 
+optional<Identifier> MSSQLCatalog::GetDefaultSchema() const {
+	// See the header: a value (not nullopt, not an empty Identifier) is what says
+	// "probe dbo for unqualified names".
+	return Identifier("dbo");
```
oluies, line 298: make this configurable?

oluies (approved):

Issue #285, opened 2026-08-29
A PR stacked on a feature branch gets no Build, Integration, C++ Unit Tests or Lint — ci.yml only triggers on PRs into main


oluies:
## Summary

`ci.yml` triggers on `pull_request: branches: [main]`, so a pull request whose **base is another branch** never runs the jobs that matter. Stacked PRs are a normal way to work here — #283 → #284 today — and they silently arrive unverified. This is the same family as #279 (PRs never compiled), one level up: there, PRs weren't built; here, *stacked* PRs still aren't.

## Evidence

Both PRs were open at the same time, same repo, same day. The only difference is the base branch.

**#283** — base `main`: **14 checks**, including `Build (linux_amd64)`, `Integration Test (linux_amd64)`, `C++ Unit Tests (linux/macos)`, `Lint`, `Windows SSPI`, smoke tests.

**#284** — base `fix/roborev-backlog`: **4 checks**, all from workflows with their own triggers:

```
Concurrency stress (scenarios 1-8)   pass
Scan PR prose for prompt injection   pass
Secret scan (TruffleHog)             pass
yamllint                             pass
```

No `Build`. No `Integration Test`. No `C++ Unit Tests`. No `Lint`.

## Why it bites harder than it looks

#284's entire change is **to `ci.yml` itself** — it deletes twelve hand-rolled test blocks on the argument that `make test-cpp-run` already covers them. That argument is precisely what a `Build` job verifies, and the PR's own checks could not verify it.

I had to dispatch the workflow by hand ([run 33250064017](https://github.com/hugr-lab/mssql-extension/actions/runs/33250064017)) to get coverage. It passed — all 18 standalone tests ran once each on both `linux_amd64` and `osx_arm64` — but that result is attached to a manual run, not to the PR, so nothing on the PR shows it and nothing would have stopped a broken version merging into its base.

The failure mode is quiet in the usual way: the PR page looks green, because four checks did run and all passed. Only counting them against another PR reveals the ten that didn't.

## Suggested fix

Options, roughly in order of cost:

1. **Drop the `branches` filter** from `ci.yml`'s `pull_request` trigger, so any PR runs it. Cost is CI minutes on stacked PRs — which is the point. The `paths-ignore` already keeps docs-only PRs free, and the Linux-only PR matrix (from #279) already bounds it.
[...]

PR #318, opened 2026-09-08
fix: the materialized scan never spilled — it was an in-memory collection all along

oluies:
One argument. v0.2.5's #239 fix does not have the memory behaviour it documents.

Found by roborev job 1475 while reviewing the forward-port in #314, which carries the identical line.

## The claim and the code

#313's description and `website/docs/writing/transactions.md` both state the materialized scan is bounded by DuckDB's buffer manager and spills when large. It was constructed from an `Allocator`:

```cpp
make_uniq<ColumnDataCollection>(Allocator::Get(context), chunk_types)
```

and in **the duckdb commit this branch pins** (`d8cdaa33`), that overload is explicitly the in-memory one:

```cpp
// src/common/types/column/column_data_allocator.cpp:13
ColumnDataAllocator::ColumnDataAllocator(Allocator &allocator)
    : type(ColumnDataAllocatorType::IN_MEMORY_ALLOCATOR) {
```

It allocates straight from the allocator — not accounted against the buffer manager, and **it cannot spill**. A large multi-scan inside an explicit transaction grows in process memory until it OOMs, having told the user in the release notes that it will not.

The `ClientContext` overload is the buffer-managed one and *defaults* to it, in that same pinned duckdb:

```cpp
// column_data_collection.hpp:38
ColumnDataCollection(ClientContext &context, vector<LogicalType> types,
                     ColumnDataAllocatorType type = ColumnDataAllocatorType::BUFFER_MANAGER_ALLOCATOR,
                     ColumnDataCollectionLifetime lifetime = ColumnDataCollectionLifetime::REGULAR);
```

so the fix is the argument, not a rewrite.

`chunk.Initialize(Allocator::Get(context), ...)` on the following line is deliberately unchanged — that one wants a plain `Allocator` and is a single `DataChunk`, not the accumulating collection.

## Bigger than the pathological case

The optimizer gate materializes **every** scan whenever a plan holds two or more of one catalog in an explicit transaction, making no attempt to decide whether they would have overlapped. So an ordinary hash join over two tables of the same catalog inside `BEGIN…COMMIT` — a query that worked before and never needed materializing — now buffers **both sides**. That is the shape most likely to be large, and it is the one that had no ceiling.

## Not built, and this branch cannot build it

Stating plainly rather than implying otherwise:

- This worktree's duckdb submodule is on `v2.0-cyanoptera` (#306), so I could not compile against 1.5.5 locally.
- **`duckdb-v1.5.5`'s `ci.yml` still carries `pull_request: branches: [main]`**, so a PR into this branch triggers no CI at all — the #285 fix that removed that filter landed on `main` only. Nothing here will go green or red; there is nothing to run it.

What I did instead was verify against the pinned duckdb commit itself rather than against a different version's headers — both the defect and the replacement signature are quoted above from `d8cdaa33`. `clang-format-14` clean.

**Someone on the 1.5.5 toolchain should compile this before the tag moves.** It is a one-argument change to a call whose overloads differ only in allocator type, so the risk is low, but "low" is not "verified".

## Related

- #314 carries the identical line and should pick this up on forward-port.
- The docs half — `transactions.md` still promising spilling, and describing a scope wider than the gate — is noted in my comment on #314 and not touched here, since this branch's copy may differ.

### DIFF
--- a/src/table_scan/table_scan.cpp
+++ b/src/table_scan/table_scan.cpp
@@ -611,7 +611,13 @@ static unique_ptr<GlobalTableFunctionState> TableScanInitGlobal(ClientContext &c
 				chunk_types.push_back(bind_data.all_types[column_ids[i]]);
 			}
 		}
-		result->materialized = make_uniq<ColumnDataCollection>(Allocator::Get(context), chunk_types);
+		// ClientContext, not Allocator::Get(context). The Allocator overload builds
+		// an IN_MEMORY_ALLOCATOR collection -- unaccounted against the buffer manager
+		// and unable to spill -- so a large materialized scan grew in process memory
+		// until it OOMed, while the docs and the release notes promised the opposite.
+		// The ClientContext overload defaults to BUFFER_MANAGER_ALLOCATOR, which is
+		// what 'bounded by the buffer manager' requires.
+		result->materialized = make_uniq<ColumnDataCollection>(context, chunk_types);
 		DataChunk chunk;
 		chunk.Initialize(Allocator::Get(context), chunk_types);
 		idx_t total = 0;


VGSML (approved):
Approving. The claim is correct, and I compiled and ran it — the verification you said was missing.

## Verified

**The defect is real and not version-specific.** You quoted `d8cdaa33` (1.5.5). It reads identically on the `v2.0-cyanoptera` pin `070ce1f6` that `main` now carries:

```cpp
// src/common/types/column/column_data_allocator.cpp:14
ColumnDataAllocator::ColumnDataAllocator(Allocator &allocator) : type(ColumnDataAllocatorType::IN_MEMORY_ALLOCATOR) {
```

and the header still labels the two overloads `//! Constructs an in-memory column data collection from an allocator` vs `//! Constructs a buffer-managed column data collection`, with the `ClientContext` one defaulting to `BUFFER_MANAGER_ALLOCATOR`. So the forward-port needs it too, which is where I applied it.

**Compiled and tested.** Against the 2.0 pin, not 1.5.5 — so your "someone on the 1.5.5 toolchain should compile this" still stands for *this* branch, and I am not claiming otherwise. What I can say is that the identical one-argument change builds clean and the full SQL suite passes on it: **159 test cases, 4925 assertions**, including `transaction_multi_scan_materialization.test`. There is nothing in the change that could behave differently across the two pins — the overload set is the same in both.

**Your CI observation is right.** `duckdb-v1.5.5`'s `ci.yml` is still `pull_request: branches: [main]`; the only checks that ran here are the prose/secret/yaml ones, which have their own triggers. That is a branch-config fix, and it is not something a PR from this account can make (the token cannot write `.github/workflows/`) — noting it so it does not get lost.

## The second half of your argument

> an ordinary hash join over two tables of the same catalog inside `BEGIN…COMMIT` now buffers **both sides**

Confirmed against the gate as written — `MaterializeSharedConnectionScans` marks every scan of a catalog the moment two appear in the plan, with no attempt to tell "would have overlapped" from "would have drained in sequence":

```cpp
for (auto &entry : by_catalog) {
    if (entry.second.size() < 2) { continue; }
    for (auto *bind_data : entry.second) { bind_data->requires_materialization = true; }
}
```

That breadth is deliberate — at plan time DuckDB does not tell us the drain order, and the failure mode of guessing wrong is a hung or torn connection rather than a slow query. But it does make the docs wrong where they say *"Only queries that would otherwise fail are affected"*, and with your fix in place the honest version is "bounded and spillable, but wider than strictly necessary". I am correcting that sentence in #314's copy rather than here, per your note.

## One thing this changes downstream

The `v0.2.5` tag is already pushed and its release run is in flight, so as it stands v0.2.5 would ship the unbounded collection **with release notes promising it spills**. That is the exact gap you are describing, and it argues for holding the community-extensions submission and moving the tag rather than releasing and patching. Raising it with @vgsml — it is a release call, not a review one.

PR #320, opened 2026-09-08
feat(#224): a non-UTF-8 text column says so once per query, not once per value

### LINKED ISSUE #224: mssql_scan() returns invalid UTF-8 for non-UTF8 collation char/varchar columns

VGSML:
The half of @oluies' #305 that costs nothing, plus a place to put it. Proposes closing #305 in favour of this.

## Where this leaves #224

#307 settled it: code-page bytes in a DuckDB `VARCHAR` are **documented, not enforced**. That decision stands, and it is worth restating why it is not "we did not care" —

A non-UTF-8 `varchar` **never reaches the binary kernel on the catalog path at all**: `NeedsNVarcharConversion` casts it to `NVARCHAR` server-side, so it arrives `0xE7` and takes the UTF-16 kernel. The only things that reach a guard there are a **UTF-8-collated** column and `VARCHAR(MAX)` with `mssql_convert_varchar_max` off. The guard would have taxed exactly the configuration #225 optimised hardest for, at ~0.03 ns/byte ASCII — scaling with **payload**, not rows.

What was missing is that the caller was never *told*. COLMETADATA is a handful of columns and `SurfaceWarnings` already runs once per stream, so the same fact costs nothing there.

```
mssql: column "cp1251" has a non-UTF-8 collation (LCID 0x0419, SortId 0); its bytes are
returned verbatim and may not be valid UTF-8. CAST it to NVARCHAR in the query to convert.
```

Reaches `duckdb_logs` at `WARNING`. No behaviour change, no refusal, nothing on the per-row path.

## Three pieces

**1. `collation_sort_id`** — the 5th wire byte, until now `offset += 5; // collation is 5 bytes but we only store 4`. @oluies' change, and here is the proof it is not redundant with the LCID:

| collation | code page | LCID | SortId |
|---|---|---|---|
| `SQL_Latin1_General_CP1_CI_AS` | CP1252 | **0x0409** | 52 |
| `SQL_Latin1_General_CP1251_CI_AS` | CP1251 | **0x0409** | 106 |

Same LCID, different code page. Without the 5th byte the first four cannot tell them apart. Applied to both parse arms, not only the `BIGCHAR` one.

**2. `ColumnMetadata::IsUtf8Collation()`** — the fUTF8 flag at `0x04000000`. #305 could not verify this; it is verified here against a live SQL Server 2025 by dumping COLMETADATA for known collations, and the evidence table is in the header:

| collation | word | fUTF8 |
|---|---|---|
| `Latin1_General_100_CI_AS_SC_UTF8` | `0x24D00409` | set |
| `Latin1_General_100_BIN2_UTF8` | `0x26000409` | set |
| `Latin1_General_CI_AS` (cp1252) | `0x00D00409` | clear |
| `Cyrillic_General_CI_AS` (cp1251) | `0x00D00419` | clear |

Each decodes without remainder (`CI` → fIgnoreCase, `AS` → fIgnoreAccent clear, `BIN2` → fBinary2), and it degrades correctly: UTF-8 collations are SQL Server 2019+, so the bit is never set on an older server.

**3. `SurfaceWarnings` actually surfaces.** It walked `info_messages_` and did nothing:

```cpp
// DuckDB doesn't have a built-in warning API, but we can log to the client context
(void)context;  // Context can be used for logging if needed
```

True when it was written; not true now — 2.0 has `DUCKDB_LOG_WARNING` and `duckdb_logs`. So `PRINT` output, `RAISERROR` below severity 11 and procedure progress notices stop being dropped on the floor. A second latent hole, found while looking for somewhere to put the first one.

## Verified

Against a live server, which is what makes the fUTF8 predicate more than a claim — four columns in one query:

| column | collation | warns |
|---|---|---|
| `w_cp1251` | `Cyrillic_General_CI_AS` | yes |
| `w_cp1252` | `Latin1_General_CI_AS` | yes |
| `w_utf8` | `Latin1_General_100_BIN2_UTF8` | **no** |
| `w_nvarchar` | NVARCHAR | **no** |

`test/sql/query/collation_metadata_warnings.test` covers that, the SortId pair above, the `PRINT` round trip on both paths, the abandoned-stream case and the no-duplicate case. Full SQL suite **157 cases / 4878 assertions**; C++ unit tests 133 cases; `clang-format-14` clean.

## Self-review found two defects — both fixed in `3591799`

Posting these rather than quietly folding them in, since they are the kind a reviewer should get to see.

**1. The warning was silent for exactly the query that needed it most.** `SurfaceWarnings` was called only at the drain, from `if (rows == 0)`. A query that stops early never gets there — the executor just stops asking for chunks:

```
SELECT v FROM mssql_scan(..., 'SELECT TOP 5000 ... COLLATE Cyrillic_General_CI_AS') LIMIT 1
warnings: 0
```

A `LIMIT` on an `mssql_scan` is how someone looks at an unfamiliar table for the first time, which is precisely when they do not yet know the column is CP1251. The same hole lost INFO messages on any early-terminating query.

The column warning depends only on COLMETADATA, known at `InitGlobal`, so it fires there **as well as** at the drain. `SurfaceWarnings` is written for the repeat: columns are one-shot behind `collations_warned_`, INFO resumes from an `info_surfaced_` cursor rather than replaying (they accumulate as the batch runs, so a cursor and not a flag). Both scans are `MaxThreads() = 1`, so neither needs a lock.

**2. My own CHANGELOG claim was wider than the code — `mssql_exec` dropped `PRINT`.** `MSSQLSimpleQuery`'s token loop read `case Info: // Ignore informational messages`. So "PRINT output now reaches `duckdb_logs`" was true for the scans and false for `mssql_exec` — the function you call to run a procedure, and therefore the one most likely to be PRINTing at you.

`MSSQLSimpleQuery` has no `ClientContext` and should not get one: every catalog metadata query goes through it. So it collects into `SimpleQueryResult` and the caller that *does* have a context logs them. A normal query sends no INFO tokens, so the vector is never allocated. They are logged **before** the error check, so a batch that reported its progress and then failed surfaces the progress instead of losing it with the exception:

```sql
SELECT mssql_exec('m','PRINT ''step one done''; RAISERROR(''boom'', 16, 1)');
-- raises "boom", and duckdb_logs holds "mssql: [0] step one done"
```

### Also checked, and now asserted rather than assumed

- **The default catalog path warns for nothing** — verified by selecting *every* column of a table holding a CP1251 `VARCHAR(50)`, a CP1252 `VARCHAR(MAX)` and an `NVARCHAR`: zero warnings, because `NeedsNVarcharConversion` has already cast them. With `mssql_convert_varchar_max = false`, exactly the `VARCHAR(MAX)` warns and the other still does not. That was in the description above as a claim; it is now measured.
- All three `SurfaceWarnings` call sites sit behind a `done` guard, so the drain call fires once per stream and the init call cannot double it. Asserted in the test, not just reasoned about.
- The 5-byte read is in bounds on both parse arms: `offset + 7 > length` before a `+2` then `[+4]`, and `offset + 5 > length` before `[+4]`.
- `MaxThreads() = 1` on both scans, so `info_surfaced_` / `collations_warned_` are not raced.

## Not taken from #305

The per-value guard, `utf8_guard.hpp`, and the codec changes — that is the part the benchmark and #307 ruled out. `DATAMODEL.md` gets @oluies' three-decoder table, rewritten to describe the behaviour that exists rather than the guard.

Closes #224. Proposes closing #305.


test/sql/query/collation_metadata_warnings.test
```diff
@@ -0,0 +1,175 @@
+# name: test/sql/query/collation_metadata_warnings.test
```
oluies, line 1: **Medium — the two new pure functions are only covered by a test CI never runs**

`ColumnMetadata::IsUtf8Collation()` (the `0x04000000` bit position) and the newly captured 5th collation byte are exercised only by this SQLLogicTest, which needs `MSSQL_TESTDB_DSN` and a live SQL Server. The integration `.test` suite is not exercised by CI in this repo, so a regression ships green — someone re-ordering the two `offset += 5` sites, or the TEXT/NTEXT arm at `src/tds/tds_column_metadata.cpp:410` drifting from the BIGVARCHAR arm at `:318`.

Both functions are DuckDB-free and consume a byte buffer, exactly like `test/cpp/test_token_parser_security.cpp`. Suggest a server-free unit test feeding synthetic COLMETADATA byte streams through `TokenParser`, asserting for **both** the `BIGVARCHAR` and the `TEXT` arm:

- `collation == 0x24D00409 && collation_sort_id == 0` -> `IsUtf8Collation()` true
- `0x00D00409` with SortId 52, and `0x00D00409` with SortId 106 -> false, and distinguishable from each other
- `IsSingleByteTextColumn()` false for `NVARCHAR` / `IMAGE`

src/query/mssql_result_stream.cpp
```diff
@@ -905,17 +906,56 @@ void MSSQLResultStream::DrainRemainingTokens() {
...
+	// when it was written and is not now.
+	// Resume where the last call stopped: a stream is visited at least twice and
+	// these accumulate as the batch runs.
+	for (; info_surfaced_ < info_messages_.size(); info_surfaced_++) {
+		const auto &info = info_messages_[info_surfaced_];
+		if (info.message.empty()) {
+			continue;
+		}
+		const std::string where = info.proc_name.empty() ? std::string() : " (in " + info.proc_name + ")";
+		DUCKDB_LOG_WARNING(context, "mssql: [%u] %s%s", info.number, info.message.c_str(), where.c_str());
```
oluies: **Low — every INFO token logs at `WARNING`; the parsed `severity` is ignored**

`TdsInfo::severity` is already parsed and available (`src/include/tds/tds_token_parser.hpp:47`) but nothing here reads it. SQL Server sends INFO for genuinely routine things — 5701 "Changed database context", 5703 "Changed language setting", "Null value is eliminated by an aggregate function", deprecation notices — so a batch containing a `USE` or a `SET` now produces `WARNING` rows on every call.

Since `WARNING` is also what the new collation check uses, the useful signal gets diluted by chatter that SQL Server itself classifies as informational (severity 0-10).

Suggested: `severity >= 11 ? DUCKDB_LOG_WARNING : DUCKDB_LOG_INFO`, and include the severity in the message text.

src/mssql_functions.cpp
```diff
@@ -413,6 +420,19 @@ static void MSSQLExecExecute(DataChunk &args, ExpressionState &state, Vector &re
...
+			// the error check below, because a batch that ultimately failed is
+			// exactly when its notices are worth reading.
+			for (const auto &info : query_result.info_messages) {
+				if (info.message.empty()) {
+					continue;
+				}
+				const std::string where = info.proc_name.empty() ? std::string() : " (in " + info.proc_name + ")";
+				DUCKDB_LOG_WARNING(client_context, "mssql: [%u] %s%s", info.number, info.message.c_str(),
+								   where.c_str());
+			}
```
oluies, line 495: **Low — this INFO-logging loop is duplicated verbatim**

Identical to `src/query/mssql_result_stream.cpp:921-928`: same `where` construction, same `"mssql: [%u] %s%s"` format string, same empty-message skip. Any later change — adding severity (see the comment on that file), a line number, or the server name — has to be made twice, or the two paths diverge in what a user sees for the same server message.

Suggested: extract a small free function, e.g. `void LogTdsInfo(ClientContext &, const tds::TdsInfo &)` next to the `TdsInfo` consumers, and call it from both.

src/query/mssql_result_stream.cpp
```diff
@@ -905,17 +906,56 @@ void MSSQLResultStream::DrainRemainingTokens() {
...
+	}
+	collations_warned_ = true;
+	for (const auto &col : column_metadata_) {
+		if (!col.IsSingleByteTextColumn() || col.IsUtf8Collation()) {
+			continue;
 		}
+		DUCKDB_LOG_WARNING(context,
+						   "mssql: column \"%s\" has a non-UTF-8 collation (LCID 0x%04X, SortId %u); its bytes are "
+						   "returned verbatim and may not be valid UTF-8. CAST it to NVARCHAR in the query to convert.",
+						   col.name.c_str(), (unsigned)(col.collation & 0xFFFFFu), (unsigned)col.collation_sort_id);
```
oluies: **Low — no off switch, and by this code's own documentation the trigger is the majority configuration**

`SQL_Latin1_General_CP1_CI_AS` is the SQL Server installation default, so every `mssql_scan()` selecting a `VARCHAR` column against such a server emits one `WARNING` per column per query — including for data that is pure ASCII and therefore already valid UTF-8. Anyone monitoring `duckdb_logs` for `WARNING` gets a permanent, unactionable stream.

Separately, the remediation text does not fit the catalog path: on a table scan the user wrote no query, so "CAST it to NVARCHAR in the query" is not a lever they have — the actual lever there is `mssql_convert_varchar_max`.

Suggested: either gate it behind a setting (e.g. `mssql_warn_non_utf8_collation`, default on), or de-duplicate per `(catalog, table, column)` for the session rather than per stream; and tailor the remediation text for the catalog path.

src/mssql_functions.cpp
```diff
@@ -160,6 +161,9 @@ unique_ptr<GlobalTableFunctionState> MSSQLScanInitGlobal(ClientContext &context,
 		auto &mssql_catalog = catalog.Cast<MSSQLCatalog>();
 		result->result_stream = mssql_catalog.RetrieveStream(bind_data.result_stream_id);
 		if (result->result_stream) {
+			// COLMETADATA is known, so warn before any rows move -- the drain-end
+			// call is missed entirely by a query that stops early (issue #224).
+			result->result_stream->SurfaceWarnings(context);
```
oluies, line 209: **Low (context for the unanchored finding in the review body)** — this init-time call is what makes the early-stop case work, but it is also the only `SurfaceWarnings` that survives a mid-stream failure. The catch block at `src/mssql_functions.cpp:235` (and its peer at `src/table_scan/table_scan_execute.cpp:65`) sets `done = true` and rethrows without surfacing, so INFO tokens that arrived after COLMETADATA are dropped exactly when they are most useful.

oluies (commented):
**Automated review (roborev, job 1479)** over `6be5581..3591799`.

No correctness defects found in the new mechanics — the `0x04000000` bit position, the bounds checks, the one-shot/cursor idempotence and the init-plus-drain call placement all check out. The five findings below are test placement and log-level/noise policy.

One finding could not be anchored inline because it lands on lines this PR does not touch:

**Low — the commit's own principle is not applied to the streaming paths** (`src/mssql_functions.cpp:235`, `src/table_scan/table_scan_execute.cpp:65`)
"the notices of a batch that failed are the ones worth reading" got applied to `mssql_exec`, but when `FillChunk` throws mid-stream both catch blocks set `done = true` and rethrow without calling `SurfaceWarnings`. Any INFO tokens that arrived after COLMETADATA — typically the ones immediately preceding the server error, which are the most useful ones — are dropped. Only INFOs seen during `Initialize()` survive, via the new init-time call. Fix: call `SurfaceWarnings(context)` in the catch before rethrowing, guarded against a nested throw from the logger.

VGSML:
All five taken, in `68db1ce`. One of them changed shape once I measured it — details below, because the suggested rule would have been a no-op and it is worth knowing why.

## Medium — pure functions covered only by a test CI never runs

Correct, and the consequence is exactly as stated: the integration suite sits behind the `workflow_dispatch` gate, so a regression there ships green.

`test/cpp/test_collation_metadata.cpp` feeds synthetic COLMETADATA through `ColumnMetadataParser` — no server, no DuckDB runtime — and is registered in `STANDALONE_TEST_SOURCES`, which both `make test-cpp` and the CI job drive off, so it gates a PR. It covers everything you listed plus two cases your list implies: that a collation-less column (`IMAGE`) does not inherit its predecessor's, and that `NTEXT` is correctly *not* single-byte text.

**Verified by mutation rather than asserted.** Both regressions you named:

| mutation | result |
|---|---|
| delete `collation_sort_id` from the TEXT/NTEXT arm | `FAIL: TEXT SortId round-trips -- the arm that drifts` |
| `0x04000000` → `0x40000000` | 4 checks fail |

One wrinkle worth recording for anyone doing this next: `make test-cpp-run` links the **prebuilt** `libmssql_extension.a`, so mutating a `.cpp` without rebuilding proves nothing. Only the inline header predicates recompile with the test. My first mutation run "passed" for that reason.

## Low — INFO at WARNING, severity ignored

The substance is right and it is taken. The suggested rule is not implementable, though, and neither half of it works:

**`severity >= 11` never fires.** Above 10 the server sends an ERROR token, not an INFO one, so nothing reaching this code can be 11.

**Severity does not separate the cases you named either.** Measured against SQL Server 2025:

```
PRINT 'x'              -> [0     sev 0]
USE master             -> [5701  sev 0]      <- sys.messages says severity 10
RAISERROR('x', 1, 3)   -> [50000 sev 1]
RAISERROR('x', 9, 5)   -> [50000 sev 9]
RAISERROR('x', 10, 6)  -> [50000 sev 0]      <- 10 is sent as class 0
```

So 1..9 round-trip faithfully — the parser is fine, the field is read correctly — but **severity 10 is sent on the wire as class 0**, which puts 5701 "Changed database context" (the chattiest one) and `PRINT` (the useful one) in the same bucket. The discriminator is the message **number**, not the severity.

So: all INFO tokens go to `LOG_INFO`, `WARNING` is left to the one message worth acting on, and both number and severity are in the text so a reader can filter on either. Recorded in `DATAMODEL.md` and in the test, since "route by severity" is the obvious idea and someone will have it again.

## Low — duplicated loop

Extracted to `LogTdsInfo` in `query/tds_info_log.hpp`, called from both paths. Your prediction came true immediately: the severity change would otherwise have had to be made twice.

## Low — no off switch, and wrong remediation for the catalog path

Both taken.

`mssql_warn_non_utf8_collation`, default `true`. Your argument is the right one — `SQL_Latin1_General_CP1_CI_AS` is the installation default, so a raw `mssql_scan()` over such a server logs one line per varchar column per query forever, pure-ASCII data included. Read from the asking session, and after the one-shot flag, so flipping it mid-session cannot make an already-visited stream warn late.

I went with the setting over per-`(catalog, table, column)` de-duplication because `mssql_scan()` rows have no table identity to key on, so the de-dup would have covered only the path that is already quiet by default.

The remediation text is now path-aware: on a catalog scan it names `mssql_convert_varchar_max`, since that being off is the only reason such a column reached the decoder there at all.

## Low — mid-stream failure drops INFO

My own stated rule, applied to `mssql_exec` and not to the streams — a fair catch. All three catch blocks (`mssql_functions.cpp`, `table_scan_execute.cpp`, and `table_scan.cpp`, which you did not list but has the same shape) now surface before rethrowing, inside a `try` that swallows: whatever the logger does, it must not replace the exception the caller is waiting for.

---

SQL suite 157 cases / 4887 assertions; 19 standalone C++ tests; `clang-format-14` clean.

oluies (dismissed):

PR #326, opened 2026-09-09
feat(#321): MSSQL_VARCHAR('MAX', 'collation') — a per-column MAX target

### LINKED ISSUE #321: MSSQL_VARCHAR(MAX, 'collation'): a per-column MAX target, without reaching for global settings
## The gap

Spec 060 gives a column two ways to state its SQL Server type, and they do not overlap:

- **Annotated** — `col::MSSQL_VARCHAR(n, 'coll')` → `varchar(n) COLLATE coll`. Per column, per statement, no session state. But `n` is capped: `BindMssqlStringType` (`src/mssql_extension.cpp:81-90`) throws `BinderException("MSSQL_VARCHAR(n): n must be between 1 and 8000 (use VARCHAR for MAX)")`, and `MAX_VARCHAR_LENGTH`/`MAX_NVARCHAR_LENGTH` are 8000/4000 (`src/include/codec/target_string_type.hpp:43-44`).
- **Unannotated** — a bare `VARCHAR` becomes `VARCHAR(MAX) COLLATE <mssql_utf8_collation>` or `NVARCHAR(MAX)` (`src/codec/string_codec.cpp:995-1005`). That is the MAX form, and spec 060 § 9 says so deliberately: *"`mssql_default_string_length = 0` is the MAX form, and it is the default."* But which of the two it is, and with which collation, is decided by `mssql_ctas_text_type` and `mssql_utf8_collation` — both `SetScope::GLOBAL`.

So **a MAX column with a stated collation can only be asked for globally**. There is no cast that says it.

## Why that is awkward

Library code that creates tables through CTAS or `COPY … (CREATE_TABLE true)` cannot set a GLOBAL setting: it would silently change the type of every unannotated string column in the user's session, including in other extensions' statements, and it has to be restored afterwards even on an exception.

The concrete case is a DuckLake metadata catalog on SQL Server ([hugr-lab/mssql-ducklake](https://github.com/hugr-lab/mssql-ducklake), which the v0.2.5 fixes came from). DuckLake inlines small inserts into catalog tables it owns; a string column there has **no length bound at all** — DuckDB dropped it, and DuckLake never had one — so MAX is the only correct target. The collation is not a preference either: DuckLake pushes string comparisons into the metadata server against `min_value`/`max_value` statistics that DuckDB computed in UTF-8 byte order, so the column must compare binary. A UTF-8 `_BIN2_` collation is exactly that; a CI/AS one prunes wrongly and drops rows from a result.

Today that leaves writing the `CREATE TABLE` by hand through `mssql_exec` and using `COPY … (CREATE_TABLE false)` — which works, and is what we will do — while the annotated path stays unavailable for precisely the columns that most need a stated collation.

## The ask

Let the length modifier say MAX:

```sql
-- either spelling, whichever fits the binder better
COPY (SELECT s::MSSQL_VARCHAR(MAX, 'Latin1_General_100_BIN2_UTF8') FROM …) TO 'mssql://db/dbo/t' (FORMAT 'bcp', CREATE_TABLE true);
CREATE TABLE db.dbo.t AS SELECT s::MSSQL_VARCHAR(-1, 'Latin1_General_100_BIN2_UTF8') FROM …;
```

→ `varchar(max) COLLATE Latin1_General_100_BIN2_UTF8`, and `MSSQL_NVARCHAR(MAX)` → `nvarchar(max)`.

## Why it looks small

The pieces are already there:

- `FormatTargetStringDdl` (`src/codec/target_string_type.cpp:60-70`) formats one integer; a sentinel would branch to `"max"` instead of `%d`.
- The write path already handles a MAX-sized target: `TargetResolver::GetTDSMaxLength` (`src/copy/target_resolver.cpp:1477-1487`) returns `0xFFFF` — the PLP form — as soon as `n * 2 > 8000`, so a sentinel just lands in the branch that exists.
- The overflow guard becomes a no-op for the column rather than needing a new rule.

Mostly it is the binder cap, plus deciding the spelling (`MAX` keyword vs `-1`) and whether `MSSQL_NVARCHAR(MAX)` comes along.

## Not urgent

`mssql-ducklake` ships against DuckDB v1.5.5 with its own `CREATE TABLE` + `COPY (CREATE_TABLE false)`, so nothing is blocked. When this lands on the 2.0 line we would switch those columns to the explicit cast and drop the hand-written DDL.

VGSML:
Closes #321.

```sql
CREATE TABLE db.dbo.docs AS
  SELECT body::MSSQL_VARCHAR('MAX', 'Latin1_General_100_BIN2_UTF8') AS body,
         notes::MSSQL_NVARCHAR('MAX')                               AS notes
  FROM local_table;
```

→ `varchar(max) COLLATE Latin1_General_100_BIN2_UTF8` and `nvarchar(max)`.

## The spelling

Three forms reach the binder and all normalize to **one stored sentinel**, so nothing downstream knows there was a choice:

| written | why it is accepted |
|---|---|
| `'MAX'` | reads like the T-SQL it stands for |
| `0` | what this extension already means by MAX — `mssql_default_string_length = 0` **is** the MAX form (spec 060 § 9). This is what gets stored. |
| `-1` | what SQL Server means by it, in `sys.columns.max_length` |

`0` over `-1` as the stored value is the deliberate part: two sentinels for one concept in one codebase is worse than either, and `ApplyDefaultStringType` already treats `length <= 0` as "MAX, leave it alone".

**A bare `MAX` keyword is not supported and cannot be.** DuckDB's *parser* refuses a non-constant type modifier before the extension is consulted:

```
D SELECT 'x'::MSSQL_VARCHAR(MAX, 'Latin1_General_100_BIN2_UTF8');
Parser Error: Expected a constant as type modifier
```

So `MSSQL_VARCHAR(MAX, …)` would need a change to DuckDB, not to this extension. Asserted in the test so it does not get re-filed as a bug.

## Where the issue's estimate was wrong

The issue expected the write path to need nothing, since `GetTDSMaxLength` already returns the PLP sentinel for an oversized target. **It does not** — the length is multiplied by 2 *first*, and neither candidate sentinel survives that:

| sentinel | `length * 2` | `> 8000`? | returned |
|---|---|---|---|
| `-1` | `-2` | no | `(uint16_t)-2` = **65534** |
| `0` | `0` | no | **0** |

`0xFFFF` is what the encoder's overflow guard reads as *"no bound"*; anything else is a bound. So a MAX column would have been declared **zero-length on the wire** and the guard would have rejected every non-empty value — while the generated DDL still said `max`. Tested before the arithmetic at both sites that do it, and asserted in the unit test.

## Tests

The DDL is the point of the annotation, and it is a pure function — so it is covered where CI can actually run it. `test/cpp/test_ctas_type_mapping.cpp` (already in `STANDALONE_TEST_SOURCES`, which the CI job drives off) gains the DDL cases and the wire-length cases. **Verified by mutation**: dropping the MAX branch from `FormatTargetStringDdl` emits `varchar(0)`, which no server accepts.

`test/sql/ctas/annotated_max_string.test` is the live-server end — CTAS, `COPY … (CREATE_TABLE true)`, the parser refusal, the binder refusals, and a **12000-byte value round-tripping exactly** through the MAX column, well past the 8000-byte inline form.

Verified against a live server: all three spellings produce `max_length = -1` in `sys.columns` with the named collation, `nvarchar(max)` comes along, and a bounded column is untouched.

## Found while testing

The inline boundary is **inclusive**, so `nvarchar(4000)` is exactly 8000 wire bytes and stays inline; 4001 goes PLP. That was already the behaviour and was not written down anywhere — my first assertion got it wrong, and it is now pinned.

## Not in this PR

The **catalog direction**. `mssql_column_info.cpp` reads `sys.columns.max_length`, sees `-1` for a MAX column, and falls through to a plain `VARCHAR` — so a `varchar(max) COLLATE X` source still loses `X`, and a CTAS from it gets whatever the session settings say. The sentinel now makes reporting `MSSQL_VARCHAR('MAX', 'X')` there possible, and it would complete the round trip `mssql_catalog_native_types` promises. It also changes the type name `DESCRIBE` prints for every `varchar(max)` column in every attached database, which is a visible change and a separate decision — say the word and I will do it.

Full SQL suite **161 cases / 4996 assertions**; 19 standalone C++ tests; `clang-format-14` clean.


VGSML:
Self-review, in `840da19` + `71cb973`. No defects in the mechanics; one bad output, and four reachable paths that nothing tested.

## The output was wrong, and visibly so

Asked what `DESCRIBE` shows for a MAX column, the answer was **`MSSQL_VARCHAR(0)`** — the internal sentinel, printed verbatim, reading as a *zero-length* column. With `-1` it would have been `MSSQL_VARCHAR(-1)`, which reads as nothing at all. Neither is something to put in front of a user, and it is visible on the cast this PR adds, not only in the catalog direction it might later enable.

The extension-info modifier now carries the word for the MAX form while the internal length stays numeric — one boundary where the sentinel becomes something a person reads. DuckDB renders it unquoted:

```
a_by_string     MSSQL_VARCHAR(MAX)
b_by_negative   MSSQL_VARCHAR(MAX)
c_by_zero       MSSQL_VARCHAR(MAX)
d_nvarchar      MSSQL_NVARCHAR(MAX)
e_bounded       MSSQL_VARCHAR(50)
```

**The asymmetry that creates is worth naming**: the printed form is exactly the form the parser refuses on input, since unquoted `MAX` is not a constant. Pasting a type name back into a cast is the obvious thing to do with it, so it is in the docs, the code comment, and pinned by the test asserting both halves.

## The riskiest change was not the feature

That rendering fix changed the modifier from `INTEGER` to `VARCHAR` — which is the **type identity**, not a display string. Checked rather than assumed:

- the cast functions are registered **by alias**, with no extension info, so the modifier's type cannot affect cast resolution;
- `mssql_extension.cpp:94` (the binder itself) is the only reader of `modifiers[0]` outside `TryGetTargetStringType` — there is no second decoder to keep in step;
- `'max'`, `'MAX'` and `-1` produce **one** type: `typeof` equal, and `UNION ALL` over two of them unifies rather than reconciling two identities;
- the type survives being stored and read back (a view), so the string modifier serializes.

The last two are now asserted. They are the expensive kind of regression — a second identity for the same type does not fail loudly, it fails at a set operation months later.

## Reachable paths that nothing tested

- **`CREATE TABLE` with a column list goes through a different translator from CTAS.** Only the CTAS one was exercised, so half the feature was unverified. Added — plus an `INSERT` into the created table, which reads the **catalog** rather than the annotation and is therefore an independent check that the column really is MAX, not a re-read of what we just declared.
- **The annotation outranks `mssql_default_string_length`.** Now asserted with both in one statement: annotated → `varchar(max)`, plain → `nvarchar(200)`.

## One comment, no behaviour change

`mssql_column_info.cpp` reads `sys.columns.max_length`, where a MAX column reports `-1`. `-1 / 2` truncates to **0** — which is now the MAX sentinel. The `< 1` test that keeps such a column on the plain-`VARCHAR` path was tidy before and is load-bearing now, so it says so. Relaxing it would otherwise start reporting `MSSQL_NVARCHAR(MAX)` by accident rather than by decision; doing that **on purpose** is the open half noted above, and the comment names the trade.

## Also verified, not worth a test

A bounded column still refuses an oversized value (server-side truncation error, unchanged); the explicit DDL path handles a non-UTF-8 collation as before; lowercase `'max'` binds.

SQL suite **161 cases / 5039 assertions**; 19 standalone C++ tests; `clang-format-14` clean.

src/copy/target_resolver.cpp
```diff
@@ -1209,7 +1209,13 @@ vector<BCPColumnMetadata> TargetResolver::GenerateColumnMetadata(const vector<Lo
...
 					// way an nvarchar declaration doubles it. An unannotated column
 					// stated no length and is varchar(max) — the PLP sentinel.
-					const bool is_max = !annotated_varchar || spec.length > MAX_INLINE_VARCHAR_BYTES;
+					// An ANNOTATED max column (issue #321) reaches here with
+					// length 0, which is not > MAX_INLINE_VARCHAR_BYTES and
+					// would otherwise be declared as varchar(0) on the wire.
+					// Only the unannotated case was max before, so the
+					// IsMaxLength arm is the new one.
+					const bool is_max =
+						!annotated_varchar || codec::IsMaxLength(spec.length) || spec.length > MAX_INLINE_VARCHAR_BYTES;
```
oluies, line 1218: **Medium — the flagship case's wire branch has no CI-runnable coverage.**

The new C++ test covers `GetTDSMaxLength`, but for `MSSQL_VARCHAR('MAX', 'collation')` — the headline example — that value is immediately *overwritten* by this block: `col.max_length = is_max ? 0xFFFF : ...` with the token switched to `TDS_TYPE_BIGVARCHAR`. So the one branch the flagship case actually takes on the wire is exercised only by `test/sql/ctas/annotated_max_string.test`, which needs a live SQL Server.

If the `IsMaxLength` arm regressed, `col.max_length` would be `0`, and `BCPColumnMetadata::GetSQLServerTypeDeclaration()` (`target_resolver.cpp:182`) would emit `varchar(0)` into the `INSERT BULK` text while the encoder's guard (`!IsPLPType() && utf8_len > col.max_length`) truncated every value — the exact silent failure this PR was written to prevent, and nothing in CI would notice. That matters more than usual here given that the integration `.test` files are not run by the PR CI lane.

`GenerateColumnMetadata` is a public static over plain vectors (`target_resolver.hpp:287`), so this is directly callable from `test_ctas_type_mapping.cpp` — which is already in `STANDALONE_TEST_SOURCES`. Assert `max_length == 0xFFFF`, `tds_type_token == TDS_TYPE_BIGVARCHAR`, and `GetSQLServerTypeDeclaration() == "varchar(max) COLLATE ..."` for a MAX-annotated varchar, with a bounded one alongside for contrast.

src/mssql_extension.cpp
```diff
@@ -78,16 +78,40 @@ static LogicalType BindMssqlStringType(BindLogicalTypeInput &input, bool unicode
...
-		throw BinderException("%s(n): n must be a non-NULL integer", name);
-	}
-	spec.length = length_val.DefaultCastAs(LogicalType::INTEGER).GetValue<int32_t>();
-	if (spec.length < 1 || spec.length > limit) {
-		// n is SQL Server's own unit for the type named: UTF-16 code units for
-		// nvarchar, BYTES for varchar. Past the limit SQL Server requires MAX,
-		// which is what a plain VARCHAR already asks for.
-		throw BinderException("%s(n): n must be between 1 and %d (use VARCHAR for MAX)", name, limit);
+	if (!length_val.IsNull() && length_val.type().id() == LogicalTypeId::VARCHAR) {
+		if (!StringUtil::CIEquals(length_val.ToString(), "max")) {
```
oluies: **Low — the MAX keyword's spelling is duplicated.**

`target_string_type.cpp:23` defines `static constexpr const char *MAX_MODIFIER = "MAX"` and uses it on both the build side (`MakeTargetStringType`) and the read side (`TryGetTargetStringType`). The binder compares against a hardcoded `"max"` literal instead.

The header already exports `MAX_LENGTH` and `IsMaxLength` precisely so the two sides cannot drift on the *sentinel* — the string was left behind. Changing `MAX_MODIFIER` would leave this binder accepting a word the codec no longer recognises, and `TryGetTargetStringType` would then return `false` for a type the binder had just built.

Exporting `codec::MAX_MODIFIER` (or an `IsMaxKeyword(const std::string &)` helper) from `target_string_type.hpp` and using it here — including in the error message on the next line — closes it.

src/catalog/mssql_column_info.cpp
```diff
@@ -92,6 +92,16 @@ LogicalType MSSQLColumnInfo::NativeDuckDBType() const {
 		// sys.columns reports these in bytes, two per UTF-16 code unit.
 		spec.unicode = true;
 		spec.length = max_length / 2;
+		// The `< 1` half is load-bearing and not merely tidy since issue #321
```
oluies: **Low — this comment points at the wrong guard.**

The claim is that the `< 1` test below is what keeps a MAX column off the native-type path, because `max_length` is `-1` and `-1 / 2` truncates to `0`. But `NativeDuckDBType()` already returns early at line 82:

```cpp
if (duckdb_type.id() != LogicalTypeId::VARCHAR || max_length <= 0 || is_cast_required || is_geometry) {
    return duckdb_type;
}
```

`max_length <= 0` catches `-1` before the `nvarchar`/`nchar` branch is ever entered, so `spec.length` is never `0` from a MAX column here and the `< 1` test is unreachable for that input.

That makes the comment actively misleading about the change it describes: someone doing the open half of #321 on purpose would relax `< 1` as instructed, see nothing change, and have to rediscover line 82. Naming line 82 as the guard that has to move (and `< 1` as the second one) would make it right — the trade-off it documents is worth keeping.

oluies (commented):
Automated review (roborev job 1488, `claude-code`) over `18ef41e..71cb973`, plus a manual read-through. **No defects in the mechanics** — I checked the three `TryGetTargetStringType` consumers this PR does *not* touch (`string_codec.cpp:995`, `mssql_catalog.cpp:964`, `target_resolver.cpp:1317`) and all of them delegate to `FormatTargetStringDdl` or never compute on `length`, so no arithmetic site was missed. `0xFFFF` really is the PLP marker (`BCPColumnMetadata::IsPLPType`, `target_resolver.hpp:152`), so the wire-length reasoning in the description holds.

Four comments, three inline. The fourth has no diff line to hang on:

**`DATAMODEL.md` was not updated (low).** Line 594 documents `TargetStringType` as carrying `{unicode, length, collation}`. This PR gives `length == 0` a *meaning* — MAX — and with it a new cross-cutting invariant: *every site that computes on `length` must test `IsMaxLength` first*. That invariant is the whole reason two call sites needed fixing here, and CLAUDE.md asks that a change altering a layer invariant update the relevant `DATAMODEL.md` section in the same PR. One sentence at line 594 would do it.

Nothing here blocks merge.

VGSML:
All four in `83dec6a`. The Medium one is a real hole and the sharper version of my own reasoning — I checked `GetTDSMaxLength` and stopped there.

## Medium — the flagship case's wire branch

Right, and the way you framed it is the point: `GetTDSMaxLength` is **not the last word** for `MSSQL_VARCHAR('MAX', 'collation')`. The UTF-8 retarget in `GenerateColumnMetadata` overwrites `max_length` with its own branch, so the number the headline example actually puts on the wire comes from a site my unit test never called — and the only thing covering it needed a live server.

Now called directly, asserting `max_length`, the type token and the `INSERT BULK` declaration text, with a bounded column and a plain `VARCHAR` alongside for contrast. **Verified by mutation**, and it fails exactly as you predicted:

```
--- dropping the IsMaxLength arm from GenerateColumnMetadata
ASSERTION FAILED at test/cpp/test_ctas_type_mapping.cpp:270
  Expected: 65535
  Actual:   0
```

`varchar(0)` in the `INSERT BULK` text with the guard truncating every value, while the `CREATE TABLE` still said `max`.

## Low — the duplicated keyword

Taken. `MAX_MODIFIER` was file-local while the binder compared against a hardcoded `"max"`, which is exactly the drift `MAX_LENGTH` / `IsMaxLength` were exported to prevent — the string had been left behind. Now `codec::MAX_MODIFIER` + `codec::IsMaxKeyword`, used on both sides and in the error message.

## Low — my comment named the wrong guard

You are right, and I verified before agreeing rather than after: `max_length <= 0` in the guard at the top of `NativeDuckDBType` returns first, so `< 1` is unreachable for a MAX column. The comment would have sent someone doing the open half of #321 to relax `< 1`, see nothing change, and rediscover line 82 themselves — worse than no comment. Rewritten to name that guard first and `< 1` as the second one that also has to move, keeping the trade-off it documents.

## Low — DATAMODEL.md

Taken, and it earned more than the one sentence: the invariant is the reason two call sites needed fixing, so it is written down with **why neither candidate sentinel survives arithmetic** (`0 * 2` is 0, `-1 * 2` is -2, neither is the `0xFFFF` PLP marker), **which two sites compute** on `length`, and why the printed modifier carries the word rather than the number.

---

Thanks for checking the three consumers this PR does not touch — that was the question I would have had to answer next, and `target_resolver.cpp:1317` delegating to `FormatTargetStringDdl` is the one I would not have thought to re-derive.

SQL suite **161 cases / 5039 assertions**; 19 standalone C++ tests; `clang-format-14` clean.

oluies (approved):

PR #330, opened 2026-09-09
fix(#317): a metadata load that fails mid-query no longer mutates the cache

### LINKED ISSUE #317: Catalog-wide metadata cache can be left empty-but-LOADED if the bulk query throws (spec 071 regression)
Found by a roborev review of `8901cc4` (#312, spec 071); job 1474. Verified against `origin/main` before filing.

## The defect

`LoadAllSchemasMetadata` (`src/catalog/mssql_metadata_cache.cpp:787-940`) clears each schema's table map from inside the row callback:

```cpp
schema.tables.clear();          // inside the ExecuteMetadataQuery row callback
```

but only re-publishes the load state **after** the query returns:

```cpp
pair.second.tables_load_state = CacheLoadState::LOADED;   // after ExecuteMetadataQuery
```

The reset lambda covers the **retry** path only. A non-retryable throw — query timeout, connection reset, any TDS error — propagates out of the function leaving already-touched schemas **emptied while `tables_load_state` still holds its previous `LOADED`**.

`EnsureTablesLoaded` then short-circuits without checking emptiness (`:1383`):

```cpp
if (schema.tables_load_state == CacheLoadState::LOADED && !IsTTLExpired(schema.tables_last_refresh, ttl_seconds_)) {
    return;                     // no emptiness check
}
```

so those tables **silently report as missing** until TTL expiry or an explicit `mssql_invalidate_cache()`. No error, no retry — the catalog just stops seeing tables it saw a moment ago.

## Why spec 071 made it worse

This is the same failure the issue #178 review fixed for `Refresh()` at `:1231-1234`. The difference is blast radius: the per-schema loop could only damage the schema it was loading. The single whole-catalog query means **a failure while listing schema `A` also wipes `B`, `C` and `D`** — every schema the callback had touched before the throw.

With `mssql_metadata_timeout` at its 300 s default and a large catalog, a timeout mid-load is not exotic.

## Fix

Either of:

1. Stage into a local `unordered_map<string, MSSQLSchemaMetadata>` and swap into `schemas_` only once the query has completed. Strictly better — nothing is destroyed until there is a replacement.
2. Wrap the body in `try/catch(...)` that sets `tables_load_state = NOT_LOADED` for every entry in `touched` before rethrowing.

(1) also removes the window where a concurrent reader sees a half-populated cache.

## Test worth having either way

An integration test that makes the bulk query fail mid-flight — a short `mssql_metadata_timeout` against a catalog large enough to exceed it — and then asserts the tables are still visible on the next query rather than silently absent.

## Related, from the same review

Not filed separately since they are in the same function's blast radius, but recorded here:

- **`BulkLoadAll`'s empty-`schema_name` branch returns before the tail that sets `schemas_load_state_` / `state_` / `last_refresh_`**, so `mssql_preload_catalog('db')` no longer marks the catalog loaded — the next listing pays a schema-discovery round trip, defeating the documented one-round-trip contract, and `IsExpired()`/`GetState()` keep reporting not-loaded. `BULK_METADATA_ALL_SQL` also inner-joins `sys.columns`, so a schema with no objects is dropped entirely where the old `SCHEMA_DISCOVERY_SQL` pass registered it.
- **`LoadAllTableMetadataForSchema` has no callers.** ~130 lines carrying a third copy of the row parse, and its doc comment says *"Caller must hold `mutex_` via the public entry"* while the body takes `std::lock_guard<std::mutex> lock(mutex_)` at `:664` — a caller that follows the documented contract self-deadlocks on the non-recursive mutex.

VGSML:
Closes #317. Thanks @oluies — the defect is real and the fix is worth having. **The severity is not what the issue says, though**, and the evidence is below rather than an assertion.

## The fix

`LoadAllSchemasMetadata` cleared each schema's table map from **inside** the row callback and only re-published `tables_load_state` after the query returned. A non-retryable throw in between left every schema the callback had touched emptied while still carrying its previous load state; the reset lambda covered the 1205 retry path only.

Staged, per your option (1): the callback fills a local map and nothing reaches `schemas_` until the query has returned. Table **maps** are staged rather than whole `MSSQLSchemaMetadata` values, so a schema keeps its other fields instead of being rebuilt from nothing. The retry-reset lambda drops to clearing the staging area — there is no longer anything published to undo.

## The symptom does not reproduce, on either build

*"those tables silently report as missing"* — I could not make that happen, before or after the fix, with a deterministic mid-load failure. Two reasons, both checkable:

- the **live** listing path, `LoadAllTableMetadata`, already tests `!schema.tables.empty()` before trusting a `LOADED` state (`mssql_metadata_cache.cpp:620`);
- the path that does not — `EnsureTablesLoaded` — is reachable only from `GetTableNames`, and `grep -rn "GetTableNames" src/` finds **no callers**. Which reinforces your own third observation about dead code in this file.

So: the mutation-on-throw is real and happens; what keeps it invisible is a guard on a different layer plus a dead function. Neither is stated where `LoadAllSchemasMetadata` can see it, and both could move — which is exactly why holding the invariant here rather than relying on downstream is worth doing. The code comment now says so, so that nobody "simplifies" the staging away after finding the same reassurance I did.

## `mssql_test_fail_metadata_after_rows` (test-only, default 0)

Added because the invariant was otherwise untestable, and I did not reach for it first. What failed:

- **A short `mssql_metadata_timeout`.** Does not fire: the bulk query runs in **0.78 s at 2039 tables / 15002 columns**, and the setting's floor is 1 second. Reaching it needs a fixture no test can build.
- **Killing the session from another connection.** Lands inside a ~0.5 s window; hits whichever query is running (`SHOW ALL TABLES` uses the *same* SQL as the preload, so the killer cannot tell them apart); and poisons the pooled connection, so the next statement fails for an unrelated reason **inside `EnsureSchemasLoaded`** — which clears `schemas_` and rebuilds everything, masking the very thing under test.

Three separate false readings came out of that before the lever existed, including one where I had warmed the cache through `GetTableMetadata`, which does not set `tables_load_state` at all. The lever is what let me falsify the severity claim rather than confirm it by construction.

Off costs nothing: `ExecuteMetadataQuery` passes the row callback through **unwrapped**, so the row loop carries no test. The seam is the thin wrapper, not the parse.

## Not in this PR

Your two related findings, both confirmed and neither touched:

- **`BulkLoadAll`'s empty-`schema_name` branch returns before the tail that sets `schemas_load_state_` / `state_` / `last_refresh_`.** Worth knowing before anyone fixes it in isolation: that missing state-set is currently what keeps **empty schemas visible**, because the next listing falls through to schema discovery. Setting the state without also fixing the `INNER JOIN sys.columns` that drops object-less schemas would make them disappear. The two halves have to land together, and the join change perturbs a query spec 071 measured — so it wants its own PR with its own numbers.
- **`LoadAllTableMetadataForSchema` has no callers**, and its doc contract (*"caller must hold `mutex_`"*) contradicts its body, which takes the lock — a caller following the docs self-deadlocks. Deleting it is right; it is unrelated surgery to a bug fix.

Happy to do either next.

Full SQL suite **162 cases / 5060 assertions**; 19 standalone C++ tests; `clang-format-14` clean.


VGSML:
Self-review, in `f85a914`. Two findings, both about my own work, and the first one matters enough that I would rather post it than let it pass.

## The regression test does not discriminate

**It passes on the pre-fix build.** Verified, not assumed — I spliced the injection lever into `origin/main`'s cache, built, and ran the committed test: 21 assertions, green. So `test/sql/regression/issue_317.test` is not a regression test for #317 and must not be read as one; the header now says that outright.

Chasing *why* produced a better explanation of the defect than the one in my own PR description, which stopped at "there is a guard downstream":

> The same deferral that leaves a schema emptied **also** defers publishing its tables' `columns_load_state`. So a touched schema comes out of a failed load either **empty** or holding tables whose columns say `NOT_LOADED` — and `LoadAllTableMetadata` rejects both (`all_columns_loaded && !tables.empty()`) and reloads. The one reader that would trust the broken state, `EnsureTablesLoaded`, is reachable only from `GetTableNames`, which has no callers.

That is a stronger claim than the one I made. The old code was safe because **three deferrals on layers that do not know about each other happened to line up**. The new code is safe because this function does not publish until it has an answer. Both the test header and the code comment now carry that, so the next reader finds the reassurance *and* the reason it is not load-bearing, rather than only the first half and a temptation to revert.

What the test pins is the **contract** a user can see — a failed metadata load leaves the catalog exactly as usable as it was, and a later load still works — and it begins discriminating the moment someone removes the downstream guard or gives `GetTableNames` a caller.

## Staging costs one extra copy: ~2 MB per 16K columns

Measured rather than reasoned, since "strictly better" deserves a number:

| scenario | pre-fix (in place) | staged |
|---|---|---|
| cold preload, 2039 tables / 16202 columns | 39.27 / 39.47 MB | 39.31 / 39.44 MB |
| **warm** cache, then 3 reloads | 67.7 / 69.3 MB | 70.3 / 70.8 MB |

Cold shows nothing, which makes sense — there is nothing to hold onto. The cost appears on a **reload**, where the old map is kept until the new one is complete: roughly **+2 MB per 16K columns**, scaling with the catalog. Recorded next to the code that pays it, so it is a known trade rather than a surprise on a 200K-object catalog.

## Also checked, and correct as written

`by_object_id` holds pointers into `staged`'s nodes. Rehashing a `std::unordered_map` invalidates iterators but **not** references to elements, so both the outer map and the inner table maps are safe to grow under those pointers; and the sort loop runs, then `by_object_id` is cleared, before the move that would dangle them.

---

SQL suite **162 cases / 5060 assertions**; 19 standalone C++ tests; `clang-format-14` clean.

src/catalog/mssql_metadata_cache.cpp
```diff
@@ -1292,9 +1323,32 @@ int MSSQLMetadataCache::GetMetadataTimeoutMs() const {
...
+	// that causes one in the wild — a metadata timeout, a reset connection, a
+	// killed session — arrives from outside and cannot be asked for from SQL.
+	// Off is the normal case and is free: the callback goes through unwrapped,
+	// so the row loop carries no test.
+	const int64_t fail_after = test_fail_after_rows_;
+	if (fail_after > 0) {
+		auto seen = make_shared_ptr<int64_t>(0);
+		auto inner = std::move(callback);
+		callback = [inner, seen, fail_after](const vector<string> &values) {
+			if (++(*seen) > fail_after) {
```
oluies: **Low — off-by-one: the lever needs N+1 rows to fire, so it silently no-ops on a result set of exactly N.**

`if (++(*seen) > fail_after)` throws on row N+1, not row N. The header contract says "throw once it has parsed this many rows". With a fixture that returns exactly `fail_after` rows the query simply succeeds, and a test written around it degrades into a no-op that still passes its non-error assertions — silently, and it gets more likely as a fixture shrinks.

```cpp
inner(values);
if (++(*seen) >= fail_after) {
    throw IOException(...);
}
```

That fires whenever the result set has at least N rows, which is what the doc comment promises.

src/catalog/mssql_metadata_cache.cpp
```diff
@@ -1292,9 +1323,32 @@ int MSSQLMetadataCache::GetMetadataTimeoutMs() const {
...
 											  MSSQLMetadataCache::MetadataRowCallback callback,
 											  MSSQLMetadataCache::MetadataResetCallback reset) {
+	// Issue #317: the only deliberate way to a mid-query failure. Everything
+	// that causes one in the wild — a metadata timeout, a reset connection, a
+	// killed session — arrives from outside and cannot be asked for from SQL.
+	// Off is the normal case and is free: the callback goes through unwrapped,
+	// so the row loop carries no test.
+	const int64_t fail_after = test_fail_after_rows_;
+	if (fail_after > 0) {
+		auto seen = make_shared_ptr<int64_t>(0);
```
oluies, line 1375: **Low — `seen` is not reset on the deadlock-victim rerun, so the lever is non-deterministic on exactly the retry path this code cares most about.**

`RunMetadataQuery` calls the caller-supplied `reset` and re-runs the query from the top (PR #308). `reset` is passed through this wrapper untouched and knows nothing about `seen`, so on attempt 2 the counter still carries attempt 1's rows and the injected failure lands earlier than the setting asked for — or immediately.

Test-only, but the retry path is where a partial-load bug is most likely to be reasoned about, so it is worth having the lever behave the same on both attempts. Wrap `reset` alongside `callback` when `fail_after > 0`, zeroing `*seen` before delegating.

src/catalog/mssql_metadata_cache.cpp
```diff
@@ -809,12 +809,41 @@ void MSSQLMetadataCache::LoadAllSchemasMetadata(tds::TdsConnection &connection,
...
+	// LoadAllTableMetadata rejects both (`all_columns_loaded &&
+	// !tables.empty()`) and reloads. The one reader that would trust the broken
+	// state, EnsureTablesLoaded, is reachable only from GetTableNames, which has
+	// no callers.
+	//
+	// So the old code was safe by three deferrals happening to line up, on
+	// layers that do not know about each other. This one is safe because this
+	// function does not publish until it has an answer. The cost is one extra
+	// copy of the table metadata for the duration of a load: measured +~2 MB
+	// peak RSS on a warm reload of 2039 tables / 16202 columns.
```
oluies: **Low — the recorded cost reads as a bound, but it is one data point at ~1% of this function's design scale.**

"+~2 MB peak RSS on a warm reload of 2039 tables / 16202 columns" is measured honestly, but the comments elsewhere in this same file size the target at 200K objects (lines 78 and 634). The staged copy is proportional to catalog size and is held while `mutex_` is held, so at the documented target scale the transient is on the order of hundreds of MB, not 2 MB.

Worth phrasing as proportional — "one extra copy of the table metadata, ~2 MB at 2039 tables; it scales with the catalog" — so the next reader does not take +2 MB as the ceiling. No objection to the trade-off itself.

test/sql/regression/issue_317.test
```diff
@@ -0,0 +1,130 @@
+# name: test/sql/regression/issue_317.test
+# description: a metadata load that fails mid-query leaves the cache exactly as it was (issue #317)
```
oluies, line 2: **Low — no discriminating coverage ships with the fix.**

The header is admirably honest that this passes on the pre-fix build, and the argument for why (the corrupted state is unobservable for `LoadAllSchemasMetadata`) is correct as far as it goes. But it is not correct for the sibling path in the Medium finding: there an emptied-but-`LOADED` column list **is** observable, through `duckdb_columns()` or a plain `SELECT *`.

So the discriminating test does exist — it just belongs to the `BulkLoadAll(schema)` path. Once that is staged: load one table's columns, arm the lever, fail `mssql_preload_catalog('i317','i317s')`, then assert the table's `column_count` is still 2. That case fails on the pre-fix build and passes after.

oluies (commented):
Automated review (roborev job 1489, `claude-code`) over `5460b8f..f85a914`. Five findings, one Medium and four Low, posted inline. I re-verified the Medium against the head commit before posting — the reviewer's line numbers were ~30 low, the corrected ones are in the comment.

The staging rewrite in `LoadAllSchemasMetadata` itself checks out: `by_object_id`'s pointers into `staged`'s nodes stay valid across rehash, the sort loop runs and `by_object_id` is cleared before the `std::move` that would dangle them, and the empty-schema pass still sees freshly published schemas through `staged.find`. Both self-safety claims in the PR body hold.

The Medium is that the same defect is left in `BulkLoadAll`'s per-schema branch, where — unlike #317 — it is genuinely observable.

src/catalog/mssql_metadata_cache.cpp
oluies, line 1: **Medium — the sibling loader still mutates the cache from inside the row callback, and there the corruption IS observable.**

The fix is scoped to `LoadAllSchemasMetadata`. `BulkLoadAll`'s non-empty-`schema_name` path — where `mssql_preload_catalog(ctx, 'schema')` routes — still writes into `schemas_` from the callback, and one of those writes is not covered by the "three deferrals happen to line up" argument the new comment rests on.

Line numbers on `f85a914` (the review's were ~30 low):

- `mssql_metadata_cache.cpp:1082` — `auto &tables = schema.tables;`, where `schema` is `schema_it->second` in the **live** `schemas_`.
- `mssql_metadata_cache.cpp:1116` — `table_it->second.columns.clear()` on a table that already exists in the cache, i.e. one whose `columns_load_state` is already `LOADED` from an earlier single-table load.
- `mssql_metadata_cache.cpp:1180-1182` — the publication that would refill it, **after** the query returns.

A throw between 1116 and 1180 leaves that table with an empty `columns` vector *and* `columns_load_state == LOADED`. The guard at `mssql_metadata_cache.cpp:620` (`all_columns_loaded && !schema.tables.empty()`) then sees a fully-loaded schema and never reloads — so it does not save this path the way it saves #317.

Reproducer using this PR's own lever:

```sql
SELECT * FROM ctx.sch.t;                      -- loads t's columns, columns_load_state = LOADED
SET mssql_test_fail_metadata_after_rows = 1;
SELECT mssql_preload_catalog('ctx','sch');    -- throws after t's group cleared its columns
-- t is now cached as a zero-column table for the rest of the session
```

The reset lambda at 1150-1165 does not cover it either: it deliberately does not clear, and its comment argues the rerun's new-table branch will re-clear — which is true for the *retry* path but says nothing about a non-retryable throw.

`LoadAllTableMetadataForSchema` has the identical unstaged shape and is worse (it clears `schema.tables` before the query is even sent), but it has no callers — dead code that reintroduces the bug the moment it is wired up. The PR body already flags it as out of scope, which is fine; this comment is only about `BulkLoadAll`.

**Suggested fix:** apply the same staging pattern to the per-schema branch — accumulate into a local `unordered_map<string, MSSQLTableMetadata>`, merge and publish only after `ExecuteMetadataQuery` returns.

VGSML:
All five in `a2339b1`. The Medium is right, and it corrects my severity analysis in the **opposite** direction from my own self-review — which had concluded the defect was latent everywhere. It is not.

## Medium — reproduced before fixing

`BulkLoadAll`'s per-schema branch, with this PR's own lever:

```sql
SELECT mssql_preload_catalog('m','i317s');     -- succeeds; schema LOADED
SET mssql_test_fail_metadata_after_rows = 1;
SELECT mssql_preload_catalog('m','i317s');     -- throws after A's clear
SELECT * FROM m.i317s.A;                       -- ONE column, not two
```

`duckdb_tables()` agreeing at `column_count = 1`, for the rest of the session.

Your diagnosis of *why* it differs from #317 is the whole thing: `columns.clear()` on a table **already in the cache and already `LOADED`**. So the "three deferrals happen to line up" argument I leaned on does not cover it — the table comes out of a failed load non-empty *and* claiming to be complete, which is exactly the state `LoadAllTableMetadata`'s guard trusts.

Staged the same way, but **per table rather than per map**, because this publication merges: the query may not cover every table the schema legitimately holds (one excluded by `table_filter`, one loaded singly), and replacing the map would drop those — the reasoning your original reset-lambda comment already recorded. Table counting still reports tables *new to the cache*: staging moved the lookup off `schema.tables`, so it asks that map directly instead of inferring it.

## This is the discriminating test I did not have

Added, and **verified failing on the un-staged build** — 1 of 30 assertions.

The setup is not the obvious one, so the test says so out loud. It needs the first preload to **succeed**; it must not be followed by a listing, because that builds DuckDB table **entries** which then answer `duckdb_columns()` from their own copy and hide the cache underneath; and the table's first touch has to come **after** the failed load. Two earlier attempts of mine passed for exactly those reasons before I worked out why.

## Low ×3

- **N+1 rows.** Right, and the failure mode you name is the bad one: a fixture returning exactly N made the lever a silent no-op that still passed everything written around it. Now `inner(values)` then `>= fail_after`.
- **`seen` not reset on the rerun.** Also right, and worse than cosmetic — that is the path a partial-load bug actually gets reasoned about. The reset is wrapped alongside the callback and zeroes the counter before delegating.
- **The memory figure read as a ceiling.** Fair: +2 MB is one point at ~1% of the 200K-object scale this file is written for. Phrased as a rate now, with the target scale named.

## Still out of scope, unchanged

`BulkLoadAll`'s empty-`schema_name` early return and `LoadAllTableMetadataForSchema`. The first still needs its two halves landing together — the missing state-set is currently what keeps empty schemas visible — and the `INNER JOIN sys.columns` change perturbs a query spec 071 measured, so it wants its own numbers.

---

Thanks for re-verifying the line numbers against the head commit before posting; that saved a round trip.

SQL suite **162 cases / 5069 assertions**; 19 standalone C++ tests; `clang-format-14` clean.

oluies (approved):

PR #332, opened 2026-09-10
fix(#323): RETURNSTATUS is fixed-length — every stored-procedure call desynced the parser

### LINKED ISSUE #323: TDS parser desyncs on RETURNSTATUS: any stored-procedure call corrupts the connection
## What happens

Calling a stored procedure through `mssql_exec` (or scanning one through `mssql_scan`) leaves the connection's token stream misaligned by two bytes. The next token the parser meets is then read at the wrong offset, and the query fails with:

```
IO Error: TDS parse error: Unknown token type: 0x0
```

The connection is not recoverable at that point — everything after it on the same connection fails or times out.

## Why

`TokenType::RETURNSTATUS` (0x79) is handled in the group of tokens that carry a two-byte length (`src/tds/tds_token_parser.cpp:155-172`):

```cpp
case TokenType::ORDER:
case TokenType::RETURNSTATUS:
case TokenType::RETURNVALUE:
case TokenType::LOGINACK:
case TokenType::TABNAME:
case TokenType::COLINFO:
    // Skip these tokens - they have a 2-byte length
    ...
    size_t token_end = static_cast<size_t>(token_length) + 3;
    ConsumeBytes(token_end);
```

Per [MS-TDS] 2.2.7.16, RETURNSTATUS is a **fixed-length** token: the type byte followed by a 4-byte return value, five bytes in total, with no length field. A procedure returning 0 puts `79 00 00 00 00` on the wire; the parser reads `00 00` as the length, consumes three bytes, and leaves two behind.

## The evidence, byte for byte

`MSSQL_DEBUG=1`, immediately after an `EXEC` of a procedure that does DML:

```
[TDS PARSER] Unknown token 0x00 at pos=419, buffer_size=434, available=15,
             hex: 00 00 fe 00 00 e0 00 00 00 00 00 00 00 00 00
```

The two leftover bytes are the tail of the return value; the real next token starts at +2 and is `fe` — DONEPROC, `status=0x0000`, `curcmd=0x00e0`, `rowcount=0`. Exactly the two-byte shift the skip above produces.

## Repro

Any procedure whose response is long enough to reach another token after the desync:

```sql
LOAD mssql;
ATTACH '<dsn>' AS srv (TYPE mssql);
SELECT mssql_exec('srv', 'CREATE OR ALTER PROCEDURE dbo.p AS BEGIN SET NOCOUNT ON; INSERT INTO dbo.t SELECT ...; END');
SELECT mssql_exec('srv', 'EXEC dbo.p');   -- desyncs; the next statement on this connection fails
```

A trivial procedure can appear to work: the two stray bytes are only fatal once a token follows them.

## Suggested fix

Give 0x79 its own arm — consume five bytes unconditionally — and leave `RETURNVALUE` (0xAC) where it is, since that one does carry a length. Worth a parser unit test with a captured `79 00 00 00 00 fe …` stream.

## Where it came from

Building the server-side commit for a DuckLake catalog on SQL Server ([hugr-lab/mssql-ducklake](https://github.com/hugr-lab/mssql-ducklake) spec 005): the commit is meant to be one call to a stored procedure. It is worked around there by sending the procedure's body as a plain batch, which produces no RETURNSTATUS — so this is not blocking, but it does mean procedures are effectively unusable through the extension today.

VGSML:
Closes #323. Spec: `specs/072-token-parser-fixed-length/spec.md` — written as reconnaissance first, then implemented in the same PR.

Filed as one bug in one token. The survey found one **group**: six token types handled by one arm on the assumption that each is `type(1) + length(2) + data`, in which two do not fit, one is registered under the wrong byte, and the parser's two consumers fail in opposite ways — the scan path loudly, the `mssql_exec` path **silently, reporting success**.

## W1 — RETURNSTATUS (0x79) is five fixed bytes

[MS-TDS] 2.2.7.16: `TokenType + Value(LONG)`, no length field. The arm read the low two bytes of the *value* as a length. Captured, `RETURN 7`:

```text
79 07 00 00 00                              RETURNSTATUS, value 7
fe 00 00 e0 00  00 00 00 00 00 00 00 00     DONEPROC
```

reads `07 00` = 7, consumes 10 — the five of RETURNSTATUS **and the first five of DONEPROC** — and lands on the rowcount. A status of 0 leaves two bytes behind instead, which is the issue's dump byte for byte. So the diagnosis in the issue is right, and the non-zero case is worse: a return value of *n* eats *n* bytes of whatever follows.

## F2 — the finding the issue did not have: `mssql_exec` lied

`mssql_scan` threw `TDS parse error: Unknown token type: 0x0` and the pool discarded the connection — loud, the failure in the issue.

`MSSQLSimpleQuery` had **no check for `ParserState::Error`**. A dead parser answers `NeedMoreData`, the loop exits, the code drains to EOM and then *forces done with success*. Every token after RETURNSTATUS was dropped. Measured:

| via `mssql_exec` | result |
| --- | --- |
| `EXEC dbo.I323N` (NOCOUNT OFF, 1-row INSERT) | `1` — **correct**, DONEINPROC precedes RETURNSTATUS |
| `EXEC dbo.I323R; RAISERROR('after', 16, 1)` | **`0`, no error** |
| `RAISERROR('control', 16, 1)` alone | `SQL Server error 50000: control` |

The common shape reported the right row count and swallowed the desync — that is how this survived. The issue's *"not recoverable"* is the scan path; the exec path was recoverable and **wrong**, the worse property.

**W4** — the exec path now records a parser error as a failure, *after* the drain to EOM that keeps the socket clean, so the connection goes back to Idle rather than being discarded; a SQL Server error already recorded keeps precedence. This is a behaviour change on purpose: any *other* latent desync on this path becomes an error where it was a success with dropped tokens.

## W2, W3, W5 — the rest of the group

- **TABNAME = 0xA4**, not 0x04. Captured from `FOR BROWSE`: `a4 15 00 02 03 00 d·b·o· 05 00 I·3·2·3·T· a5 03 00 01 01 00 d1 …` — it *does* carry a 2-byte length and *does* belong in the group; only the enum byte was wrong, and the arm was never reached. Any `FOR BROWSE` result failed with `Unknown token type: 0x164`.
- **RETURNVALUE (0xAC) leaves the group** and fails by name. No length field either (2.2.7.17) — only knowable by parsing `TypeInfo`. Also **unreachable**: sent only for RPC, which this extension does not issue; verified, an OUTPUT parameter over `SQL_BATCH` comes back as an ordinary row (`d1 04 2a 00 00 00` = 42). A named error beats a silent mis-skip if that ever changes, and that arm is where TypeInfo parsing goes if RPC is added.
- **Hex in the message.** `"0x" + std::to_string(byte)` printed decimal: `0x164` above is 0xA4. The issue's `0x0` is the same in both bases.

## Verified failing first

`test/cpp/test_token_parser_tokens.cpp` — server-free, in `STANDALONE_TEST_SOURCES` so it gates a PR, every stream a live capture — **fails 18 checks on `main`'s parser**: RETURN 0/7/1/256/−1, the token split across packets, a procedure with rows, FOR BROWSE, RETURNVALUE, the hex message. Passes with the fix.

`test/sql/regression/issue_323.test` on `main` fails at the **F2 case** — the swallowed RAISERROR — which is the assertion that will not regress quietly. Everything before it *passes on main*, including the DML `EXEC` and its follow-up on one pinned connection, because the exec path hid the desync.

Two expectations I got wrong writing it, both caught by running it: `mssql_exec('SELECT 1')` returns 1 (DONE's row count), and a NOCOUNT-OFF procedure inserted a fourth row that the SELECTing procedure then returned.

`fuzz/corpus/tds_tokens/returnstatus_doneproc` — the RETURN 7 capture with the leading mode byte the README requires. On `main` this is a wrong parse, not a crash, which the fuzzer cannot see: the unit test is the gate, the seed is coverage under mutation.

## Not in this PR

Parsing RETURNVALUE (needs RPC), and a token-length-class self-check from [MS-TDS] 2.2.4.1 — it would have flagged W1, but DONE class-decodes to 8 bytes and is 13, and RETURNVALUE class-decodes to "USHORT length" and has none: two exceptions in a twenty-entry table is a trap, not an invariant (spec § 3).

SQL suite **163 cases / 5103 assertions**; 20 standalone C++ tests; `clang-format-14` clean.


VGSML:
Self-review, in `HEAD`. The mechanics held; two things around them did not, and three claims are now verified rather than assumed.

## Found and fixed

**W4 handed parse errors to a formatter that blames the server.** `mssql_exec` renders every failure as `MSSQL execution error: SQL Server error %d: %s`, and `error_number` is 0 for anything the server did not say — a query timeout before, a TDS parse error now. *"SQL Server error 0: TDS parse error"* blames the wrong party. `error_number == 0` now formats without the prefix. No test pinned the old text, so nothing else moves.

**Where `0x04` came from.** `docs/tds-protocol.md`'s token table had no TABNAME row, so the enum byte had nothing to be checked against — and line 92 of the same file is `TABULAR_RESULT | 0x04`, the **packet** type. The table now carries TABNAME, COLINFO and RETURNVALUE with their real layouts, RETURNSTATUS marked fixed-length, and both traps written next to them, including why the [MS-TDS] 2.2.4.1 length class is deliberately not used as a self-check.

## Verified, correct as written

- **The exec loop cannot spin.** The RETURNVALUE arm returns `None`, and `MSSQLSimpleQuery`'s loop treats `None` as "loop again" — but `TryParseNext` answers `NeedMoreData` for any call once `state_ == Error`, so the second call breaks out. Same contract the `default` arm has always relied on.
- **W4's blast radius on the catalog.** `RunMetadataQuery` → `ExecuteWithCallback`, and `!success` there is already `throw IOException("Metadata query failed: %s")`. So a parse error in a metadata query becomes an error where it used to be a silently truncated result — the widening spec 072 § 4 predicted, and the correct answer.
- `<cstdio>` is included for `snprintf` rather than reached transitively.

SQL suite **163 cases / 5103 assertions**; 20 standalone C++ tests; `clang-format-14` clean.

oluies (commented):
## roborev review — `fix/323-returnstatus-token` (2 commits, `5fadf8a..73e6da3`)

**Verdict: Fail** — 1 Medium, 2 Low. Job `1518`, agent `claude-code`.

The protocol work is correct and well evidenced. RETURNSTATUS is re-classified as a fixed five-byte token against [MS-TDS] 2.2.7.16 with a live capture to prove the old arm consumed ten; RETURNVALUE leaves the length-prefixed group for a named refusal rather than a silent mis-skip; the TABNAME enum byte is corrected to `0xA4` with a `FOR BROWSE` capture behind it; and the decimal-after-`0x` message bug is fixed. **F2 is the real find** — `MSSQLSimpleQuery` had no `ParserState::Error` check at all, so `mssql_exec` drained to EOM and forced success while dropping every token after the desync. Reporting the right row count in the common shape is exactly how that survived. The server-free byte-level test in `STANDALONE_TEST_SOURCES` is the right gate for it.

The findings are all in the **supporting** material, not the parser:

- The live regression test's connection-pinning premise **is not actually enforced** — `SET mssql_connection_limit = 1` runs after `ATTACH`, and pool config is captured at ATTACH. The one assertion that would catch a returning desync is the one not pinned.
- The `"a parse error is not SQL Server error 0"` fix landed at one of the **two** call sites that format it.
- The test reads a `mssql_exec`-created table through the catalog with no invalidation, against the repo's own convention in three other files.

<sub>Posted by `roborev` via Claude Code. Line anchors verified against `73e6da3`; the reviewer's citation of `issue_323.test:47-49` is corrected to 50-52 here.</sub>

test/sql/regression/issue_323.test
```diff
@@ -0,0 +1,173 @@
...
+
+statement ok
+SELECT mssql_exec('i323', 'CREATE OR ALTER PROCEDURE dbo.I323O @x INT OUTPUT AS BEGIN SET NOCOUNT ON; SET @x = 42; END');
+
+statement ok
+SELECT mssql_exec('i323', 'CREATE OR ALTER PROCEDURE dbo.I323N AS BEGIN INSERT INTO dbo.I323T VALUES (4); END');
+
+# One connection, so every statement below reuses the one the EXEC ran on.
+statement ok
+SET mssql_connection_limit = 1;
```
oluies: **Medium — the pool limit is set after `ATTACH`, so the test's central premise is not actually imposed.**

`LoadPoolConfig()` is read **once, at ATTACH time** (`src/mssql_storage.cpp:1652` → `:1726`), and `ConnectionPool` never re-reads `connection_limit` afterwards — `src/tds/tds_connection_pool.cpp:172` uses the captured `config_`. `ATTACH` is at line 24; this `SET` lands at line 52, twenty-eight lines too late.

So the comment on line 50 — *"One connection, so every statement below reuses the one the EXEC ran on"* — describes a constraint this file does not create. The pool is running with the default limit of **64**, and the three `total_connections, idle_connections = 1 1` assertions hold only **incidentally**, because `TableScanGlobalState::MaxThreads()` happens to return 1.

That matters more here than it usually would: the property under test is *desync-on-statement-N+1*, which only exists if statement N+1 lands on the same physical connection as the `EXEC`. Right now nothing pins that. A future change that opens a second connection anywhere in this flow does not fail as *"the desync came back"* — it fails as a pool-stats mismatch with a misleading message, and the regression this PR exists to prevent silently stops being covered.

**Fix**: move `SET mssql_connection_limit = 1;` **above** the `ATTACH` on line 24, so the pool is constructed with the limit.

src/mssql_functions.cpp
```diff
@@ -498,6 +498,12 @@ static void MSSQLExecExecute(DataChunk &args, ExpressionState &state, Vector &re
 			ConnectionProvider::ReleaseConnection(client_context, catalog, std::move(connection));
 
 			if (!query_result.success) {
+				// error_number 0 is not a server error -- it is a query timeout or,
+				// since issue #323, a TDS parse error the exec path used to swallow.
+				// "SQL Server error 0: TDS parse error" blames the wrong party.
+				if (query_result.error_number == 0) {
+					throw InvalidInputException("MSSQL execution error: %s", query_result.error_message);
+				}
```
oluies: **Low — this fix has a twin that did not get it.**

The second commit's reasoning is right and worth applying twice: `error_number == 0` means the server never spoke, so `"SQL Server error 0: ..."` blames the wrong party. But there are **two** call sites that format `MSSQLSimpleQuery::Execute`'s failure this way, and only this one was changed:

```
src/mssql_functions.cpp:508      MSSQL execution error: SQL Server error %d: %s   <- fixed here
src/catalog/mssql_catalog.cpp:1083  MSSQL DDL error: SQL Server error %d: %s      <- unchanged
```

The catalog DDL path runs the same `Execute` and inherits the same widened failure surface this PR deliberately opened (W4). A TDS parse error there now surfaces as:

    MSSQL DDL error: SQL Server error 0: TDS parse error: ...

— exactly the attribution the new branch above was added to prevent. The same applies to the pre-existing `error_number = 0` producers: query timeout and socket failure.

(`mssql_catalog.cpp` is outside this diff, which is why this is anchored here.)

**Fix**: apply the same `error_number == 0` branch at `mssql_catalog.cpp:1083` — or, since both sites now need an identical rule, factor the two into one shared formatter.

test/sql/regression/issue_323.test
```diff
@@ -0,0 +1,173 @@
...
+# whole response had fit one packet.
+# =============================================================================
+
+query I
+SELECT mssql_exec('i323', 'EXEC dbo.I323P');
+----
+0
+
+query I
+SELECT count(*) FROM i323.dbo.I323T;
```
oluies, line 79: **Low — this catalog-path read is not preceded by a cache invalidation, and passes by luck of ordering.**

`dbo.I323T` is created through `mssql_exec` at line 33, then read here through the catalog. `mssql_exec_invalidate_cache` defaults to **`false`**, so the catalog is never told the object exists; this works only because nothing has populated the lazy cache for schema `dbo` yet in this session. Reorder anything above it — or add an earlier catalog read — and this line starts failing for a reason that has nothing to do with issue #323.

Every other test in the repo that creates objects via `mssql_exec` and then reads them through the catalog refreshes first:

- `test/sql/insert/insert_types.test:49`
- `test/sql/insert/insert_server_defaults.test:33`
- `test/sql/integration/issue178_concurrent_same_table.test:41`

**Fix**: add `SELECT mssql_invalidate_cache('i323');` after the DDL block (line 48), before this read.

VGSML:
All three in `HEAD`. The parser work stood; these were the supporting material, and the Medium was a real hole in what the test proves.

**Medium — the premise was not imposed.** Right: the pool reads `connection_limit` once, at ATTACH, so the `SET` twenty-eight lines later changed nothing and the three `1 1` assertions held only because `TableScanGlobalState::MaxThreads()` is 1. The `SET` now precedes the `ATTACH`, with a comment saying why it must — a returning desync should fail as *"the desync came back"*, not as a pool-stats mismatch.

**Low — the twin.** `mssql_catalog.cpp`'s DDL path formatted the same failure the same way and inherited the same widened surface W4 opened. Rather than a second `error_number == 0` branch, one `SimpleQueryResult::DescribeError()` now serves both callers: it names the server only when the server spoke. A DDL rejected by the server still reads `MSSQL DDL error: SQL Server error 448: …`; a parse error there no longer says `SQL Server error 0`.

**Low — the catalog read.** `mssql_invalidate_cache('i323')` after the DDL block, per the convention in the three files you cited.

Re-run with the limit actually imposed: `issue_323.test` green; 20 standalone C++ tests.

src/query/mssql_simple_query.cpp
```diff
@@ -324,6 +324,24 @@ SimpleQueryResult MSSQLSimpleQuery::ExecuteWithCallback(tds::TdsConnection &conn
...
+		// issue #323 that was indistinguishable from "waiting for the next
+		// packet": the code below forced done and RETURNED SUCCESS, with every
+		// token after the desync dropped -- later result sets, later row counts,
+		// and a SQL Server ERROR token following a stored-procedure call, which
+		// is exactly the one a caller most needs. Record it as the failure it is.
+		// The drain to EOM still happens below, on purpose: that is what leaves
+		// the socket clean for the next statement on this connection, so the
+		// connection can go back to Idle rather than be discarded. A SQL Server
+		// error already recorded keeps precedence -- it says more than this does.
+		if (parser.GetState() == tds::ParserState::Error && result.success) {
```
oluies, line 343: **Low — the drain keeps feeding a dead parser, and this comment now blesses it**

This block makes "parser in `Error`, drain to EOM" a first-class, reported outcome, and the comment explicitly endorses the drain. That part is right — draining is what leaves the socket clean so the connection can return to `Idle` instead of being discarded.

The problem is what the drain does on the way. It still calls `parser.Feed(payload)` for every remaining packet (the `Feed` at ~`:236-240`), and `TokenParser` only reclaims memory from `CompactBuffer()`, which runs solely out of `ConsumeBytes()`. Once `state_ == Error` nothing is ever consumed, so `buffer_` grows by the **full remaining response** before the error is returned.

**Failure scenario:** a desync early in a large `mssql_exec` batch — or a garbled/hostile server that emits one bad token followed by a large message — buffers the entire tail in memory for nothing.

The behaviour predates this change, but W4 turns it from an accident into a supported path, which is why it is worth closing here.

**Fix:** skip the `Feed` once the parser is in `ParserState::Error`. Keep calling `ReceivePacket` until `is_eom` so the socket is still left clean, but stop accumulating bytes no one will parse.

oluies (commented):
## Automated review (roborev job 1537)

Range reviewed: `5fadf8a..1f4564b` — the full PR against `main`.

The core change is right: `RETURNSTATUS` (fixed five bytes) and `RETURNVALUE` (no length field, now refused by name) are correctly split out of the USHORT-length skip group, `TABNAME` is fixed to `0xA4`, and `MSSQLSimpleQuery` no longer reports a post-desync batch as a success. The byte-level test built from live captures plus the fuzz seed carrying the required mode byte is the right shape of coverage.

Two **Low** findings. One is inline below; the other is in a file this PR does not touch, so it cannot be anchored:

---

### Low — the hex-prefix/decimal-digits bug survives one file over

`src/tds/tds_column_metadata.cpp:87`

F5/W5 fixed `"hex prefix, decimal digits"` in the parser's unknown-token message, but `ColumnMetadata::GetTypeName()` still does:

```cpp
"UNKNOWN(0x" + std::to_string(type_id) + ")"
```

It reaches users through `tds_column_metadata.cpp:431` ("Unsupported SQL Server type: …") and `tds_row_reader.cpp:410` / `:906` ("Unsupported type in RowReader: …"). A TDS type byte of `0xAF` is reported as `UNKNOWN(0x175)`, sending the reader to look up a type id that does not exist.

These messages name a *type* byte the parser did not recognise — which makes them exactly the diagnostics someone will paste into the next issue like #323. Worth carrying the fix across in this PR while the context is loaded.

**Fix:** same treatment as the parser arm — `snprintf(buf, sizeof(buf), "UNKNOWN(0x%02X)", type_id)`, or `StringUtil::Format`.

---

Neither finding blocks the fix; both are follow-on polish on paths this PR legitimises.

src/query/mssql_simple_query.cpp
VGSML: Fixed in e36edcc: `Feed` is skipped once the parser is in `Error`; the packets are still read to EOM so the socket stays clean. Details in the PR comment.

VGSML:
Both Low findings from job 1537 are fixed in e36edcc; each was verified against the code before touching it.

**`ColumnMetadata::GetTypeName()` — hex prefix, decimal digits.** Confirmed: `"UNKNOWN(0x" + std::to_string(type_id)` at `tds_column_metadata.cpp:87`, reached through "Unsupported SQL Server type" and "Unsupported type in RowReader". Now `snprintf("UNKNOWN(0x%02X)")`, the same treatment as the parser's message. `test_token_parser_tokens.cpp` gains `TestUnknownTypeNameIsHex` (type byte `0x99` must print `UNKNOWN(0x99)` and must not contain `153`); it fails on the previous commit.

**The drain kept feeding a parser in `Error`.** Confirmed: `Feed()` only appends (`tds_token_parser.cpp:54-59`), `CompactBuffer()` runs from `ConsumeBytes()` alone, and nothing is consumed once `state_ == Error`, so the whole remaining response accumulated. `ExecuteWithCallback` now skips `Feed` when `parser.GetState() == ParserState::Error` and still reads every packet to EOM, so the socket is left clean exactly as before. No test for this one: the buffer growth of a dead parser is not observable without a fake server, and the mechanism is the three lines above.

`issue_323.test`, `mssql_exec.test` and the standalone C++ list (20/20) are green on the branch.

PR #333, opened 2026-09-10
fix(#302): resolve the Azure AD token when a connection is created, not when the catalog was (spec 073 + implementation)

### LINKED ISSUE #302: Attached Azure SQL catalog cannot reconnect after long-running DuckDB work
## Summary

I have successfully attached and interacted with SQL Server. Then I do long running work in DuckDB. After the long running work, I try to write those results to SQL Server, and get this error:
```text
IO Error: MSSQL UPDATE failed: batch 1: {"exception_type":"IO","exception_message":"MSSQL: Failed to acquire connection from pool (timeout)"}
```

## Environment

- DuckDB v1.5.5
- mssql-extension v0.2.4
- Windows
- Azure SQL Managed Instance
- Authentication through `AZURE_SECRET`
- One in-memory DuckDB connection

## Temporary configurations I have tried during failures
- `mssql_connection_timeout = 600`
- `mssql_acquire_timeout = 600`
- `mssql_query_timeout = 600`
- `mssql_connection_cache = true`
- `mssql_idle_timeout = 300`
- `mssql_min_connections = 0`
- `threads = 1`

## Observed behavior

It occurred on different statement types, including a scan, a join against attached MSSQL tables, remote `INSERT` and remote `UPDATE` statements.

In one run, `DETACH` followed by `ATTACH` immediately before a large `UPDATE` allowed the update to complete successfully.

My script often runs for more than 80 minutes.

I have Entity Framework and Dapper running through C#. Those also connect to SQL Server and I have never seen a connection pool error from them.

## Sample DuckDb Script
The full script is much larger, but the relevant pattern looks like this:

```sql
SET mssql_connection_timeout = 600;
SET mssql_acquire_timeout = 600;
SET mssql_query_timeout = 600;
SET mssql_idle_timeout = 0;
SET threads = 1;

ATTACH 'Server=<server>;Database=<database>'
    AS mssql_db (TYPE mssql, AZURE_SECRET azure_mssql_secret);

-- Read millions of source rows and materialize intermediate results in DuckDB.
CREATE TEMP TABLE LocalUniqueRecords AS
SELECT ImportRecordId, HashKey, PlanSponsorId
FROM read_parquet('<input files>');

-- Earlier attached-table queries work and leave an established connection in the pool.
CREATE TEMP TABLE LocalExistingRecords AS
SELECT input.ImportRecordId, remote.CommonRecordId
FROM LocalUniqueRecords input
JOIN mssql_db.import.RecordHashKeys remote
    ON remote.HashKey = input.HashKey;

-- Materialize new rows locally, then insert them into MSSQL through BCP.
CREATE TEMP TABLE LocalNewRecords AS
SELECT ImportRecordId, PlanSponsorId
FROM LocalUniqueRecords input
WHERE NOT EXISTS (
    SELECT 1
    FROM LocalExistingRecords existing
    WHERE existing.ImportRecordId = input.ImportRecordId
);

COPY LocalNewRecords TO 'mssql_db.import.CommonRecords' (FORMAT 'bcp', CREATE_TABLE false, FLUSH_ROWS 2000);

-- The real script performs additional local transformations and remote writes.
-- By the time it reaches the next query, the catalog may have been attached
-- for more than two hours.

-- This query references two attached MSSQL tables. The existing connection remains idle and open, but the pool attempts to create another connection.
CREATE TEMP TABLE LocalNewHashKeys AS
SELECT records.CommonRecordId, input.HashKey
FROM LocalUniqueRecords input
JOIN mssql_db.import.CommonRecords records
    ON records.ImportRecordId = input.ImportRecordId
LEFT JOIN mssql_db.import.RecordHashKeys existingHash
    ON existingHash.CommonRecordId = records.CommonRecordId
WHERE existingHash.CommonRecordId IS NULL;

-- Other phases apply locally staged changes to MSSQL. This kind of update can issue thousands of pool acquisitions when many rows are updated.
CREATE TEMP TABLE LocalRowsToUpdate AS
SELECT existing.CommonRecordId, input.PlanSponsorId
FROM LocalExistingRecords existing
JOIN LocalUniqueRecords input
    ON input.ImportRecordId = existing.ImportRecordId;

UPDATE mssql_db.import.CommonRecords remote
SET PlanSponsorId = changes.PlanSponsorId
FROM LocalRowsToUpdate changes
WHERE remote.CommonRecordId = changes.CommonRecordId;
```

The `LocalNewHashKeys` statement is representative of the statement that failed on September 5. Other failures have occurred during remote `COPY` and `UPDATE` operations.
It is not waiting for all 64 connections to become available: pool statistics show one idle connection immediately before the statement. The failure occurs when the query needs the pool to create an additional physical connection.

## 2026-09-05 reproduction

I added explicit `DETACH` / `ATTACH` refreshes. Then used `select * from mssql_pool_stats()` to record pool statistics around the long-running phases.

### Run 1 - In the failed run:

- The script started at 12:35:58.
- I refreshed the attached catalog at 13:06:43 with `mssql_idle_timeout = 0`.
- Immediately after the refresh, the pool had one idle connection: 
  `total_connections=1`, `connections_created=1`, and
  `connections_closed=0`.
- At 15:29:52, the pool still had one idle connection with no closed connections or acquisition timeouts.
- The next statement joined local DuckDB results against two attached MSSQL tables, requiring the pool to grow beyond its existing connection.
- At 15:30:23, approximately 31 seconds later, it failed with `MSSQL: Failed to acquire connection from pool (timeout)`.

That connection was requested approximately 2 hours 23 minutes after the catalog refresh.

### Run 2 - immediately reran the same workload and is succeeded:

- I refreshed the catalog at 15:42:55.
- The same two-table statement began at approximately 16:04:27, about 21 minutes after the refresh.
- It ran for approximately 78 minutes and completed successfully.
- After it completed, pool statistics showed two idle connections,
  `connections_created=2`, `connections_closed=0`, and
  `acquire_timeout_count=0`.
- A later refresh occurred immediately before an update of 1,286,009 rows.
  That update ran for approximately 33 minutes and completed after 2,583 pool acquisitions with no acquisition timeout.
- The complete script finished successfully after approximately 4 hours 53 minutes.

This appears consistent with the pool trying to create an additional physical connection after the token captured by `ATTACH` had aged, rather than all 64 pool slots being busy.

## Source analysis provided by GPT-5.6 Sol (High reasoning)

At `ATTACH` time, `src/mssql_storage.cpp` calls
`BuildFedAuthExtension()` and passes the resulting UTF-16LE token bytes to
`MSSQLCatalog`.

`src/catalog/mssql_catalog.cpp` stores those bytes in
`fedauth_token_utf16le_` and captures them by value in the `AZURE_AD` /
`MANUAL_TOKEN` connection factory:

```cpp
auto token = fedauth_token_utf16le_;
factory = [..., token, ...]() {
    // ...
    conn->AuthenticateWithFedAuth(database, token, encrypt, app_name);
};
```

This factory does not consult `TokenCache`, check expiration, or call
`AcquireToken()` when creating a replacement connection.

The repository already has token-refresh infrastructure:

- `TokenCache::GetToken()` rejects expired cached tokens.
- `FedAuthStrategy::IsTokenExpired()` checks cache validity.
- `AuthStrategyFactory::BuildTokenAcquirer()` obtains a fresh token when the
  cache has expired.

That infrastructure does not appear to be used by the catalog pool's
`AZURE_AD` factory.

There is also a diagnostic issue:

- The factory returns `nullptr` for both connection and authentication
  failures without preserving `TdsConnection::GetLastError()`.
- `ConnectionProvider` reports only a generic pool timeout instead of the
  underlying connection or authentication error.

Additionally, the pool factory calls `conn->Connect(host, port)` without
passing `pool_config_.connection_timeout`, so pool refills appear to use the
compiled 30-second default even when `mssql_connection_timeout=600`. The
approximately 31-second failure above is consistent with that default.

VGSML:
Closes #302. Spec 073 and its implementation in one PR — #335 carried the implementation and was merged into this branch, so the spec, the live reproduction and the code are reviewed together.

An attached Azure SQL catalog could not open a **new** connection once the script had run past an hour. The pool factory captured the FEDAUTH bytes at `ATTACH` and presented them for every refill; an Azure AD token lives **60 minutes**. Measured on the reporter's line (stock 1.5.5 + community 0.2.5, `acquire_timeout = 600`, 75-minute sleep): `Failed to acquire connection from pool (timeout)` after **600.3 s** and two login attempts, then `DETACH`/`ATTACH` and it worked.

## What the reconnaissance established (`specs/073-fedauth-token-refresh/spec.md`)

- **F1** — the factory holds the ATTACH-time token bytes forever. Token lifetime measured against Azure AD: 60 min. The report's timeline (21 min OK, 2 h 23 fail) is exactly that.
- **F2** — a failed creation was treated as an **exhausted pool**: `Acquire` waited for a `Release` that could not come, said "(timeout)", and discarded the reason the connection had recorded.
- **F3** — `mssql_connection_timeout` never reached a refill: no timeout to `Connect`, and eight login-phase reads with the constant spelled out.
- **F4** — the refresh path the issue points at (`BuildTokenAcquirer`) captures a `ClientContext` by reference and is **unreachable**: its only caller `CreateFedAuth` needs a context no live `Create` call site supplies. A pool factory outlives the ATTACH context, so wiring it up would be a use-after-free.
- **F5** — the reporter's "31 s" is one 30 s login read (F3) plus a deadline already passed — `acquire_timeout` was 30 in that session, not 600. The code with 600 waits 600.3 s, measured.
- **F7** — **Azure SQL does not say "expired"; it hangs up.** Inside the 600 s the gateway dropped the connection twice — after `LOGIN7`, after `FEDAUTH_TOKEN` — and never sent an ERROR token. Expiry has to be diagnosed on the client, from the token's own `exp`, before dialing.

## What changed

**W1 — the token is resolved when a connection is created.** The `AZURE_AD` factory holds the secret's **name**, the tenant and the `DatabaseInstance` — never the token, never the client secret — and calls a new `AcquireToken(DatabaseInstance &, …, allow_interactive)` that reads the secret through `SecretManager::Get(db)` under `GetSystemTransaction(db)`. `MANUAL_TOKEN` (a token handed to ATTACH) keeps its bytes — they *are* the credential — and gains the `exp` check. **Every provider's cached expiry is now the JWT's own `exp`**; `access_token`, `cli` and `env` used to get a fabricated `now + 3600`, so a cache hit proved the entry was young, not that the token was valid. A fixed token past its `exp`, and an interactive chain from a worker thread, fail **by name** before any socket:

```text
Azure AD access token in secret 'x' expired at 2026-09-10 08:49:00 UTC;
a fixed token cannot be refreshed -- DETACH and ATTACH with a new one
```

**W2 — a failed creation is not a full pool.** Factories **throw** with the reason; `CreateNewConnection` keeps it (`ErrorData(e).RawMessage()` — a 2.0 exception's `what()` is JSON). `Acquire`: nothing active → return at once; others active → wait the caller's budget, retry creation on a backoff (250 ms → 4 s). One renderer, `ConnectionPool::DescribeAcquireFailure()`, used by the provider and every catalog-internal caller; `mssql_pool_stats()` gains `creation_failures` and `last_create_error`.

```text
MSSQL: Failed to acquire connection: pool 'bad' could not create a connection:
Authentication failed (error 18456, state 1): Login failed for user 'sa'.
```

**W3 — `mssql_connection_timeout` reaches a refill**: passed to `Connect`, and all eight login-phase reads use it.

## Verified failing first

| test | pre-fix | now |
| --- | --- | --- |
| `test/cpp/test_pool_creation_failure.cpp` — server-free, `STANDALONE_TEST_SOURCES` | does not compile | 4 cases: fail-fast with reason (not a timeout); silent `nullptr` gets a reason; one active → waits, **2–4 attempts in 800 ms**, reason kept; success resets backoff |
| `test/sql/regression/issue_302.test` — wrong password, `lazy_validation`, `acquire_timeout 30` | `(timeout)`, **92 s** for three statements | `could not create a connection: Login failed for user` at once, pool 0/0 |
| `test/sql/azure/fedauth_expired_access_token.test` — `exp = 1000000000` | — | refused at ATTACH, `expired at 2001-09-09` (azure lane) |
| `test/sql/attach/attach_validates_credentials.test` | expected the old "(timeout)" text | expects the reason |

The 60-minute path itself is the spec's § 0 script; Azure AD does not issue short-lived tokens, so it stays a documented manual run.

## Review of the spec (job 1519), all addressed in the spec text and — where they were about mechanism — already in the code

- F4 restated: one caller, itself unreachable; § 3 lists what removing it would drag along and why that is its own PR.
- `TokenCache` is not an expiry oracle → every provider records the JWT's `exp` (W1).
- `MANUAL_TOKEN` was silently dropped → split `case`, keeps its bytes, gets the `exp` check.
- W2 had no channel → factories throw; no `ConnectionFactory` signature change, so the two existing test files' factories stand.
- W3 covered 2 of the reads → all 8.
- `test_connection_pool.cpp` is dead source → W4 says why a new server-free file rather than reviving it (that one dials a real server).
- The risk named the cleanup thread → restated as concurrent `Acquire` callers; F6/F7 in order; the intro counts seven findings.

## Docs

`AZURE.md` / `website/docs/connection/azure.md` gain *Token Lifetime and Long-Running Sessions*; `CLAUDE.md` rows for both timeouts; `DATAMODEL.md` pool invariants.

**No backport** — releases now come from `main` on the duckdb 2.0 line only; v0.2.5 was the last from `duckdb-v1.5.5`.

SQL suite 163 cases; 20 standalone C++ tests; `clang-format-14` clean.


VGSML:
The live repro has run — stock DuckDB 1.5.5 + community 0.2.5, service-principal secret, `acquire_timeout = 600`, `idle_timeout = 60`, 75-minute sleep:

```text
T0        pool 1/1, created 1, closed 0; two FEDAUTH logins OK
T+75min   SELECT mssql_exec('az','SELECT 1')
          IO Error: MSSQL: Failed to acquire connection from pool (timeout)
          Run Time (s): real 600.319
          pool 0/0, created 1, closed 1, acquire_timeout_count 1
          DETACH; ATTACH; retry  ->  1
```

**F5 is settled.** 600.3 s is exactly what the code predicts, so the reporter's 31 s is one ~30 s creation attempt followed by a deadline that had already passed — `acquire_timeout = 30`, not 600. Their `SET` did not reach the attaching session (host-side; DuckDB.NET, judging by the C# mention). The fix does not depend on it.

**New F7, and it changes W1.** Inside the wait the pool made two creation attempts, and on neither did the server send an ERROR token: it dropped the connection — once after LOGIN7, once after `FEDAUTH_TOKEN`. So `last_error_` would only ever say *"Failed to receive LOGINACK after FEDAUTH_TOKEN"*, after F3's hard-coded 30 s read. Azure will not tell us "expired"; the client has to diagnose it from the token's `exp` **before opening a socket** — which is where W1's check now sits — and W2's "surface the server's reason" is necessary but not sufficient for this case.

Spec updated in `HEAD`.

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+(`tds_connection.cpp:147`) exists and the routing-hop code uses it — *"the
+caller's timeout, not the compiled-in default"* — but no factory calls it.
+
+Then `DoLogin7WithFedAuth` reads both login responses with the constant
+spelled out — `:597` and `:674`, `ReceiveMessage(response,
+DEFAULT_CONNECTION_TIMEOUT * 1000)` — so a login that the server holds or drops
+costs 30 s regardless of any setting. The reporter's `mssql_connection_timeout
+= 600` governed ATTACH-time validation and nothing after it.
+
+### F4 — The refresh infrastructure exists, has no callers, and cannot be called from where it is needed
```
oluies: **Medium — F4's headline claim is the one thing here the code contradicts (and the truth is stronger).**

`BuildTokenAcquirer` does have a caller: `src/tds/auth/auth_strategy_factory.cpp:349`, inside `CreateFedAuth`. What is actually true is sharper — **`CreateFedAuth` itself is unreachable**: both live `AuthStrategyFactory::Create` call sites (`src/catalog/mssql_catalog.cpp:193`, `src/mssql_storage.cpp:1412`) are the spec-068 integrated-auth hop factories and pass no `ClientContext`, so the `use_azure_auth` branch at `auth_strategy_factory.cpp:321-327` never runs.

This matters for the removal plan in §3 (`:278`): deleting `BuildTokenAcquirer` alone leaves `CreateFedAuth`, `FedAuthStrategy::SetTokenAcquirer` / `GetFedAuthToken` and the `TokenAcquirer` typedef dangling — while the reader has been told there is nothing to unhook.

**Fix**: restate F4 as *"`BuildTokenAcquirer`'s only caller is `CreateFedAuth`, which is itself unreachable — the Azure branch of `Create` requires a `ClientContext` no live call site supplies"*, and extend §3's removal list to `CreateFedAuth` + the `FedAuthStrategy` acquirer plumbing.

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+time, and the reason preserved when creation fails.
+
+## 2. The work
+
+### W1 — Resolve the token at creation time, not at ATTACH
+
+The factory captures `DatabaseInstance &`, the secret name and the tenant
+override — never the token, never the client secret — and on each creation:
+
+1. asks `TokenCache` (keyed by `DatabaseInstance`, spec 047 FR-012) for the
```
oluies: **Medium — `TokenCache` is not an expiry oracle; its `expires_at` is real for exactly one provider.**

W1 leans on the cache to refuse a stale token *"within `IsTokenExpired`'s 300 s margin ... before any socket is opened"*. But `AcquireToken` stores a **fabricated** `now + DEFAULT_TOKEN_LIFETIME_SECONDS` (3600 s) for `provider='access_token'` (`src/azure/azure_token.cpp:444`) and for the CLI/env path (`:387`); only the service-principal path records the server's own `expires_in` (`:339`), and even the cache-hit return at `:427` labels it *approximate expiry*.

So a `GetToken` hit proves the entry is **young**, not that the JWT is **unexpired**. A token minted 55 minutes before it was handed to `CREATE SECRET` reads as valid and walks straight back into F7's silent hangup — the exact failure this spec exists to diagnose.

The same gap invalidates W4's azure-lane premise at `:256`: `PROVIDER access_token` with an already-expired token is *not* rejected client-side today, so that test is not deterministic as written.

**Fix**: have W1 diagnose expiry from the JWT's own `exp` via the existing `mssql::azure::ParseJwt` / `IsTokenExpired` (`src/azure/jwt_parser.cpp:235`) rather than from cache membership — and/or record the parsed `exp` at `SetToken` on the `access_token`/CLI paths instead of `now + 3600`. Make it an explicit W1 sub-item so W4 has something to assert against.

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+
+`mssql_catalog.cpp:167-200`, the integrated-auth factory, builds a fresh
+authenticator **per connection** so a `kinit`-refreshed ticket is picked up,
+and on failure prints `conn->GetLastError()` before returning `nullptr`. It is
+the shape the Azure factory should have had: credentials resolved at creation
+time, and the reason preserved when creation fails.
+
+## 2. The work
+
+### W1 — Resolve the token at creation time, not at ATTACH
```
oluies, line 222: **Medium — W1 silently drops `MANUAL_TOKEN`, and implemented literally it deletes that path's only credential.**

W1 is written for the secret-backed path alone (*"the factory captures `DatabaseInstance &`, the secret name and the tenant override — never the token"*). But the factory it replaces serves **two** auth methods: `src/catalog/mssql_catalog.cpp:136-137` falls `AuthMethod::AZURE_AD` and `AuthMethod::MANUAL_TOKEN` into the same `case`.

For `MANUAL_TOKEN` (`access_token=` in the connection string, set at `mssql_storage.cpp:229` / `:1547`) there is no secret name to re-resolve — **the captured bytes *are* the credential**, to the point that `~MSSQLCatalog` `OPENSSL_cleanse`s them (`mssql_catalog.cpp:87-88`).

**Fix**: state explicitly that `MANUAL_TOKEN` keeps the captured token, gains the client-side `exp` check, and produces the cannot-refresh message; only `AZURE_AD` switches to creation-time resolution. Split the `case` accordingly.

oluies (commented):
## roborev review — `spec/073-fedauth-token-refresh` (2 commits, `5fadf8a..92b8490`)

**Verdict: Fail** — 4 Medium, 3 Low. Job `1519`, agent `claude-code`.

The reconnaissance lands: the never-refreshed ATTACH-time FEDAUTH token (F1) and the pool's *"a failed creation looks like an exhausted pool"* behaviour (F2) are both correctly identified and measured, and F7's finding — that Azure SQL answers an expired token by **dropping the connection rather than sending an ERROR token** — is the kind of thing only the live repro could have produced.

What the findings below share is a single shape: **the work items rest on mechanisms the code does not currently provide.**

- **W1** treats `TokenCache` as an expiry oracle, but its `expires_at` is fabricated (`now + 3600`) for every provider except service-principal — so a cache hit proves *young*, not *unexpired*, and the spec's own F7 failure survives it.
- **W1** also covers only the secret-backed path, while the `case` it replaces also serves `MANUAL_TOKEN`, where the captured bytes **are** the credential.
- **W2** promises to surface the factory's reason through a `ConnectionFactory` signature that has nowhere to put it.
- **W3** fixes two of the four hard-coded login reads on the FEDAUTH path.

And **F4's headline** — *"`BuildTokenAcquirer` has no callers"* — is the one claim here the code contradicts. It has one; the caller is itself unreachable. The correction strengthens the finding but widens §3's removal list.

None of this is fatal to the plan — it is a spec-only PR, and every item is a wording or scoping change made now rather than discovered mid-implementation.

<sub>Posted by `roborev` via Claude Code. Line anchors verified against `92b8490`.</sub>

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+3. builds the FEDAUTH bytes from the string — `BuildFedAuthExtension` gains a
+   form that takes the token rather than the context.
+
+Providers that cannot mint silently — `access_token` (a fixed string), and the
+interactive ones after their first use — fail with a message that says what
+happened and what to do: *"Azure AD token for secret 'x' expired at <time>;
+DETACH and ATTACH to authenticate again"*. That is the same outcome as today,
+minus the thirty-one seconds and the word "timeout".
+
+### W2 — A failed creation is a failure, not a full pool
```
oluies, line 265: **Medium — W2 has no channel to carry the reason it promises to surface.**

W2 requires the pool to keep the factory's creation error (*"the provider's message carries the reason"*), but `ConnectionFactory` is `std::function<std::shared_ptr<TdsConnection>()>` (`src/include/tds/tds_connection_pool.hpp:49`) and `CreateNewConnection` merely returns `factory_()` (`tds_connection_pool.cpp:323-326`). The `TdsConnection` holding `last_error_` is destroyed inside the closure — nothing reaches `pool_mutex_`-guarded state.

As written the work item is **not implementable** without a signature change the spec never mentions, and that change has four call sites: `mssql_catalog.cpp` ×3, plus `test/cpp/test_connection_pool.cpp` and `test/cpp/test_tls_connection.cpp`.

**Fix**: put the contract in W2 — e.g. `using ConnectionFactory = std::function<std::shared_ptr<TdsConnection>(std::string &error_out)>` — and list the call sites that must move, the two test files included.

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+  released, so waiting is pointless — `Acquire` returns at once;
+- when others are active, it keeps waiting as now, since a release may still
+  serve the request, but re-tries creation with a backoff rather than on every
+  wake;
+- on any `nullptr` return the provider's message carries the reason:
+  `Failed to acquire connection from pool 'x' after 5 s: Azure AD
+  authentication failed (error 18456): Login failed for user '<token-identified
+  principal>'`.
+
+### W3 — Plumb `mssql_connection_timeout` where it was meant to go
```
oluies, line 316: **Low — W3 fixes two of the four hard-coded login reads, so its own acceptance criterion would still not hold.**

W3 scopes the timeout fix to *"`DoLogin7WithFedAuth` reads its responses"* (`tds_connection.cpp:597`, `:674`). But the FEDAUTH path gets there through `DoPreloginWithFedAuth` (`:473`), which spells out the same constant twice more: the PRELOGIN response read at `:500` and the TLS handshake at `:531`.

A gateway that accepts the TCP connection and then goes quiet still costs up to 60 s of un-configurable wait after W3 lands — so *"`mssql_connection_timeout` bounds a pool refill's connect and login reads"* (F3, `:115`) would not actually be true.

**Fix**: extend W3 to `:500` and `:531`, and note the equivalent `:214` / `:245` on the SQL-auth path, which the same criterion covers.

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+
+Every factory calls `conn->SetConnectTimeout(pool_config.connection_timeout)`
+before `Connect`, and `DoLogin7WithFedAuth` reads its responses with
+`connect_timeout_seconds_ * 1000`, as the routing-hop path already does.
+
+### W4 — Tests
+
+- **Pool, server-free, gates a PR**: a factory that fails; with nothing active,
```
oluies: **Low — the server-free pool test W4 proposes already exists, and is wired into nothing.**

`test/cpp/test_connection_pool.cpp` is present but appears in **no** `STANDALONE_TEST_SOURCES` list (`Makefile:594-613`) and in no CMake target — it is dead source today. An implementer following W4 will either add a second parallel file or rediscover this and stall. It also constructs `ConnectionPool` directly, which makes it the first casualty of the `ConnectionFactory` signature change W2 needs.

**Fix**: say whether W4 revives `test/cpp/test_connection_pool.cpp` into `STANDALONE_TEST_SOURCES` or adds a new file, and note the link surface the standalone lane must satisfy (`tds_connection_pool.cpp` pulls `TdsConnection`, hence OpenSSL/vcpkg).

specs/073-fedauth-token-refresh/spec.md
```diff
@@ -0,0 +1,305 @@
...
+  a *new* one.
+- **Fixing `BuildTokenAcquirer` in place.** Its contract — a `ClientContext`
+  captured by reference — is the problem, not a detail. It is replaced by the
+  `DatabaseInstance`-bound acquirer and removed, along with the strategy-side
+  `IsTokenExpired` overrides if the `mssql_open` path turns out to be their only
+  consumer.
+
+## 4. Risks
+
+- **Secret access from a worker thread.** `SecretManager::Get(db)` with a
```
oluies, line 404: **Low — the named risk points at a thread that never calls the factory.**

This risk asks for *"an explicit check for a `SecretManager` lock the pool's **cleanup thread** could contend on"*, but `CleanupThreadFunc` (`tds_connection_pool.cpp:341-380`) only reaps idle connections and never touches the factory — `CreateNewConnection` is reachable from `Acquire` alone (`:174`). The real contention question is between **concurrent `Acquire` callers on request threads**.

Separately (§1 ordering, `:156`): the findings run F1–F5, then **F7**, then **F6**, so a reader scanning for F6 finds it after F7 and may think one is missing — and the intro still describes *"four hold, one does not explain..., and one thing it did not say"* for what are now seven findings.

**Fix**: restate the risk in terms of concurrent `Acquire` callers, and renumber so §1 reads in order.

VGSML:
All seven addressed in `HEAD`, and the PR is now the spec **and** the implementation — #335 was merged into this branch, so the text and the code are reviewed together. The shape you identified was right: the draft rested on mechanisms the code did not have. The implementation had built most of them by the time this landed; the text now describes what was built, and one of your findings turned into a code change.

## The one that changed code

**`TokenCache` is not an expiry oracle.** Correct, and only half-fixed by the implementation: it parsed the JWT for `access_token` but `cli` and `env` still cached `now + 3600`. `AcquireToken` now records the JWT's own `exp` for **every** provider; the fabricated lifetime survives only for a token that is not a JWT. That is also what makes the azure-lane test deterministic, as you noted it was not as written.

## The rest, in the text

- **F4** — restated: `BuildTokenAcquirer` has one caller, `CreateFedAuth`, and *that* is unreachable (the Azure branch of `Create` needs a `ClientContext` no live site supplies). § 3 now lists what removing it drags along — `CreateFedAuth`, `SetTokenAcquirer` / `GetFedAuthToken`, the typedef — and says why that is its own PR rather than a rider on a bug fix.
- **`MANUAL_TOKEN`** — split `case`; it keeps its bytes (they *are* the credential, which is why `~MSSQLCatalog` cleanses them) and gets the `exp` check per creation.
- **W2's channel** — the factories **throw** `ConnectionException`; `CreateNewConnection` catches and records. No `ConnectionFactory` signature change, so the two existing test files' factories stand. And `ErrorData(e).RawMessage()`, because a 2.0 exception's `what()` is JSON — the first cut pasted it into the user's error.
- **W3** — all eight hard-coded reads, not two: `:214`, `:245`, `:500`, `:531`, `:597`, `:674`, `:865`, `:1002`.
- **W4** — a new server-free file, and the text says why `test_connection_pool.cpp` is not revivable into the standalone lane: its factory dials a real server with real credentials.
- **Risk / ordering** — concurrent `Acquire` callers, not the cleanup thread; F6 before F7; the intro counts seven.

## What the merge also caught

The full suite on the merged branch failed **one** pre-existing test: `attach_validates_credentials.test` expected `Failed to acquire connection from pool` for a lazily-validated ATTACH to an unresolvable host. The new message is `pool 'att_lazy' could not create a connection: TCP connect to unreachable.invalid:1433 failed: Failed to resolve hostname` — the test now expects the reason, which is what W2 is for.

SQL suite **163 cases / 5077 assertions**; 20 standalone C++ tests; `clang-format-14` clean.

src/tds/tds_connection_pool.cpp
```diff
@@ -168,13 +170,20 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms) {
...
 			lock.unlock();
 			conn = CreateNewConnection();
 			lock.lock();
 
 			if (conn) {
+				// A success resets the backoff so a transient failure does not
+				// slow the next refill.
+				create_backoff_ms_ = CREATE_BACKOFF_INITIAL_MS;
+				next_create_allowed_ = std::chrono::steady_clock::time_point{};
 				uint64_t id = next_connection_id_++;
```
oluies, line 206: **Medium — `last_create_error_` is never cleared, so one transient failure poisons every later message**

This success branch resets `create_backoff_ms_` and `next_create_allowed_`, but not `last_create_error_`. And `DescribeAcquireFailure()` (`:398`) prefers that string whenever it is non-empty:

```cpp
if (!last_create_error_.empty()) {
    return "pool '" + context_name_ + "' could not create a connection: " + last_create_error_;
}
```

So the first creation failure a pool ever suffers permanently rewrites every subsequent acquire failure as a creation failure with a stale reason — suppressing the `"timed out (N active of M, limit L)"` diagnostic that this very function was added to produce.

**Failure scenario:** one refused dial at 09:00 while the server restarts records `TCP connect to host:1433 failed: Connection refused`. The next attempt succeeds; the pool runs healthy all day. At 14:00 the pool is at `connection_limit` with all 64 connections busy and `Acquire` times out on genuine exhaustion — the user is told `could not create a connection: Connection refused`. That is the same wrong-diagnosis class this PR set out to eliminate.

It also crosses threads: thread A's exhaustion timeout can report thread B's old login error. And `mssql_pool_stats.last_create_error` shows a long-dead error forever, with `creation_failures` unable to distinguish "recovered" from "still failing".

**Fix:** clear `last_create_error_` here, under `pool_mutex_`, alongside the backoff reset. Better still, have `Acquire` record whether *this* call's attempt failed in the factory and pass that into the renderer, rather than reading pool-global state after the fact. A "last failure timestamp" in stats would also make a stale reason recognisable.

src/catalog/mssql_catalog.cpp
```diff
@@ -133,28 +135,94 @@ void MSSQLCatalog::Initialize(bool load_builtin) {
...
+			auto token_result = mssql::azure::AcquireToken(*db, secret_name, tenant, /*allow_interactive=*/false);
+			if (!token_result.success) {
+				throw ConnectionException("Azure AD token for secret '%s': %s", secret_name,
+										  token_result.error_message);
+			}
+			auto fedauth = mssql::azure::BuildFedAuthData(token_result.access_token);
+			auto conn = std::make_shared<tds::TdsConnection>();
+			conn->SetRequestedPacketSize(tds_packet_size);
+			conn->SetRequestUtf8Support(utf8_support);
+			if (!conn->Connect(host, port, connect_timeout)) {
```
oluies, line 177: **Medium — `SET mssql_connection_timeout = 0` now breaks every pooled connection**

`mssql_connection_timeout` is registered with `ValidateNonNegative` (`src/connection/mssql_settings.cpp:73`), so `0` is a legal value — and several sibling timeouts in this codebase document `0` as "no timeout" (`mssql_metadata_timeout`, `mssql_attach_validation_timeout`, `mssql_catalog_cache_ttl`). A user setting `0` today is reasonably expecting "wait as long as it takes".

Before this change the factories ignored the setting entirely and every login read used the compiled-in 30 s, so `0` was harmless. Now `pool_config_.connection_timeout` flows through here into `Connect()`, is stored in `connect_timeout_seconds_`, and reaches all eight login-phase reads (`tds_connection.cpp:214,245,500,531,597,674,865,1002`). There it becomes `ReceiveMessage(response, 0)` → `poll(fd, 1, 0)` (`tds_socket.cpp:865`), which returns immediately → `"Socket timeout waiting for data"`.

**Failure scenario:** `SET mssql_connection_timeout = 0;` then any query on an attached catalog. Every pooled connection fails instantly with a bogus timeout, on a configuration that worked before this PR. W3 is the right change; this is its unguarded edge.

**Fix:** clamp at the boundary — treat `timeout_seconds <= 0` as `DEFAULT_CONNECTION_TIMEOUT` in the factories or in `TdsConnection::Connect`, or map `0` to an infinite poll (`-1`) if "no timeout" is the intended semantic. Then say which it is in the CLAUDE.md row, since the neighbouring rows set the opposite expectation.

The same `connect_timeout` reaches the other three factories at `:220`, `:242` and `:304`.

src/connection/mssql_diagnostic.cpp
```diff
@@ -264,6 +264,13 @@ unique_ptr<FunctionData> MSSQLPoolStatsFunction::Bind(ClientContext &context, Ta
 
 	names.emplace_back("pinned_count");
 	return_types.emplace_back(LogicalType::BIGINT);
+	// Issue #302: the two numbers that were missing when the reporter stared at
+	// "1 idle, 0 closed" for two hours -- how many times the pool tried to
+	// create a connection and could not, and what the last attempt said.
+	names.emplace_back("creation_failures");
+	return_types.emplace_back(LogicalType::BIGINT);
+	names.emplace_back("last_create_error");
+	return_types.emplace_back(LogicalType::VARCHAR);
```
oluies, line 273: **Low (security gate) — the SC-005 redaction test no longer covers what it claims to**

`last_create_error` is a second VARCHAR on `mssql_pool_stats`, populated from `pool.GetLastCreateError()` — arbitrary factory and server error text, queryable by anyone who can reach the catalog.

`test/sql/tds_connection/pool_stats_no_credentials.test` is the SC-005 gate, and its header states the invariant it rests on verbatim:

> Today's pool_stats schema is `db` (VARCHAR — catalog alias + GetDBPath form `mssql://host:port/database`) plus 8 BIGINT counters. Numeric columns physically cannot contain a string sentinel; the `db` column is what we have to grep.

That is exactly what this PR changes, and the gate still greps only `db`. Today's error texts do not appear to embed credentials — but the gate no longer *proves* that, and it will not catch a future one that does.

**Failure scenario:** a later factory error, or a bare `last_error_` from the auth layer, interpolates a connection string or a token into its message. `mssql_pool_stats()` exposes it, and `pool_stats_no_credentials.test` still passes because it never looks at the new column.

**Fix:** extend the gate to every VARCHAR column (`WHERE db LIKE … OR last_create_error LIKE …`), update the header comment, and ideally add a case that provokes a creation failure so the new column is actually non-empty when grepped — otherwise the assertion passes against an empty string.

*(This one is carried forward from the review of the earlier range, job 1536; it was not re-reported by 1538, but the code is unchanged between them.)*

src/tds/tds_connection_pool.cpp
```diff
@@ -187,6 +196,28 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms) {
...
+
+			// Creation failed. Issue #302: this used to fall through to the same
+			// wait an exhausted pool does -- for a Release that, with nothing
+			// active, cannot come -- and report "(timeout)" after the full
+			// acquire_timeout, with the reason discarded. Measured: an expired
+			// Azure AD token cost 600 s and two login attempts, and read
+			// identically to a wrong password or an unreachable host.
+			stats_.creation_failures++;
+			next_create_allowed_ = std::chrono::steady_clock::now() + std::chrono::milliseconds(create_backoff_ms_);
+			create_backoff_ms_ = std::min(create_backoff_ms_ * 2, CREATE_BACKOFF_MAX_MS);
```
oluies: **Low — `std::min` odr-uses `CREATE_BACKOFF_MAX_MS`, which has no out-of-line definition**

`std::min` takes both arguments by `const&`, which odr-uses `CREATE_BACKOFF_MAX_MS` (`tds_connection_pool.hpp:128-129`). In-class `static constexpr` members are only implicitly inline from C++17 onward.

This repo's CLAUDE.md has an explicit "ODR errors on Linux" section warning that the extension must build against DuckDB's C++ standard and must not force C++17 — and `src/` contains no `if constexpr`, structured bindings, or `std::optional`, consistent with that rule. If this TU is compiled pre-C++17, an unoptimised build (`make debug`, where the reference is not folded away) fails to link with an undefined reference.

**Fix:** avoid the reference bind entirely —

```cpp
create_backoff_ms_ = create_backoff_ms_ * 2 > CREATE_BACKOFF_MAX_MS
                         ? CREATE_BACKOFF_MAX_MS
                         : create_backoff_ms_ * 2;
```

— or add an out-of-line `constexpr int ConnectionPool::CREATE_BACKOFF_MAX_MS;` definition.

src/tds/tds_connection_pool.cpp
```diff
@@ -205,6 +236,12 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms) {
 		}
 
 		int remaining = timeout_ms - static_cast<int>(elapsed);
+		// Wake for the next creation attempt if that comes first (issue #302).
+		if (next_create_allowed_ > now && stats_.total_connections < config_.connection_limit) {
+			auto until_create =
+				std::chrono::duration_cast<std::chrono::milliseconds>(next_create_allowed_ - now).count();
+			remaining = std::min<int>(remaining, static_cast<int>(until_create) + 1);
+		}
```
oluies, line 278: **Low — the backoff defeats W2's "nothing active → fail at once" contract from the second call onward**

The fast-fail contract holds only for the *first* failing `Acquire`. On the next one the gate at `:176-177` skips creation entirely (the backoff has not elapsed), so control falls through to this exhausted-pool wait and sleeps `min(remaining, until_create + 1)` — waiting on a `Release` that cannot come, because nothing is active. It then retries and fails with the same error.

**Failure scenario:** a user running failing statements back to back against a catalog with bad credentials sees 250 ms → 500 ms → 1 s → 2 s → 4 s of pure stall before each identical error. `test/sql/regression/issue_302.test` only asserts the *first* failure is immediate, so it passes.

There is a second effect: `next_create_allowed_` is pool-global, so a thread whose creation might have succeeded (a refreshed token, a server that came back) is penalised by an unrelated thread's failure.

**Fix:** in the wait path, when `stats_.active_connections == 0` and `idle_connections_` is empty, return `nullptr` immediately and let the caller render `last_create_error_`. Keep the backoff for the others-are-active case it was actually written for.

src/include/azure/azure_token.hpp
```diff
@@ -177,6 +178,22 @@ class TokenCache {
...
+TokenResult AcquireToken(DatabaseInstance &db, const std::string &secret_name, const std::string &tenant_id_override,
+						 bool allow_interactive);
+
+//! The same acquisition from a DatabaseInstance, for a caller with no
+//! ClientContext -- the connection-pool factory refreshing a token for a new
+//! connection (issue #302, spec 073). `allow_interactive = false` refuses the
+//! device-code / interactive chain with a message naming the way out rather
+//! than prompting a user from inside a query on a worker thread.
+TokenResult AcquireToken(DatabaseInstance &db, const std::string &secret_name, const std::string &tenant_id_override,
+						 bool allow_interactive);
```
oluies: **Low — this declaration and its doc comment are a verbatim duplicate of lines 181-187**

Both the six-line doxygen block and the `AcquireToken(DatabaseInstance &, const std::string &, const std::string &, bool)` declaration appear twice, introduced together in this diff. Harmless to the compiler, but it invites drift: a later signature change (say, defaulting `allow_interactive`) applied to one copy yields either a redeclaration mismatch or a silently stale doc comment.

**Fix:** delete this second copy.

oluies (commented):
## Automated review (roborev job 1538)

Range reviewed: `5fadf8a..39f7351` — the full PR against `main`, including the stacked #335 merge.

The shape of the change is sound: token resolution moves from ATTACH into the pool factory, a factory failure fails fast carrying the server's reason instead of masquerading as pool exhaustion, and `mssql_connection_timeout` finally reaches the login-phase reads. Confirmed while reviewing that `39f7351` closed the earlier gap where only the `access_token` provider read the JWT `exp` — `cli` and `env` now do too.

**Two Medium**, one of which is a regression on a legal setting, plus three Low and one carried forward. All inline.

The carried-forward one is worth calling out separately, because it is a security gate this PR silently widens: `mssql_pool_stats` gains a second VARCHAR column, and the SC-005 redaction test's own header states the invariant it depends on — *"Today's pool_stats schema is `db` (VARCHAR …) plus 8 BIGINT counters. Numeric columns physically cannot contain a string sentinel; the `db` column is what we have to grep."* That is no longer true. See the comment on `mssql_diagnostic.cpp`.

VGSML:
All six findings from job 1538 are addressed in e16c376. Each was verified against the code before it was touched; what changed, per finding:

**Medium — `last_create_error_` never cleared.** Confirmed, and the fix is the "better still" variant: the reason is now **per call**. `Acquire(timeout_ms, std::string *failure)` renders why *this* call failed — `pool 'x' could not create a connection: <reason>` or `pool 'x' timed out (N active of M, limit L)`, the latter carrying this call's own creation failure when it made one on the way (with others active that is what explains why the pool did not simply grow). `DescribeAcquireFailure()` is gone; all nine throwing callers pass the out-parameter (the three that swallow a nullptr silently are unchanged). A successful creation clears `last_create_error_`, so `mssql_pool_stats` shows a live problem, not history; `CreateNewConnection` hands the reason back instead of taking the mutex itself. `test_pool_creation_failure.cpp` gains the scenario from the finding — one refused dial, then a healthy pool at its limit — and asserts the exhaustion is reported as `timed out (1 active of 1, limit 1)` without the morning's `Connection refused`; it fails on 39f7351. No "last failure timestamp" column: with the error cleared on success, a non-empty `last_create_error` is by construction the *current* problem.

**Medium — `mssql_connection_timeout = 0`.** Confirmed live: with the setting at 0 the first pooled connection failed at once with `TCP connect to localhost:1433 failed: Connection timed out`. Clamped at the one boundary every path crosses, `TdsConnection::Connect`: `<= 0` becomes `DEFAULT_CONNECTION_TIMEOUT`. I chose "0 = the 30 s default" rather than "no timeout", deliberately unlike the metadata/query timeouts next to it: a dial or a login read that never completes must not hang a pool refill forever. The setting's description, the CLAUDE.md row and spec W3 say so. `test/sql/regression/connection_timeout_zero.test` fails before the clamp and passes after.

**Low (security gate) — SC-005.** The gate now greps `last_create_error` as well as `db`, its header no longer claims `db` is the only VARCHAR, and a new case provokes a creation failure with a sentinel password through `lazy_validation`, asserts the column is non-empty (so the grep is not vacuous), and asserts the sentinel is absent from both columns.

**Low — `std::min` odr-use.** Replaced with a ternary. For the record, on the 2.0 line the extension compiles as C++17 — DuckDB 2.0 refuses `CMAKE_CXX_STANDARD=11` at configure — so the reference was already folded, but the ternary costs nothing and the CLAUDE.md rule stands.

**Low — the backoff on an empty pool.** Confirmed. The backoff now applies only while others are active (the case it was written for); with nothing active every `Acquire` dials once and fails at once with its own reason. Test: three failing acquires back to back, each under 200 ms, each carrying its own attempt's reason. The second effect you name — a pool-global `next_create_allowed_` penalising a thread whose creation might have succeeded while others are active — remains; that is what a backoff is, and the alternative (per-thread schedules) would let N waiters dial N times per wakeup against a server that hangs up on each.

**Low — duplicated `AcquireToken` declaration.** Removed.

Spec 073 (W2, W3, W4), CHANGELOG and the two CLAUDE.md rows describe the mechanics as they now are. On the branch: full suite 177 cases / 5549 assertions, standalone C++ list green.

src/tds/tds_connection_pool.cpp
VGSML: Fixed in e16c376: the reason is per call (`Acquire(timeout, &why)`), `DescribeAcquireFailure()` is gone, and a successful creation clears `last_create_error_`. The scenario from this finding is a unit test that fails on 39f7351. Details in the PR comment.

src/catalog/mssql_catalog.cpp
VGSML: Fixed in e16c376: `TdsConnection::Connect` maps `<= 0` to `DEFAULT_CONNECTION_TIMEOUT`, documented as "0 = the default, not no-timeout" in the setting, CLAUDE.md and spec W3. `connection_timeout_zero.test` fails before the clamp.

src/connection/mssql_diagnostic.cpp
VGSML: Fixed in e16c376: the gate greps `last_create_error` too, the header is corrected, and a sentinel-password creation failure makes the column non-empty before it is grepped.

src/tds/tds_connection_pool.cpp
VGSML: Fixed in e16c376: ternary instead of `std::min`. (On the 2.0 line the TU is C++17 already; the ternary costs nothing.)

src/tds/tds_connection_pool.cpp
VGSML: Fixed in e16c376: the backoff applies only while others are active; with nothing active every Acquire dials once and fails at once. Three back-to-back failures under 200 ms each is a unit test now.

src/include/azure/azure_token.hpp
VGSML: Removed in e16c376.

oluies (dismissed):

PR #338, opened 2026-09-11
spec 074: Encrypt, TrustServerCertificate and HostNameInCertificate as the Microsoft drivers define them; the deprecated mssql_open API goes

VGSML:
Spec 074 and its implementation, in one PR, plus the removal the spec's W6 names.

## What changes

The three TLS options now mean what they mean in the Microsoft drivers (ODBC 18, `Microsoft.Data.SqlClient` 4.0, JDBC 10.2, `go-mssqldb`), spelling for spelling and default for default:

| Option | Default | Meaning |
|---|---|---|
| `Encrypt` | `true` | Encrypt the session. `false` sends no TLS at all and the two below are ignored. Unchanged. |
| `TrustServerCertificate` | `false` | `false`: the server's certificate chain must validate against the platform trust store, and its subject must match the host connected to. `true`: any certificate is accepted; the channel is still encrypted. |
| `HostNameInCertificate` (new) | the host connected to | The name the certificate must carry when it differs from the address dialled: an IP, an SSH tunnel to `localhost`, an alias. After a login-time routing hop the routed host is checked, unless this is set, in which case it applies to every hop (as in go-mssqldb). |

`TrustServerCertificate` used to be an alias of `Encrypt`, and giving both with different values was an ATTACH error; `Encrypt=true;TrustServerCertificate=false` is now simply the default. URI spellings `trustservercertificate` / `hostnameincertificate`, secret fields `trust_server_certificate` / `host_name_in_certificate`.

**Breaking.** A connection to a server on a self-signed certificate — the docker image, an on-prem instance with none installed — now fails and says so:

```text
TLS handshake failed: certificate verification failed for localhost: self-signed certificate. Set TrustServerCertificate=yes to accept this server's certificate without verification, or HostNameInCertificate=<name> if the certificate is valid but issued for a different name
```

The reason is OpenSSL's own (`self-signed certificate`, `unable to get local issuer certificate`, `hostname mismatch`, `certificate has expired`), so it can be searched for. Adding `TrustServerCertificate=yes` is what `sqlcmd -C` does; every test DSN in the Makefile carries it, and the CHANGELOG entry is under **Changed / Breaking**.

**Trust store.** The platform's, loaded the way the Azure OAuth client (cpp-httplib) already loads it: Windows `ROOT` + `CA`, macOS keychain trust settings plus OpenSSL's default paths (both, always — httplib skips the paths once the keychain yields anything, which silently disables `SSL_CERT_FILE` there), other Unix OpenSSL's default paths. `SSL_CERT_FILE` / `SSL_CERT_DIR` override on every platform; a private CA goes there or into the platform store. No per-connection CA-file option (spec § 3).

**Matching** is OpenSSL's (`X509_VERIFY_PARAM_set1_host` with `NO_PARTIAL_WILDCARDS`; an IP literal, recognised by `a2i_IPADDRESS`, goes through `set1_ip_asc` and matches `iPAddress` SANs). No hand-written matcher.

**Plumbing.** `tds::TlsOptions { verify_certificate, expected_host }` travels `MSSQLConnectionInfo::GetTlsOptions()` → `TdsConnection::SetTlsOptions` (a setter, like `SetRequestedPacketSize`; the three `Authenticate*` signatures do not grow) → `TdsSocket::EnableTls` → `TlsImpl::Initialize`. The expected name is resolved at `EnableTls` time from `host_`, which after a routing hop is the routed host, so D3's hop rule falls out of spec 068 with no code of its own. Every site that opens a connection — the four pool factories, the four ATTACH-validation paths — sets it.

**W6 — the deprecated diagnostic-handle API is removed.** `mssql_open` / `mssql_close` / `mssql_ping` / `mssql_close_all` and the `MSSQLConnectionHandleManager` singleton behind them, marked `[DEPRECATED]` in spec 047 (v0.2.1) for removal at the next major boundary, which the first release on the duckdb 2.0 line is. Removing them rather than teaching them the new options leaves exactly the eight sites above opening connections. `mssql_pool_stats` stays.

## Tests

- **Server-free**, `test/cpp/test_tls_verification.cpp` in `STANDALONE_TEST_SOURCES`: an OpenSSL server in-process on a loopback socket with a certificate minted at runtime; the client is `TlsTdsContext` driven as `TdsSocket::EnableTls` drives it. Seven cases: the self-signed certificate is rejected with `self-signed certificate` and the expected name; trusted, it is accepted with a cipher negotiated; with the certificate in the trust store (`SSL_CERT_FILE`) verification passes; a different expected name fails with `hostname mismatch`; an IP literal matches the `iPAddress` SAN; a different IP fails with `IP address mismatch`; under trust the expected name is ignored.
- **Docker**, `test/sql/integration/tls_server_certificate.test`: the default fails with the reason and both hints, `Encrypt=true;TrustServerCertificate=false` is a certificate failure and not a parse error, and a trusted session reports `encrypt_option = TRUE`. `attach_trust_cert.test` becomes `attach_tls_options.test` with its "Conflicting values" cases inverted.
- **Azure lane**, `test/sql/azure/tls_host_name_in_certificate.test`: Azure SQL's public chain passes the default; `HostNameInCertificate=wrong.invalid` fails with `hostname mismatch`. Both were also run live from this branch, along with the docker cases.
- The Makefile's DSNs, the CI lane's own copy of them in `scripts/ci/integration_test.sh`, the CI smoke secret in `scripts/sql/smoke_test.sql`, and the ten test files that build strings or secrets from `MSSQL_TEST_HOST` carry `TrustServerCertificate=yes`; `catalog_filter_sources` appends its URI filters with `&`. The macOS lanes that compile the TLS TU by hand (`test-login-routing-hops`, CI's routing step) link `Security` + `CoreFoundation`; Windows already links `crypt32`.

Full suite and the standalone C++ list are green on the branch (after merging `main` with #332 and #333, whose factory changes this PR re-applies `SetTlsOptions` on top of).

## Verified live, beyond CI

The whole Azure lane was run from this branch with an `azure` extension built against the same duckdb pin (there is no prebuilt one for it): **26 test cases, 276 assertions, none skipped** — service-principal and env-provider FEDAUTH logins, the manual-token login, the expired-token message, and **Fabric Warehouse through its login-time routing hop** (`fabric_types.test`), all under the default verification. The routed worker's certificate matches the routed host name, so D3's default holds on the one path the spec listed as a risk.

Left as is: `test/cpp/test_tls_connection.cpp` is built by no target.


VGSML:
Self-review round, in a0305f1:

- **`TdsConnection`'s move constructor and move assignment did not carry `tls_options_`** — nor, since they were added, `requested_packet_size_` (spec 055) and `request_utf8_support_` (#225). Nothing in `src/` or `test/` moves a `TdsConnection` today, so this is hygiene, but a moved connection would have verified against the host with default options instead of the policy the caller set. All three travel now; the initialiser list is in declaration order.
- **The spec said "four outcomes"** for the server-free test; it has seven (the four D2 × D3 cells, the IP-literal pair for D5, and the ignored name under trust). W4 and the acceptance list now say so.

Also checked and left alone: no CI workflow builds a docker DSN without `TrustServerCertificate` (they all go through the Makefile); the Kerberos stack's examples already carry it; `test/cpp/test_tls_connection.cpp` is built by no target. A Windows-enabled CI run is dispatched on this branch (run 34574373300) because the MinGW/MSVC compile of the `wincrypt.h` store enumeration is otherwise exercised by nothing on a PR.

VGSML:
`Build (osx_arm64)` on the dispatched run 34574373300 failed in `make test-cpp-run`: `test_tls_verification` exited non-zero with **no output at all**, which is how a process ends when a signal kills it. macOS has no `MSG_NOSIGNAL` (the TLS layer's direct `send()` passes 0 there), and the server side of a *rejected* handshake closes first, so the client's alert or `close_notify` lands on a closed peer and the default `SIGPIPE` ended the binary before it could print. Locally the client finishes writing before the server closes, which is why it passed here.

e759558: `SO_NOSIGPIPE` on the test's raw client socket (as `TdsSocket` sets on its own sockets), `SIGPIPE` ignored in `main` as belt and braces, and an unbuffered progress line per case so a future death still says where. The PR's own Build job re-runs the standalone list on this head.

VGSML:
The Fabric routing hop — the one item the spec's § 4 and this PR listed as not verified — now is. There is no prebuilt `azure` extension for the 2.0 pin, so I built duckdb-azure `main` against our own duckdb / extension-ci-tools submodules (its duckdb pin is an ancestor of ours; clean compile) and ran the full Azure lane from this branch: **26 test cases, 276 assertions, none skipped**. That includes service-principal and env-provider FEDAUTH logins, the manual-token login, the expired-token message, and `fabric_types.test` through the Fabric Warehouse's login-time redirect, all under the default verification: the routed worker's certificate matches the routed host name, so D3's default holds there. The PR body's "not verified" section is updated accordingly.

VGSML:
The dispatched full run (34576818817) came back: **Build Windows (MSVC) and (MinGW), both Windows smoke tests, and Build (osx_arm64) with the standalone list — including `test_tls_verification` after the SIGPIPE fix — all green.** One failure: `Integration Test (linux_amd64)`, on the first statement of the CI smoke test: `certificate verification failed for localhost: self-signed certificate` on `ATTACH '' AS mssql_smoke (SECRET smoke_test_secret)`.

The cause is a second copy of the test DSNs I had missed: `scripts/ci/integration_test.sh` builds the suite's DSN/URI defaults itself ("mirroring the Makefile verbatim", by its own comment), and `scripts/sql/smoke_test.sql` creates its secret without `trust_server_certificate`. Both now trust the CI server's self-generated certificate, in e972ffc; the spec and CLAUDE.md no longer claim the Makefile is the only place CI gets its DSNs from. Verified by running `scripts/ci/integration_test.sh` locally exactly as the job does: smoke test passes, 175 cases / 5570 assertions.

VGSML:
Site docs in 3d56e1c: the settings reference rows for `mssql_connection_timeout` / `mssql_acquire_timeout` now describe the spec 073 semantics (0 = the default; a failed creation reports at once), every `localhost` example on the site carries `TrustServerCertificate=yes` with the secret example saying why, the landing feature line matches the README, and the Azure page gains a Certificates note (Azure SQL / Fabric pass the default with nothing added, including through the redirect; `HostNameInCertificate` for private endpoints and tunnels).

VGSML:
The `Concurrency stress` job failed for the same reason as the Linux integration job before it: a **third** copy of the test DSNs. `make test-concurrent-reads` — and the four spec-047 isolation / transaction / attach-loop C++ tests — assemble `Server=...;Password=...` in their own `Dsn()` builders. All five now say `TrustServerCertificate=yes` (1f90441); verified by building `test_concurrent_reads` against the release library and running all scenarios against docker. Same commit: `AddDerCertificate` is compiled on Windows and macOS only — GCC flagged it as unused on Linux, where OpenSSL's default paths do the loading.

oluies (approved):

PR #341, opened 2026-09-11
spec 075: read path off the pinned connection — describe at bind, execute at init, parameters

### LINKED ISSUE #239: Materialise a scan of our own catalog before the sink, instead of documenting it as a limitation

### LINKED ISSUE #329: mssql_scan holds the pinned connection while streaming, so two scans cannot share one statement in a transaction
Inside an explicit transaction the extension pins one pooled connection. A statement containing two `mssql_scan` calls then fails:

```
IO Error: Failed to execute SQL batch: Cannot execute: connection not in Idle state (current: Executing)
```

The DuckDB postgres extension runs the same shape without complaint, and the interesting part is that it is **not** because it uses more connections.

## The comparison

```sql
-- postgres: works, returns 6270
ATTACH '…' AS pg (TYPE postgres);
BEGIN TRANSACTION;
WITH a AS NOT MATERIALIZED (SELECT * FROM postgres_query('pg','SELECT 1 AS x')),
     b AS NOT MATERIALIZED (SELECT * FROM postgres_query('pg','SELECT 2 AS x'))
SELECT a.x + b.x + (SELECT count(*) FROM pg.public.ducklake_snapshot) FROM a, b;
COMMIT;

-- mssql: connection not in Idle state
ATTACH '…' AS srv (TYPE mssql);
BEGIN TRANSACTION;
WITH a AS NOT MATERIALIZED (SELECT * FROM mssql_scan('srv','SELECT 1 AS x')),
     b AS NOT MATERIALIZED (SELECT * FROM mssql_scan('srv','SELECT 2 AS x'))
SELECT a.x + b.x + (SELECT count(*) FROM srv.dbo.ducklake_snapshot) FROM a, b;
COMMIT;
```

Both postgres scans report the **same** `pg_backend_pid()` (28396), so postgres pins one connection for the transaction exactly as this extension does. What differs is that libpq buffers a result set fully client-side, so the connection is back to idle before the second scan binds; the TDS path streams and leaves the connection `Executing` until the scan is drained.

The scan also runs its batch at **bind** time, to learn its result columns, and every source of a statement is bound before any of them is executed - so the two collide before either has a chance to be consumed. `MATERIALIZED` does not help, tried at the top level and nested (`WITH x AS (WITH s AS MATERIALIZED (…) SELECT * FROM s)`); materialisation reorders execution and the constraint is in binding.

Measured behaviour, for the record:

| | one `mssql_scan` beside a catalog scan | two `mssql_scan`s |
| --- | --- | --- |
| autocommit | works | works |
| explicit transaction | works | fails |

Autocommit is fine because each scan takes its own pooled connection.

## Why it matters

From [mssql_ducklake](https://github.com/hugr-lab/mssql-ducklake). DuckLake's postgres metadata manager overrides `GetLatestSnapshotQuery` and `GenerateFileColumnStatsCTEBody` to read through `postgres_query` instead of through the attached catalog, and the second one is where the rows are: one per file per filtered column. Measured on a synthetic catalog, the same filtered read through the catalog against a single server-side statement:

| files | stats rows | filter returns | through the catalog | one server-side scan | ratio |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 10,000 | 410,000 | 6 rows | 0.043s | 0.014s | 3.07x |
| 30,000 | 1,230,000 | 11 rows | 0.038s | 0.005s | 7.0x |
| 30,000 | 1,230,000 | 30,000 rows | 0.039s | 0.035s | 1.1x |

The catalog path costs what the catalog holds; the server-side form costs what the answer holds. A lake read is normally selective, so the shape favours the second - but a filter over two columns means two `mssql_scan`s in one statement, and DuckLake reads its metadata inside its own transaction, so that override cannot be written today.

## What would fix it

Either of these, and the first looks closest to machinery that already exists:

1. **Buffer a scan's result when the connection is pinned**, the way spec 003's R2 already materialises *catalog* scans inside a transaction - the same treatment applied to `mssql_scan` would return the connection to idle before the next source binds.
2. **Defer `BEGIN TRANSACTION` until the first write.** A read-only transaction does not need server-side isolation for a snapshot-addressed catalog like DuckLake's, and reads could then take pooled connections. This is the bigger change but it also removes the pin from the read path entirely, which is most of a read workload.

### LINKED ISSUE #334: Per-table metadata queries are one ad-hoc text per table: every first touch of a table pays a plan compile (~35 ms + ~48 ms)
## What happens

`SINGLE_TABLE_METADATA_SQL_TEMPLATE` (`src/catalog/mssql_metadata_cache.cpp:108`) and `PK_DISCOVERY_SQL_TEMPLATE` (`src/catalog/mssql_primary_key.cpp:38`) put the table name into the query text — `WHERE o.object_id = OBJECT_ID('[dbo].[x]')`. SQL Server caches ad-hoc plans by text, so each table is its own plan for each of the two shapes, and the first execution of every such text compiles a join over five catalog views. Measured: **28–37 ms** for the metadata query and **45–51 ms** for the PK query; the second execution of the same text is **0–1 ms**.

The cache miss path itself is fine — a table the cache has not seen is loaded with one single-table query, and `mssql_invalidate_cache(catalog, schema, table)` keeps the schema's table list (both verified in the `MSSQL_DEBUG=2` stream of the flow below: 24 single-table loads, zero re-listings). The cost is the compile, not any re-discovery.

## Evidence

All on one catalog: a DuckLake catalog of 1000 lake tables in `dbo` (~1200 tables with the inlined-data ones), SQL Server 2025-CU8 in Docker on a 2 GB VM, extension v0.2.5 on duckdb v1.5.5.

- `sys.dm_exec_query_optimizer_info` (`optimizations`) around the exact metadata text through `mssql_scan`: first run **+1, 34 ms**; second run **+0, 1 ms**. A different table name: +1 again.
- `SET STATISTICS TIME ON` (sqlcmd): `parse and compile time: CPU 28 ms, elapsed 28 ms`; execution 1 ms.
- The extension's own path (`MSSQL_DEBUG=1`, `RunMetadataQuery: completed in`), three untouched tables in a fresh session: **37 / 32 / 32 ms**; the same table after `mssql_invalidate_cache(…, table)`: 1 ms.
- Same query through `EXEC sp_executesql N'… OBJECT_ID(@n) …', N'@n sysname', @n = N'[dbo].[x]'` (sqlcmd, plain batch): the first table compiles once (29 ms), the following tables **0–1 ms** — one plan serves every name.
- `ALTER DATABASE lake_meta SET PARAMETERIZATION FORCED`: the extension's unchanged loads become **38 / 1 / 1 ms** — the server parameterizes the literal itself. (A TEMPLATE plan guide from `sp_get_query_template` + `OPTION (PARAMETERIZATION FORCED)` did not match in my one attempt.)
- Why it is paid again and again: the whole plan cache on this server holds ~53 plans (611 MB total server memory), so per-table plans do not survive to the next session. On a larger server they live longer, but with `optimize for ad hoc workloads` on — common in production — the first execution stores a stub and the second compiles again, so a touch-each-table-once pattern compiles for every table regardless.

## Where it lands (mssql-ducklake measurements)

- ATTACH + first read on the 1000-table catalog: 24 single-table loads × ~37 ms ≈ **0.9 s of 1.9 s**.
- The first inlined INSERT into each freshly created table: its new `ducklake_inlined_data_*` table is loaded on the miss, 40 ms — that is the entire premium of the first commit over the second (57 vs 37 ms per commit across 1000 tables; the postgres backend: 6 vs 10 ms).
- A table whose inlined data spans 154 schema versions: 154 first touches on one read.
- `mssql_preload_catalog('cat', 'dbo')` is not a way out here: it costs 0.95 s itself over ~1200 tables and saves ~0.5 s on the reads that follow.

## Options

1. **Parameterize both templates**: `EXEC sp_executesql N'<template with OBJECT_ID(@n)>', N'@n sysname', @n = N'[schema].[table]'`, sent as an ordinary SQL_BATCH — no RPC needed; one plan per shape for all tables (the sp_executesql line above). Blocked on #323 / #332: today the RETURNSTATUS desync makes `sp_executesql` through `mssql_scan` fail with `Unknown token type: 0x0` (reproduced on v0.2.5 with this exact query), and on the simple-query path the tokens after RETURNSTATUS are dropped silently. With #332 in, this is a template change plus the metadata path accepting the procedure's RETURNSTATUS/DONEPROC tail.
2. Outside the extension, documented rather than coded: `ALTER DATABASE … SET PARAMETERIZATION FORCED` on the catalog database gives the same effect database-wide (numbers above), at the price of a database-level setting and ALTER permission. Worth a line in the performance docs for catalogs with many tables.

## Repro

```sql
LOAD mssql;
ATTACH 'mssql:Server=localhost,7433;Database=lake_meta;User Id=sa;Password=…' AS m;
.timer on
-- optimizer counter, warmed twice so its own compile is out of the way
FROM mssql_scan('m', 'SELECT occurrence FROM sys.dm_exec_query_optimizer_info WHERE counter = ''optimizations''');
FROM mssql_scan('m', 'SELECT occurrence FROM sys.dm_exec_query_optimizer_info WHERE counter = ''optimizations''');
-- the SINGLE_TABLE_METADATA_SQL_TEMPLATE text with '[dbo].[ducklake_snapshot]' substituted, twice
FROM mssql_scan('m', '<template>');     -- ~34 ms, counter +1
FROM mssql_scan('m', '<template>');     -- ~1 ms,  counter +0
-- any other table name: +1 and ~34 ms again; under PARAMETERIZATION FORCED: +0 and ~1 ms
```

Filed from the specs/009 reconnaissance in hugr-lab/mssql-ducklake; the numbers there come from `make bench-scale --tables 1000` and the `MSSQL_DEBUG=2` stream of one CREATE TABLE + two inlined INSERTs.

https://claude.ai/code/session_01SANcUJwr9mKaY3xjr5jKgX

### LINKED ISSUE #336: mssql_scan: decouple bind from execution — metadata at bind (FMTONLY / sp_describe_first_result_set), the query at Init
## What happens

`mssql_scan` learns its result's column types by **executing the query at bind time** and keeping the result on the transaction's pinned connection until the scan consumes it. That is what makes the v0.2.5 rule *"inside a transaction, `mssql_scan` must be the sole source of its query"* — a plan mixing it with a catalog scan, or two `mssql_scan`s, fails depending on execution order (measured on 20,000-row results: `catalog JOIN mssql_scan` passes when the plan drains the stream first, the same pair as a CTE fails). #314 addresses it on the 2.0 line by draining the result at bind — correct, but it materialises every `mssql_scan` result in full at bind, whether or not the plan reads it, and keeps the bind-executes-the-query coupling.

## The proposal: decouple bind from execution

At bind, ask the server for the result's metadata **without executing**; execute in the operator, when the plan reads. Two ways to get the metadata, both plain `SQL_BATCH` (no RPC), measured on SQL Server 2025 against a 5000-row table:

| bind does | cost | returns |
| --- | ---: | --- |
| `SET FMTONLY ON; <query>; SET FMTONLY OFF` | ~0 ms | COLMETADATA of the result, zero rows — no procedure call, so it parses on today's token parser |
| `EXEC sp_describe_first_result_set N'<query>', NULL, 0` | 4 ms (parse and bind only, no optimisation) | name, `system_type_name` with length/precision/scale, nullability, collation — after #332, since it is a procedure |
| executing the query (what bind pays today) | 3 ms for 5000 rows + the materialisation + the connection held | |

`FMTONLY` is deprecated but present everywhere; `sp_describe_first_result_set` is SQL Server 2012+, Azure SQL, Synapse, Fabric. Either gives the bind exactly what it builds the DuckDB schema from.

What it changes:

- **The sole-source rule's bind half disappears**: nothing is open on the pinned connection after bind, so the ordering dependence is gone and a `mssql_scan` can sit beside catalog scans in a plan.
- **Execution is paid once, when read**, streaming outside a transaction. Inside a transaction one TDS connection holds one open result set, so the operator materialises at `Init` under the connection's lock — the same memory as #314's drain-at-bind, but later, and only for results the plan reads; two sources on one pinned connection serialise at `Init` instead of failing.
- A query with side effects (a user's `EXEC` through `mssql_scan`) no longer runs twice or at plan time.

A name like `mssql_scan_prepare` for the bind step, or the same `mssql_scan` doing it internally — the behaviour is the point: bind = metadata, `Init` = execute.

## Where it lands (mssql-ducklake)

The DuckLake metadata manager for SQL Server uses `mssql_scan` for its T-SQL reads — the latest snapshot, the conflict check, the inlined-deletion probe (hugr-lab/mssql-ducklake specs/005 D13, 007, 008) — and follows the sole-source rule for each. With metadata-only binds, reads that today stay on the catalog path because they cannot share a plan with one another could move to T-SQL, and the manager's rule would go with the 2.0 bump.

Related: #314 (drain at bind), #323 / #332 (procedure calls through the parser — what `sp_describe_first_result_set` needs), #334 (the per-table metadata query compiled once).

https://claude.ai/code/session_01SANcUJwr9mKaY3xjr5jKgX

VGSML:
Closes #336, closes #329, closes #239, closes #334.

Spec: `specs/075-read-path-pinned-connection/spec.md` (§ 0 is what was measured before anything was designed). Spec and implementation in one PR, as agreed.

## What changes

**`mssql_scan` no longer executes at bind (#336).** Bind asks `sp_describe_first_result_set` for the first result set's shape and maps the rows through the catalog's type mapping; the query runs at InitGlobal. A `DESCRIBE`/`EXPLAIN` of `INSERT ...; SELECT ...` inserts nothing; a statement with a side effect runs once. A statement the server cannot describe (a batch that reads a `#temp` table it creates, error 11525) still runs at bind exactly as before, stream registered on the catalog for InitGlobal. InitGlobal compares the executed COLMETADATA against the bound shape and fails loudly on drift rather than serving rows of another shape.

**`mssql_scan(..., prepared := true)`.** `sp_prepare` at bind: the server compiles once and answers with the statement's COLMETADATA (read exactly as the stream reads it), execution is `EXEC sp_execute <handle>` on the session that holds the handle — a pooled connection held from bind until the bind data dies in autocommit, the pinned connection in a transaction. No `sp_unprepare`: the session reset frees the handle. A prepare without a shape (a multi-statement batch) degrades to the describe.

**Two reads of one catalog inside a transaction (#329).** Every `mssql_scan` drains at init into a `ColumnDataCollection` under the catalog's `MaterializeMutex`, taken **before** the batch is sent (the order `table_scan.cpp` keeps). The optimizer's #239 rule now counts raw scans beside catalog scans, so a catalog scan next to an `mssql_scan` materialises too. Both of #329's shapes are in the tests.

**A sink reading the catalog it writes to, in a transaction (#239).** The optimizer collects the catalogs the plan sinks into (`LogicalCopyToFile` with the `bcp` function, `LogicalInsert` into an mssql catalog) and materialises every scan of those catalogs. Found while testing: DuckDB initialises the COPY sink on another thread while the source drains, so `BCPCopyInitGlobal` now waits on the same mutex in a transaction (held through its CREATE TABLE). INSERT needs nothing extra — it acquires the connection only when it sends a batch, after the source produced rows.

**Metadata queries parameterised (#334).** Table list, columns, PK, row count and the bulk-load query send `@s` / `@t` through `sp_executesql`; `OBJECT_ID(QUOTENAME(@s) + N'.' + QUOTENAME(@t))`. One text per shape, one server plan for every table. Measured on three fresh tables: `RunMetadataQuery` 0 / 1 / 1 ms, against ~30 ms per first touch before. The hand-rolled quote escaping in `mssql_statistics.cpp` goes with it.

**`mssql_scan_params` / `mssql_exec_params`.** `{'id': 42, 'since': TIMESTAMP '...'}` becomes

```sql
DECLARE @id int = 42, @since datetime2(6) = CAST('2024-01-02T00:00:00.000000' AS DATETIME2(7));
EXEC sp_executesql N'<statement>', N'@id int, @since datetime2(6)', @id = @id, @since = @since
```

Declarations derived per W5's table (`VARCHAR` → `nvarchar(4000)` / `nvarchar(max)` past 4000 bytes — SqlClient's two buckets; `MSSQL_VARCHAR(n)` exact; `TIMESTAMP` → `datetime2(6)`, `_S/_MS/_NS` → `datetime2(0/3/7)`; …) or given by the caller (`'@ts datetime, @c varchar(8)'`, both directions checked). The DECLARE line is why a CAST literal can travel: `EXEC proc @p = <expr>` accepts constants and variables only. Bare `NULL`, LIST/STRUCT, a key that is not an identifier: refused at bind with the fix named. `mssql_exec_params` runs one batch per row with the plan shared across rows and sessions.

## Found on the way

- **A refused `Initialize` closed the pinned connection.** `~MSSQLResultStream` closed any non-Idle connection; a stream whose `ExecuteBatch` was refused (pinned connection busy) closed the connection under the stream that was using it — SIGSEGV in `SSL_read` on the worker thread once scans moved to init. It now closes only a connection it sent its batch on (`batch_sent_`).
- **Catalog vs stream type mapping.** The catalog reports `datetime2(3/7)` as TIMESTAMP_MS/NS (spec 045, for the catalog scan); the stream reads every datetime2 into TIMESTAMP. The describe folds datetime2 to TIMESTAMP so `mssql_scan` keeps reporting what it always has; `mssql_scan_describe_types.test` holds the two mappings equal for every type the extension reads, and the init check is what fails if they drift.
- **Not fixed, noted:** a catalog read of a table created inside the same transaction loads its metadata on a *pooled* session and blocks on the schema lock (`copy_transaction.test` already reads such a table through `mssql_scan` for this reason). Pre-existing; the new sink test does the same.

## Tests

- New (W6): `query/mssql_scan_bind_describe`, `query/mssql_scan_describe_types`, `query/mssql_scan_prepared`, `query/mssql_params`, `transaction/sink_reads_own_catalog`, `catalog/metadata_query_parameterized`; `test/cpp/test_sql_params.cpp` (standalone, in `STANDALONE_TEST_SOURCES`).
- Full SQL suite locally: 197/199, the two being `azure/*` with error 40613 (serverless database resuming). Standalone C++: 23/23.

## Docs

CLAUDE.md (function table, read-path concept), DATAMODEL.md, `docs/query-execution.md`, `docs/TESTING.md`, `website/docs/reference/functions.md`, CHANGELOG (Added / Changed / Fixed).


VGSML:
Self-review, after CI (16 green, Windows gated). What I would raise as a reviewer, and where each stands:

1. **The datetime2 fold in `DescribeFirstResultSet`.** The catalog's mapping (`MapSQLServerTypeToDuckDB`) reports `datetime2(3/7)` as TIMESTAMP_MS/NS, the stream reads every datetime2 into TIMESTAMP. The fold keeps `mssql_scan` reporting what it always has; the alternative — teaching the stream the scale — would change `mssql_scan`'s result types for every user and is out of scope. `mssql_scan_describe_types.test` holds the two mappings equal for 38 columns, and InitGlobal fails loudly on any other divergence rather than serving rows of another shape.
2. **`prepared := true` in autocommit holds a pooled connection from bind until the bind data dies.** By design: the handle lives in that session. A statement with many prepared scans holds that many connections for its duration; the release goes through `~MSSQLPreparedSession` with the pool's `weak_ptr`, so a torn-down catalog degrades to closing the socket, as `~MSSQLResultStream` does. Documented in the spec (W2) and the functions reference.
3. **`~MSSQLResultStream` closing a busy pinned connection** was a latent bug before this PR (reachable only if two streams ever shared a pinned connection); moving scans to init made it reachable and it crashed the process in `SSL_read`. `batch_sent_` scopes the close to the stream that sent something. Worth a look on its own.
4. **W3's rule fires only for scans the optimizer can see.** A COPY whose source is a DuckDB-side table needs nothing; a COPY from the same catalog materialises the source. `BCPCopyInitGlobal` waits on `MaterializeMutex` because DuckDB initialises the sink on another thread while the source drains (seen in the trace, not reasoned). INSERT acquires the connection only when it sends a batch, so it needs nothing.
5. **Not fixed, noted in the sink test:** a catalog read of a table created inside the same transaction loads its metadata on a pooled session and blocks on the schema lock. Pre-existing (`copy_transaction.test` already reads such a table through `mssql_scan`).
6. **Small:** the `is_hidden` skip in the describe is moot with browse mode 0 and stays as a guard; `RunExecBatch`'s "Unknown context" text says `mssql_exec` for the params variant too — same function, one message.

Spec 076 (parameterised filter pushdown for catalog scans, one round trip per first touch) is drafted on top of this and waits for the merge.

VGSML:
Two commits on top, from the review of #343 (the findings it left to me):

- `9b8bf8f` **the describe reports the stream's type, by construction.** The described shape went through the catalog's `MapSQLServerTypeToDuckDB`, the executed one through `TypeConverter::GetDuckDBType`, and they disagree on more than the datetime2 scales the first draft folded by hand: a `rowversion` column described as VARCHAR and arrived as BLOB (a working scan failed the init check), and `sql_variant` / `hierarchyid` / `geography` — refused at bind with "Unsupported SQL Server type" before this PR — were refused at init with "Connection closed while waiting for COLMETADATA". Each describe row is now turned into the COLMETADATA the server would send and mapped by `GetDuckDBType`, so the two paths agree by construction, the unsupported types fail at bind with their old message, and the fold is gone. Self-review item 1 above is superseded. `mssql_scan_describe_types.test` gains rowversion and the two refusals.
- `c60fc3f` **the bulk load opens its stream on the first chunk, not when the sink is created.** `BCPCopyInitGlobal` sent INSERT BULK + COLMETADATA, and DuckDB runs it from an open job scheduled when the sink state is created — concurrently with the source's InitGlobal, which inside a transaction drains on the same pinned connection. So the COPY of #239 worked when the source won the race (15 of 15 runs here) and would have found the connection Executing when the sink did. `StartBulkStream()` now opens the stream under `write_mutex` on the first chunk that reaches the shared writer, and FlushToServer reopens it the same way; Finalize closes only a stream that was opened. Parallel writers are untouched.

Local: copy 28 cases / 878 assertions, ctas 11 / 437, transaction 9 / 282, insert 7 / 171, the spec-075 scan tests, and `sink_reads_own_catalog.test` 10 of 10 in a loop — all green. The third #343 finding (`Copy()` sharing the prepared session) is unreachable: DuckDB copies a `LogicalGet`'s bind data only for late materialization, which our functions do not enable.

VGSML:
@oluies ready for review at `2fd25b1`. Since roborev 1543 ran: your #343 is merged in (function name in the params messages, the shared `decimal(38,0)` rule, the case-collision backstop), and the two structural findings it left are done — the describe now maps through the stream's `TypeConverter::GetDuckDBType` (`9b8bf8f`), and the bulk load opens its stream on the first chunk instead of at sink creation (`c60fc3f`). CI is green on this head, including the Kerberos lane.

PR #342, opened 2026-09-11
fix: the four DML response loops kept both bugs #323 fixed in the fifth

oluies:
Follow-up to roborev reviews 1536 / 1541 / 1542 against `main` as merged, finished after the review here and the decision on #344. #332 and #333 had landed most of what those reviews found; four things had not, and the review of this PR found four more.

### The drain guard went into one loop of five

`MSSQLSimpleQuery::ExecuteWithCallback` stopped feeding a parser already in `ParserState::Error`. The INSERT batch, `INSERT ... RETURNING`, UPDATE and DELETE response loops are copies of the same loop and kept feeding it: `Feed()` appends unconditionally, `CompactBuffer()` runs only from `ConsumeBytes()`, and `TryParseNext()` returns at once in Error — so `buffer_` grew by the entire remaining response. All four now skip the feed and still read to end-of-message, which is what leaves the socket clean.

### And they reported a desync as success — or hung

The three executors forced `done` at EOM and fell through to "no error message, therefore fine", dropping every token after the desync — including a SQL Server ERROR following a procedure call. `INSERT ... RETURNING` was worse: its loop had no EOM exit, so it blocked on the next `ReceivePacket` for the full 30 s, reported the socket rather than the parse error, and left the connection `Executing`.

Each loop now records the parse error (a SQL error already captured keeps precedence) and the RETURNING loop leaves at EOM. Per #344 (decided: option A, as an interim until spec 062 / the UPDATE-DELETE rework remove the client-side batches), the message says what a desync means — the server **executed** this batch, the failure is in reading its answer — and how far the statement had got:

```
UPDATE failed: TDS parse error: ...; the server executed this batch (batch 3), and 1000 row(s) from the 2 batch(es) before it are applied
INSERT failed at statement 0 (rows 0-999): TDS parse error: ...; the server executed this statement, and 0 row(s) from the statements before it are applied
```

A parse error no longer renders as `[0]` (the ruling in `73e6da3`). Found on the way: the INSERT error named the statement *after* the failing one with an empty row range (`statement 1 (rows 1000-999)`), because the batch builder had already moved past the batch when it ran; `ExecuteBatch` now takes the batch and reports its own index and range.

### Three CTAS `Acquire` callers threw reasonless — four, with the cleanup DROP

`mssql_catalog.cpp` and `mssql_table_set.cpp` use `Acquire(-1, &why)`, the per-call reason from spec 073. The CTAS existence checks, its bulk load and its cleanup DROP did not, so a CTAS against a catalog whose Azure AD token has expired said `Failed to acquire connection to check table existence` and nothing more. All four carry the reason now.

### `mssql_pool_stats.last_create_error` could outlive the failure — or be blanked by a reuse

A pool at its limit recovers by `Release`, not by creation, so the error recorded by a failed creation stayed for the pool's life. The first draft cleared it on reuse, which is wrong twice: a reuse says nothing about whether **login** works (an expired token beside warm connections would blank a live failure on every idle serve), and the same lines reset the #302 creation backoff, so a failing factory was dialled on every wakeup. The reuse path now touches neither; the error carries its age instead — `last_create_error_age_ms`, NULL when there is none — so a reader can tell a live failure from an outlived one.

### Test

`test/sql/regression/issue_344_dml_desync.test` drives all four loops through a new test-only setting, `mssql_test_fail_parse_after_tokens`, which puts the response parser into Error after N tokens the way a desync would (a desync cannot be induced from SQL since spec 072; same shape as `mssql_test_fail_metadata_after_rows`, #317). Each case checks the statement fails naming the desync and what was executed, that the rows the server did apply are there, and that the connection is usable afterwards. The RETURNING case is the one that used to take 30 s.

Verified locally: regression / dml / insert / ctas / tds_connection / transaction / copy suites and `make test-cpp-run` green; details in the last comment.

### DIFF
--- a/src/dml/insert/mssql_insert_executor.cpp
+++ b/src/dml/insert/mssql_insert_executor.cpp
@@ -186,9 +186,15 @@ idx_t MSSQLInsertExecutor::ExecuteBatch(const string &sql) {
 
 			bool is_eom = packet.IsEndOfMessage();
 
-			// Feed packet payload to parser
+			// Feed packet payload to parser -- but not into a parser already in
+			// Error. From there TryParseNext returns at once, so ConsumeBytes and
+			// with it CompactBuffer never run, and every fed byte is retained;
+			// a desync early in a batched response buffers the whole tail for
+			// nothing. The drain to EOM below still runs: that is what leaves the
+			// socket clean for the next statement. MSSQLSimpleQuery already does
+			// this (issue #323); these four loops are copies of the same loop.
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
 				parser.Feed(payload);
 			}
 
@@ -226,6 +232,18 @@ idx_t MSSQLInsertExecutor::ExecuteBatch(const string &sql) {
 					// Skip other tokens
 					break;
 				}
+			}
+
+			// A parser stuck in Error answers NeedMoreData forever, so the token
+			// loop above has already exited and the EOM branch below forces done
+			// and reports SUCCESS -- dropping every token after the desync,
+			// including a SQL Server ERROR token following a stored-procedure
+			// call, which is the one a caller most needs. Record it as the
+			// failure it is. A SQL error already captured keeps precedence.
+			if (parser.GetState() == tds::ParserState::Error && error_message.empty()) {
+				error_number = 0;
+				error_message = "TDS parse error: " + parser.GetParseError();
+				INSERT_DEBUG(1, "ExecuteBatch: %s", error_message.c_str());
 			}
 
 			// Handle EOM without done token
--- a/src/dml/insert/mssql_returning_parser.cpp
+++ b/src/dml/insert/mssql_returning_parser.cpp
@@ -130,8 +130,11 @@ unique_ptr<DataChunk> MSSQLReturningParser::Parse(tds::TdsConnection &connection
 				return nullptr;
 			}
 
+			// Not into a parser already in Error -- see the DML executors: from
+			// there nothing consumes, so buffer_ grows by the whole remaining
+			// response (issue #323).
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
 				parser.Feed(payload);
 			}
 
--- a/src/dml/update/mssql_update_executor.cpp
+++ b/src/dml/update/mssql_update_executor.cpp
@@ -280,9 +280,15 @@ idx_t MSSQLUpdateExecutor::ExecuteBatch(const string &sql) {
 
 			bool is_eom = packet.IsEndOfMessage();
 
-			// Feed packet payload to parser
+			// Feed packet payload to parser -- but not into a parser already in
+			// Error. From there TryParseNext returns at once, so ConsumeBytes and
+			// with it CompactBuffer never run, and every fed byte is retained;
+			// a desync early in a batched response buffers the whole tail for
+			// nothing. The drain to EOM below still runs: that is what leaves the
+			// socket clean for the next statement. MSSQLSimpleQuery already does
+			// this (issue #323); these four loops are copies of the same loop.
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
 				parser.Feed(payload);
 			}
 
@@ -318,6 +324,18 @@ idx_t MSSQLUpdateExecutor::ExecuteBatch(const string &sql) {
 					// Skip other tokens
 					break;
 				}
+			}
+
+			// A parser stuck in Error answers NeedMoreData forever, so the token
+			// loop above has already exited and the EOM branch below forces done
+			// and reports SUCCESS -- dropping every token after the desync,
+			// including a SQL Server ERROR token following a stored-procedure
+			// call, which is the one a caller most needs. Record it as the
+			// failure it is. A SQL error already captured keeps precedence.
+			if (parser.GetState() == tds::ParserState::Error && error_message.empty()) {
+				error_number = 0;
+				error_message = "TDS parse error: " + parser.GetParseError();
+				UPDATE_DEBUG(1, "ExecuteBatch: %s", error_message.c_str());
 			}
 
 			// Handle EOM without done token
--- a/src/tds/tds_connection_pool.cpp
+++ b/src/tds/tds_connection_pool.cpp
@@ -173,6 +173,18 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms, std::stri
 		// Try to get an idle connection
 		auto conn = TryAcquireIdle();
 		if (conn) {
+			// A usable connection is evidence the pool is working, however it was
+			// obtained. The creation path clears the recorded error for exactly
+			// that reason; a pool at or near its limit recovers by REUSE and
+			// never reaches that path, so without this last_create_error_ stays
+			// set for the pool's life. The thrown message no longer depends on
+			// it (own_create_error carries this call's reason), but
+			// mssql_pool_stats.last_create_error does -- and it would show a
+			// long-dead reason for a healthy pool, which is the opposite of what
+			// that column was added for.
+			create_backoff_ms_ = CREATE_BACKOFF_INITIAL_MS;
+			next_create_allowed_ = std::chrono::steady_clock::time_point{};
+			last_create_error_.clear();
 			auto elapsed =
 				std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count();
 			stats_.acquire_wait_total_ms += elapsed;
--- a/src/dml/ctas/mssql_ctas_executor.cpp
+++ b/src/dml/ctas/mssql_ctas_executor.cpp
@@ -232,10 +232,10 @@ bool CTASExecutionState::TableExists(ClientContext &context) {
 						   MSSQLDDLTranslator::EscapeStringLiteral(target.table_name));
 
 	auto &pool = catalog->GetConnectionPool();
-	auto conn = pool.Acquire();
+	std::string why;
+	auto conn = pool.Acquire(-1, &why);
 	if (!conn) {
-		throw IOException("Failed to acquire connection to check table existence: " +
-						  pool.DescribeAcquireFailure());
+		throw IOException("Failed to acquire connection to check table existence: " + why);
 	}
 
 	try {
@@ -257,10 +257,10 @@ bool CTASExecutionState::SchemaExists(ClientContext &context) {
 										  MSSQLDDLTranslator::EscapeStringLiteral(target.schema_name));
 
 	auto &pool = catalog->GetConnectionPool();
-	auto conn = pool.Acquire();
+	std::string why;
+	auto conn = pool.Acquire(-1, &why);
 	if (!conn) {
-		throw IOException("Failed to acquire connection to check schema existence: " +
-						  pool.DescribeAcquireFailure());
+		throw IOException("Failed to acquire connection to check schema existence: " + why);
 	}
 
 	try {
@@ -543,10 +543,10 @@ void CTASExecutionState::ExecuteBCPInsert(ClientContext &context) {
 	// complete — it did not exist before the statement — and needs no shared
 	// transaction. That is what mssql_ctas_drop_on_failure does.
 	auto &pool = catalog->GetConnectionPool();
-	connection = pool.Acquire();
+	std::string why;
+	connection = pool.Acquire(-1, &why);
 	if (!connection) {
-		throw IOException("CTAS BCP: Failed to acquire connection from pool: " +
-						  pool.DescribeAcquireFailure());
+		throw IOException("CTAS BCP: Failed to acquire connection from pool: " + why);
 	}
 
 	try {
--- a/src/dml/ctas/mssql_ctas_executor.cpp
+++ b/src/dml/ctas/mssql_ctas_executor.cpp
@@ -234,7 +234,8 @@ bool CTASExecutionState::TableExists(ClientContext &context) {
 	auto &pool = catalog->GetConnectionPool();
 	auto conn = pool.Acquire();
 	if (!conn) {
-		throw IOException("Failed to acquire connection to check table existence");
+		throw IOException("Failed to acquire connection to check table existence: " +
+						  pool.DescribeAcquireFailure());
 	}
 
 	try {
@@ -258,7 +259,8 @@ bool CTASExecutionState::SchemaExists(ClientContext &context) {
 	auto &pool = catalog->GetConnectionPool();
 	auto conn = pool.Acquire();
 	if (!conn) {
-		throw IOException("Failed to acquire connection to check schema existence");
+		throw IOException("Failed to acquire connection to check schema existence: " +
+						  pool.DescribeAcquireFailure());
 	}
 
 	try {
@@ -543,7 +545,8 @@ void CTASExecutionState::ExecuteBCPInsert(ClientContext &context) {
 	auto &pool = catalog->GetConnectionPool();
 	connection = pool.Acquire();
 	if (!connection) {
-		throw IOException("CTAS BCP: Failed to acquire connection from pool");
+		throw IOException("CTAS BCP: Failed to acquire connection from pool: " +
+						  pool.DescribeAcquireFailure());
 	}
 
 	try {
--- a/src/dml/delete/mssql_delete_executor.cpp
+++ b/src/dml/delete/mssql_delete_executor.cpp
@@ -244,9 +244,15 @@ MSSQLDMLResult MSSQLDeleteExecutor::ExecuteBatch(const MSSQLDMLBatch &batch) {
 
 			bool is_eom = packet.IsEndOfMessage();
 
-			// Feed packet payload to parser
+			// Feed packet payload to parser -- but not into a parser already in
+			// Error. From there TryParseNext returns at once, so ConsumeBytes and
+			// with it CompactBuffer never run, and every fed byte is retained;
+			// a desync early in a batched response buffers the whole tail for
+			// nothing. The drain to EOM below still runs: that is what leaves the
+			// socket clean for the next statement. MSSQLSimpleQuery already does
+			// this (issue #323); these four loops are copies of the same loop.
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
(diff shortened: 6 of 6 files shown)


src/tds/tds_connection_pool.cpp
```diff
@@ -173,6 +173,18 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms, std::stri
...
+			// A usable connection is evidence the pool is working, however it was
+			// obtained. The creation path clears the recorded error for exactly
+			// that reason; a pool at or near its limit recovers by REUSE and
+			// never reaches that path, so without this last_create_error_ stays
+			// set for the pool's life. The thrown message no longer depends on
+			// it (own_create_error carries this call's reason), but
+			// mssql_pool_stats.last_create_error does -- and it would show a
+			// long-dead reason for a healthy pool, which is the opposite of what
+			// that column was added for.
+			create_backoff_ms_ = CREATE_BACKOFF_INITIAL_MS;
```
VGSML: These two lines change the backoff contract, not the stats column: a reuse by any thread re-arms immediate creation for the thread that is failing to create. Under an expired token with live connections that is a dial per wakeup — what the backoff in #302 was for. Suggest keeping only `last_create_error_.clear()` here (if at all; see the review body) and leaving `create_backoff_ms_` / `next_create_allowed_` to the creation-success path.

src/dml/insert/mssql_returning_parser.cpp
```diff
@@ -130,8 +130,11 @@ unique_ptr<DataChunk> MSSQLReturningParser::Parse(tds::TdsConnection &connection
 				return nullptr;
 			}
 
+			// Not into a parser already in Error -- see the DML executors: from
+			// there nothing consumes, so buffer_ grows by the whole remaining
+			// response (issue #323).
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
```
VGSML, line 163: The guard is right, but this loop is the fourth copy and it still reports a desync as success: after the loop `Parse` returns by `row_count_` and never checks `parser.GetState()`, so `error_message_` stays empty and `ExecuteWithReturning` (which decides by `HasError()`) hands back the rows parsed so far. Before the `SetChildCardinality` / return:

```cpp
if (parser.GetState() == tds::ParserState::Error) {
	error_message_ = "TDS parse error: " + parser.GetParseError();
	return nullptr;
}
```

VGSML (dismissed):
Built and run locally at `72a130f`: insert / dml / ctas / regression / tds_connection / transaction — 40 cases, 1213 assertions, 2 skipped, all green. The four Feed guards and the three executors' parse-error recording read right and match the `MSSQLSimpleQuery` reference. Three things keep this from approve:

**1. The pool hunk resets the creation backoff on reuse, not just the recorded error** (inline). The description argues about `mssql_pool_stats.last_create_error`; the code also zeroes `create_backoff_ms_` / `next_create_allowed_` every time an idle connection is handed out. With a broken factory (expired token) and live connections being reused, a caller that finds no idle one dials on every wakeup — each other thread's reuse re-arms it — which is the behaviour the backoff was added to stop (#302, review 1538). Keep the `last_create_error_.clear()` if you want it; leave the backoff alone on the reuse path.

On the clear itself, a judgement call rather than a defect: reuse proves the pool works, not that the factory does. With an expired token and live connections the column reads empty most of the time and flickers on each failed creation. Spec 073's contract was "cleared by a creation success"; the stale-after-Release case might be better served by an age (timestamp) than by clearing.

**2. The RETURNING parser still reports a desync as success** (inline). `mssql_returning_parser.cpp` gets only the Feed guard. After the loop, `Parse` returns the chunk by `row_count_` and never looks at the parser's state; `error_message_` stays empty, and `MSSQLInsertExecutor::ExecuteWithReturning` decides by `parser.HasError()`. So `INSERT ... RETURNING` after a desync returns the rows parsed so far as a success — the fourth copy of the loop keeps exactly the bug the other three lose here. One check before the return closes it.

**3. Text and coverage.**
- The description still says `mssql_catalog.cpp` and `mssql_table_set.cpp` "render `DescribeAcquireFailure()`"; they use `Acquire(-1, &why)`, as the CTAS callers now do.
- `mssql_ctas_executor.cpp:352` (the cleanup DROP) is a fourth reasonless `Acquire()`; `cleanup_error = "no connection available for cleanup DROP"` could carry `why` the same way.
- `test/sql/regression/issue_323.test` drives `mssql_exec` / `mssql_scan`, i.e. `MSSQLSimpleQuery` and the result stream — not these four loops, so "already drives end to end" does not hold for them. A desync cannot be induced from SQL any more (spec 072 taught the parser RETURNSTATUS; RETURNVALUE is RPC-only), and the project's answer to that shape was a test-only lever (`mssql_test_fail_metadata_after_rows`, #317). At minimum the text; better, a lever that puts the parser in Error after N tokens and a test that sees success-with-dropped-rows before this fix and an error after.

src/tds/tds_connection_pool.cpp
```diff
@@ -173,6 +173,18 @@ std::shared_ptr<TdsConnection> ConnectionPool::Acquire(int timeout_ms, std::stri
...
+			next_create_allowed_ = std::chrono::steady_clock::time_point{};
+			last_create_error_.clear();
```
oluies: **Two defects, and the whole 3-line reset should come out.** Both skeptics returned SURVIVES on different mechanisms.

**1. It disarms the #302 backoff — single-threaded, inside one `Acquire()`:**
1. iteration 1: no idle → `CreateNewConnection` fails (expired AAD token / `Login failed`) → arms `next_create_allowed_ = now + 250ms`, `create_backoff_ms_ = 500`; `active_connections != 0` so it waits
2. another caller `Release()`s → `available_cv_` wakes it
3. iteration 2: `TryAcquireIdle()` succeeds → **these lines reset the backoff to 250ms and `next_create_allowed_` to the epoch** → the *next* caller dials immediately

That is precisely the "others are active" case the backoff exists for (`tds_connection_pool.hpp:129-135`). An idle reuse is evidence a previously-created connection still works — it says nothing about whether **login** works.

*What the skeptic did refute:* the empty-pool path. There `active_connections == 0 ||` short-circuits and `next_create_allowed_` is never read, so main already re-dials with no delay, deliberately (`TestBackToBackFailuresStayFast`). That half is pre-existing by design.

**2. Clearing `last_create_error_` blanks the diagnostic in the exact #302 scenario.** An expired Azure AD token leaves established TDS sessions fully usable while every new LOGIN7 fails. Warm connections keep serving, so **nothing throws** and pool stats are the only channel carrying the reason — and every idle serve blanks it. The operator sees `creation_failures` climbing with `last_create_error = ''`. `CreateNewConnection`'s log of the reason is `MSSQL_POOL_DEBUG_LOG(1, …)`, off by default, so once blanked it is unrecoverable. The stale-error problem this fixes is *recoverable* (sample `creation_failures` twice); a blanked live error is not.

**3. These lines are dead code under `make test-cpp`.** All six cases in `test/cpp/test_pool_creation_failure.cpp` build connections with `make_shared<TdsConnection>()`, which starts `Disconnected`, so `Release()` closes them and nothing ever enters `idle_connections_` — `TryAcquireIdle()` returns `nullptr` in every iteration of every test. The green run proves nothing here.

**Fix:** leave `create_backoff_ms_` and `next_create_allowed_` untouched on the reuse path, and either timestamp the error (`last_create_error_at_`, age it out) or gate the clear on `next_create_allowed_ <= now`.

src/dml/insert/mssql_returning_parser.cpp
oluies, line 163: **This file got half the fix, and half is worse than none.**

It has the `Feed` guard but **not** the error block the three executors got — and `Parse()`'s loop is `while (!done)` with no `IsEndOfMessage()` check at all. `done` is set only by a DONE-final token (`:105-108`), which a parser in `Error` can never produce.

So on a desync during `INSERT … OUTPUT INSERTED`:
- remaining packets are read and discarded fast (the guard skips `Feed` — the memory fix works)
- the next `socket.ReceivePacket(packet, recv_timeout)` at `:128` blocks for the whole remaining budget, since the deadline check at `:66` has not expired
- returns false → `error_message_ = "Failed to receive TDS packet: " + socket.GetLastError()` → `nullptr`

Call site is `mssql_insert_executor.cpp:331`, `parser.ParseResponse(*connection, 30000)`. **User-visible: a ~30 s hang, then an error naming the socket instead of the desync** — while `parser.GetParseError()` holds the real cause and is never read. It never reaches the deadline branch, so no ATTENTION is sent either, and the connection is left `Executing` so every such statement burns a pooled connection.

**Fix:** give this loop the same error block, or an `is_eom` check, so it terminates fast and reports `GetParseError()`.

src/dml/update/mssql_update_executor.cpp
```diff
@@ -320,6 +326,18 @@ idx_t MSSQLUpdateExecutor::ExecuteBatch(const string &sql) {
...
 
+			// A parser stuck in Error answers NeedMoreData forever, so the token
+			// loop above has already exited and the EOM branch below forces done
+			// and reports SUCCESS -- dropping every token after the desync,
+			// including a SQL Server ERROR token following a stored-procedure
+			// call, which is the one a caller most needs. Record it as the
+			// failure it is. A SQL error already captured keeps precedence.
+			if (parser.GetState() == tds::ParserState::Error && error_message.empty()) {
+				error_number = 0;
+				error_message = "TDS parse error: " + parser.GetParseError();
```
oluies, line 346: **Partial-application hazard in autocommit — the judgement call on this PR.** Same applies to `mssql_delete_executor.cpp:301`.

`defer_execution_` is set **only** when `!context.transaction.IsAutoCommit()` (`:47-60`). In autocommit — the default — `Execute()` flushes every `mssql_dml_batch_size` (500) rows as an **independent, already-committed** statement (`:90-96`), each on its own pooled connection.

A TDS desync is a **client-side framing** failure: the server executed and committed the batch. Before this change: silent success. After: batch K throws out of `Sink`, and batches 1..K — **including the one that "failed"** — remain permanently applied. There is no savepoint or compensation path in `src/dml` (`GetNextSavepointName()` at `mssql_transaction.cpp:197` is never called from DML). The user gets an error plus a half-applied UPDATE with no way to tell how far it got.

Under an explicit transaction this is safe: deferred batches flush on the pinned connection inside the open `BEGIN TRANSACTION` and rollback covers them. The hazard is specific to autocommit.

This may still be the right trade — a visible failure beats a silent one — but it is a behaviour change that deserves a deliberate decision, a CHANGELOG entry, and ideally a message that says how many batches were applied.

src/dml/insert/mssql_insert_executor.cpp
```diff
@@ -228,6 +234,18 @@ idx_t MSSQLInsertExecutor::ExecuteBatch(const string &sql) {
...
 
+			// A parser stuck in Error answers NeedMoreData forever, so the token
+			// loop above has already exited and the EOM branch below forces done
+			// and reports SUCCESS -- dropping every token after the desync,
+			// including a SQL Server ERROR token following a stored-procedure
+			// call, which is the one a caller most needs. Record it as the
+			// failure it is. A SQL error already captured keeps precedence.
+			if (parser.GetState() == tds::ParserState::Error && error_message.empty()) {
+				error_number = 0;
+				error_message = "TDS parse error: " + parser.GetParseError();
```
oluies, line 255: **This re-introduces the `[0]` mislabelling that commit `73e6da3` removed one commit earlier in the same chain.**

That commit's ruling: *"a parse error is not 'SQL Server error 0' … `error_number` 0 now formats without the prefix."* But `MSSQLInsertError`'s formatter (`mssql_insert_error.hpp:49-54`) prints the code unconditionally:

```cpp
StringUtil::Format("INSERT failed at statement %d (rows %d-%d): [%d] %s", …, sql_error_number, sql_error_message);
```

So setting `error_number = 0` here renders:

```
INSERT failed at statement 3 (rows 2000-2999): [0] TDS parse error: Unknown token type: 0xE4
```

— the same server-blaming `[0]` that ruling stripped from `mssql_exec`. UPDATE/DELETE escape it only because `MSSQLDMLResult::FormatError` happens not to print `error_code`.

**Fix:** make `MSSQLInsertError`'s formatter omit the `[%d]` when `sql_error_number == 0`, matching what `73e6da3` did for the other path.

src/dml/insert/mssql_insert_executor.cpp
```diff
@@ -186,9 +186,15 @@ idx_t MSSQLInsertExecutor::ExecuteBatch(const string &sql) {
...
+			// Feed packet payload to parser -- but not into a parser already in
+			// Error. From there TryParseNext returns at once, so ConsumeBytes and
+			// with it CompactBuffer never run, and every fed byte is retained;
+			// a desync early in a batched response buffers the whole tail for
+			// nothing. The drain to EOM below still runs: that is what leaves the
+			// socket clean for the next statement. MSSQLSimpleQuery already does
+			// this (issue #323); these four loops are copies of the same loop.
 			const auto &payload = packet.GetPayload();
-			if (!payload.empty()) {
+			if (!payload.empty() && parser.GetState() != tds::ParserState::Error) {
```
oluies, line 201: **Over-engineering lens:** this ~7-line comment is pasted verbatim into three files (`insert:189`, `update:283`, `delete:247`), and the ~5-line comment + 6-line error block is pasted into three more. That is ~45 of the 85 added lines.

`mssql_returning_parser.cpp:133` already does it right — a 2-line pointer to "the DML executors" rather than the full explanation. That is the model, not the exception.

**Suggested:** hoist the explanation once onto `TokenParser::Feed` / `ParserState::Error` in `src/include/tds/tds_token_parser.hpp`, leave `// Skip Feed once errored — see TokenParser::Feed doc (issue #323).` at each site, and extract `bool RecordParserError(TokenParser&, int32_t&, std::string&)` for the error block.

Worth doing here specifically because this PR's own commit message names the cost of the duplication: the four loops kept a bug the fifth had fixed, for multiple releases.

oluies (commented):
## Adversarial self-review — 5 lenses + 2 skeptics

Ran a hostile multi-lens review over this diff (correctness, security, blast-radius, project convention, over-engineering), then a refutation pass whose job was to disprove each finding. 4 survived on this PR. **I would not merge this as written.**

Inline below. Two things that have no single line to anchor to:

### No test reaches any of the four changed loops

`git diff --name-only 2e6e949...HEAD -- test/` is empty, and `test/sql/regression/issue_323.test` drives **only** `mssql_exec()` and `mssql_scan()` — i.e. `MSSQLSimpleQuery`, the *fifth* loop, which was already fixed. The only `INSERT` text in that file is inside procedure *bodies*, executed server-side. It never issues a DuckDB-native `INSERT`/`UPDATE`/`DELETE`/`RETURNING` against an attached table.

`test/cpp/test_token_parser_tokens.cpp` is byte-level parser tests with no executor. So every behavioural claim in this PR rests on reading the code, not on running it. The suite passing locally (23/23 standalone, 181 integration cases) is **not** evidence for these lines.

### Well-formed responses can now throw where they previously succeeded

`ParserState::Error` is confirmed terminal, and a normal SQL Server ERROR token (0xAA) never transits through it — so the guard cannot block legitimate errors. But five sites *do* set `Error` on responses the server is entitled to send:

- unhandled token bytes: `SESSIONSTATE (0xE4)`, `FEATUREEXTACK (0xAE)`, `ALTMETADATA (0x88)`, `ALTROW (0xD3)` — the last two arrive from `COMPUTE BY` inside a trigger
- `ParseColMetadata`/`ParseRow`/`ParseNBCRow` decode exceptions: a DML target with an `AFTER` trigger that `SELECT`s emits COLMETADATA + ROW into the DML response. The three DML loops skip those tokens, but the parser still **decodes** them, and any column type `RowReader` can't decode now throws where it previously skipped.

`grep -rl TRIGGER test/` is empty, so none of this is exercised.

### What held up

The connection-lifetime concern did **not** reproduce. `Release` closes any non-`Idle` connection, so every throw path drops rather than poisons it — no #191-class leak, no poisoned pooled connection, no transaction-state corruption on the pinned path. The new block also cannot spin, and `GetParseError()` cannot be empty.

oluies:
The autocommit partial-application question from the review is now **#344** — it needs a semantics decision rather than a patch, so I did not want to settle it inside this PR.

The other findings here do not depend on it and I can act on them now:

- revert the `create_backoff_ms_` / `next_create_allowed_` half of the pool hunk, and timestamp `last_create_error_` instead of clearing it
- give `mssql_returning_parser.cpp` the error block so `INSERT … RETURNING` fails fast naming the desync instead of hanging 30 s naming the socket
- make `MSSQLInsertError` omit `[%d]` when the code is 0, matching the ruling in `73e6da3`
- hoist the duplicated comment onto `TokenParser::Feed` and extract `RecordParserError()`
- add DML-path coverage, which is absent today

Holding until #344 lands, since option C would remove the error-reporting block entirely and make two of those moot.

VGSML:
@oluies #344 is decided (option A as an interim: throw, applied count in the message, no compensation or setting; the batches themselves go away with spec 062 / the 065 rework), so this can move. Your own list, trimmed to what outlives 062/065 — the three executors' response loops are on borrowed time, so no hoisting or `RecordParserError()` needed for them:

1. **Pool:** revert the `create_backoff_ms_` / `next_create_allowed_` half of the reuse hunk; keep `last_create_error_` and date it rather than clear it.
2. **RETURNING:** check the parser state (and EOM) in `Parse()` so a desync fails fast naming `GetParseError()` instead of hanging 30 s on the socket.
3. **The three executors:** throw as you have it, with how many batches / rows were applied before the failing one in the message, and no `[%d]` for code 0 (`73e6da3`'s ruling).
4. **Text:** the description's `DescribeAcquireFailure()` line; `mssql_ctas_executor.cpp:352` with a reason like the other three.
5. **Coverage:** a test for the RETURNING hang if there is a cheap way to put the parser in Error; if not, say so in the PR rather than pointing at `issue_323.test`, which drives only `mssql_exec` / `mssql_scan`.

No deadline from our side. If it sits for a couple of days I can finish it on this branch with your commits kept.

VGSML:
Finished on this branch, your two commits kept. What changed since `72a130f`, per the list above: the RETURNING loop now names the desync and leaves at EOM instead of blocking 30 s on the socket; the three executors' messages say the server executed the batch and how many rows the batches before it applied (#344, option A as an interim); no `[0]` for a parse error; the pool's reuse path leaves the backoff and the recorded error alone and the error carries `last_create_error_age_ms`; the cleanup DROP names its reason; and `mssql_test_fail_parse_after_tokens` drives all four loops in `test/sql/regression/issue_344_dml_desync.test`. Found on the way and fixed: an INSERT error named the statement after the failing one with an empty range (`statement 1 (rows 1000-999)`).

Local, this head: regression 5 cases / 103 assertions (the new one 28, in 1 s), dml 7 / 235, insert 7 / 171, ctas 11 / 437, tds_connection 3 / 34, transaction 8 / 258, copy 28 / 878; `make test-cpp-run` 29/29. The description is updated to match.

PR #343, opened 2026-09-11
review(075): the three findings that are a line each, not a design change

oluies:
Stacked on `spec/075-read-path-pinned-connection`. Addresses the three findings from roborev 1543 that are a line each; the structural ones are left to you, with evidence, below.

### `mssql_exec_params` reported failures under the wrong function name

`RunExecBatch` hard-coded `"mssql_exec:"` in all four of its messages — including the read-only refusal and the unknown-context help, which are the two a caller is most likely to hit. `BindExecContextName` was parameterized with `function_name` for exactly this reason; `RunExecBatch` was not. It now takes one too.

```
SELECT mssql_exec_params('nope', 'SELECT 1', {'a': 1})
  before: mssql_exec: Unknown context 'nope'...
  after:  mssql_exec_params: Unknown context 'nope'...
```

### W5's duplicate-parameter refusal was never implemented

The `seen` map is built, commented as the thing that catches two keys differing only in case, and then read by nothing on that path. T-SQL variable names are case-insensitive, so `{'a': 1, 'A': 2}` declared `@a` twice and the server answered `The variable name '@A' has already been declared` — which does not say which struct key to change. Refused client-side now, naming both spellings.

### `HUGEINT` / `UHUGEINT` were declared `decimal(38,0)`, which is narrower

`decimal(38,0)` holds ±(10^38 − 1). `HUGEINT` reaches ~1.7e38 and `UHUGEINT` ~3.4e38, so the widest values became literals the server rejects with an arithmetic overflow naming neither the parameter nor the cause. Range-checked on the rendered digits — exact, since that text is the literal we emit — and refused by name, as the `LIST` and `NULL` arms already do.

---

## Not fixed here — the High, with the decisive evidence

**`copy_function.cpp:328-338`. The `materialize_lock` gives mutual exclusion, not ordering, and this is not a maybe.**

DuckDB's `Pipeline::Reset()` calls `ResetSink()` (line 221) and only then `ResetSource(false)` (line 229). So `BCPCopyInitGlobal` **always** runs first. And it does not merely find the connection Idle: while holding the lock it runs `INSERT BULK` (`:517`) and transitions `Idle → Executing` (`:523`), then writes COLMETADATA (`:556`). The flagged source scan's `InitGlobal` then acquires the freed mutex and hits the exact failure W3 exists to remove:

```
Cannot execute: connection not in Idle state (current: Executing)
```

Only source-first works, and nothing enforces it — so `test/sql/transaction/sink_reads_own_catalog.test`'s COPY half passes or fails on init-order luck.

Repro: `BEGIN; COPY (SELECT id, v FROM sk75.dbo.sk75_src) TO 'mssql://sk75/dbo/sk75_copy' (FORMAT 'bcp');`

The fix is structural — defer `INSERT BULK` + `WriteColmetadata()` out of `BCPCopyInitGlobal` to the first `BCPCopySink()` call, so the connection is genuinely Idle when the source initializes — and it touches the parallel-writer path, which is yours to make rather than mine to guess at.

**Two Mediums also left to you**, both needing a call on intended behaviour: the describe-vs-execute type divergence (`sp_describe_first_result_set` reports `rowversion` as `timestamp` → `MapSQLServerTypeToDuckDB`'s VARCHAR catch-all, while the wire type is `BIGVARBINARY` → BLOB, so the new strict shape check turns a working scan into a hard error); and `MSSQLPreparedSession` being shared by `Copy()` while owning one connection with one `sp_prepare` handle.

---

`src/query/mssql_sql_params.cpp` and `src/mssql_functions.cpp` pass `clang-format-14 --dry-run -Werror`. Not built or tested locally — the `duckdb` submodule is not checked out in the authoring worktree.

### DIFF
--- a/src/mssql_functions.cpp
+++ b/src/mssql_functions.cpp
@@ -798,7 +798,7 @@ static int64_t RunExecBatch(ClientContext &client_context, const string &context
 		throw InvalidInputException(
 			"%s: Unknown context '%s'. Attach a database first with: ATTACH '' AS %s (TYPE mssql, SECRET "
 			"...)",
-			context_name, context_name);
+			function_name, context_name, context_name);
 	}
 	auto &catalog = *catalog_ptr;
 	if (catalog.IsReadOnly()) {
--- a/src/mssql_functions.cpp
+++ b/src/mssql_functions.cpp
@@ -782,13 +782,13 @@ static bool ExecSqlMayChangeSchema(const string &sql) {
 // text for the DDL heuristic -- for mssql_exec_params the batch wraps it in
 // sp_executesql, whose name would otherwise trip the EXEC keyword every time.
 static int64_t RunExecBatch(ClientContext &client_context, const string &context_name, const string &sql,
-							const string &statement) {
+							const string &statement, const char *function_name) {
 	// Get the MSSQL catalog (Spec 047: per-catalog ownership)
 	MSSQLCatalog *catalog_ptr = nullptr;
 	try {
 		auto &raw_catalog = Catalog::GetCatalog(client_context, Identifier(context_name));
 		if (raw_catalog.GetCatalogType() != "mssql") {
-			throw InvalidInputException("mssql_exec: Context '%s' is attached as a non-MSSQL catalog (type: %s)",
+			throw InvalidInputException("%s: Context '%s' is attached as a non-MSSQL catalog (type: %s)", function_name,
 										context_name, raw_catalog.GetCatalogType());
 		}
 		catalog_ptr = &raw_catalog.Cast<MSSQLCatalog>();
@@ -796,13 +796,13 @@ static int64_t RunExecBatch(ClientContext &client_context, const string &context
 		throw;
 	} catch (const std::exception &) {
 		throw InvalidInputException(
-			"mssql_exec: Unknown context '%s'. Attach a database first with: ATTACH '' AS %s (TYPE mssql, SECRET "
+			"%s: Unknown context '%s'. Attach a database first with: ATTACH '' AS %s (TYPE mssql, SECRET "
 			"...)",
 			context_name, context_name);
 	}
 	auto &catalog = *catalog_ptr;
 	if (catalog.IsReadOnly()) {
-		throw InvalidInputException("Cannot execute mssql_exec: catalog '%s' is attached in read-only mode",
+		throw InvalidInputException("Cannot execute %s: catalog '%s' is attached in read-only mode", function_name,
 									context_name);
 	}
 
@@ -810,7 +810,7 @@ static int64_t RunExecBatch(ClientContext &client_context, const string &context
 	auto connection = ConnectionProvider::GetConnection(client_context, catalog);
 
 	if (!connection) {
-		throw IOException("mssql_exec: Failed to acquire connection from pool for '%s'", context_name);
+		throw IOException("%s: Failed to acquire connection from pool for '%s'", function_name, context_name);
 	}
 
 	// Execute the SQL.
@@ -889,7 +889,7 @@ static void MSSQLExecExecute(DataChunk &args, ExpressionState &state, Vector &re
 
 		MSSQL_FN_DEBUG_LOG(1, "mssql_exec: context=%s, sql=%s", context_name.c_str(), sql.c_str());
 
-		return RunExecBatch(state.GetContext(), context_name, sql, sql);
+		return RunExecBatch(state.GetContext(), context_name, sql, sql, "mssql_exec");
 	});
 }
 
@@ -928,7 +928,7 @@ static void MSSQLExecParamsExecute(DataChunk &args, ExpressionState &state, Vect
 		auto params = mssql::BuildSqlParams(params_val, decl_val.IsNull() ? string() : decl_val.ToString());
 		string batch = params.ExecuteSqlBatch(statement);
 		MSSQL_FN_DEBUG_LOG(1, "mssql_exec_params: context=%s, batch=%s", context_name.c_str(), batch.c_str());
-		out[i] = RunExecBatch(client_context, context_name, batch, statement);
+		out[i] = RunExecBatch(client_context, context_name, batch, statement, "mssql_exec_params");
 	}
 }
 
--- a/src/query/mssql_sql_params.cpp
+++ b/src/query/mssql_sql_params.cpp
@@ -112,6 +112,30 @@ static std::string StringDeclaration(const LogicalType &type, const Value &value
 	return "nvarchar(4000)";
 }
 
+// decimal(38,0) is the widest exact numeric T-SQL has: +/-(10^38 - 1), i.e. at
+// most 38 decimal digits. HUGEINT reaches ~1.7e38 and UHUGEINT ~3.4e38, so both
+// can carry a 39-digit value that no SQL Server numeric can hold. Comparing the
+// rendered digits keeps this free of DuckDB's hugeint internals and is exact:
+// the literal we would emit is the same text.
+static bool ValueFitsDecimal38(const Value &value) {
+	if (value.IsNull()) {
+		return true;
+	}
+	std::string digits = value.ToString();
+	if (!digits.empty() && (digits[0] == '-' || digits[0] == '+')) {
+		digits.erase(0, 1);
+	}
+	// A hugeint renders as an integer, so anything else is not ours to judge.
+	for (size_t i = 0; i < digits.size(); i++) {
+		if (digits[i] < '0' || digits[i] > '9') {
+			return true;
+		}
+	}
+	size_t first = digits.find_first_not_of('0');
+	const size_t significant = (first == std::string::npos) ? 1 : digits.size() - first;
+	return significant <= 38;
+}
+
 std::string DeclarationForValue(const std::string &name, const LogicalType &type, const Value &value) {
 	switch (type.id()) {
 	case LogicalTypeId::SQLNULL:
@@ -134,6 +158,17 @@ std::string DeclarationForValue(const std::string &name, const LogicalType &type
 		return "decimal(20,0)";
 	case LogicalTypeId::HUGEINT:
 	case LogicalTypeId::UHUGEINT:
+		// decimal(38,0) holds +/-(10^38 - 1); HUGEINT reaches ~1.7e38 and UHUGEINT
+		// ~3.4e38, so the widest values render as literals the server cannot
+		// convert -- "Arithmetic overflow error converting numeric to data type
+		// numeric", which names neither the parameter nor the cause. Refuse here
+		// instead, the way the LIST and NULL arms already do.
+		if (!ValueFitsDecimal38(value)) {
+			throw InvalidInputException(
+				"parameter '%s' does not fit T-SQL decimal(38,0), the widest exact numeric SQL Server has; "
+				"cast it to VARCHAR and convert server-side",
+				name);
+		}
 		return "decimal(38,0)";
 	case LogicalTypeId::FLOAT:
 		return "real";
@@ -281,6 +316,21 @@ SqlParamSet BuildSqlParams(const Value &params, const std::string &declarations_
 		// DuckDB already refuses a STRUCT whose keys differ only in case, which is
 		// the one collision T-SQL would see.
 		std::string key = StringUtil::Lower(name);
+		// W5: T-SQL variable names are case-insensitive, so two struct keys that
+		// differ only in case DECLARE the same variable twice. Refuse here and
+		// name both spellings -- otherwise the batch reaches the server and comes
+		// back as "The variable name '@A' has already been declared", which does
+		// not say which struct key to change. The map was already built for this;
+		// nothing was reading it.
+		{
+			auto dup = seen.find(key);
+			if (dup != seen.end()) {
+				throw InvalidInputException(
+					"parameter names '%s' and '%s' differ only in case; T-SQL variable names are "
+					"case-insensitive, so both declare @%s",
+					dup->second, name, key);
+			}
+		}
 		seen[key] = name;
 		SqlParam p;
 		p.name = name;
--- a/src/query/mssql_sql_params.cpp
+++ b/src/query/mssql_sql_params.cpp
@@ -3,6 +3,7 @@
 #include <cctype>
 #include <map>
 
+#include "codec/integer_codec.hpp"
 #include "codec/literal_format.hpp"
 #include "codec/target_string_type.hpp"
 #include "duckdb/common/exception.hpp"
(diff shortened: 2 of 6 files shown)


src/mssql_functions.cpp
```diff
@@ -782,35 +782,35 @@ static bool ExecSqlMayChangeSchema(const string &sql) {
...
 		}
 		catalog_ptr = &raw_catalog.Cast<MSSQLCatalog>();
 	} catch (const InvalidInputException &) {
 		throw;
 	} catch (const std::exception &) {
 		throw InvalidInputException(
-			"mssql_exec: Unknown context '%s'. Attach a database first with: ATTACH '' AS %s (TYPE mssql, SECRET "
+			"%s: Unknown context '%s'. Attach a database first with: ATTACH '' AS %s (TYPE mssql, SECRET "
 			"...)",
 			context_name, context_name);
```
oluies: **BLOCK — the format string takes three arguments and gets two.**

I added the `%s:` prefix for `function_name` and never added it to the argument list. The correct sibling is 70 lines up at `:730-732`:

```cpp
throw BinderException(
    "%s: Unknown context '%s'. … AS %s (TYPE mssql, SECRET ...)",
    function_name, context_name, context_name);   // 3 placeholders, 3 args
```

**Reachability.** `BindExecContextName` validates at bind time but only `if (argument.IsFoldable())` (`:722`). A constant context is foldable, so the existing tests (`test/sql/mssql_exec.test:17-20`, `test/sql/query/error_handling.test:43-46`) are answered by that correct `BinderException` and **never reach this line** — they give false comfort. It is reached with a non-foldable context, or if the catalog is detached between bind and execute:

```sql
SELECT mssql_exec(c, 'SELECT 1') FROM (VALUES ('nope')) t(c);
SELECT mssql_exec_params(c, 'SELECT 1', {}) FROM (VALUES ('nope')) t(c);
```

**Symptom.** Release: `duckdb_fmt::vsprintf` throws `format_error("argument index out of range")`, caught and rethrown as `InternalException` — the user gets an `INTERNAL Error` with the "please file a bug report" banner and an invalidated transaction, from a typo'd catalog name. Debug: `Exception::ConstructMessageRecursive` counts placeholders first and throws `"Expected 3 parameters, received 2"`.

Neither the build nor clang-format catches fmt arity, and no test at any level covers this path.

**Fix:** add `function_name` as the first argument.

src/query/mssql_sql_params.cpp
```diff
@@ -112,6 +112,30 @@ static std::string StringDeclaration(const LogicalType &type, const Value &value
...
+	// A hugeint renders as an integer, so anything else is not ours to judge.
+	for (size_t i = 0; i < digits.size(); i++) {
+		if (digits[i] < '0' || digits[i] > '9') {
+			return true;
+		}
+	}
+	size_t first = digits.find_first_not_of('0');
+	const size_t significant = (first == std::string::npos) ? 1 : digits.size() - first;
+	return significant <= 38;
+}
```
oluies: **This is a fork of a helper the repo already has, done by a weaker method.**

`src/codec/integer_codec.cpp:85-105` defines `MaxDecimal38()`, `CheckHugeintFitsDecimal38()` and `CheckUhugeintFitsDecimal38()` — exact int128/uint128 range comparisons, already used at four call sites (`:270`, `:276`, `:324`, `:330`) on the BCP/insert write path. They are simply not declared in any header, which is presumably why I re-derived them instead of reusing them.

So the same rule is now enforced two ways: by exact range comparison on the write path, and by string-length counting here. A future change to one will not track the other.

**A trap worth naming.** The obvious shortening —

```cpp
return Hugeint::Abs(value.GetValue<hugeint_t>()) < Hugeint::POWERS_OF_TEN[38];
```

— is a **known fixed bug in this repo**. `integer_codec.cpp:91-95` documents it: `Hugeint::Negate` on HUGEINT min (−2^127) is itself out of int128 range, so abs-then-compare let that one value slip the guard. The existing helper bounds both ends instead (`Negate(max)`, not `Negate(value)`). My string version happens to sidestep it by stripping the sign textually, but anyone "simplifying" this later will walk straight into it.

**Fix:** export the existing helpers from `integer_codec` and call them, rather than keeping the fork.

oluies (commented):
## Adversarial self-review — 5 lenses + refutation pass

Ran a hostile multi-lens review (correctness, security, blast-radius, project convention, over-engineering) then a pass whose job was to disprove each finding.

**One blocking bug, found independently by three lenses and verified by hand — see the inline comment on `mssql_functions.cpp`.** The rest are quality.

### What the security lens refuted (worth recording)

- `parser.GetParseError()` **cannot** leak wire bytes or row data: the only hex dump goes to `TDS_PARSER_DEBUG` on stderr, never into `parse_error_`, which holds a fixed 4-char token-type string. The COLMETADATA/ROW arms propagate only structural facts — declared length, type id, column name.
- No format-string injection in the new duplicate-key throw: `IsSqlIdentifier(name)` throws *before* that block, so both spellings are already `[A-Za-z_][A-Za-z0-9_]*`, and all three are passed as `%s` **arguments**, matching every other call site in the file.
- `ValueFitsDecimal38`'s fail-open branch is not an injection vector — the rendered literal is digits-only on every path, and `SerializeDecimal` can never emit a quote, semicolon, or keyword.

### Convention checks that passed

`Acquire(-1, &why)` matches the header default so behaviour is unchanged, and the message shape matches `mssql_catalog.cpp:425` exactly. No C++17-only constructs. `ValueFitsDecimal38` is correctly unprefixed for `duckdb::mssql` per CLAUDE.md's namespace table, and `static` matches its sibling helpers.

### Not addressed

No CHANGELOG entry, and `test/cpp/test_sql_params.cpp` is already in `STANDALONE_TEST_SOURCES` with a `CHECK_THROWS_WITH` macro and `Value::STRUCT` construction — three one-line additions there would cover all three changes in this PR. None were added.

src/query/mssql_sql_params.cpp
```diff
@@ -134,6 +158,17 @@ std::string DeclarationForValue(const std::string &name, const LogicalType &type
...
+		// ~3.4e38, so the widest values render as literals the server cannot
+		// convert -- "Arithmetic overflow error converting numeric to data type
+		// numeric", which names neither the parameter nor the cause. Refuse here
+		// instead, the way the LIST and NULL arms already do.
+		if (!ValueFitsDecimal38(value)) {
+			throw InvalidInputException(
+				"parameter '%s' does not fit T-SQL decimal(38,0), the widest exact numeric SQL Server has; "
+				"cast it to VARCHAR and convert server-side",
+				name);
+		}
```
oluies, line 164: **Two scope problems with this guard.**

**1. The UHUGEINT half is unreachable.** Two lines after this, `:347` calls `codec::FormatSqlLiteral`, which for `LogicalTypeId::UHUGEINT` throws at `integer_codec.cpp:372-375`:

```cpp
case LogicalTypeId::UHUGEINT:
    throw NotImplementedException("codec::integer::FormatSqlLiteral: UHUGEINT literal not yet implemented");
```

No UHUGEINT parameter has ever reached the server, in range or out. The commit message's "UHUGEINT reaches ~3.4e38" rationale describes a path that does not exist — so half this code cannot be tested, and the reasoning behind it was not checked against the codebase.

**2. The `declarations` override bypasses it entirely.** `DeclarationForValue` — and therefore this check — is only called in the `else` branch at `:337-343`. Pass `'@p decimal(38,0)'` as the fourth argument and a 39-digit HUGEINT goes straight to the server and returns the raw "Arithmetic overflow" this change exists to prevent. That is the form documented for type control, so it is the likely one for a caller pushing big integers.

Worth deciding explicitly whether the override means "I know what I'm doing" (fine, but say so in the comment) or is an oversight.

src/query/mssql_sql_params.cpp
```diff
@@ -281,6 +316,21 @@ SqlParamSet BuildSqlParams(const Value &params, const std::string &declarations_
...
+		// nothing was reading it.
+		{
+			auto dup = seen.find(key);
+			if (dup != seen.end()) {
+				throw InvalidInputException(
+					"parameter names '%s' and '%s' differ only in case; T-SQL variable names are "
+					"case-insensitive, so both declare @%s",
+					dup->second, name, key);
+			}
+		}
```
oluies: **The commit message's justification for this block is factually wrong on both halves.**

It says *"The map was already built for this; nothing was reading it."* `seen` **was** being read — at `:352`, `if (!seen.count(d.first))`, the reverse "declaration with no value" check.

And the pre-existing comment three lines above at `:316-317` — which I left standing, so the file now contradicts itself — is the accurate one for the normal path. DuckDB 2.0's `Identifier` hashes and compares case-insensitively, and both struct-literal construction sites use `identifier_set_t`, so `{'a':1,'A':2}` is already refused at bind with `Duplicate struct entry name`. Confirmed in `struct_pack.cpp`'s `StructPackBind`.

**The block is not dead, though.** A STRUCT `Value` built outside `struct_pack` — a direct `Value::STRUCT`, an inferred-schema file read, another extension — is not covered by that bind-time check, so this is a legitimate backstop. It also cannot regress previously-working input: any case-colliding pair reaching here before would already have failed server-side with the confusing `"variable name '@A' has already been declared"`.

**Fix:** keep the check, correct the commit rationale, and reconcile the two comments so the file stops arguing with itself. The bare `{ }` scope around `dup` is also unnecessary — nothing after the block uses it.

Testable today: `test/cpp/test_sql_params.cpp` is in `STANDALONE_TEST_SOURCES` and its `TestBuildSqlParamsRefusals` already builds structs via `Struct({{Identifier(...)}})`, which bypasses `struct_pack`'s dedupe — the one place this check is reachable.

oluies:
All four findings from the adversarial review are fixed in `d2916b0` + `44ff7fc`.

- **the blocking one** — `function_name` added to the argument list, and `test/sql/query/mssql_exec_unknown_context.test` covers both shapes: a constant context (answered at bind, which is why every existing test missed it) and a column reference (non-foldable, so it reaches `RunExecBatch`), for `mssql_exec` and `mssql_exec_params` separately
- **the fork** — the range rule is now exported from `codec::integer` as `HugeintFitsDecimal38` / `UhugeintFitsDecimal38`; the BCP-side `Check*` wrappers call it and keep their column-worded messages, `mssql_sql_params` words its own for a parameter. `ValueFitsDecimal38`'s string scan is gone. I left a comment naming the trap, because the obvious shortening (`Hugeint::Abs(v) < POWERS_OF_TEN[38]`) is the bug #177 already fixed
- **the override bypass** — the guard runs on the `declarations` path too; naming a type does not make the value fit it
- **the contradicting comments** — reconciled into one that states what is actually true, and the bare `{ }` is gone

Tests: decimal(38,0) boundary at both signs, `POWERS_OF_TEN[38]` refused, HUGEINT min refused (the value abs-then-compare misses), the case-collision backstop via `Value::STRUCT` (the one construction that bypasses `struct_pack`'s dedupe), and the override-path overflow.

Verified locally in a Linux/GCC container against a live SQL Server, not just by reading: build clean, `make test-cpp` 23/23, integration 182 cases / 5856 assertions. The first attempt at the new unit test did not compile — this file imports names one at a time rather than `using namespace duckdb` — which is the sort of thing the loop exists to catch.

*(Edited: the second SHA was wrong in the original.)*

src/query/mssql_sql_params.cpp
```diff
@@ -278,9 +306,23 @@ SqlParamSet BuildSqlParams(const Value &params, const std::string &declarations_
...
+		// case DECLARE the same variable twice. A STRUCT built by the parser is
+		// already safe -- Identifier compares case-insensitively and both
+		// struct-literal sites use identifier_set_t, so {'a':1,'A':2} is refused
+		// at bind with "Duplicate struct entry name". This is the backstop for a
+		// STRUCT that did not come from there: a direct Value::STRUCT, an
+		// inferred-schema file read, another extension. Without it the batch
+		// reaches the server and returns "The variable name '@A' has already been
+		// declared", which does not say which key to change.
 		std::string key = StringUtil::Lower(name);
+		auto dup = seen.find(key);
```
VGSML, line 319: Unreachable: DuckDB refuses a STRUCT with two names differing only in case at type construction (`Duplicate STRUCT type argument name`), so no value with `a` and `A` can arrive here. Please drop this block; the `seen` map stays for the override completeness check below it.

VGSML (changes requested):
Thanks — two of the three go in as they are; the third does not hold, with the check below.

**Take:** the `function_name` plumbing through `RunExecBatch` (the read-only refusal and the unknown-context help were the two a caller hits first, and both said `mssql_exec`), and the `decimal(38,0)` range check for `HUGEINT` / `UHUGEINT` — the overflow the server raised named neither the parameter nor the cause, and refusing by name is what the `LIST` and `NULL` arms already do.

**Drop:** the duplicate-key refusal. DuckDB cannot build the STRUCT this guards against:

```
D SELECT {'a': 1, 'A': 2};
Binder Error: Duplicate named argument "A" in function call to '"struct_pack"'
D SELECT {'a': 1}::STRUCT(a INT, A INT);
Binder Error: Duplicate STRUCT type argument name "A"
```

The type constructor itself rejects two names that differ only in case, so no STRUCT value with `a` and `A` reaches `BuildSqlParams`, from a literal, a cast, or a reader. That is why the map's second use was removed in the spec-075 branch: the check was dead, and the comment above it says so. The description's `The variable name '@A' has already been declared` was not observed — the PR notes it was not built or run — and cannot be.

**On the three findings left to me,** verified on the branch and going into #341 as two commits: (1) describe-vs-stream type divergence is real and wider than `rowversion` — `sql_variant`, `hierarchyid` and `geography` used to fail at bind with `Unsupported SQL Server type` and now fail at init with `Connection closed while waiting for COLMETADATA`; the fix builds a `tds::ColumnMetadata` from the describe row and asks `TypeConverter::GetDuckDBType`, the stream's own mapping, so the two agree by construction and the datetime2 fold goes away. (2) The COPY init order: in this DuckDB `copy_to_initialize_global` runs from an open job that `RequestFileState` schedules asynchronously when the sink state is created, so it is a race with the source's init rather than sink-first (15 consecutive runs of `sink_reads_own_catalog.test` pass, source wins); the deferral of `INSERT BULK` + COLMETADATA to the first shared-writer Sink is still the right fix and goes in. (3) `Copy()` sharing the prepared session: DuckDB copies a `LogicalGet`'s bind data only for late materialization, which requires `function.late_materialization`; ours does not set it, and `EXPLAIN` of a prepared scan under `ORDER BY ... LIMIT` shows one scan. Left as is, with a comment.

Once the dup-check hunk is out I will merge this into the 075 branch.

VGSML (approved):
Approving at `44ff7fc`. The range rule shared from `codec::integer` with the #177 BCP check is the right home for it, the override path is covered, and the two new tests pin both. I withdraw the request to drop the case-collision check: it is unreachable from a struct literal or a cast, as shown, but a STRUCT built outside the parser (a direct `Value::STRUCT`, an inferred-schema file read) can carry two such keys, and the check is one map lookup at bind.

Verified here as the merge result with the current 075 head (`c60fc3f`): build clean; `mssql_params` 87, `mssql_exec_unknown_context` 3, `mssql_scan_describe_types` 102, `mssql_scan_prepared` 26, `mssql_scan_bind_describe` 24, transaction 9 cases / 282, `copy_hugeint_bcp` 21; `make test-cpp-run` 23/23. Merging into the 075 branch now.

Issue #324, opened 2026-09-09
Pool warm-up: a way to open connections up front, in parallel (each TDS login is ~175ms and they are paid serially)


VGSML:
Filed from measurements taken while benchmarking [mssql_ducklake](https://github.com/hugr-lab/mssql-ducklake) against the DuckLake postgres backend. Everything below is on loopback, SQL Server 2025-CU8 in docker, so the numbers are a floor — a real network makes each one worse.

## What a login costs, and how many an ATTACH performs

Counted from the server's own `Logins/sec` performance counter.

| ATTACH | time | logins |
| --- | ---: | ---: |
| default | 0.510s | 3 |
| `lazy_validation true` | 0.359s | 2 |
| `lazy_validation true, catalog false` | 0.004s | 0 |

One login on loopback is ~0.175s, and 3 × 0.175 ≈ 0.51s — so the logins are essentially the whole of what an attach costs, and they appear to be opened **serially**.

For comparison on the same machine, the DuckDB postgres extension's `ATTACH` is 0.009s.

## The good news: it is warm-up, and it is paid once

The connections are reused for the rest of the session. A session doing 41 commits performs the same number of logins as one doing a single commit:

| session | logins |
| --- | ---: |
| attach + 1 commit | 4 |
| attach + 21 commits | 4 |
| attach + 41 commits | 4 |

So this is a fixed startup cost that the pool then serves everything from, not a per-operation one. No complaint about that part — it is the right shape.

## The request

**A function to warm the pool explicitly**, e.g. `mssql_warm_pool(catalog [, n])`, that opens `n` connections up front and — the part that matters — opens them **concurrently**. At ~175ms per login, four connections cost ~0.7s serially and ~0.18s in parallel.

Two reasons it would help:

1. **A workload that goes parallel pays the login latency at its first parallel step**, in the middle of the work, rather than at attach. Warming up front moves a known cost to a known place.
2. **Attach itself would get cheaper** if the connections it needs were opened concurrently rather than one after another.

## Related: `mssql_min_connections` appears to be inert

The setting reads as if it would pre-open connections, but changing it makes no difference to attach time:

| `mssql_min_connections` | attach + one scan |
| ---: | ---: |
| 0 | 0.178s |
| 2 | 0.171s |
| 4 | 0.174s |
| 8 | 0.176s |

[...]

PR #345, opened 2026-09-14
spec 076: catalog scans on one plan — parameterised filters, one round trip per first touch

VGSML:
Spec `specs/076-catalog-scan-parameters/spec.md`; § 0 is what was measured before anything was designed. Builds on spec 075's `query/mssql_sql_params` and per-result-set COLMETADATA.

## What changes

**W1 — pushed filters are parameterised.** The filter encoder gets a parameter sink: every constant of a pushed filter becomes `@pN` of an `sp_executesql` batch, and the statement text is one fixed string per filter shape. Measured on the local docker server, forty catalog scans with forty distinct two-predicate filters:

| | plans in `sys.dm_exec_cached_plans` | bytes |
|---|---:|---:|
| literal (`mssql_scan_parameterize_filters = false`) | 40 Adhoc | 2 293 760 |
| parameterised (default) | 1 Prepared | 65 536 |

Wall time is unchanged locally (a trivial plan compiles in well under a millisecond); the gain is what the server keeps, and what compiles once the predicates stop being trivial.

Each parameter is declared from the **column it is compared with** (`FilterEncoder::DeclarationForColumn`) — which the catalog scan knows and `mssql_scan_params` does not — and the rule is *the column's kind, never narrower than the constant*: varchar stays varchar so an index on it is still seekable; the width is at least the constant's, so a 25-character constant against `varchar(20)` is `varchar(25)` and can never be truncated into a false match; integers and decimals take the wider of column and constant; `datetime2` / `time` / `datetimeoffset` the wider scale; `float`, `real`, `datetime` and `smalldatetime` follow the constant (declaring a DOUBLE as `real` would round it). A non-ASCII constant goes as `nvarchar` whatever the column's collation — a varchar *variable* takes the database's code page, not the column's, and the spec's UTF-8 exemption sent `?` for Cyrillic (`annotated_max_string.test` caught it). `IN` lists, `IS NULL`, NULL constants and types with no parameter form (xml, sql_variant, UDTs, rowversion) stay literal. The peer lookup sees through the spec-060 no-op cast; without it a LIKE on an `MSSQL_NVARCHAR(10)` column declared `nvarchar(4000)`.

Plan-time (complex) parameters ride on the bind data (`complex_filter_params`, serialised with the clause as properties 107/108); init copies them, appends the simple filters' own, and wraps the statement. TOP N and ORDER BY pushdown are inside the wrapped text and unchanged. `mssql_scan_parameterize_filters = false` restores literal SQL — the escape hatch for parameter sniffing, and what a reviewer flips to compare plans.

**W2 — one round trip per first touch.** The primary-key discovery statement rides in the same batch as the single-table metadata query (a second result set off the same `@s` / `@t`; PK rows have 8 values, table rows 12) and the entry is constructed with its key. A fresh table's first touch is one 1864-byte metadata batch where it was two (1132 + 808) on two pooled connections.

**W2 — ATTACH initialised the catalog twice.** The "collation queried twice per ATTACH" of § 0.2 turned out to be `MSSQLCatalog::Initialize` running twice — from the storage extension's attach callback and again from DuckDB's `AttachedDatabase::Initialize` — building a second pool, throwing away the logged-in connection, logging in again and asking again. It is a no-op once the pool exists; ATTACH makes two logins instead of three (two of the three #324 counted).

## Tests

- `test/sql/catalog/scan_parameterized_filters.test`: every row of the declaration table with the setting on and off, expecting the same rows; the F3 cases (25 chars vs `varchar(20)`, BIGINT vs `int`, wider decimal scale); LIKE, BETWEEN, IN, IS NULL, rowid; UPDATE/DELETE through the parameterised rowid scan.
- `test/cpp/test_filter_encoder.cpp`: the text with a sink (`([name] = @p0)`, peer on either side, no peer inside an expression, NULL stays literal, the rendered batch) and the declaration table.
- Full SQL suite locally 200/202 (the two are `azure/*` on a sleeping serverless database, 40613); `make test-cpp-run` 30/30.

## Out of scope by decision

DML — spec 062 and the UPDATE/DELETE rework replace the batches that would benefit.


VGSML:
Self-review at `55f36de`, what I would raise as a reviewer:

1. **Where the declaration rule can still differ from DuckDB's comparison.** Two places are value-driven on purpose and one is a known #272-class divergence. `float` / `real` columns declare from the constant (a DOUBLE as `real` would round and turn "not equal" into "equal"); `datetime` / `smalldatetime` declare from the constant, as the literal form (a CAST to `datetime2`) always did. A `char(n)` column gets `varchar(k)`: the server's padding semantics on the comparison are unchanged from the literal form and are #272's subject, not this PR's.
2. **The non-ASCII rule was wrong in the spec and is fixed in the code.** The spec exempted UTF-8 varchar columns; a varchar *variable* takes the database's code page, so `@p varchar(max) = N'ы…'` arrived as `?` against a UTF-8 column. `annotated_max_string.test` (#321) caught it; any non-ASCII constant now goes as nvarchar. Spec § W1 updated to say so.
3. **What stays literal, deliberately:** `IN` lists (N parameters for N values, and the 2100 cap), `IS NULL`, NULL constants, `rowversion` (`is_cast_required` — the scan casts the column, so a BLOB parameter would compare against text), xml / sql_variant / UDTs, and everything past `SqlParamSet::MAX_PARAMS` (2000): the encoder falls back to the literal rather than fail.
4. **`Equals` on the catalog bind data** compares only catalog / schema / table, as before. Two scans of one table with different filters already compared equal before this PR (the clause was never in it); parameterising does not change that. Not touched here.
5. **TOP N / ORDER BY pushdown** (`mssql_order_pushdown = true`) is inside the wrapped text: `EXEC sp_executesql N'SELECT TOP 3 [n], [id] FROM … WHERE ([v] = @p0) AND ([n] = @p1) ORDER BY [id] DESC', …` — same rows as the literal form. OFFSET and multi-key ORDER BY are not pushed in either mode, as before.
6. **The `Initialize` guard** keys on `connection_pool_` being set. Nothing else in `Initialize` has a side effect a second call should repeat (the collation is read into the cache on the first pass). `catalog_enabled = false` still builds the pool and returns before the collation, so the guard holds there too.
7. **W2's row-width dispatch** (8 values = PK row, 12 = table row) is explicit in both templates and commented at the callback; a column added to either query changes the constant next to it.

Verified locally: full SQL suite 200/202 (the two are `azure/*` on a sleeping serverless database), `make test-cpp-run` 30/30; plan cache 40 → 1 for forty distinct filters; ATTACH 3 → 2 logins; first touch two metadata batches → one.

VGSML:
Bench against the 2.0 baseline, as asked — same-session interleaved A/B, two rounds, min per cell: **A** = `main` at the spec-069 merge (`1d74ae2`, the last state measured on the 2.0 line), **B** = this branch at `55f36de`. Full tables are now § 6 of the spec.

- **Reads:** client CPU per value at parity across all 22 families (B/A 0.92–1.09). Wall clock is 0.15–0.18 s shorter on B on *every* step — the same amount everywhere, and the amount of one login: the `startup` control (invocation + ATTACH) went 0.561 → 0.384 s, which is W2's double-`Initialize` fix, not the codec.
- **Writes:** wall 0.78–0.95 for the same reason; CPU at parity where it is measurable (strings), timer-resolution noise on the narrow families (0.001–0.05 s over a million values, swinging both ways).
- **Wide-write matrix:** `threads=1` parity (0.98–0.99 on all four cells); `threads=4` swings up to 1.7× both ways between rounds — four parallel writers on a shared docker, not a difference.
- One apparent outlier, `read_int` CPU 2.0×, did not survive three more interleaved pairs of that step nor the per-phase counters (parse and process identical, the delta all in socket wait): noise.

What the bench cannot see is W1's own effect: every step runs one filter shape, and a trivial plan compiles in under a millisecond either way — the 40 → 1 plans of § 0.1 is the measurement for that. Side note: both bench scripts' DSNs needed `TrustServerCertificate=yes` for the post-074 binary; they carry it now.

src/catalog/mssql_metadata_cache.cpp
```diff
@@ -477,6 +480,12 @@ bool MSSQLMetadataCache::GetTableMetadata(tds::TdsConnection &connection, const
 			// header). It has to stay — MSSQLCatalogScanCardinality reads the
 			// catalog's copy before anything else, so removing it plans every direct
 			// query at ~1 row.
+			// The second result set: a primary-key column (8 values), see
+			// PrimaryKeyInfo::DiscoverySqlTemplate. The table rows have 12.
+			if (values.size() == 8) {
+				mssql::PrimaryKeyInfo::AppendColumnFromRow(table_meta.pk_info, values, database_collation_);
+				return;
+			}
```
oluies: **Medium — the two result sets are told apart by column count, and the failure mode is silent.**

`values.size() == 8` → PK row, `< 12` → ignore, else table row. The comment above it states the coupling (*"The table rows have 12"*) but nothing enforces it, and nothing ties either number to the SQL that produces it.

If `PK_DISCOVERY_SQL_TEMPLATE` ever gains a ninth column — `ic.is_descending_key` for ORDER BY pushdown, `kc.name` for better error text — every PK row fails the `== 8` test, then falls through `< 12`, and is dropped. `pk_loaded` is still set to `true` further down, so `MSSQLTableEntry::EnsurePKLoaded` will not re-discover. **Every table in the catalog silently loses its primary key, taking `rowid`, UPDATE and DELETE with it, with no error anywhere.** The same happens in reverse if the table query is ever trimmed to 8 columns.

That is a lot of blast radius resting on two integer literals that a reasonable future edit would change without thinking about this site.

**Fix:** discriminate on the result-set ordinal, not the arity. `MSSQLSimpleQuery::ExecuteWithCallback` already pushes one entry into `result.result_sets` per COLMETADATA token, so the index is available — route on `set_index == 1`. If that is more surgery than this PR wants, a `D_ASSERT` that the PK statement's COLMETADATA really has 8 columns at least makes a future edit fail loudly instead of silently.

src/table_scan/filter_encoder.cpp
```diff
@@ -255,13 +259,306 @@ std::string FilterEncoder::ValueToSQLLiteral(const Value &value, const LogicalTy
...
+	case LogicalTypeId::UHUGEINT:
+		return 6;
+	default:
+		return -1;
+	}
+}
+
+const char *IntegerNameOfRank(int rank) {
+	static const char *names[] = {"bit", "tinyint", "smallint", "int", "bigint", "decimal(20,0)", "decimal(38,0)"};
+	return names[rank];
```
oluies, line 338: **Refuted — the tables do not diverge on `UINTEGER`.** Recording it so the claim doesn't get actioned.

The review said `UINTEGER` is ranked 3 (`int`) here but declared `bigint` by `DeclarationForValue`. It isn't:

```
:323  case LogicalTypeId::UINTEGER:
:324  case LogicalTypeId::BIGINT:
:325      return 4;
:337  names[] = {"bit","tinyint","smallint","int","bigint","decimal(20,0)","decimal(38,0)"};
```

Rank 4 → `bigint`, which is exactly what `mssql_sql_params.cpp:153-155` returns. Walking all seven ranks against `DeclarationForValue`, they agree on every one: bit / tinyint / smallint / int / bigint / decimal(20,0) / decimal(38,0). No `DECLARE @p0 int = 3000000000` is reachable from this.

**The other half of that finding does stand**, though: `IntegerNameOfRank(6)` returns `decimal(38,0)` unconditionally, so a HUGEINT constant outside ±(10^38−1) is declared without the range check `DeclarationForValue` applies — it now goes to the server and returns a bare *"Arithmetic overflow converting numeric to numeric"* instead of the named client-side error. Worth routing rank 6 through the shared `codec::integer::HugeintFitsDecimal38` guard.

A unit assertion pairing each `IntegerRankOfValue` arm with `DeclarationForValue`'s answer would keep the two from drifting later — they agree today, and nothing enforces that they keep agreeing.

src/table_scan/filter_encoder.cpp
```diff
@@ -255,13 +259,306 @@ std::string FilterEncoder::ValueToSQLLiteral(const Value &value, const LogicalTy
...
+		}
+		int precision = column.precision > DecimalType::GetWidth(type) ? column.precision : DecimalType::GetWidth(type);
+		int scale = column.scale > DecimalType::GetScale(type) ? column.scale : DecimalType::GetScale(type);
+		if (precision < scale) {
+			precision = scale;
+		}
+		if (precision > 38) {
+			precision = 38;
+		}
+		return "decimal(" + std::to_string(precision) + "," + std::to_string(scale) + ")";
```
oluies: **Low, confirmed — precision and scale are maximised independently, which under-allocates integer digits.**

```cpp
precision = max(col_p, val_p);
scale     = max(col_s, val_s);
```

Integer digits are `precision - scale`, so taking the two maxima separately does not preserve *either* side's integer width. A `decimal(10,8)` column against a `DECIMAL(11,1)` constant gives `precision = 11`, `scale = 8` → `decimal(11,8)`, i.e. **three** integer digits for a constant that needs ten. The `DECLARE` fails with *"Arithmetic overflow error converting numeric to data type numeric"*.

The comment describes the intent as "never narrower than either", and this rule does not implement that.

DuckDB's comparison-type promotion usually brings both sides to a common width first, which is why `scan_parameterized_filters.test` doesn't hit it — but the rule is reachable whenever the column's scale exceeds the constant's.

**Fix:** derive precision from integer digits:
```cpp
scale = max(col_s, val_s);
precision = min(38, max(col_p - col_s, val_p - val_s) + scale);
```

website/docs/reference/settings.md
```diff
@@ -26,6 +26,7 @@ sidebar_position: 1
 
 | Setting                            | Type    | Default | Range | Description                           |
 | ---------------------------------- | ------- | ------- | ----- | ------------------------------------- |
+| `mssql_scan_parameterize_filters` | `true` | Send the constants of a pushed filter as `sp_executesql` parameters, declared from the column's SQL Server type, so SQL Server keeps one plan per filter shape instead of one per distinct value set. `false` restores literal SQL (the escape hatch for parameter sniffing). |
```
oluies: **Low, confirmed — three cells against a five-column header, so the published table renders wrong.**

The header is `Setting | Type | Default | Range | Description`; this row supplies `Setting | true | Description`. Rendered, that puts **Type = `true`**, **Default = the entire description**, and leaves Range empty.

```markdown
| `mssql_scan_parameterize_filters` | BOOLEAN | true | - | Send the constants of a pushed filter as `sp_executesql` parameters, declared from the column's SQL Server type, so SQL Server keeps one plan per filter shape instead of one per distinct value set. `false` restores literal SQL (the escape hatch for parameter sniffing). |
```

Also worth checking: this row was added under the **Statistics Settings** heading, next to `mssql_enable_statistics` and `mssql_statistics_level`. A filter-pushdown setting probably belongs with the scan/query settings instead.

oluies (commented):
## Automated review (roborev job 1593), verified

Range `d5d2346..8e51b10`. Two Medium and five Low. I checked each against the branch rather than relaying it — **one Low is factually wrong and I've marked it refuted below**, the rest hold.

Worth saying first: the declaration rules are the careful part of this change and they are unit-tested (`test_filter_encoder.cpp` +130). The findings are about what surrounds them.

### Verified summary

| Finding | Verdict |
|---|---|
| Merged-batch result sets separated by column arity | **Confirmed** — inline |
| W2 (PK folded into the metadata batch) ships untested | **Confirmed** — the spec names `first_touch_round_trips.test`; it is not in the diff |
| `UINTEGER` ranked `int` vs `bigint` | **Refuted** — the two tables agree exactly; see inline |
| Rank-6 arm skips the `decimal(38,0)` range check | **Stands** (the other half of that finding) |
| Constants registered then refused stay in the `DECLARE` | Plausible, not run |
| Rowid *inequality* declares from the DuckDB type, not the PK column | Plausible, not run |
| Decimal precision/scale maxima taken independently | **Confirmed** — inline |
| `settings.md` row has 3 cells against a 5-column header | **Confirmed** — inline |

On the missing test specifically: `git diff --name-only d5d2346..HEAD -- test/` returns only `bench_live_server.sh`, `bench_wide_write.sh`, `test_filter_encoder.cpp` and `scan_parameterized_filters.test` — and that last one exercises W1. The spec's own W3 names `first_touch_round_trips.test` by filename. W2 is the larger behavioural change of the two (PK discovery folded into the batch, `pk_loaded` seeded, `MSSQLCatalog::Initialize` becoming a no-op once the pool exists), and it is the half with no coverage.

VGSML:
Thanks — all six that hold are addressed in `98d984d`, per finding:

- **Result sets told apart by width (Medium):** routed by the result-set **ordinal** now. `MSSQLSimpleQuery::ExecuteWithSetCallback` hands the callback the index of the COLMETADATA a row came from (the old `ExecuteWithCallback` wraps it), the cache gets `ExecuteMetadataQuerySets` on top of the same `RunMetadataQuery` core (the #317 lever and the 1205 retry unchanged), and `GetTableMetadata` reads set 1 as the key and set 0 as the table. Two literals with a catalog-wide silent failure mode became one `== 1` that a changed query cannot fool.
- **W2 untested (Medium):** `test/sql/catalog/first_touch_round_trips.test` — `mssql_pool_stats` makes both halves observable: `connections_created` is **1** right after ATTACH (one pool, one login through it), and a fresh table with a key costs exactly **two** `acquire_count` (metadata + key in one batch, then the scan) where it cost three; rowid-based UPDATE and DELETE work on the entry as created, for a scalar and a composite key; a table without a key still refuses `rowid`.
- **Rank 6 skips the range check:** a HUGEINT / UHUGEINT constant against an integer column now goes through `DeclarationForValue`, i.e. the shared `HugeintFitsDecimal38` guard, and a value past ±(10^38−1) is refused by name at bind instead of the server's bare overflow. Unit test: `POWERS_OF_TEN[38]` names `decimal(38,0)`; plus the pairing assertion you suggested — every integer rank against a `bit` column declares what `DeclarationForValue` declares, so the two tables cannot drift silently.
- **Registered-then-refused constants:** `EncodeExpression` is now a wrapper that marks the sink's size and truncates back to it when the body refuses, so a comparison whose other side does not encode leaves nothing in the DECLARE line. Unit test: a constant against an unmapped function refuses and the sink is unchanged.
- **Rowid inequality declared from the DuckDB type:** `ColumnInfoOf` resolves `COLUMN_IDENTIFIER_ROW_ID` to the scalar PK column, so `rowid > @p` is declared from that column (a varchar key stays varchar), same as the equality path already was.
- **Decimal precision/scale maxima taken independently:** integer digits are preserved — `scale = max(s)`, `precision = min(38, max(p−s) + scale)`. Your case `decimal(10,8)` vs `DECIMAL(11,1)` now declares `decimal(18,8)`; unit tests updated (`decimal(9,2)` vs `DECIMAL(10,4)` is `decimal(11,4)`, not `(10,4)`).
- **settings.md row:** five cells, under its own *Filter Pushdown Settings* heading rather than *Statistics*.

The refuted `UINTEGER` finding stays refuted (ranks agree, and the new pairing assertion pins it). Spec W2/W3 text updated to match.

oluies (approved):
lgtm

PR #346, opened 2026-09-14
chore: bump the duckdb pin to v2.0-cyanoptera d673cf9ab4 (v2.0.0-dev65284)

VGSML:
`070ce1f695` (2026-09-07) → `d673cf9ab4` (2026-09-14), 556 commits on `v2.0-cyanoptera`.

The header drift in what the extension includes is additive — new `TableFunction` callbacks (`combine_schema`, `claim_batch`, `finish_batch`), `TableFunctionBindInput`'s expected schema, `Catalog::GetEntryInCatalog` — except one break: a parsed `ConstantExpression` is now a `Literal` (the query text's atom) and no longer holds a `Value`. The one reader we have, the `WITH (...)` table-option parser in `mssql_table_options.cpp`, takes `GetLiteral()` and binds it with `ToValue()` for the string it applies.

Verified locally at `v2.0.0-dev65284`: full SQL suite 200/201 (the one is `azure/*` on a sleeping serverless database), `make test-cpp-run` 30/30.


src/catalog/mssql_table_options.cpp
```diff
@@ -82,11 +82,13 @@ void MSSQLTableOptions::ApplyWithClause(const case_insensitive_map_t<unique_ptr<
...
-		auto &constant = entry.second->Cast<ConstantExpression>();
-		if (constant.GetValue().IsNull()) {
+		// DuckDB v2.0 (d673cf9): a parsed constant is a Literal -- the query
+		// text's atom -- not a Value; ToValue() binds it for the string we need.
+		auto &literal = entry.second->Cast<ConstantExpression>().GetLiteral();
+		if (literal.IsNull()) {
 			throw InvalidInputException("MSSQL: table option '%s' must not be NULL", entry.first);
 		}
-		ApplyOption(entry.first, constant.GetValue().ToString());
+		ApplyOption(entry.first, literal.ToValue().ToString());
```
oluies, line 91: **Checked both concerns against `d673cf9ab4` itself rather than inferring — both are fine.**

**1. The NULL guard moved from the bound `Value` to the unbound `Literal`. Equivalent.**

`src/parser/literal.cpp:70`:
```cpp
bool Literal::IsNull() const { return kind == LiteralKind::NULL_LITERAL; }
```
and `:161-164`:
```cpp
Value Literal::ToValue() const {
    switch (kind) {
    case LiteralKind::NULL_LITERAL: return Value();   // a NULL Value
```

So `IsNull()` means exactly "this atom is the SQL `NULL` keyword" — the semantics the old `constant.GetValue().IsNull()` asserted — and it agrees with `ToValue().IsNull()`. `WITH (table_kind = NULL)` still reaches the dedicated *"must not be NULL"* message.

Also worth correcting: the suggested rewrite was said to avoid "the second `ToValue()` binding pass". There isn't one — `IsNull()` reads `kind` and does not bind, so this code already binds exactly once. Binding first and testing the `Value` is a fair style preference, but it is not cheaper and not a behaviour fix.

**2. `ToValue()` throwing on an oversized numeric. Not for the example given.**

`NumberToValue` (`src/parser/literal.cpp:93-158`) is a non-throwing `TryCast` chain: `int32` → `int64` → `hugeint` → `uhugeint` → `Bignum::VarcharToBignum` (with `ConversionException` **caught** and fallen through) → decimal → double. The cited case, `WITH (data_compression = 99999999999999999999999)`, lands on BIGNUM and returns a `Value`; it does not throw.

The tail is `Cast::Operation<string_t, double>`, which is throwing in principle — but its input is a `NUMERIC`/`INTEGER` literal the parser already accepted, so it is parseable as a double by construction. I could not build a case that reaches it from this call site.

---

One genuine (non-blocking) observation the review did not make: the comment says *"`ToValue()` binds it for the string we need"*. For a `STRING` literal `ToValue()` is `Value(text)` — the raw parsed text — so for the options this function reads (`table_kind`, and friends) `.ToString()` is a pure round-trip and the migration is inert. Worth knowing if a future option ever takes a numeric or boolean, where `ToValue()` does apply real type inference and `.ToString()` would render the *inferred* form rather than the source text.

oluies (commented):
## Automated review (roborev job 1594) — **both findings refuted on evidence**

Reviewed `c8202f7`. The reviewer raised two Low findings about the `GetLiteral()` / `ToValue()` migration but explicitly could not confirm the contracts, because the `duckdb` submodule is not checked out in a review worktree. I checked them against the actual source at the new pin (`d673cf9ab4`) and **neither holds**. Detail inline.

It also confirmed the thing most worth confirming about a pin bump: a tree grep shows this was the **only** parsed-`ConstantExpression` consumer needing adaptation — `filter_encoder.cpp` uses `BoundConstantExpression`, which is unaffected — and both call sites of `ApplyWithClause` are exercised by `test/sql/catalog/table_options.test`.

No blocking issues from this review.

oluies (approved):
## Approve

Both findings from the automated review are refuted on evidence, and I checked the thing a pin bump most needs checking: that the adaptation is complete.

**The migration is correct.** `ConstantExpression::GetLiteral()` / `Literal::ToValue()` is the right shape at `d673cf9ab4`, and the NULL guard is preserved — `Literal::IsNull()` is `kind == LiteralKind::NULL_LITERAL` (`src/parser/literal.cpp:70`) and `ToValue()` returns `Value()` for that kind (`:163`), so `WITH (table_kind = NULL)` still reaches the dedicated message rather than falling through to `ApplyOption`.

**It is the only site.** `filter_encoder.cpp` consumes `BoundConstantExpression`, which is unaffected, and both call sites of `ApplyWithClause` are covered by `test/sql/catalog/table_options.test`.

**Nothing here can throw where it previously couldn't.** `NumberToValue` (`src/parser/literal.cpp:93-158`) is a non-throwing `TryCast` chain — int32 → int64 → hugeint → uhugeint → bignum, with `ConversionException` caught — so an oversized numeric option value lands on BIGNUM rather than escaping as a raw conversion error.

One thing to keep in mind rather than change: for a `STRING` literal `ToValue()` is `Value(text)`, so `.ToString()` is a pure round-trip and this is inert for today's options. It stops being inert the first time an option takes a numeric or boolean, where `ToValue()` applies type inference and `.ToString()` renders the inferred form rather than the source text. Not this PR's problem.

Not verified by me: that the 556-commit bump itself is clean beyond this one call site — CI's build is the authority on that, not this review.

Issue #327, opened 2026-09-09
IDENTITY columns are not surfaced to DuckDB: an INSERT without a column list demands a value for them


VGSML:
The catalog's column metadata does not record whether a column is an IDENTITY, so DuckDB binds one as an ordinary column. An `INSERT ... VALUES` without a column list — legal T-SQL against such a table, and the form SQL Server itself expects — is then rejected by DuckDB's binder before it ever reaches the server.

## Repro

```sql
LOAD mssql;
ATTACH 'Server=…;Database=…;User Id=…;Password=…' AS srv (TYPE mssql);

SELECT mssql_exec('srv', '
  CREATE TABLE dbo.idtest(a BIGINT, b VARCHAR(50));
  ALTER TABLE dbo.idtest ADD rid BIGINT IDENTITY(1,1) NOT NULL;
  ALTER TABLE dbo.idtest ADD CONSTRAINT pk_idtest PRIMARY KEY NONCLUSTERED (rid);');

-- works: raw T-SQL, SQL Server assigns rid
SELECT mssql_exec('srv', 'INSERT INTO dbo.idtest VALUES (1, ''x'');');

-- fails: DuckDB sees three columns and two values
INSERT INTO srv.dbo.idtest VALUES (2, 'y');
```

```
Binder Error: table idtest has 3 columns but 2 values were supplied
```

## Where it comes from

The bulk metadata query selects `column_name`, `column_id`, `type_name`, `max_length`, `precision`, `scale`, `is_nullable`, `collation_name`, `index_type`, `partition_count` — but not `sys.columns.is_identity`. `MSSQLTableEntry` therefore has no way to mark the column, and DuckDB's insert binder counts it among the insertable ones.

`src/dml/ctas/mssql_ctas_executor.cpp` already carries the right vocabulary — `has_identity_column`, `identity_column_index`, `col.is_identity` — but hard-codes them to `false`, so the shape for this exists and is simply never populated.

## Suggested fix

Add `c.is_identity` to the metadata query, carry it on the column entry, and exclude identity columns from the default insert column list (an explicit insert into one should still error unless `SET IDENTITY_INSERT` is in play, which matches SQL Server).

## Why it matters here

[...]

Issue #296, opened 2026-09-02
Asking to add support for TIMESTAMP/ROWVERSION data types


test32443:
These are effectively binary(8) or could be treated as BIGINT datatype, but currently not supported. Also with the current implementation, DuckDB considers TIMESTAMP data type to be of datetime type and tries to convert it to varchar, which fails.

VGSML:
Yes we can support this in the next release

PR #348, opened 2026-09-14
spec 062: INSERT via BCP — one session type, atomic loads, the statement path under the auto-parameterisation line

### LINKED ISSUE #344: DML desync in autocommit: silent-but-complete vs honest-but-partial — which semantics do we want?
## The question

A TDS parser desync during a DML batch is a **client-side framing** failure — by the time we notice, SQL Server has already executed and committed the batch. PR #342 makes that desync a thrown error instead of a silent success. In an **explicit transaction** that is unambiguously right. In **autocommit** — the default — it exposes a partial-application state that has no compensation path, and I do not want to pick the semantics unilaterally.

## Why autocommit is different

`defer_execution_` is set only when *not* autocommit:

- `src/dml/update/mssql_update_executor.cpp:47-60`
- `src/dml/delete/mssql_delete_executor.cpp:50-63`

So inside a transaction, every batch flushes on the pinned connection inside the open `BEGIN TRANSACTION`, and rollback covers all of them.

In autocommit, `Execute()` flushes every `mssql_dml_batch_size` (default 500) rows as an **independent, already-committed** statement, each on its own pooled connection:

- `src/dml/update/mssql_update_executor.cpp:90-96`
- `src/dml/delete/mssql_delete_executor.cpp:92-99`

There is no savepoint or compensation path in `src/dml` — `MSSQLTransaction::GetNextSavepointName()` (`src/catalog/mssql_transaction.cpp:197`) is never called from the DML path.

## What changes for a user

Take `UPDATE mssql.dbo.T SET x = 1` over 2000 rows in autocommit — four batches of 500 — where batch 3's response desyncs.

| | today (main) | with PR #342 |
|---|---|---|
| Batches 1-3 | committed server-side | committed server-side |
| Batch 4 | runs, committed | never runs |
| Client sees | success | `MSSQL UPDATE failed: batch 3: … TDS parse error` |
| Rows actually updated | 2000 | 1500 |
| Can the user tell? | no — it silently lied | no — the error names no count |

So today's behaviour is *wrong but complete*; the new behaviour is *honest but partial*, and the user still cannot tell how far it got. Neither is good. The question is which is less bad, and what the error should say.

## Options

**A. Ship the throw, and put the applied-batch count in the message.**
Something like `MSSQL UPDATE failed after 1500 of 2000 rows (batch 3 of 4): TDS parse error: …`. A visible failure beats a silent one, and naming the boundary makes the state recoverable by hand. Cost: a statement that previously "worked" now errors, which is a breaking change for anyone who was unknowingly relying on it.

**B. Throw only inside a transaction; keep autocommit's current behaviour.**
Conservative — no behaviour change on the default path. Cost: keeps a known silent-data-loss bug in the commonest case, which is hard to defend now that we know it is there.

**C. Keep only the memory guard, drop the error reporting.**
PR #342 also stops feeding a parser that is already in `ParserState::Error` (it never consumes, so `buffer_` grew by the whole remaining response). Nothing has argued against that half. This ships the leak fix and defers the semantics question entirely.

## Related

Found during an adversarial review of #342 — full write-up in the inline comments there. This is the one finding I did not want to resolve by choosing.

Two smaller things in the same area, mentioned for context but not blocking this decision:

- `src/dml/insert/mssql_returning_parser.cpp` currently hangs the full 30 s on the same class of desync and then reports a socket error rather than the parse error, because it never checks `IsEndOfMessage()`. Fixing that is uncontroversial and is in #342's review.
- Whichever option is chosen, the four DML response loops are currently untested for this path — `test/sql/regression/issue_323.test` drives only `mssql_exec()` / `mssql_scan()`, never a DuckDB-native `INSERT`/`UPDATE`/`DELETE`.

## Ask

Which of A / B / C, and if A, should the count go in the exception message, in `duckdb_logs`, or both?

VGSML:
Spec `specs/062-insert-via-bcp/spec.md`. Everything the spec lists is in, one commit per step; the spec's § 6 carries the measurements.

**What changes for a user**

- `INSERT` into an attached table loads through the BCP protocol (`INSERT BULK`) above `mssql_insert_bcp_threshold` rows (default 1000, measured crossover: ~300 rows locally, ~1000 at a 20 ms RTT). 1M rows × 3 columns into an existing table: **1.8 s** on one writer, **0.5 s** on four, against 74 s before. Per family at 1M rows: 3–18× on wall, 6–17× on client CPU (§ 6.3). `RETURNING`, an explicitly named identity column and `mssql_insert_use_bcp = false` keep the statement path; the decision between "few rows" and "many" is made by counting them, never by a plan estimate.
- **Every DML statement is atomic in autocommit** — INSERT on both paths, UPDATE and DELETE (closes #344): one connection per statement, its batches bracketed in a server transaction, `rolled back` in the message on failure. Tables are left exactly as they were; the connection goes back to the pool.
- The statement path never carries more than 1000 constants per statement (W1b): SQL Server auto-parameterises a multi-row `VALUES` INSERT only up to that line, and the old 1000-row statements sat past it for every table with two or more columns — that, not bytes, was the 74 s. (`RETURNING` does not gain: an `OUTPUT INSERTED` statement costs ~70 µs a row at any size — measured 69.8 s → 70.2 s at 1M rows.)
- `is_identity` in the catalog metadata (W4, the metadata half of #327).

**What the measurements changed on the way**

1. A bulk load ignores CHECK constraints, does not fire triggers and writes a DEFAULT where the stream says NULL — a 1000-row INSERT with a CHECK violation in row 950 loaded every row. The INSERT's `INSERT BULK` carries `CHECK_CONSTRAINTS, FIRE_TRIGGERS, KEEP_NULLS`; `COPY` keeps bcp's contract.
2. Parallel writers each holding their own transaction deadlock **client-side** when their locks conflict (B waits on A's uncommitted rows, the server stops reading B, B blocks in `send()`, A's commit waits for B's `Combine`) — no server deadlock, nothing times out: a hang past ten minutes on a heap on row locks, a 30 s BCP timeout on a clustered rowstore. Writers fan out only on a heap under TABLOCK or a clustered columnstore without it; anything else loads on one writer.
3. `sp_executesql` with a `DECLARE` block for the sub-threshold statement was measured 16× slower than the literal form and is recorded as rejected; the server's own auto-parameterisation is what the 1000-constant cap buys.

**Structure:** W0 moved COPY's and CTAS's shared writer onto `BulkLoadSession` first (two move commits, no test changed), so the INSERT sink is the third consumer of one session type rather than a fourth copy; `LoadTransaction`, `MSSQLStatementConnection`, `MSSQLStagedRows` and `BCPColumnMetadata::FromServerColumn` are the seams spec 066 (UPDATE/DELETE via `#temp`) takes over (§ 2.9).

**Verification:** full integration suite 193 cases / 6196 assertions green; C++ 31/31; eight new SQL tests (paths, types row-by-row on both paths, transaction, atomicity, statement semantics, parallel writers per shape, plan cache, statement-path atomicity for INSERT/UPDATE/DELETE); wide-write bench within noise after W0 (§ 6.1).

Closes #344.


VGSML:
**Self-review** (`98ab3ee`): three independent passes over the diff — sink concurrency/lifetime, the statement path's transaction, plan side + catalog + docs. Seven findings, all fixed in that commit, suites green locally (insert 15/15 with the new cases, dml, copy, ctas, transaction, regression; C++ 31/31):

1. **Sink, race** — a thread that passed the error check and then waited on `write_mutex` while the holder abandoned the shared session would write into a session with no writer (null deref) or open a second stream on the closed connection. The check is repeated under the lock; `Write` on a closed session throws.
2. **Sink, leak of an open transaction** — `TryStart`'s fallback after `Adopt` had sent `BEGIN TRANSACTION` returned the connection to the pool mid-transaction, and the session's destructor then sent `ROLLBACK` down a connection the pool may have handed elsewhere. The fallback abandons (rollback + clear) before the connection goes back.
3. **Statement path, partial commit through the destructor** — `~MSSQLInsertExecutor` flushed pending rows "in case the caller forgets" and, with W1c, committed them: on an unwind that had not gone through `FailStatement` (a row over `mssql_insert_max_sql_bytes`, an interrupt) that was exactly the partial application W1c removes. No I/O from the destructor now.
4. **Statement path** — `Fail` closed the transaction's *pinned* connection when it was not Idle, which may be another consumer's stream; the pinned connection is dropped as it is, as before.
5. **LoadTransaction** — a BEGIN whose response hides the descriptor is undone before the failure is reported; an unconfirmed ROLLBACK closes the connection instead of returning it in an unknown state.
6. **Writer rule read the cached shape** — a table cached as a heap and given a clustered index since (`mssql_exec` does not invalidate) would fan out into the client-side deadlock the rule exists to prevent. The shape is queried live when the stream opens (`TargetResolver::QueryTableShape`); `insert_bcp_parallel.test` loads that stale-cache case.
7. **Routing** — a view, and a column the bulk wire cannot carry (geometry, hierarchyid, sql_variant), stay on the statement path; `insert_bcp_paths.test` covers both.

Also found on the way, not in this PR: `INSERT` of a `GEOMETRY` value fails on main too (the codec renders WKB as `0x…`; the server wants `geometry::STGeomFromWKB`), and a table with a `rowversion` column is unreadable through the catalog (cast to `nvarchar(max)`, error 529 — that is #296). Both go into a small types PR right after this one.

VGSML:
## How to review this PR

It is 12 commits and ~4 800 lines, but it is built to be read **commit by commit** — each one is a self-contained step of the spec, and three of them are moves with no test changed. Suggested order and what to look at in each.

### 1. The design — `0118564` (spec only)
`specs/062-insert-via-bcp/spec.md`. Read § 0 (the measurements the design rests on) and § 1 (findings) first; § 2 is the work, one W per commit below; § 6 the numbers after. Everything a reviewer might want to contest is stated as a decision there, not buried in code:
- the two paths and the runtime threshold (W1), and why not a plan estimate (F3);
- atomicity in autocommit via one server transaction per statement / per writer (§ 0.2, W1c, W2);
- **parallel writers only on a heap under TABLOCK or a columnstore without it** — because transactional writers with conflicting locks deadlock *client-side* (W2, measured);
- **`INSERT BULK` with `CHECK_CONSTRAINTS, FIRE_TRIGGERS, KEEP_NULLS`** — a bulk load ignores all three by default, an INSERT may not; COPY keeps bcp's contract (W2);
- the 1000-constant cap on statements, and the rejected `sp_executesql` form (§ 0.5, W1b);
- what spec 066 (UPDATE/DELETE via `#temp`) takes over (§ 2.9).

### 2. Two moves — `db3a070` (COPY), `b85b18f` (CTAS)
`BulkLoadSession::Adopt` + the shared writer of COPY and CTAS moved onto the session. **No test changed.** The one behaviour change is named in the CHANGELOG: the success-path release now sets the reset flag from `mssql_reset_connection`, as every other release path did. Check: `src/copy/bulk_load_session.{hpp,cpp}` (Adopt / TryStart / Write / Finish), then confirm `copy_function.cpp` and `mssql_ctas_executor.cpp` lost code and gained none. Proof: full suite green after each.

### 3. Small, independent — `04282c5` (W4), `a0cd7e7` (W1b), `077be7a` (W3)
- W4: `c.is_identity` appended **last** to the four column-metadata queries, parsed at five sites by index (`src/catalog/mssql_metadata_cache.cpp`); `MSSQLColumnInfo::is_identity`. Check the indices against each SELECT list.
- W1b: `MSSQLInsertConfig::RowsPerStatement` = `min(batch_size, 1000 / inserted columns)`; three lines in the batch builder.
- W3: `BCPColumnMetadata::FromServerColumn` is the body of the old loop in `GetExistingTableColumnMetadata`, verbatim (`git show main:src/copy/target_resolver.cpp` to diff by eye); `test/cpp/test_bcp_column_metadata.cpp` pins the type table.

### 4. The statement path's transaction — `6e44027` (W1c, #344)
New: `copy/load_transaction.{hpp,cpp}` (BEGIN with the ENVCHANGE descriptor captured — error 3989 without it — COMMIT/ROLLBACK clearing it) and `dml/mssql_statement_connection.{hpp,cpp}` (one connection per statement: Acquire / Commit / Fail / a destructor without a ClientContext). Then the three executors: every `ReleaseConnection` per batch became `Fail` on error paths and the success release went away; `Commit` in `Finalize`. **What to check:** every early return/throw in the three `ExecuteBatch` bodies reaches `Fail` exactly once (the self-review pass did this; a second pair of eyes is welcome). Proof: `test/sql/regression/issue_344_dml_desync.test` (rewritten: tables unchanged after each failure) and `test/sql/dml/dml_statement_atomic.test`.

### 5. The sink — `3eb9105` (W2) — the one to spend time on
`src/dml/insert/mssql_physical_insert.{hpp,cpp}` (~500 lines) and `PlanInsert` in `src/catalog/mssql_catalog.cpp` (the routing and `MSSQLInsertBulkPlan`). Read in this order: `PlanInsert` → `MSSQLInsertGlobalSinkState` → `Sink` (staging under `write_mutex`, then `OpenSharedStream`, then the COPY-style claim loop) → `Combine` (a thread's session moves into `finished_sessions`) → `Finalize` (close, commit all, release all). Concurrency questions to ask: who holds which mutex (`write_mutex` → `MaterializeMutex` → `finished_mutex`, never inverted), what happens on a throw from each of the three phases, and whether the first-failure protocol (`has_error` re-checked under the lock, since `98ab3ee`) covers the thread that was waiting. Proof: `insert_bcp_atomic.test` (a CHECK violation in batch 5 of 5 → 0 rows, the batch named), `insert_bcp_transaction.test`, `insert_bcp_parallel.test` (fan-out by shape, and the stale-cache case).

### 6. Tests and docs — `acf226d`, `b2c2714`, `5b5231d`
Eight new `.test` files; each header says what it pins and why that way (the path is observed through the statement path's test lever, since EXPLAIN cannot be queried from sqllogictest). `5b5231d` fixes a pre-existing bug the cap made visible: `INSERT … RETURNING` returned only the last statement's rows. Docs: `website/docs/writing/dml.md` is the user-facing statement of all of the above.

### 7. Self-review — `98ab3ee`
Seven findings, listed in the comment above; read the commit message for the scenario behind each. If you have time for one thing here, it is the `TryStart` fallback and the `~MSSQLInsertExecutor` change.

### Not worth review time
The bench group in `test/bench/bench_live_server.sh` (additive), CHANGELOG/CLAUDE.md/DATAMODEL wording, the metadata-cache comment fixes.

### Verification already done
Full integration suite green after every commit that touches code (193 cases at the end); C++ 31/31; CI green on the head (Concurrency stress and Kerberos e2e included); wide-write bench within noise after the moves (§ 6.1); INSERT bench per family (§ 6.3): 3–18× on wall, 6–17× on client CPU at 1M rows.

src/dml/insert/mssql_physical_insert.cpp
```diff
@@ -9,89 +36,483 @@ namespace duckdb {
...
+	// for B's Combine. The server sees no deadlock -- one side is a client --
+	// and nothing times out (measured: a hang, 10 minutes and counting, on a
+	// heap without TABLOCK; a 30 s BCP timeout on a clustered rowstore). Only
+	// two shapes let concurrent transactional bulk loads coexist: a heap under
+	// TABLOCK (BU locks are mutually compatible, and nothing escalates) and a
+	// clustered columnstore without it (each session fills its own rowgroups;
+	// measured 1M rows in 0.49 s and 0.80 s at four writers). Everything else
+	// -- a clustered rowstore index, a heap on row locks, a columnstore under
+	// a table lock -- gets one writer.
+	const bool locks_compatible = (live_shape == MSSQLIndexKind::HEAP && bulk.tablock) ||
```
oluies: **High — this gate cannot see a nonclustered index, and the hang it rules out is the one it enables.**

`live_shape == HEAP && bulk.tablock` rests on "BU locks are mutually compatible" (comment above). That holds only for a heap with **no indexes at all** — SQL Server's rule is *"if the table has no indexes and TABLOCK is specified, the table can be loaded concurrently by multiple clients"*. Add a nonclustered index and the bulk load takes an exclusive table lock instead.

`MSSQLIndexKind` can't distinguish the two. Both the catalog query (`WHERE i.index_id IN (0, 1)`) and the new `QueryTableShape` (`src/copy/target_resolver.cpp:442,447,479,486` — `i.index_id <= 1`) filter nonclustered indexes out, so a heap with `PRIMARY KEY NONCLUSTERED` or any `CREATE INDEX` still reports `HEAP`.

Because spec 062 defers every writer's COMMIT to `Finalize`, the result is exactly the client-side deadlock lines 215-226 describe and rule out — *"the server sees no deadlock … and nothing times out"*:

```sql
CREATE TABLE dbo.T (id INT NOT NULL PRIMARY KEY NONCLUSTERED, n INT);  -- a heap
SET threads = 4;
INSERT INTO mssql.dbo.T SELECT ... ;  -- a few million rows, autocommit
```

`mssql_copy_tablock = auto` → heap → TABLOCK on; `live_shape == HEAP` → `locks_compatible` true → up to 4 transactional writers. Writer B blocks on the table lock held by A's uncommitted load → the server stops draining B → B blocks in `send()` → A's commit never runs, because `Finalize` is waiting on B's `Combine`. Indefinite hang, no deadlock victim, no timeout.

COPY is safe today only because its writers commit per batch (`own_transaction = false`); this path is the first to hold the transaction open across the fan-out.

**Suggested fix:** make the rule "heap with no indexes at all". `QueryTableShape` is already a dedicated round trip on the load connection, so it can also return `EXISTS (SELECT 1 FROM sys.indexes WHERE object_id = … AND index_id > 1)`; force `parallel_writer_limit = 1` when a heap target has any nonclustered index.

src/dml/insert/mssql_physical_insert.cpp
```diff
@@ -9,89 +36,483 @@ namespace duckdb {
...
+		gstate.shared.Commit();
+		{
+			std::lock_guard<std::mutex> finished_lock(gstate.finished_mutex);
+			for (auto &session : gstate.finished_sessions) {
+				session->Commit();
+			}
+		}
+	} catch (std::exception &e) {
+		const string msg = StringUtil::Format("INSERT via BCP failed to commit: %s; %s", e.what(),
+											  MSSQLRowsBeforeOutcome(gstate.transaction_pinned, 0));
```
oluies: **Medium — this message claims a rollback that did not happen, and the comment above it says so.**

`MSSQLRowsBeforeOutcome(gstate.transaction_pinned, 0)` renders *"rolled back, the 0 row(s) from the batches before it included"*. But the multi-writer case is the opposite: `gstate.shared.Commit()` and each `finished_sessions[i]->Commit()` are separate server round trips, so a failure at writer *k* leaves writers *0..k-1* committed.

**Failure scenario:** a 4-writer autocommit INSERT of 4M rows; shared writer and two parallel writers commit, the fourth's `COMMIT TRANSACTION` fails (connection reset). The user reads `… rolled back, the 0 row(s) … included`, then finds ~3M rows in the table — the worst shape for this error, since it invites a blind retry onto a partially loaded table.

**Suggested fix:** only the single-writer / pinned case can honestly claim a clean rollback. In the multi-writer case, count the writers that did commit and report the rows actually in.

src/dml/insert/mssql_physical_insert.cpp
```diff
@@ -9,89 +36,483 @@ namespace duckdb {
...
+		materialize_lock = std::unique_lock<std::mutex>(catalog.MaterializeMutex());
+	}
+
+	auto connection = ConnectionProvider::GetConnection(context, catalog);
+	if (!connection) {
+		throw IOException("INSERT via BCP: failed to acquire a connection");
+	}
+	if (connection->GetState() != tds::ConnectionState::Idle) {
+		if (!gstate.transaction_pinned) {
+			catalog.GetConnectionPool().Release(connection);
```
oluies, line 182: **Low — this release path (and the one at line 198) skips the `mssql_reset_connection` contract.**

Both non-pinned early exits in `OpenSharedStream()` — connection not `Idle` here, `QueryTableShape` throwing at line 198 — hand the connection back with a raw `catalog.GetConnectionPool().Release(connection)`. Every other release path in the extension goes through `ConnectionProvider::ReleaseConnection` / `SetNeedsReset(...)` and honours the setting (issue #189); these two ship whatever reset bit the connection happened to carry.

**Failure scenario:** with `SET mssql_reset_connection = false`, a bulk INSERT whose `QueryTableShape` fails returns the connection with the prior flag intact, so the next user of that pooled connection may take a session reset the setting disabled — losing `SET` options, isolation level, `CONTEXT_INFO`.

**Suggested fix:** `connection->SetNeedsReset(gstate.reset_on_release)` before `Release`, or route both through `ConnectionProvider::ReleaseConnection`. `gstate.reset_on_release` is already captured for exactly this.

src/include/dml/insert/mssql_insert_executor.hpp
```diff
@@ -93,8 +98,15 @@ class MSSQLInsertExecutor {
 
 	// State
 	bool finalized_;
+	//! A batch failed: the transaction is rolled back and the connection gone,
+	//! so nothing pending may be sent (the destructor used to flush it).
+	bool failed_ = false;
```
oluies, line 106: **Low — `failed_` is written twice and read never; its comment describes a guard that does not exist.**

Assigned in `FailStatement` and `CommitStatement`, consulted nowhere. The declaration comment presents it as what stops pending rows being sent after a batch failure (*"the destructor used to flush it"*), but the destructor now only logs. The actual protection is `~MSSQLStatementConnection` rolling back — a different mechanism in a different class.

**Why it matters:** a later change that reintroduces a flush in `~MSSQLInsertExecutor`, or adds a post-failure `Execute` path, reads this comment and assumes enforcement that isn't there.

**Suggested fix:** either delete `failed_` and point the comment at `~MSSQLStatementConnection`, or make it real — have `Execute` / `ExecuteWithReturning` / `Finalize` throw `InternalException` when `failed_`.

test/sql/insert/insert_returning_batches.test
```diff
@@ -0,0 +1,67 @@
...
+2500
+
+# The same through a plain sequence, to pin what the hash stands for.
+query I rowsort
+SELECT i FROM range(1, 2501) t(i);
+----
+2500 values hashing to 9394fe7df1c8c49a465d18f8c783d03f
+
+# RETURNING * with a server-filled column, over one statement boundary
+# exactly (334 rows = 333 + 1).
```
oluies: **Medium — this case does not cross the statement boundary it is named for.**

The annotation says *"over one statement boundary exactly (334 rows = 333 + 1)"*, but 333 is the per-statement cap for a **3-column** insert. This insert names two columns (`(id, v)`), so `insert_column_indices.size() == 2` and `MSSQLInsertConfig::RowsPerStatement(2)` yields `1000 / 2 = 500`. 334 rows is a single statement.

So the case would pass unchanged if `ExecuteWithReturning` still kept only the last statement's OUTPUT chunk for the `RETURNING *` / partial-column-list shape — which is the regression it exists to catch. The 2500-row case above it does cross boundaries, but it uses `RETURNING id` with a full column list, so this shape is currently untested.

**Suggested fix:** size it to the real cap for a 2-column list (501 rows), or add a third inserted column so 334 genuinely is 333 + 1.

oluies (commented):
Findings from a roborev branch review (job 1632, `claude-code`) over all 12 commits, `cc5652d..98ab3ee`. Verdict: fail — 1 high, 2 medium, 2 low, inline below.

The one I would merge-block on is the **parallel-writer gate**: `index_id <= 1` in both the catalog query and the new `QueryTableShape` makes a heap-with-a-nonclustered-index indistinguishable from a bare heap, so the fan-out rule admits exactly the client-side hang the comment above it rules out. The two mediums are a commit-failure message that misreports what is in the table and a `RETURNING` test that doesn't exercise its boundary; both lows are latent.

I have not run these against a live server — the high in particular deserves a `PRIMARY KEY NONCLUSTERED` + `SET threads=4` repro before anyone takes my word for it.

src/dml/insert/mssql_physical_insert.cpp
oluies: Fixed in 918697f. `QueryTableShape` now returns a `TableLoadShape` — `kind` plus `has_nonclustered`, from a `CASE WHEN EXISTS (… index_id > 1)` subquery on the same round trip — and the gate requires a *bare* heap. A short result row reports `has_nonclustered = true`, which costs a writer and never a hang.

Left deliberately alone: the `CLUSTERED_COLUMNSTORE && !tablock` arm. A nonclustered index on a CCI is the same mechanism in principle, but it was neither reported nor measured, and narrowing it would give up the fan-out spec 062 measured. Your call whether that wants its own look.

`insert_bcp_parallel.test` gains a `PRIMARY KEY NONCLUSTERED` heap case, and `DATAMODEL.md:777` is updated — it stated the old rule.

src/dml/insert/mssql_physical_insert.cpp
oluies: Fixed in 918697f. `committed_writers` is counted across `shared.Commit()` and the per-session loop; only the single-writer case still uses `MSSQLRowsBeforeOutcome`, and the multi-writer message now names the writers that committed and warns the rows were **not** rolled back.

src/dml/insert/mssql_physical_insert.cpp
oluies: Fixed in 918697f — both paths (:178 and :198) now call `connection->SetNeedsReset(gstate.reset_on_release)` before `Release`. `gstate.reset_on_release` was already populated at :161, so nothing new had to be threaded in.

src/include/dml/insert/mssql_insert_executor.hpp
oluies: Fixed in 918697f by making it real rather than deleting it: `Execute`, `ExecuteWithReturning`, `Finalize` and `FinalizeWithReturning` all refuse once `failed_` is set. Safe because every `FailStatement` call site rethrows, so no working path reaches those entry points afterwards. The comment now credits `~MSSQLStatementConnection` with the rollback.

test/sql/insert/insert_returning_batches.test
oluies: Fixed in 918697f — resized to 501 rows (500 + 1 for a two-column list).

The hash was not guessed: I first reproduced the existing 334-row hash exactly (`ef7f091a…`) to confirm the rowsort/md5 convention, then computed the 501-row one from real DuckDB CLI output — `1b17b822d7fd363f2539dcd132ae4676`, 1503 values.

VGSML:
Reviewed your five findings and `918697f`, and took the high one to the server rather than taking your word for it — you were right to ask.

## The writer rule: confirmed, and the mechanism is sharper than the finding says

`sys.dm_tran_locks` for the target, read while the load's transaction was still open (`INSERT BULK ... WITH (TABLOCK)`, 5000 rows into each):

| target | lock granted |
| --- | --- |
| heap, no indexes | **BU** |
| the same heap + `PRIMARY KEY NONCLUSTERED` | **Sch-M** |

Not merely "an exclusive table lock": **Sch-M**, which is incompatible with everything including another Sch-M. Two transactional writers on that table cannot both be granted, exactly as you argued.

What it costs, 400k rows into a heap with a nonclustered PK, `threads = 4`, `mssql_copy_parallel_writers = 4`, on a build with the gate as it was before your commit against one with it:

| gate | wall | writers used |
| --- | --- | --- |
| `live.kind == HEAP` | 30.97 s | 1 of 4 |
| `bare_heap` | **0.55 s** | 1 of 1 |

**One correction to the finding, in your favour and mine:** it is not the indefinite hang. Each extra writer blocks inside `TryStart` — on its own `INSERT BULK`, before a single row — and that read carries the 30 s BCP timeout, after which `TryStart` reports a failed claim. A failed claim is deliberately not an error ("a load must not fail because it could not go faster"), so the load completes on the shared writer having paid 30 s of dead time and silently lost the parallelism it asked for. The unbounded version needs a writer that got *past* `INSERT BULK` and then blocks mid-stream in `send()`, which is the heap-on-row-locks shape I measured during development. Bounded by luck rather than by design, and 56× is damage enough. I have written this into the spec (§ 2 W2 and a new § 6.5) with both tables.

The other four are right and I have kept them as they are: the commit-failure message was claiming a rollback the code above it says does not happen, the two raw `pool.Release()` calls did skip the `mssql_reset_connection` contract, `failed_` was decoration, and the 334-row `RETURNING` case really did assert nothing (the cap for a two-column list is 500 — I should have caught that when I wrote it).

## What I added on top

- **The docs that still stated the old rule**: `CLAUDE.md`, the CHANGELOG entry, `website/docs/reference/settings.md` and `website/docs/writing/dml.md` all said "a heap under TABLOCK". You fixed `DATAMODEL.md`; these four are the rest of it, plus the spec's own statement of the rule and the measurements above.
- **`MSSQLRawMessage(e)` = `ErrorData(e).RawMessage()`** at the seven error sites spec 062 added. `e.what()` on a DuckDB exception carrying extra info returns the serialized envelope, so the messages were nesting `{"exception_type":"Invalid Input","exception_message":"MSSQL: BCP failed: ..."}` inside themselves. I noticed it in your new pool-exhaustion message on #349 and then found it in my own.

## Your deliberate omission

The `CLUSTERED_COLUMNSTORE && !tablock` arm — you left it and said it was my call. Measured, and it wants narrowing more than the heap arm did — it is the worse of the two. 400k rows into a clustered columnstore carrying one nonclustered index, `threads = 4`, `mssql_copy_parallel_writers = 4`:

| clustered columnstore target | wall | rows landed |
| --- | --- | --- |
| no nonclustered index | 0.63 s (4 writers) | 400000 |
| + one nonclustered index, gate as you left it | 30.81 s | **0 — the INSERT failed and rolled back** |
| + one nonclustered index, gate narrowed | 0.99 s (1 writer) | 400000 |

The difference from the heap case is where the extra writer blocks. On the heap it blocks *at* its `INSERT BULK`, inside `TryStart`, so the 30 s timeout is reported as a failed claim and the load survives. On the columnstore it gets past `INSERT BULK` and times out reading its BCP response mid-stream — that is an error, so the whole statement rolls back and lands nothing. `has_nonclustered` now gates both arms (`4d904cd`), `insert_bcp_parallel.test` carries both shapes, and the numbers are in spec § 6.5 beside yours.

## #349

Approved, verified live, and merged into this branch (`41dc1d9`) rather than separately, per the maintainer. I pushed the one thing it was missing — a regression test for the case-sensitive sink match, which reproduces cleanly — to your branch as `6dfb5c6`. The two notes there: the prepared-session claim is defensive rather than a live fix (the only planner site that copies bind data is late materialization, which we never opt into), and the same JSON-envelope wart as above.

## Verification

Full SQL suite on the merged branch: **194 cases, 6225 assertions, green**. C++ 31/31. CI green on `918697f` (16 pass).

VGSML:
**Measured against main, end to end** (`ddec241` vs this branch, one fixture, both binaries, interleaved in the same session, two rounds, min per cell, no settings touched — spec § 6.8):

| cell | main | this PR | ratio |
| --- | --- | --- | --- |
| `INSERT … SELECT`, 1M rows | 74.71 s | **0.51 s** | **0.01** |
| 20 × `INSERT … SELECT` of 500 rows (under the threshold) | 0.729 s | **0.219 s** | 0.30 |
| `INSERT … RETURNING`, 100k rows | 7.44 s | 7.15 s | 0.96 |
| `COPY … (FORMAT 'bcp')`, 1M rows | 0.739 s | 0.520 s | 0.70 |
| `CREATE TABLE … AS SELECT`, 1M rows | 0.855 s | 0.745 s | 0.87 |
| `UPDATE` 5000 rows by rowid | 0.318 s | 0.312 s | 0.98 |
| `DELETE` 5000 rows by rowid | 0.214 s | 0.223 s | 1.04 |

145× on the headline and 31× on client CPU (1.604 s → 0.051 s); 3.3× on small inserts, which is W1b's constant cap alone. COPY and CTAS are the W0 move and are **not** claimed as an improvement — 0.70 and 0.87 sit inside the band this fixture swings over two rounds on a shared docker server, and nothing in that move touches the wire.

The line I had not measured before: **W1c's per-statement server transaction costs nothing** — UPDATE 0.98, DELETE 1.04, the latter ~9 ms over 5000 rows, which is the `BEGIN` and `COMMIT` round trips at a local RTT. At a 20 ms RTT that is ~40 ms per statement, once, whatever its size — the price of the statement no longer being applied in part (#344).

oluies (approved):

oluies:
**Integration suite now run — green.** The fix commit (918697f) had been verified only by build + C++ tests; it is now verified against a live SQL Server 2022, so the "compiled, not run" caveat in that commit message is discharged.

```
All tests passed (5 skipped tests, 6220 assertions in 194 test cases)
integration-test: 194 cases ran (floor 150)
```

199 files, zero failures. The 5 skips are env-gated (spatial, Kerberos, WinSSPI, named-instance, counters).

The two tests this commit touched both matter here:

| Test | Result |
|---|---|
| `insert_bcp_parallel.test` | **3.785s** — carries the new `heap + PRIMARY KEY NONCLUSTERED` case. Its regression signature is a *hang*, so finishing in under 4 s is the assertion. |
| `insert_returning_batches.test` | 0.254s — runs the resized 501-row case, which independently confirms the hash computed for it (`1b17b822d7fd363f2539dcd132ae4676`). |

Environment, since it cost some false signal on the way: the suite must not run against host port 1433 on a machine where another project already has a SQL Server there, and `docker/docker-compose.yml` needs a local `.env` (copy `.env.example`) or `make docker-up` fails outright. A stale long-lived container also produced two failures that vanish on a fresh one — `ctas_large.test` failed with a BCP timeout there and passes in 0.54 s on a clean server.

Still not covered: the nonclustered-heap deadlock is prevented rather than demonstrated — nothing here proves the *old* code hangs on that shape, only that the new code loads it correctly on one writer.

PR #349, opened 2026-09-15
fix: three live findings from roborev reviews of specs 075/076

oluies:
Follow-up to roborev jobs **1543** (spec 075) and **1594**. Of the ten findings across both, **seven were already fixed on main** by the follow-up commits — several citing the review by name (`"review of #345"`) — and **one is disproved**. These are the three that were still live, plus one cosmetic.

## What changes

**`prepared := true` shared one session between every global state over the same bind data.** `FunctionData::Copy` shares `prepared_session`, and that session owns exactly one connection carrying exactly one `sp_prepare` handle. A plan that duplicates a `LogicalGet` over a prepared scan (delim join, duplicated CTE) sent `EXEC sp_execute` down a connection the first stream was still reading — `connection not in Idle state`, on a query the default path serves fine. The session is now claimed with a CAS; a loser runs `fallback_sql`, the ad-hoc batch kept aside before the handle form replaced `execute_sql`.

Two subtleties the claim has to respect:

- A session with **no** connection is the in-transaction case, where the handle lives on the *pinned* connection and `execute_sql` must stay the `sp_execute` form. Only a genuinely **lost** claim falls back — otherwise this would have quietly disabled `prepared := true` inside every transaction.
- The claim is released in `~MSSQLScanGlobalState` **after** `result_stream.reset()`: the stream borrows the connection, and a sticky claim would send every later `EXECUTE` of a prepared statement down the ad-hoc path forever.

**Pool exhaustion while binding a prepared scan said nothing about the cause.** Each prepared scan holds a pooled connection from Bind until the bind data dies, so a plan with several can exhaust `mssql_connection_limit` and fail during binding with a bare acquire timeout. The message now names `prepared := true` and the setting.

**`CollectSinkCatalogs` matched catalog names case-sensitively.** The sink side is `MSSQLCopyBindData::catalog_name`, verbatim from the URI's first path segment; the scan side is the attached catalog's own name. DuckDB resolves aliases case-insensitively, so `mssql://SK75/dbo/dst` is a legal sink into `sk75` — and the mismatch silently found no sink and skipped materialization, reintroducing #239 with no diagnostic. Now `case_insensitive_set_t`.

**`ApplyWithClause`'s `ToValue()`** can throw for a number too wide even for a bignum; it now reports under the same `MSSQL: table option ...` wording as every other rejection there.

## Deliberately not done

Capping how many sessions one statement may hold (the "ideally" half of the connection-reservation finding). That is a design decision about connection ownership, not a bug fix — and the finding's "at minimum, degrade instead of failing" cannot work as written, since describing the statement needs a connection too. **roborev job 1543 is left open** for that call.

## Disproved, no change

The claim that moving the NULL guard from the bound `Value` to `Literal::IsNull()` may have changed semantics. It did not: `Literal::IsNull()` is `kind == LiteralKind::NULL_LITERAL`, and that enumerator *is* the SQL `NULL` keyword. The same finding's "second `ToValue()` binding pass" does not exist either — there is exactly one call. (The reviewer flagged this as unverifiable because the `duckdb/` submodule was not checked out; it is now.)

## Verification

Release build green; **30/30** standalone C++ tests pass (`make test-cpp-run`). The **SQL suite was not run** — it needs a live SQL Server. No new SQL tests here: the prepared-session race needs a plan that duplicates a `LogicalGet`, which I could not construct and assert without a server.


VGSML (commented):
Reviewed and **verified against a live server** — the part your PR text says you could not do. Build green, C++ 30/30 as you report, and I ran the full SQL suite on `9523cf3` (result at the end). Three of the four changes I can confirm behave as described; the fourth is sound but, as far as I can tell, cannot currently be reached.

## Verified live

**`CollectSinkCatalogs` case sensitivity — real, and the nastiest of the three.** Reproduced on a build without the fix, catalog attached as `sk75`:

```sql
BEGIN;
COPY (SELECT id, v FROM sk75.dbo.cs_src) TO 'mssql://sk75/dbo/cs_dst' (FORMAT 'bcp', CREATE_TABLE false);  -- 211 rows
COPY (SELECT id, v FROM sk75.dbo.cs_src) TO 'mssql://SK75/dbo/cs_dst' (FORMAT 'bcp', CREATE_TABLE false);
-- Invalid Input Error: MSSQL COPY: Connection is busy executing another query ...
```

One character of case, and #239 is back with the error message that issue was closed on. With your fix the second form copies its 211 rows. Your reading of the two sides is right: `TargetResolver::ResolveURL` does `target.catalog_name = parts[0]` and never normalises it, while the scan side is keyed by the attached catalog's own name.

**The one thing this PR is missing is a test for exactly that** — and unlike the prepared race, it is trivial to write: `test/sql/transaction/sink_reads_own_catalog.test` already attaches as `sk75` and is the file this belongs in. I wrote and ran it; happy to push it to your branch, or take it:

```
# --- the sink's catalog spelled in another case
statement ok
SELECT mssql_exec('sk75', 'TRUNCATE TABLE dbo.sk75_copy');

statement ok
BEGIN;

statement ok
COPY (SELECT id, v FROM sk75.dbo.sk75_src) TO 'mssql://SK75/dbo/sk75_copy' (FORMAT 'bcp', CREATE_TABLE false);

statement ok
COMMIT;

query I
SELECT count(*) FROM sk75.dbo.sk75_copy;
----
5000
```

**Pool exhaustion while binding prepared scans — reproduced.** `mssql_connection_limit = 2` with three `prepared := true` scans in one statement, before: `IO Error: MSSQL: Failed to acquire connection: pool 't' timed out (2 active of 2, limit 2)`. After: the sentence naming the option. Worth having.

**Prepared paths still work.** Plain, `EXECUTE` twice over a `PREPARE`, two prepared scans inside one transaction, and `mssql_scan_params(..., prepared := true)` — all correct on your branch. The in-transaction subtlety you call out in the PR body is handled: an empty `connection` never claims and never falls back, so `prepared := true` still uses the handle on the pinned connection.

## The prepared-session race: the fix is sound, the trigger is not reachable

I could not build the race either, and I think there is a structural reason. `MSSQLScanBindData::Copy()` is called by exactly one planner-side site in the pinned DuckDB — `LateMaterializationHelper::CreateLHSGet` (`src/optimizer/late_materialization_helper.cpp:9`) — and `LateMaterialization` gates on `get.function.late_materialization` (`late_materialization.cpp:228`). Nothing in `src/` sets `late_materialization` or `get_row_id_columns` on a `TableFunction`; our rowid support lives on `MSSQLTableEntry`, not on the scan function. The other `bind_data->Copy()` callers are `multi_file_reader` and the aggregate-export path, neither of which we go through.

So a delim join or a duplicated CTE gets two *binds* (two sessions, two handles), not two global states over one bind data. I tried the three shapes that looked most likely on a pre-fix build — an inlined CTE self-joined, a correlated subquery, a `MATERIALIZED` CTE read twice — and all three pass; no `connection not in Idle state`.

That makes the new `else if (lost_claim …)` branch unreachable by any query either of us can write today, which is worth saying out loud in the PR text: it is **defensive**, and the cost (one CAS, one string kept) is small enough that I would keep it. But I would reword the justification — as written it claims a live failure — and if you do know a shape that copies the bind data, that shape is the test this needs.

## One nit, and it is mine too

The new message nests DuckDB's JSON envelope:

```
IO Error: mssql_scan: could not acquire a connection for 't' while binding a `prepared := true` scan:
{"exception_type":"IO","exception_message":"MSSQL: Failed to acquire connection: pool 't' timed out ..."}. Each prepared scan holds ...
```

`e.what()` on a DuckDB exception carrying extra info returns the serialized form; `ErrorData(e).RawMessage()` gives just the text. Spec 062's bulk-load error messages have the same wart — I am fixing mine in #348, so this is a note rather than a request.

## Verification

Full SQL suite on `9523cf3` against SQL Server 2025: **185 cases, 5965 assertions, all green** (5 skipped — the cloud lanes). C++ 30/30. Plus the repros above and the new test case, run on your branch.

VGSML (approved):
Approved. Verified live: the case-sensitivity finding reproduced and fixed (COPY into `mssql://SK75/...` from a catalog attached as `sk75`, inside a transaction, failed with "Connection is busy" before and copies its rows after), the pool-exhaustion message reproduced and improved, every prepared path still correct, full SQL suite green on your branch (185 cases, 5965 assertions) and C++ 30/30.

I pushed the missing regression test for the case-sensitivity fix to this branch (`6dfb5c6`) — `sink_reads_own_catalog.test`, the file that owns that invariant.

Per the maintainer's call these commits go in through #348 rather than separately, so this PR will be closed as landed once #348 merges — its branch stays put until then. The two notes from my review that are not blockers: the prepared-session claim is defensive rather than a live fix (the only planner site that copies bind data is late materialization, which we never opt into), and the nested `e.what()` renders DuckDB's JSON envelope — I fixed the same wart on spec 062's side with `ErrorData(e).RawMessage()`.

VGSML:
Heads-up on the test I pushed here (`6dfb5c6`): its first CI run failed, the rerun of that same job passed, and I could not reproduce the failure anywhere. Recording what it was and what I ruled out, so it is not a mystery if it ever comes back.

The failure was `sink_reads_own_catalog.test:130` — the uppercase-sink COPY — with `Failed to execute SQL batch: Cannot execute: connection not in Idle state`. That message comes from the result stream, so it is a **scan** colliding with the COPY's init on the pinned connection: the signature of a scan that was NOT materialized.

Ruled out:

- **The fix works.** With `MSSQL_DEBUG=1` the optimizer says `sink: COPY into catalog 'SK75'` → `catalog 'sk75': 1 catalog scan(s), 0 raw scan(s), sink=1` → `TableScanInitGlobal: materializing`. Before your change the same statement fails; after it, it materializes and copies.
- **Not the build.** That run's Build job compiled for 28 minutes, so the artifact carries the fix.
- **Not the runner mode.** CI runs the suite as a file list without `--force-reload`; I reproduced that exact invocation locally (199 files, `-f`) and the test ran and passed at position 187. Also 5/5 alone with `--force-reload` and 20/20 without, plus full suites on both branches.
- **Not branch-specific.** The identical file passed in CI on #348, which carries these commits merged in.

What is left unexplained is a one-in-many timing window on a 2-core runner against SQL Server 2022 (local is 2025). If it shows up again, the thing to instrument is whether `requires_materialization` survives from the optimizer pass to the physical plan — a later pass re-binding the `LogicalGet` would drop it silently, and that would look exactly like this.

VGSML:
Landed in main as part of #348, squashed as `b84e259`. Your three fixes and the regression test I added are in that commit, verified on main: `CollectSinkCatalogs` takes a `case_insensitive_set_t`, the uppercase-sink case sits in `sink_reads_own_catalog.test`, and the pool-exhaustion message and the prepared-session guard came along with them. Closing this one as merged there rather than separately, per the repo owner's call. Thanks for the review and for pushing the fixes rather than only describing them.

oluies:
**Integration suite now run — green.** The PR description said the SQL suite had not been run; it has now, against a fresh SQL Server 2022 container.

```
All tests passed (5 skipped tests, 5965 assertions in 185 test cases)
integration-test: 185 cases ran (floor 150)
```

190 files, zero failures. The 5 skips are env-gated (spatial, Kerberos, WinSSPI, named-instance, counters).

The files that exercise the changed paths:

| Test | Result | Covers |
|---|---|---|
| `mssql_scan_prepared.test` | 0.088s | the `prepared := true` path, i.e. the session-claim change |
| `mssql_scan_bind_describe.test` | 0.219s | `BindDescribedScan`, where the acquire error was reworded |
| `mssql_params.test` | 0.156s | `mssql_scan_params` over the same bind path |
| `sink_reads_own_catalog.test` | 0.436s | `CollectSinkCatalogs` / sink-vs-scan matching |
| `table_options` via CTAS/COPY `table_kind` cases | pass | `ApplyWithClause`'s `ToValue()` wrap |

An honest limit on that coverage: **none of these tests reproduce the bug the session-claim fixes.** That needs a plan which duplicates a `LogicalGet` over a `prepared := true` scan (delim join or duplicated CTE); I could not construct one that reliably produces two global states over the same bind data. So the suite proves the change does not *regress* the prepared path — including inside a transaction, which my first version of the fix would have silently broken — but not that it repairs the duplicate-init case. Worth a targeted test if you can think of a plan shape that forces it.

PR #350, opened 2026-09-16
spec 077: IDENTITY columns, and a rowid that does not need a primary key

VGSML:
Spec and implementation in one PR, per the standing rule. The spec (`specs/077-identity-columns/spec.md`) was reviewed on this PR in draft; the six findings of that review are addressed in it, and the code follows it one commit per work item.

## What changes for a user

- **`rowid`, and with it `UPDATE`/`DELETE`, no longer require a primary key** (W1). A table with a usable unique index — not filtered, not disabled, key columns NOT NULL — gets its rowid from it; `BIGINT IDENTITY … UNIQUE` is the common shape. A usable primary key still wins.
- **A primary key that cannot address a row falls through instead of being taken.** A `DATETIME` key produced an `UPDATE` that reported success and changed nothing (#358 — no `datetime2` literal at any precision equals a `datetime` column, measured), and a `SQL_VARIANT` key does the same through its lossy read (#354). Such a table now uses another unique index if it has one and otherwise refuses **by name** — the index, the column, the reason, the issue. A deliberate behaviour change: what it replaces is a statement that did nothing and told nobody.
- **An `INSERT` that supplies identity values works** (W2). The statement's connection is bracketed with `SET IDENTITY_INSERT … ON`/`OFF`; `OFF` is guaranteed on every way out, and a connection on which it was not confirmed is discarded. The server's refusals of the `ON` are explained (1088 needs ALTER, 8106 stale catalog, 8107 another table). A NULL in a named identity column is refused before anything is sent — DuckDB hands `DEFAULT` and `NULL` to the extension identically. It stays on the statement path; **loading many rows with their identity values is `COPY`** (W4, pinned both ways, no new code).
- **#327 closes as won't-fix** with the reason on the issue (W5): the count check is DuckDB's binder's and runs first; hiding the column was specified and rejected because it does not even give the issue what it asks for.

## What was measured, not assumed

- The widened discovery query returns every unique index and is **cheaper** than the primary-key-only one on the test database (151 → 38 µs; two joins gone) and within 6 µs of a join-less primary-key-only form on a 3000-table, 9200-index catalog (49 vs 43 µs; 58 as shipped). Tens of microseconds per table, once per cache load.
- Why `OFF` is explicit: with it deliberately leaked, the pool's session reset clears it in autocommit with the default setting — and **only** there. Inside a transaction the next plain INSERT fails with 545; with the reset off, likewise; a second table on the session is 8107.
- Both sides of every guard: the rowid test's `DATETIME`-plus-unique case changes the row that main's build silently skipped; the identity leak test fails with the `OFF` removed; the unit tests pin the choice order and every rejection reason, and the bracket text with its three explained errors.

## Tests

`test/sql/rowid/rowid_unique_index.test` (nine table shapes), `test/sql/insert/insert_identity.test` (66 assertions: the leak test with pool limit 1 set before `ATTACH`, reset off and `@@SPID` compared; a failing batch; ROLLBACK on the pinned connection; `DEFAULT`/`NULL`/mixed; above the BCP threshold; `DECIMAL(38,0)`, `TINYINT`, `IDENTITY(0,-1)`; a view), `test/sql/copy/copy_identity.test`, two header-only C++ unit tests (`make test-rowid-key-choice`, `make test-identity-insert`). Two existing tests asserted the old "explicit identity value is refused" contract and are updated; `no_pk_update_delete.test` asserts the widened message. Full SQL suite 209 green locally.

`PrimaryKeyInfo` is `RowIdKeyInfo` in its own mechanical commit; `pk_info` fields and `GetPrimaryKeyInfo` keep their names on purpose.


oluies:
Design review of `91e802c` (roborev job 1635, `claude-code`, `--type design`). **Verdict: fail** — the approach is sound and the spec's claims check out against the code; what fails is a set of decisions left ambiguous enough that two implementers would build different things.

I spot-checked the three cheapest-to-verify findings against the tree myself; those are marked **verified** below. The rest are the reviewer's, unverified by me.

## High

**1. W5 contradicts itself on whether #327 closes.** *(verified)*
- `spec.md:292` — "Nothing in this spec changes that, and the issue **stays open**."
- `spec.md:315` — "**The issue does not stay open.**"
- The section header and W5b both say it closes.

Whoever ships this could reasonably do either. Suggested fix: change line 292 to "…and #327 is closed as won't-fix (below)."

**2. W2 never says where `IDENTITY_INSERT OFF` is sent when a statement fails.**

W2 requires `OFF` "on every failure path before the connection is released". But the only cleanup hook that exists is `~MSSQLStatementConnection() noexcept` → `ReleaseWithoutContext()` → `transaction_.Rollback()` → `ReleaseBcpConnectionOnError(...)`, which by design runs with no `ClientContext`, possibly on a worker thread (the #178 contract).

So an implementer must choose between sending a network batch out of a `noexcept` destructor, leaking `ON` onto a pooled connection, or inventing a third thing — and each choice has a different failure mode. Suggested: record "identity-insert is on for T" on `MSSQLStatementConnection`; send `OFF` inside `ReleaseWithoutContext` after the rollback, catching everything; if `OFF` fails or the connection is not Idle, **close** it rather than pool it (`ReleaseBcpConnectionOnError` already closes non-Idle connections — make that part of the stated guarantee). Say separately what happens for a connection pinned to a transaction, which cannot be closed.

## Medium

**3. W5b's claim that error 8107 can only mean the extension leaked `ON` is false.** A user can run `SET IDENTITY_INSERT dbo.Other ON` through `mssql_exec` themselves — on the pinned connection inside a transaction, or on a pooled connection with `mssql_reset_connection = false`. The message would then blame the extension for the user's own action. Related: W2's unconditional `OFF` silently undoes session state the user set deliberately. Both belong in the wording.

**4. The W1 tie-break ranks identity below byte width.** Identity is step 4, after "fewest key columns" and "narrowest in bytes", so a `BIGINT IDENTITY UNIQUE` never beats an `INT UNIQUE`. But the spec's own reason for preferring identity is that its values do not change under UPDATE — a correctness property, ranked beneath a size heuristic. Either move it above width or say why width wins. Also worth stating: does "fewest key columns" mean `is_included_column = 0` only, and are unique indexes on indexed views candidates?

**5. The "usable index" test ignores key columns that do not read back exactly.** A unique index can sit on `sql_variant`, `hierarchyid` and other `is_cast_required` types, which the read path turns into NVARCHAR — and § 8 of this very spec shows that conversion is lossy for `sql_variant`. A `rowid` built from such a key produces a VALUES JOIN that can miss rows, so UPDATE/DELETE silently affects zero rows on tables that previously gave a clean error. Suggest adding "no key column is `is_cast_required`" to the usable criteria (and deciding about `float` and `datetime`), listing the rejection in W5b, and noting that the primary-key path already carries this exposure.

**6. W2 does not define explicit `NULL` or `DEFAULT` for the identity column** — `INSERT INTO t VALUES (DEFAULT, 1, 'x')`, or `INSERT … SELECT` where the source column is NULL. Whether DuckDB keeps the column in `insert_col_indices` for `DEFAULT` is not stated; if it does, the INSERT runs with `ON` and a NULL and the server rejects it, where omitting the column would have worked.

**7. W2 does not say what happens when the cached `is_identity` is stale.** § 5.1 accepts a stale cache for `rowid`, and W2 leans on the same flag. After DDL via `mssql_exec`, the `ON` batch raises error 8106 ("table does not have the identity property"), which is not in W5b's table. Worth mapping to a message that points at `mssql_invalidate_cache()`.

## Low

**8.** The view rule depends on a `GetObjectType() == VIEW` check the spec never names — the routing code already has it at `mssql_catalog.cpp:733`. Without it, `SET IDENTITY_INSERT <view> ON` produces an error other than the "honest" one W2 promises.

**9.** Clarity: an untranslated Russian phrase at `spec.md:224` (*verified*); the header says "Spec and implementation in one PR" while the commit says spec only; W5b says "Four places" but row 3 is caught nowhere, so it is three plus a documented gap.

## Tests

- **W6 omits an existing test that W5b breaks.** *(verified)* `test/sql/dml/no_pk_update_delete.test` asserts the text `requires a table with a primary key` at lines 43 and 50. W5b rewrites that message, so either the suite fails or the message is frozen to keep it green. List the update in W6/W7 and keep a stable substring.
- **The leak test can pass without testing anything.** It runs "a second statement on the same connection", but the pool does not guarantee the same connection. Force it — `SET mssql_connection_limit = 1`, or run inside a transaction — then assert that an explicit identity value via `mssql_exec` on that session fails with 544.
- **The view-with-identity test states no expected outcome**; W2 says the server's own error surfaces, but W6 does not say which.
- **The 1088 permission path has no test.** A live test with a second login may be impractical, but the 1088 → rewritten-message mapping could be a pure function with a C++ unit test — which would test which server errors are recognised, not source text.

## Missing considerations

- **Interrupt / ATTENTION mid-statement**: the connection is not Idle and gets closed. State that closing counts as clearing `IDENTITY_INSERT`.
- **Error 1088 inside a statement-level transaction**: `transaction_.Begin` has already run — confirm the rewritten error still rolls back and releases cleanly.
- **DATAMODEL.md**: W7 mentions it, but the new session-state invariant ("a statement connection that turned `IDENTITY_INSERT` on must turn it off or discard the connection") belongs in the connection-pool lifetime section, not only the key-discovery diagram.
- **Security**: W2 turns `IDENTITY_INSERT` on unasked. It only works where the caller already holds ALTER, so it grants nothing new — but the spec should say that out loud.

---

The three that actually block implementation: the W5 contradiction, where `OFF` is sent on failure, and the 8107 claim. The rest can be resolved in review.

VGSML:
Verdict accepted, and the fixes are in `4f66a61`. I checked everything checkable against the tree and the server before agreeing; nothing fell over, and one finding turned out worse than reported.

## The two that blocked implementation

**The #327 contradiction** — my editing. W5 was rewritten when the owner ruled out a DuckDB patch, and the earlier sentence was left standing. It closes as won't-fix and every mention says so now.

**Where `OFF` goes on failure.** You were right that "on every failure path" is not an instruction, and right about the shape of the hook. Now named: a flag on `MSSQLStatementConnection`; `OFF` sent inside `ReleaseWithoutContext` after `transaction_.Rollback()` with everything caught; the connection **closed rather than pooled** if that throws or it is not Idle, which makes `ReleaseBcpConnectionOnError`'s existing behaviour part of the guarantee; and the pinned case called out separately, because a pinned connection cannot be closed — there `OFF` is attempted and a failure logged, since destroying a transaction the user still owns is worse. Interrupt/ATTENTION folds into the same close path, and closing the session clears the setting by construction.

## Your best finding, and it is worse than you wrote

Medium 5. Confirmed that a unique index **and a primary key** can both sit on `sql_variant` and `hierarchyid` — I created both. Then I reproduced the consequence with no spec 077 code involved:

| what the server holds | what DuckDB reads |
| --- | --- |
| `2026-09-16T10:11:12.1234567` | `Sep 16 2026 10:11AM` |
| `2026-09-16T10:11:59.7654321` | `Sep 16 2026 10:11AM` |

Two distinct `SQL_VARIANT` primary keys, one string. An `UPDATE … WHERE payload = 'a'` through the extension then **reported no error and changed nothing** — the VALUES JOIN compares a string against a sql_variant column and matches no row. A silent zero-row UPDATE, in today's build.

So your "the primary-key path already carries this exposure" was not an aside, it is a live data-integrity bug. It is now on #354 with the reproduction, W1 gains `no key column is is_cast_required` so this spec fences the hole at primary keys instead of widening it to every unique index, and § 4 says plainly that the criterion is a fence around something already broken.

`float` and `datetime` are deliberately **not** excluded, and the spec now says why: both round-trip exactly through the literal renderer. The criterion is about a lossy READ, not about types that are awkward to compare.

## The rest, all taken

- **Tie-break:** identity moves **above** byte width. You were right that a correctness property should not rank under a size heuristic — a `BIGINT IDENTITY UNIQUE` should beat an `INT UNIQUE` the application maintains by hand. Also stated: key columns only (`is_included_column = 0`), and indexed-view indexes are not candidates.
- **8107:** the message no longer accuses us. A user can set `IDENTITY_INSERT` themselves through `mssql_exec`, so it names both possibilities and blames neither — and `OFF` is now sent only if this statement turned it on, so we stop clearing state someone set deliberately.
- **8106** maps to a message naming `mssql_invalidate_cache()`.
- **Explicit `NULL` / `DEFAULT`** for the identity column is defined: `DEFAULT` means "server decides" so the column is dropped and no `ON` is sent; an explicit `NULL` is a value, is sent, and the server's refusal stands unrewritten. Which one DuckDB actually produces is named as the first thing to check when the code is written.
- **The VIEW rule** names the check at `mssql_catalog.cpp:733`.
- **1088 inside the statement transaction:** the rewritten error is thrown, not swallowed, so the rollback-and-release path stays intact. Stated.
- **DATAMODEL** gains the pool-lifetime invariant, not only the key diagram.
- **Security** is stated out loud: it succeeds only where the caller already holds ALTER, so it grants nothing they could not do through `mssql_exec`.
- **Russian phrase** gone. **"Four places"** is now five, four catchable and one not, which is the honest count.

## Tests

- `test/sql/dml/no_pk_update_delete.test` — good catch, W6 and W7 now say it is updated in the same commit as the message, with a stable substring to match.
- **The leak test** is forced onto one connection, `mssql_connection_limit = 1` plus `mssql_reset_connection = false`, asserting 544 from a later `mssql_exec`. You were right that it proved nothing as written.
- **The view test** states its expected outcome, asserted by the server's error number rather than any wording of ours.
- **The 1088/8106/8107 mapping** gets a C++ unit test, for exactly your reason: it pins which server errors are recognised, not source text.

One thing I did not change: the header's "spec and implementation in one PR". That is the repo's rule rather than a contradiction with the commit — the PR is a draft precisely because the code lands beside the spec. I made the header say that outright.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+| `2026-09-16T10:11:12.1234567` | `Sep 16 2026 10:11AM` |
+| `2026-09-16T10:11:59.7654321` | `Sep 16 2026 10:11AM` |
+
+`UPDATE … SET payload = 'CHANGED' WHERE payload = 'a'` through the extension
+then **reported no error and changed nothing**: the rowid VALUES JOIN compares
+a string against a sql_variant column and matches no row. A silent zero-row
+UPDATE, today. Filed as a consequence on the sql_variant issue; this criterion
+is what stops spec 077 widening it from primary keys to every unique index.
+
+`float` and `datetime` are deliberately NOT excluded: both round-trip exactly
```
oluies: **High — `datetime` keys do not match exactly, and this is the one claim W1 rests on.**

The text says a `datetime` key round-trips exactly because the value goes back through a `datetime2` cast "that only widens". SQL Server does not compare these types that way:

1. A `datetime` is stored in 1/300-second ticks, so `.003` is really `.00333…`.
2. The read path decodes it to DuckDB's microsecond `TIMESTAMP` (`src/codec/datetime_codec.cpp:744`), which keeps `.003333`.
3. The literal renderer then sends `CAST('….003333' AS DATETIME2(7))`.
4. Since compatibility level 130 (SQL Server 2016), comparing `datetime` with `datetime2` converts the `datetime` side **exactly**, to `.0033333` — which is not equal to `.0033330`.

So a unique key on a `datetime` column whose milliseconds are not a multiple of 10 can **never** match. The rowid VALUES JOIN finds no row and the UPDATE or DELETE silently changes nothing — exactly the failure `is_cast_required` exists to prevent, except W1 would carry it from primary keys to every unique index.

**Fix:** either exclude `datetime` (and `smalldatetime`) from the usable key types, or render the rowid literal for a `datetime` column as `CAST(… AS DATETIME)` so both sides are the same type. Add a W6 test with a key value such as `10:00:00.003`, and correct the claim in this paragraph — it currently reads as a positive assurance.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+  and names `mssql_invalidate_cache()`.
+- **`IDENTITY_INSERT` is turned off only if this statement turned it on.** A
+  user can set it themselves through `mssql_exec` — on a pinned connection, or
+  on a pooled one with `mssql_reset_connection = false` — and clearing it
+  unconditionally would undo session state they set deliberately.
+- The target of `SET IDENTITY_INSERT` is a **table**. The VIEW case is decided
+  by the `GetObjectType() == MSSQLObjectType::VIEW` test the routing already
+  makes at `mssql_catalog.cpp:733`; a view target is left alone, and the server
+  refuses an explicit identity value through a view on its own terms.
+- **An explicit `NULL` or `DEFAULT` for the identity column.** `DEFAULT` is the
```
oluies: **Medium — undefined for a multi-row INSERT.**

Dropping the identity column from the generated list when DuckDB passes `DEFAULT` only works when there is one row. Consider:

```sql
INSERT INTO t VALUES (DEFAULT, 1), (42, 2);
```

- one statement mixes "let the server assign" and an explicit value in the same column;
- the column list is shared by every row in the batch, so it cannot be dropped for row 1 and kept for row 2;
- with `IDENTITY_INSERT ON` there is no way to spell `DEFAULT` for an identity column, so row 1 either fails or comes out wrong.

The spec defines neither the per-row decision nor the resulting error.

**Fix:** choose a rule and write it down — refuse a mixed `DEFAULT`/explicit identity column with a message naming the way out, or split the statement into separate batches. Add a W6 case either way.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+  statement on that connection silently accept identity values, and a second
+  target would fail with 8107.
+- Inside an explicit transaction the connection is pinned and shared, so the
+  same discipline applies to the pinned connection, and the `OFF` must not wait
+  for COMMIT.
+- The `ON` and `OFF` go as **batches of their own**, never wrapped in `EXEC`:
+  § 0.3 measured that a SET inside a nested batch is restored when that batch
+  ends, so a wrapped one would silently do nothing and the INSERT would fail
+  with 544.
+- **Where `OFF` is sent when the statement fails**, named precisely, because
```
oluies, line 372: **Medium — this `OFF` batch has no timeout.**

`OFF` is sent over the network from inside `~MSSQLStatementConnection() noexcept` → `ReleaseWithoutContext()`. By the issue #178 contract that path has no `ClientContext`, so no session-level query timeout reaches it. If the server or the network is stalled, a destructor on a worker thread blocks with no bound — and so does query teardown behind it.

"Everything caught" covers exceptions, not hangs.

**Fix:** require a bounded timeout for the `OFF` batch, taken from a value captured at `Acquire` — the same way `reset_on_release` is carried down to `ReleaseBcpConnectionOnError`. On expiry, treat it as a failure and close the connection rather than waiting.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+
+**Usable** means all of:
+
+| Condition | Read from | Why |
+|---|---|---|
+| `is_unique = 1` | `sys.indexes` | otherwise values repeat |
+| `has_filter = 0` | `sys.indexes` | a filtered unique index covers a subset of rows; the rest are unaddressable |
+| every key column `is_nullable = 0` | `sys.index_columns` joined to `sys.columns`, `is_included_column = 0` | measured: a unique index on a nullable column accepts exactly one NULL row and refuses the second, so NULL is neither unique nor addressable |
+| `is_disabled = 0`, `is_hypothetical = 0` | `sys.indexes` | a disabled index enforces nothing |
+| no key column is `is_cast_required` | `MSSQLColumnInfo` | **the review's best finding, and it is a live bug on the PK path already — see below** |
```
oluies, line 191: **Medium — two gaps in how these criteria are applied.**

**Where does the `is_cast_required` filter run?** Every other criterion in this table is evaluable server-side in `PK_DISCOVERY_SQL_TEMPLATE`, but `is_cast_required` exists only client-side, in `MSSQLColumnInfo`. The spec doesn't say which side does the filtering, and the two answers have different consequences:

- filtered inside the SQL → W5b cannot report a rejected index *with its reason*, because the rejected index never comes back;
- filtered client-side → the query must return **every** candidate index with its key columns, not just the one it picked.

**Step 1 takes the primary key unconditionally.** So a PK on `sql_variant` still beats a perfectly usable unique index on the same table. §4 (Risks) admits the primary-key path is already broken this way; step 1 as written keeps it broken even where a safe alternative is sitting right there.

**Fix:** state that the query returns all candidates and that the choice runs in the pure C++ function W6 already plans to unit-test; and decide explicitly whether a cast-required PK falls through to step 2.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+UPDATE, today. Filed as a consequence on the sql_variant issue; this criterion
+is what stops spec 077 widening it from primary keys to every unique index.
+
+`float` and `datetime` are deliberately NOT excluded: both round-trip exactly
+through the literal renderer (`float` as its shortest round-trip decimal,
+`datetime` through a `datetime2` cast that only widens), so their keys match.
+The criterion is about columns whose READ is lossy, not about types that are
+awkward to compare.
+
+**Choice, in order** (deterministic, so two sessions never disagree):
```
oluies, line 286: **Low — step 2 can yield more than one index.**

SQL Server permits duplicate indexes, so a table can carry several usable unique indexes on its single identity column. The rule is still deterministic — `index_id` breaks the tie — but the text reads as if step 2 always produces exactly one index and steps 3–5 only run when it produces none.

**Fix:** say that steps 3–5 order the candidates *within* each step, not only after step 2 has come up empty.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,629 @@
...
+
+### W6 — tests
+
+New file `test/sql/insert/insert_identity.test`:
+
+- explicit list without the identity column — server assigns, values ascend;
+- explicit list with it — lands verbatim, and a later statement is not silently
+  in `IDENTITY_INSERT` mode. **The leak test must force the same connection**,
+  or it proves nothing: the pool does not promise to hand back the one just
+  released. Pin it with `SET mssql_connection_limit = 1`, run it with
```
oluies, line 568: **Low — this test can pass without ever pinning a connection.**

Two things the wording leaves open:

- the pool's limit is fixed when the catalog's pool is created, so `SET mssql_connection_limit = 1` has to be applied **before** `ATTACH`. If it isn't, the pool keeps its old limit and the test passes without pinning anything;
- with a limit of 1, the `mssql_exec` probe also depends on the statement connection having been released first. A leaked, un-released connection makes the probe fail on an **acquire timeout**, not with error 544 — so the test fails for a reason that doesn't name the leak.

**Fix:** set the limit before `ATTACH` (or use a dedicated alias), and have the test additionally confirm it is the same session — comparing `@@SPID` before and after is the cheap version.

oluies (dismissed):
Approving the spec — it resolves the earlier review's blocking points, and the `is_cast_required` observation is a good catch that stands on its own. **One High inline that I'd treat as blocking for W1**: the claim that `datetime` keys round-trip exactly is not how SQL Server compares `datetime` with `datetime2`, and acting on it would spread silent zero-row UPDATE/DELETE from primary keys to every unique index. Three Mediums are unspecified mechanics rather than errors.

**Summary of findings**

| Severity | Where |
| --- | --- |
| **High** | W1 — `datetime` keys do **not** match exactly; widening rowid to them causes silent zero-row DML |
| Medium | W2 — `DEFAULT` mixed with explicit values in a multi-row INSERT is undefined |
| Medium | W2 — the destructor-time `IDENTITY_INSERT OFF` batch has no timeout |
| Medium | W1 — where the `is_cast_required` filter runs is unstated; step 1 takes an unusable PK |
| Low | W6 — the leak test can pass without pinning a connection |
| Low | W1 — step 2 reads as if it always yields exactly one index |


---
*Posted from a [roborev](https://github.com/) review (job 1642), relayed via Claude Code.*

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,757 @@
...
+naive float formatting — was created and all three rows were found and updated
+through the extension. The renderer emits the shortest round-trip decimal and
+both sides are the same type, so they compare equal.
+
+So the criterion is no longer "columns whose READ is lossy". It is **columns
+whose rowid literal cannot be made to match the column they came from**, which
+covers the lossy reads (`sql_variant`, `hierarchyid`) and the mixed-type
+comparison (`datetime`) alike.
+
+**Where the filtering runs.** Every criterion above except the last two is
```
oluies, line 274: **Medium — this paragraph and W6 can't both be satisfied.**

This says the SQL still drops rows with `has_filter`, and gives the reason: a candidate filtered out inside the SQL *"cannot be reported with its reason"*. But:

- **W5b** (line ~508) requires the refusal to name each rejected index with its reason, and gives *"filtered"* as one of the examples;
- **W6** (line ~565) asserts *"a **filtered** unique index — refused, and the message says why"*.

A filtered unique index removed by the query never reaches the C++ chooser, so that message cannot be produced and the W6 case cannot pass as written. The same applies to disabled and hypothetical indexes if W5b is meant to cover them (line 190 filters those in SQL too).

**Fix — pick one:**

1. Move `has_filter` (and optionally `is_disabled` / `is_hypothetical`) out of the SQL and into the C++ criteria, so those candidates come back and can be named; or
2. Drop "filtered" from the W5b reasons and from the W6 case, and state that such indexes are silently ignored.

(1) is the better user experience — "there is a unique index but it's filtered" is exactly the thing someone needs told — but (2) is honest and cheaper.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,757 @@
...
+**Usable** means all of:
+
+| Condition | Read from | Why |
+|---|---|---|
+| `is_unique = 1` | `sys.indexes` | otherwise values repeat |
+| `has_filter = 0` | `sys.indexes` | a filtered unique index covers a subset of rows; the rest are unaddressable |
+| every key column `is_nullable = 0` | `sys.index_columns` joined to `sys.columns`, `is_included_column = 0` | measured: a unique index on a nullable column accepts exactly one NULL row and refuses the second, so NULL is neither unique nor addressable |
+| `is_disabled = 0`, `is_hypothetical = 0` | `sys.indexes` | a disabled index enforces nothing |
+| no key column is `is_cast_required` | `MSSQLColumnInfo` | **the review's best finding, and it is a live bug on the PK path already — see below** |
+| no key column is `datetime` or `smalldatetime` | `MSSQLColumnInfo::sql_type_name` | the rowid literal cannot be made to match such a column at all — measured, see below. Lifted when [#358](https://github.com/hugr-lab/mssql-extension/issues/358) fixes the renderer |
```
oluies: **Medium — the new criterion is right, but it's applied to three types when it describes more.**

The criterion is now *"columns whose rowid literal cannot be made to match the column"*, which is the correct generalisation. But it's only applied to `sql_variant`, `hierarchyid` and `datetime`, and at least two more types are in the same class for the same reason:

- **`DATETIMEOFFSET`** always decodes to DuckDB's microsecond `TIMESTAMP_TZ` (`src/codec/datetime_codec.cpp:786-790`). A `datetimeoffset(7)` key whose 100 ns digit is non-zero is truncated on read, the literal sent back carries the truncated value, and it matches no row — the same silent zero-row UPDATE this change exists to close.
- **`time(7)`** almost certainly the same, since DuckDB's `TIME` is also microsecond.

The commit argues, rightly, for measuring rather than assuming — these two were neither measured nor listed.

**Fix:** measure `datetimeoffset(7)` and `time(7)` keys whose last digit is non-zero (and `datetime2(7)`, if the catalog can read it at microsecond precision). Then either add them to the exclusion or record why they're safe, and add the matching W6 cases. If they do belong, #358 should cover them too.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,757 @@
...
+  refuses an explicit identity value through a view on its own terms.
+- **An explicit `NULL` or `DEFAULT` for the identity column.** `DEFAULT` is the
+  form that means "let the server decide", so if DuckDB leaves the column in
+  `insert_col_indices` for it, the column is dropped from the generated list
+  and no `IDENTITY_INSERT` is sent — the same outcome as not naming it. An
+  explicit `NULL` is a value, so it is sent, and the server refuses it; that
+  refusal is the honest one and is not rewritten. Which of the two DuckDB
+  actually produces is the first thing to check when the code is written, and
+  the test pins whichever it is.
+- **`DEFAULT` and an explicit value for the identity column in the same
```
oluies, line 446: **Medium — the rule is stated, but not how the case is detected.**

This says to refuse `VALUES (DEFAULT, 1), (42, 2)` by name. By the time the sink sees the rows, though, DuckDB's binder has already replaced `DEFAULT` with the column's default expression — for an MSSQL identity column, probably a NULL constant. The earlier W2 text already admits it isn't known which of the two outcomes DuckDB produces.

If a per-row `DEFAULT` arrives as NULL, it is indistinguishable from an explicit NULL, and the rule as written cannot be enforced — the mixed batch would instead fail on the server, with NULL into a NOT NULL identity column under `IDENTITY_INSERT ON`.

**Fix:** make this rule conditional on the same "check what DuckDB actually produces" step W2 already schedules, and state the fallback explicitly: if `DEFAULT` turns out to be indistinguishable from NULL, the W6 case pins the server's NOT NULL error instead of our own message.

specs/077-identity-columns/spec.md
```diff
@@ -0,0 +1,757 @@
...
+`PK_DISCOVERY_SQL_TEMPLATE` (`src/catalog/mssql_primary_key.cpp`) filters
+`kc.type = 'PK'` and nothing else. Widen it to consider every index on the
+object and pick one deterministically.
+
+**Usable** means all of:
+
+| Condition | Read from | Why |
+|---|---|---|
+| `is_unique = 1` | `sys.indexes` | otherwise values repeat |
+| `has_filter = 0` | `sys.indexes` | a filtered unique index covers a subset of rows; the rest are unaddressable |
```
oluies, line 188: **Low — `smalldatetime` is excluded as "measured", but nothing shown measures it.**

Line 192 excludes `datetime` *or* `smalldatetime` and calls the exclusion measured — every measurement in the section below uses `datetime`. And `smalldatetime` has minute resolution with no fractional seconds, so it very likely converts to `datetime2` exactly and matches fine.

It's also inconsistent downstream: the W5b row (~508) and the behaviour-change lists in §3 and §4 name only `sql_variant`, `hierarchyid` and `datetime`. As written, `smalldatetime` primary keys change behaviour — falling through or being refused — with no measurement behind it and no release note.

**Fix:** either measure a `smalldatetime` key and drop it from the exclusion if it matches, or keep it and add it to the W5b, §3 and §4 lists and to W6.

oluies (commented):
The `datetime` High is properly closed — measuring it rather than reasoning about it was the right move, and #358 for the renderer is a good way to keep the exclusion temporary. The `IDENTITY_INSERT OFF` bound, the PK falling through the same criteria as a unique index, and the tightened leak test all look right too.

Holding the approval on three Mediums, all of the same shape: the new criterion and the new filtering split are each correct where they're applied, but the spec now disagrees with itself about where they apply.

| Severity | Where |
| --- | --- |
| Medium | "Where the filtering runs" vs W5b and W6 — a filtered index removed in SQL can't be reported, but W6 asserts it is |
| Medium | W1 line 192 — the new criterion misses `datetimeoffset(7)` and `time(7)`, same class, same cause |
| Medium | W2 — the mixed `DEFAULT`/explicit rule doesn't say how the case is detected |
| Low | `smalldatetime` excluded as "measured", but every measurement shown uses `datetime` |

Happy to approve once these are settled — most of them may just be wording.


---
*roborev job 1646 (re-review of `893f082` only), relayed via Claude Code.*

oluies:
## roborev review — 6 findings (3 Medium, 3 Low)

Branch diff `556f15f..bdbdcf4` (10 commits) against `main`, roborev job 1654, agent `claude-code`.

### Medium

**1. `src/include/catalog/mssql_rowid_key_choice.hpp` — `IsLiteralMismatchType` / `Unusable` misses `time(7)` and `datetimeoffset(7)`**

`MapSQLServerTypeToDuckDB` (`src/catalog/mssql_column_info.cpp:249-279`) reads `datetimeoffset` as microsecond `TIMESTAMP_TZ` and `time` as microsecond `TIME` whatever the declared scale — the comment there already admits the trailing digit is dropped at scale 7. A key of either type whose 100 ns digit is non-zero therefore produces a rowid that matches no row: the same silent zero-row UPDATE/DELETE this work exists to fence off. This range also widens the exposure, because any unique index on such a column is now a rowid source where before only a primary key was. Raised in the earlier review and still open.

*Fix:* treat `time` and `datetimeoffset` with `scale == 7` as unusable (`scale` is already on `RowIdKeyColumn`); add a `test_rowid_key_choice.cpp` case and a live test with a non-zero seventh digit.

**2. `mssql_rowid_key_choice.hpp:100` — `smalldatetime` is excluded without a measurement**

Every measurement in the spec is for `datetime`. `smalldatetime` has minute resolution and its read-back value converts to `datetime2` exactly, so its rowid literals very likely do match — which would make this a regression: tables keyed by `smalldatetime` that update correctly today would start refusing UPDATE/DELETE. It is also undocumented; CHANGELOG, CLAUDE.md and `dml.md` name only `DATETIME`/`SQL_VARIANT`.

*Fix:* measure a `smalldatetime` key end to end. If it matches, drop it from the exclusion and the unit test; if it does not, name it in the user-facing docs and the CHANGELOG.

**3. `Makefile` + `.github/workflows/ci.yml` — the two new unit tests never run in CI**

`test-rowid-key-choice` and `test-identity-insert` got Makefile targets but no CI steps. This family of tests runs only as explicit steps (`test_index_kind` at `ci.yml:225`, `test_load_policy` at `ci.yml:304`), so the tie-break order and the 1088/8106/8107 error mapping — the two things the spec says must not drift — are unguarded on PRs.

*Fix:* add matching compile-and-run steps next to `test_index_kind`.

### Low

**4. `src/catalog/mssql_table_entry.cpp`, `EnsurePKLoaded` catch / no-connection branch — a transient failure is cached for good**

Pool exhaustion or a server error sets `discovery_error` with `exists = false`, and `pk_loaded_` is then published as true. Every later UPDATE/DELETE on that table keeps reporting "indexes could not be read" until the cache is invalidated, even after the server recovers. Caching the failure predates this change, but the new message now tells the user it was a lookup failure while the code never retries the lookup.

*Fix:* do not publish `pk_loaded_` when `discovery_error` is set, so the next bind retries — or have the message name `mssql_invalidate_cache()` as the way to retry.

**5. `src/include/catalog/mssql_primary_key.hpp` — stale comment on `RowIdKeyInfo`**

"The struct keeps its name for now; the rename to RowIdKeyInfo is its own mechanical commit" — that rename landed in this range (4e908d2). Remove the sentence.

**6. `test/sql/insert/insert_bcp_paths.test` (`3000	3500`) — brittle expectation**

The expected `max(rid)` depends on the injected-desync hook firing only after the first 500-row statement reached the server, on the statement batch size derived from `mssql_dml_max_parameters`, and on identity-seed behaviour under rollback. A change to batch sizing or to where the desync fires will move this number for reasons unrelated to what the test covers.

*Fix:* assert `count(*) = 3000` and `max(rid) >= 3000`, or reseed/recreate the table before the bulk INSERT so the value is deterministic.

VGSML:
All six taken, each measured on the docker server before acting; one commit on top of a merge with `main` (pin bump #365, specs #364).

**Medium 1 — `time(7)` / `datetimeoffset(7)`.** Confirmed through the extension, not by reasoning: three rows keyed on `.0000001` / `.1234567` / `.123456`, `UPDATE … WHERE n = 0` updated **one** (the row whose seventh digit is zero) and reported nothing. `time(6)`, `datetimeoffset(6)` and `datetime2(7)` (read as `TIMESTAMP_NS`) updated every row. `IsLiteralMismatch(type, scale)` now refuses `time` / `datetimeoffset` at scale 7 by name (`… has type time(7), which no rowid literal can match (see issue #358)`), with unit cases for all six shapes and a live test, `rowid_key_types.test`, that pins the two refusals and the four round trips.

**Medium 2 — `smalldatetime`.** You were right that it was excluded on `datetime`'s evidence alone. Measured: a `smalldatetime` key matches `CAST('2026-01-01 10:00:00.0000000' AS DATETIME2(7))` (and the (6) and (0) forms) — it has no fraction to lose — while the `datetime` control does not. It is usable again; the spec's W1 criterion reads `datetime`, or `time` / `datetimeoffset` at scale 7, and a new paragraph records the family's measurement. CHANGELOG names all three excluded types and says `smalldatetime` / `datetime2(7)` stay usable.

**Medium 3 — CI.** Both header-only tests run on every PR now, as steps beside the load-policy one.

**Low 4.** A discovery failure is no longer cached: `pk_loaded_` is not published when `discovery_error` is set, so the statement is refused with the error it met and the next bind retries. **Low 5.** Sentence removed. **Low 6.** `count(*) = 3000` and `max(rid) >= 3000`, with the comment saying what the exact figure depended on.

#358 will get a note that `time(7)` / `datetimeoffset(7)` belong to it.

oluies:
## roborev review of 3702c20 — 3 findings (1 High, 1 Medium, 1 Low)

Follow-up commit only (`3702c20`), roborev job 1656, agent `claude-code`. The fixes for the `time(7)`/`datetimeoffset(7)` refusal, `smalldatetime`, and the CI steps look right. The problem is in the fix for the earlier Low about discovery failures being cached: not caching them introduces a race.

### High — `src/catalog/mssql_table_entry.cpp:358-364` (`EnsurePKLoaded`), and the lock-free readers of `pk_info_` (`:151`, `:260`, `:377`, `:395`, `:401`, `:460`)

A failed discovery now returns with `pk_loaded_` still false. Before, once `EnsurePKLoaded` returned, nothing wrote `pk_info_` again, and that is what made the unlocked reads after it safe. That guarantee is gone:

- Thread A hits a discovery error, leaves the lock, and reads `pk_info_` without it — `exists`, the `discovery_error` inside `RowIdRefusal`, `columns` in `GetScanFunction`.
- Thread B takes `pk_load_mutex_`, sees `pk_loaded_ == false`, and runs `pk_info_ = Discover(...)`.
- That move-assign frees the strings and vectors thread A is reading.

This is the use-after-free / double-free spec 052 added the mutex to prevent. `GetRowIdColumns` is the worst case, because it reads `pk_info_` precisely when `!pk_loaded_`. (I checked the code at `3702c20`; the `return` before the release-store on the `discovery_error` path is as described.)

*Fix:* keep lock-free readers away from a `pk_info_` that can still be overwritten. Either build the result in a local and return or throw the failure by value instead of storing it in `pk_info_`, or keep failures in a separate lock-protected field and have callers copy what they need under `pk_load_mutex_` whenever `pk_loaded_` is false.

### Medium — `mssql_table_entry.cpp:151` (`GetScanFunction` → `EnsurePKLoaded`)

`EnsurePKLoaded` runs on every scan bind, not only on UPDATE/DELETE. Now that a failure is not cached, every plain `SELECT` on the table repeats the discovery round trip for as long as the failure lasts. Some failures are persistent, for example `VIEW DEFINITION` denied on `sys.indexes` or a metadata timeout. Under pool exhaustion, every bind can also block in `pool.Acquire()` for up to `mssql_acquire_timeout`. Before this change that cost was paid once.

*Fix:* retry only on the DML/rowid paths (`BindUpdateConstraints`, `GetRowIdType`), or cache the failure with a short backoff/TTL. When discovery has failed, a scan that did not ask for rowid should skip the lookup rather than trigger another one.

### Low — `test/sql/rowid/rowid_key_types.test`

The new live test covers the refusals and the round trips, but not the new retry behaviour. `mssql_test_fail_metadata_after_rows` (or a similar hook) could fail discovery once. The test would then check that the refusal names the error, and that the next UPDATE succeeds without `mssql_invalidate_cache()`.

VGSML:
Taken — and thank you for catching what the previous fix broke.

**High / Medium.** Both right, and they share a cause: the readers of `pk_info_` are lock-free by the spec 052 contract ("nothing writes `pk_info_` once `pk_loaded_` is published"), so a failure must be published like a result. It is again. The refusal now names the way out instead: `… could not be read, so whether it has one is unknown: <error>. The lookup is not retried on its own: mssql_invalidate_cache('lim', 'dbo', 'DiscoveryFail') drops the cached answer so the next statement reads the indexes again.` — `RowIdRefusal` takes the catalog name for it. The comment at the publish site records why a retry-on-next-bind is wrong there, so it is not re-proposed.

**Low.** `rowid_discovery_failure_cached.test` — the failure is real rather than injected (the metadata hook does not reach the discovery's own query): a pool of one connection, pinned by a transaction's scan of another table, and the discovery's pool acquire timing out after `mssql_acquire_timeout = 1`. It asserts the refusal text, that the same refusal comes back after ROLLBACK with the pool free (cached), and that `mssql_invalidate_cache('lim', 'dbo', 'DiscoveryFail')` makes the next UPDATE succeed (inside a transaction — with a pool of one an autocommit UPDATE cannot run at all, its scan and its batches want a connection each; that is the rowid path's shape, not this change's).

oluies:
## roborev review of 0383262 — 4 findings (1 Medium, 3 Low)

Follow-up commit only (`0383262`), roborev job 1660, agent `claude-code`. Caching the failure again removes the High race from the `3702c20` review for every reader but one (see the third finding). The refusal naming `mssql_invalidate_cache(...)` also resolves the retry question. The main remaining issue is that a scan can now cache a refusal for DML.

### Medium — `src/catalog/mssql_table_entry.cpp:326-369` (`EnsurePKLoaded`, reached from `GetScanFunction`): a short pool shortage during a plain SELECT becomes a permanent DML refusal

`EnsurePKLoaded` still runs on every scan bind. A plain `SELECT` that hits a full pool therefore caches the failure — for example one that runs while a transaction holds the only connection, or while the pool is saturated under load. After that, every UPDATE/DELETE on the table is refused until someone runs `mssql_invalidate_cache()` by hand, even though nobody tried to modify the table while the pool was short. The same thing happened before `3702c20`. The difference is that the behaviour is now deliberate and documented, so where the failure gets cached matters more.

*Fix:* run discovery only on the rowid/DML paths (`BindUpdateConstraints`, `GetRowIdType`, `PlanUpdate`/`PlanDelete`), so a scan that doesn't need rowid neither waits in `Acquire` nor caches a failure. Alternatives: give `EnsurePKLoaded` a short `Acquire` timeout, or cache only failures that aren't pool exhaustion. Either way, build the result in a local and publish it once so the lock-free contract holds.

### Low — `mssql_table_entry.cpp:328`, `:346-349`: the reason for a creation failure is discarded

`pool.Acquire()` is called without the out-parameter, so a creation failure (login failed, expired Azure token, refused dial) turns into the generic "no connection could be acquired from the pool". Now that the text is cached, a user can't tell a timeout apart from a persistent failure, and in the persistent case invalidating the cache won't help. *Fix:* call `pool.Acquire(timeout, &why)` and put `why` into `discovery_error` when it isn't empty.

### Low — `mssql_table_entry.cpp:463` (`GetRowIdColumns`): one unlocked read is left

`if (!pk_loaded_.load(...) || !pk_info_.exists) throw BinderException(pk_info_.RowIdRefusal(...))` reads `pk_info_` exactly when `pk_loaded_` is false. At that moment another thread can be inside `EnsurePKLoaded`, holding `pk_load_mutex_` and move-assigning `pk_info_`. This is the same kind of race the commit removes everywhere else, and it is now the only remaining one. (Checked at `0383262`; the code is as described.) *Fix:* don't touch `pk_info_` in the `!pk_loaded_` branch. Either throw a fixed message, or take `pk_load_mutex_` and copy what the message needs.

### Low — `test/sql/rowid/rowid_discovery_failure_cached.test`

The test assumes two things: that the UPDATE's bind discovers the key before the pinned connection comes into play, and that `mssql_acquire_timeout = 1` governs discovery's default `Acquire()`. If either assumption breaks, the test either hangs for the default timeout or passes for the wrong reason. It also never checks that a plain SELECT still works while the failure is cached. *Fix:* after the cached refusal, add `SELECT count(*) FROM lim.dbo.DiscoveryFail`, which should succeed without a new discovery attempt, and add a one-line comment saying which timeout governs discovery's `Acquire`.

oluies:
## Adversarial review — PR head `0383262` (branch vs merge-base `634b485`, 38 files)

**Method.** Six parallel reviewers, each with one focus:
- correctness of the rowid-key choice;
- correctness of IDENTITY_INSERT;
- security and C++11/ODR;
- project conventions and consistency with the spec;
- blast radius;
- over-engineering.

Every finding and every all-clear then went to skeptic agents told to refute it. Several findings were also checked against a live SQL Server (the repo's docker image) running the debug build of this head.

**Tally:** 30 candidates, 13 left after refutation. That is **1 BLOCK**, no CONSIDER, 12 NOTE.

### BLOCK

**`UPDATE`/`DELETE` through a `varchar` key under a `SQL_` collation changes rows it was not given.** Reproduced live.
- **Where.** `src/include/catalog/mssql_rowid_key_choice.hpp` (`Unusable`), plus the ON clauses in `mssql_update_statement.cpp` and `mssql_delete_statement.cpp`.
- **Mechanism.** Every rowid value goes out as an `N'…'` literal (`string_codec.cpp`), so the VALUES column is nvarchar. `t.[k] = v.[k]` therefore compares a `varchar` key column under its collation's **Unicode** rules. Under a `SQL_` collation the unique index enforced the **non-Unicode** SQL sort order instead, where `'Straße'`/`'Strasse'` and `'Æ'`/`'AE'` are distinct keys. They compare equal once converted to nvarchar.
- **Reproduction.** `varchar(50) COLLATE SQL_Latin1_General_CP1_CI_AS NOT NULL` with `CREATE UNIQUE INDEX`, no PK. That collation is the installation default.
  - `DELETE … WHERE v = 2` (the `'Straße'` row) also deleted `'Strasse'`.
  - `UPDATE … SET v = 40 WHERE v = 4` (the `'Æ'` row) also set `'AE'` to 40.
  - Neither statement raised an error.
- **Scope.**
  - The same defect already exists on `main` for a `varchar` primary key. A `char(10)` PK reproduces it the same way.
  - This PR widens it: a table whose only key is such a unique index had no DML before and now gets it.
  - `Unusable()` carries `collation_name` but never reads it.
- **Why it survived refutation.** The skeptic could not find any CAST or COLLATE on either side of the join, and the live run confirmed the over-match. Severity 5, confidence 1.0.
- **Fix.** For `char`/`varchar` key columns under a `SQL_` collation, compare against the sent value converted back to the column's own type and collation: `t.[k] = CAST(v.[k] COLLATE <col collation> AS varchar(n))`. The COLLATE has to come before the CAST, because the conversion takes its code page from the input's collation. The value was read from that column, so the round trip is lossless. Windows and UTF-8 collations apply the same rules to varchar and nvarchar and need no change. As a side benefit, the comparison no longer puts `CONVERT_IMPLICIT` on the column, so it can seek.

### NOTE (low risk)

1. **CHANGELOG overstates #358.** "no `datetime2` literal at any precision equals a `datetime` column" is false. A value on a whole 10 ms (1/300 s ticks divisible by 3) converts exactly and matches, and the spec's own table records that the `11:00:00.000` row works. Refusing the whole key is still correct. The sentence should say "a value with a 1/300 s fraction". The same wording appears in comments in `mssql_rowid_key_choice.hpp` and two `.test` headers.
2. **CHANGELOG overclaims the OFF guarantee.** "a connection on which it could not be confirmed is discarded rather than returned to the pool" holds only for unpinned connections. A pinned one cannot be closed. It relies on the transaction-end reset, which happens only under the default `mssql_reset_connection = true`. The code comment is accurate; the release note isn't.
3. **The Commit-path error lies on a pinned connection.** `mssql_statement_connection.cpp` `Commit` says "the statement was rolled back and the connection discarded". When `transaction_pinned_` is set, `Fail` does neither: `LoadTransaction::Rollback` is a no-op there and `Close` is skipped.
4. **The 1088 explanation always blames a missing ALTER permission.** The server raises 1088 for a missing object too, e.g. after `DROP` through `mssql_exec` with the default no-invalidate setting. The server text is appended, which softens this, but the diagnosis should also name `mssql_invalidate_cache()`.
5. **The spec over-promises "turned off only if this statement turned it on".** The server has no readable IDENTITY_INSERT state, and `ON` for a table that is already `ON` succeeds silently. So a user's own `ON` for the same table, set through `mssql_exec` on a pinned session, is cleared by our `OFF`. There's no fix, but the spec should state the limit.
6. **`GetRowIdColumns` (`mssql_table_entry.cpp:463`) reads `pk_info_` when `pk_loaded_` is false.** At base, that branch threw a constant message. Now it calls `RowIdRefusal`, which reads the new string and vector fields, and those can be under move-assignment on another thread. This is **latent**: `GetScanFunction` publishes first on every DML bind. It still breaks the spec 052 contract that this PR invokes, and the unloaded refusal would also say "has neither" for a table that may have a key.
7. **`EnsurePKLoaded` calls `pool.Acquire()` without the `failure` out-param.** A creation failure (login failed, expired token) gets cached as the generic "no connection could be acquired from the pool". Invalidating then won't help, and the message gives no hint of that. The same branch existed at base; this is a missed improvement, not a regression.
8. **The "has neither" and "has no usable one" refusals tell the user to add an index but never mention `mssql_invalidate_cache()`.** Indexes can only be added through `mssql_exec`, which does not invalidate by default, so the advice alone leaves the user stuck with the same refusal. This predates the PR, but the PR rewrote these messages.
9. **Nullable-key reason.** "NULL is neither unique nor addressable" is half wrong: a unique index allows exactly one NULL, as the test comment itself says. What's true is only that `= NULL` matches nothing.
10. **Spec provenance.** The status line cites the rename as `cd015d1`, which isn't in the repository. The rename commit is `4e908d2`.
11. **`GetDmlDebugLevel` reads only `MSSQL_DML_DEBUG`.** Every sibling DML executor reads `MSSQL_DEBUG`, so a user who set the general switch doesn't see the new IDENTITY_INSERT leak warnings.
12. **The unwind path can now block for up to about 60 s instead of about 30 s.** A 30 s ROLLBACK is now followed by a 30 s OFF inside a `noexcept` destructor. This only happens when both calls confirm slowly and IDENTITY_INSERT was on, and the bound is documented.

**Correction to my earlier comment on `0383262`.** That comment rated a Medium, saying a pool shortage during a plain SELECT "becomes a permanent DML refusal". Measured against the merge base, that isn't new. At `634b485`, `EnsurePKLoaded` already published `pk_loaded_` on failure (as "no primary key") from every scan bind. Only the intermediate commit `3702c20` changed that. Relative to `main`, `0383262` is an improvement, because the refusal now names the failure and the retry. Please disregard that Medium.

### Refuted or cleared (checked, not issues)
- **The C++11/ODR concern is cleared.** Both header-only files are fully `inline` and are compiled by the real C++11 extension build, not only by the `-std=c++17` unit tests.
- **Identifier quoting is sound.** `]` is doubled on every `SET IDENTITY_INSERT` path, including the destructor. Discovery sends schema and table as `sp_executesql` parameters.
- **The identity routing holds.** An INSERT that names the identity column is forced onto the statement path whatever the row count, BY NAME, or RETURNING, so `ON` never reaches bulk. COPY is unchanged by design.
- **The NULL/DEFAULT check covers every row of every chunk.** No route to a partial commit was found; later chunks roll back with the statement.
- **Discovery SQL is correct.** It produces no per-partition duplicates, excludes included and partition columns, keeps `key_ordinal` order, and keeps the one-round-trip batch from spec 076.
- **The datetime / time(7) / datetimeoffset(7) refusal is documented as deliberate.** That includes whole-second values, which are now refused. The CHANGELOG and `dml.md` say so.
- **A rowversion key is not a problem in practice.** An int key beats it on width, and without a clustered rowversion index there's no Halloween re-read.
- **Stale key after index DDL is generic stale-metadata behaviour.** It predates the PR and is documented.

### Blast radius / if it goes sideways
- **What it touches:** every UPDATE, DELETE and `SELECT rowid` (key choice and refusals); every table metadata load (discovery SQL); every INSERT that names an identity column (now needs ALTER); and `MSSQLStatementConnection` teardown.
- **What to watch:**
  - "rowid cannot address a row" from tables with datetime-family keys that used to work;
  - 545/8107 errors in sessions with `mssql_reset_connection = false`;
  - 1088 "needs ALTER" reports;
  - slow cancellation of identity INSERTs against an unresponsive server.
- **Rollback:** revert the PR. It changes no persisted format and no settings.

### Over-engineering (surviving cuts)
- `src/catalog/mssql_primary_key.cpp:L270`: yagni: `ChoiceWith` builds a throwaway `RowIdKeyChoice` only because `DescribeRejections` takes one and reads nothing but `.rejections`. Pass the vector instead and delete `ChoiceWith`.
- `src/include/catalog/mssql_rowid_key_choice.hpp:L220`: delete: `RowIdKeyChoice::index_id` is written and never read. Remove the field and its assignment.
- net: -7 lines possible.

### Unrelated, found while testing
`test/sql/catalog/first_touch_round_trips.test:69` (`SELECT k1, k2, rowid` over a composite key with a string column) fails in a **debug** build with the internal assertion `source_p.GetType() == target.GetType()` in `PopulateRowIdVector`. It fails identically at this head **and** at `main` (`634b485`), so it is not from this PR. It's worth its own issue. Release builds don't assert, and I'd guess the cause is the rowid STRUCT child typed `VARCHAR` while the scan column is `MSSQL_VARCHAR(n)`.

**Verdict:** 1 blocking issue. I recommend fixing it before merge.

oluies:
Pushed `274014a` onto this branch. It fixes the findings from the adversarial review above.

- **The BLOCK:** the key join now compares a `char`/`varchar` key column under a `SQL_` collation against `CAST(v.[k] COLLATE <column collation> AS varchar(n))`. `rowid_unique_index.test` adds both shapes, a unique index and a `char` primary key. I ran it against the unfixed build first: it failed, then passed with the fix.
- **NOTEs 1–11:** each is fixed as described in the commit message. The debug-variable fix is a fallback to `MSSQL_DEBUG`.
- **Not changed:**
  - NOTE 12, the unwind bound. It is documented, so I left it as is.
  - The two over-engineering cuts.

**Tests:** run on a debug build against the docker server.

| Suite | Assertions | Result |
|---|---|---|
| rowid | 256 | green |
| dml | 258 | green |
| insert | 560 | green |
| transaction | 287 | green |
| copy | 1129 | green |
| query | 644 | green |
| ctas | 476 | green |
| catalog | 42 of 43 cases | one failure, below |

The one catalog failure is `first_touch_round_trips.test:69`, the debug-only `PopulateRowIdVector` assertion. It fails identically on `0383262` and on `main`. `test_rowid_key_choice` and `test_identity_insert` pass.

VGSML:
Thank you for the adversarial round and for `274014a` — I read the fix against the mechanism before taking it, and it is the right one: `COLLATE` before the `CAST` so the conversion runs in the column's own code page, the conversion on the VALUES side so the column stays seekable, and only `SQL_` collations, where `varchar` and `nvarchar` genuinely compare under different rules (a Windows or UTF-8 collation's unique index could not have held `'Straße'` and `'Strasse'` as two keys in the first place). One thing worth knowing about the other side of the same coin: since #361 (merged today) the scan's `rowid = 'Straße'` filter also goes as a `varchar` parameter when the column's code page holds the value — which a value read from that column always is — so the filter and the key join now compare under the same rules.

On top of your commit, two small things:

- **The branch is merged with `main`** (`d9a1419`: #361, #366/#367, the pin bump and the specs) — `.PHONY` and the CHANGELOG were the only conflicts.
- **Your two over-engineering cuts** (`170aa8c`): `DescribeRejections` takes the rejection list, `ChoiceWith` is gone, and so is the unread `RowIdKeyChoice::index_id`; −9 lines.

Your correction on the `0383262` Medium is taken as well: I had a further commit that made pool failures transient (not published, retried on the next bind) and it is dropped — caching them, with the reason named, is the shipped behaviour improved, not a new hazard.

The debug-build assertion in `PopulateRowIdVector` on a composite key with a string column is filed as #369, as its own fix — it is the rowid STRUCT child typed `VARCHAR` against a projected `MSSQL_VARCHAR(n)`, exactly your guess.

Local after the merge: rowid 256, dml 258, insert 560, catalog 1824 (43 files), transaction 287 assertions green; both unit tests pass.

PR #352, opened 2026-09-16
fix: rowversion columns are readable, and the 2025 JSON type is read uncast (#296)

### LINKED ISSUE #296: Asking to add support for TIMESTAMP/ROWVERSION data types

VGSML:
Closes #296.

Three type names the catalog did not know, and one write path that was broken because of the third. Read commit by commit: each is a self-contained step and the third is a different subsystem.

## 1. A table with a `ROWVERSION` column could not be read at all

Neither that name nor SQL Server 2025's `JSON` was in the catalog's type table, so both took the unknown-type route, `CAST(col AS NVARCHAR(MAX))` — and for rowversion the server refuses that cast outright:

```
Invalid Input Error: SQL Server error [529, severity 16]: Explicit conversion
from data type timestamp to nvarchar(max) is not allowed.
```

It is `BIGBINARY(8)` on the wire (`sp_describe_first_result_set` reports `tds_type_id` 173), and the describe path in `mssql_functions.cpp` has always mapped it that way, so only the catalog disagreed with the stream.

**BLOB rather than the BIGINT the issue also suggested**, for two measured reasons. The counter is unsigned, so a BIGINT would report negative values past 2^63. And the bytes are big-endian: three consecutive writes gave `0x…1fcc`, `1fcd`, `1fce`. A BLOB therefore compares bytewise in exactly the order the counter advances, so `WHERE rv > <last seen>` works with no conversion on either side and pushes down to the server.

**Not a TIMESTAMP either, and not because of rounding.** The counter is database-wide and advances once per write to *any* table. Three rows, the second written two seconds after the first, the third ten milliseconds after the second with three writes to a different table in between:

| row | counter | actually written at |
| --- | ------- | ------------------- |
| 1   | 8328    | 10:11:45.351        |
| 2   | 8329    | 10:11:47.351        |
| 3   | 8333    | 10:11:47.361        |

Two seconds moved it by one; ten milliseconds moved it by four. As a DuckDB `TIMESTAMP` those read `1970-01-01 00:00:00.008328` — an invented time, not an approximate one. Microsoft deprecates the `timestamp` spelling for the same reason; both are accepted here.

**`JSON` → `VARCHAR`.** It arrives as `varchar(max)` under `Latin1_General_100_BIN2_UTF8`, so the bytes are already what a DuckDB VARCHAR wants. Naming it only stops a pointless server-side conversion of every value.

## 2. Self-review, which found the two write paths disagreeing

Making `json` known left INSERT into a JSON column working while COPY into the same column was refused at bind: the COPY compatibility table lists `xml` for a VARCHAR source but not `json`. Strictly that predates this branch, but the changelog now claims json is supported and a split like that is not something a user can guess.

Three boundaries were assumed and are now measured and pinned: writing a rowversion is refused identically on **all three** paths and always by the server's own message (273 from a statement, the same text from `INSERT BULK`, the same through COPY); a COPY that does not name the column loads and the server fills it in; CTAS from such a table creates `varbinary` and carries the values.

One test was deleted rather than kept: the JSON section guarded its assertions with `CASE WHEN … THEN 1 ELSE 1 END`, true whatever happens. The table is now created with a JSON column where the server has one and NVARCHAR(MAX) where it does not, so every assertion runs on both; the file header says plainly that a green run on 2022 is not coverage of the JSON half.

## 3. Writing a `GEOMETRY` into a `geometry` / `geography` column

Broken on main too, and it is the same class of problem: the value went as a bare `0x…` literal, which SQL Server reads as its own Spatial Type Binary Format rather than as the OGC WKB a DuckDB GEOMETRY carries. The heads differ — WKB `0x01 01000000 <16 bytes>` against internal `0xe6100000 01 0c <the same 16 bytes>` — so no cast bridges them. `STGeomFromWKB` is the server's reader for the form we have, and INSERT and UPDATE now wrap the literal in it.

**The SRID is an assumption, documented as one.** `.STAsBinary()` carries none, so a value read from SQL Server has already lost it: a geometry target gets 0, a geography target gets 4326, because geography refuses 0 with error 24204.

**No routing change was needed, and that was verified rather than assumed.** Spec 062 already keeps an INSERT naming a spatial column on the statement path. Proved with the parse-failure lever that exists only there: 20 rows at a threshold of 5 naming the column fail on the injected error, and the same 20 without it go through `INSERT BULK` untouched. Both directions are in the test, which also covers a point, a linestring and a polygon — three branches of the server's WKB reader, so a wrapper that only worked for points would fail it.

## 4. `INSERT … RETURNING` against a table holding an undecodable column

Found in the second self-review, older than this branch, and visible now only because writing a geometry works at all. The generated `OUTPUT` clause carries **every** column of the table rather than the ones the user named, because DuckDB's RETURNING projection sits above the insert and expects the table's full width. So one undecodable column failed the whole statement:

```sql
-- on a table that also has a geometry column
INSERT INTO t (n) VALUES (5) RETURNING id, n;
-- COLMETADATA parse error: Unsupported SQL Server type: UDT
```

The same held for `hierarchyid` and `sql_variant`.

The fix is not a new rule. The read path solved this for exactly these columns, and `OUTPUT` accepts exactly those expressions — verified against the server before any code was written. The OUTPUT list is now built the way the scan builds its SELECT list: `.STAsBinary()` for a spatial UDT, a CAST to NVARCHAR(MAX) for the types with no decodable wire form, the plain column otherwise, each aliased back to its name so order and count stay what `MSSQLReturningParser` matches on. INSERT is the only path that builds an OUTPUT clause.

## Verification

| | |
| --- | --- |
| integration suite | 197 cases, 6345 assertions, 5 env-gated skips |
| standalone C++ tests | all 31, including `test_filter_encoder` |
| server | SQL Server 2025 RTM-CU8 locally; the JSON half is exercised as nvarchar on 2022 |

Two knock-on effects are pinned by tests: `test_filter_encoder` asserted that rowversion declares **no** parameter *because* it was cast-required, and now declares `varbinary(8)` — reaching a branch `DeclarationForColumn` has always had; and writing a rowversion is refused on both DML paths, so a later change that lets the bulk path swallow the value fails here.

## Found alongside, deliberately not fixed here

One more thing the self-reviews changed: `geometry` and `geography` were spelled out in four places, and two of them have to agree or the failure is silent — the catalog sets `is_geometry` from that pair and that is what keeps such an INSERT on the statement path, while the renderer decides from the same pair whether to wrap the value. They are now one predicate, `MSSQLColumnInfo::IsSpatialType`.

- **`COPY` into a table with a CLR UDT column drops that column** — silently when it is nullable, with a misleading NULL error when it is not. `GetExistingTableColumnMetadata` joins `sys.types` on `system_type_id`, which no `sys.types` row satisfies for a UDT, so the column never enters the target metadata. The fix is verified as a query already: make it a LEFT JOIN and add a second one on `user_type_id` for the declared name, which keeps alias-type resolution working and makes an empty base name the signal. Own PR, because it is a different subsystem and a different failure semantic.
- **The `sql_variant` read path loses data silently.** A `datetime2(7)` holding 10:11:12.1234567 arrives as `Sep 16 2026 10:11AM`; a `varbinary` arrives as mojibake. The cast is style 0. Spec 077 § 8 records the measurements.


src/dml/insert/mssql_insert_statement.cpp
```diff
@@ -39,7 +40,28 @@ void MSSQLInsertStatement::InitializeCache() const {
...
+			// `INSERT … RETURNING id` that never mentioned it — measured, a table
+			// with a geometry column answered
+			// `COLMETADATA parse error: Unsupported SQL Server type: UDT`.
+			//
+			// The read path already solved this for the same columns, and OUTPUT
+			// accepts the same expressions: a method call for the spatial UDTs,
+			// whose bytes then arrive as the OGC WKB a DuckDB GEOMETRY wants, and
+			// a CAST to NVARCHAR(MAX) for the types with no decodable wire form
+			// (sql_variant, hierarchyid, a CLR UDT). Both were tried against the
+			// server before this was written.
```
oluies: **Low — "the same expressions the read path does" isn't quite what this builds.**

The CHANGELOG and the comment here both say the `OUTPUT` list uses the same expressions as the read path, but this copies only two of the scan's rewrites: `.STAsBinary()` and the `is_cast_required` CAST.

`BuildColumnExpression` (`src/table_scan/table_scan.cpp:165`) also rewrites:

- `ntext` → `NVARCHAR(MAX)` and `image` → `VARBINARY(MAX)`;
- non-UTF-8 `char`/`varchar`/`text` → `NVARCHAR(n)`.

Those types are *known*, so they aren't `is_cast_required` and still go out as bare `INSERTED.col`. Since `text` has no decodable raw wire form (the comment at `table_scan.cpp:135` says exactly this), `INSERT … RETURNING id` on a table carrying a `text`, `ntext` or `image` column still fails the whole statement — which is the failure this commit is meant to fix.

**Fix:** build the `OUTPUT` expressions from the same helper the scan uses — lift `BuildColumnExpression` into a shared function taking a qualifier, and call it with `INSERTED.`. That keeps the two lists in step by construction rather than by comment. Please add a RETURNING case with a `text` or `ntext` column to `insert_returning_udt_columns.test`.

test/sql/catalog/rowversion_and_json_types.test
```diff
@@ -0,0 +1,267 @@
...
+
+query I
+SELECT count(*) FROM rvt.dbo.RowVersionTypes WHERE id = 200;
+----
+1
+
+statement error
+COPY (SELECT 201 AS id, '\x01\x02'::BLOB AS rv, 'x' AS note) TO 'mssql://rvt/dbo/RowVersionTypes' (FORMAT 'bcp', CREATE_TABLE false);
+----
+Cannot insert an explicit value into a timestamp column
```
oluies, line 183: **Low — this assertion depends on ordering that isn't guaranteed.**

The test expects the server's `Cannot insert an explicit value into a timestamp column` (error 273). But the BLOB arm of `IsTypeCompatible` in `src/copy/target_resolver.cpp` accepts only `varbinary`, `binary` and `image`, and `sys.types` names this column `timestamp`. Bind therefore admits `rv` as all-NULL only (`null_only_columns`), and the encoder's own type-mismatch error fires on the first chunk carrying a value.

Which of the two errors the user actually sees depends on whether the server rejects the `INSERT BULK` column list before that chunk is encoded — and that ordering isn't pinned by anything, so this assertion can flip when COPY initialisation order changes.

**Fix:** pick the intended behaviour and pin it. Either add `timestamp`/`rowversion` to the BLOB arm so error 273 always wins, or refuse the column at bind with a message of our own — then assert that specific message.

oluies (dismissed):
Approving — two low-severity comments inline, neither blocking. The first one is worth a look though: the `OUTPUT` list may not cover the case the change set out to fix.

**Summary of findings**

| Severity | Where |
| --- | --- |
| Low | `mssql_insert_statement.cpp` — `OUTPUT` copies only two of the scan's four rewrites, so `text`/`ntext`/`image` still fail |
| Low | `rowversion_and_json_types.test` — the asserted error depends on unguaranteed ordering |


---
*Posted from a [roborev](https://github.com/) review (job 1641), relayed via Claude Code.*

test/sql/insert/insert_returning_udt_columns.test
```diff
@@ -0,0 +1,139 @@
...
+----
+1	hello text	hello ntext	5
+
+# --- image too: it takes the VARBINARY(MAX) rewrite, not the NVARCHAR one, so
+# it comes back as a BLOB rather than mojibake.
+statement ok
+SELECT mssql_exec('retu', 'INSERT INTO dbo.RetLob (im, n) VALUES (CONVERT(IMAGE, 0x0102FF), 6)');
+
+query III
+INSERT INTO retu.dbo.RetLob (n) VALUES (7) RETURNING id, im, n;
```
oluies: **Medium — this case would still pass with the fix removed.**

The `image` block never sends a value through `OUTPUT INSERTED`. The row holding `0x0102FF` is inserted with `mssql_exec` on the line above, and *this* statement inserts a row whose `im` is NULL — so what comes back is `NULL`, and the expected output says so.

The test's own header makes the point: a NULL isn't enough, because the row reader only dispatches on the type when a value is present. Drop the `VARBINARY(MAX)` rewrite from the OUTPUT path and this assertion is unchanged — which means the commit's headline fix for legacy LOBs has no guard.

The comment below ("RETURNING * over the whole width, every rewrite at once") also sits above a plain `SELECT`, which exercises the scan path, not the OUTPUT path.

**Fix:** put the value in the RETURNING statement itself, e.g.

```sql
INSERT INTO retu.dbo.RetLob (im, n) VALUES ('\x01\x02\xFF'::BLOB, 6) RETURNING id, im, n;
```

and assert the bytes come back. Then either make the whole-width check a real `RETURNING *`, or correct the comment.

src/include/dml/insert/mssql_insert_config.hpp
```diff
@@ -124,7 +129,8 @@ struct MSSQLInsertColumn {
 	uint8_t scale;
 
 	// Default constructor
-	MSSQLInsertColumn() : is_identity(false), is_nullable(true), has_default(false), precision(0), scale(0) {}
+	MSSQLInsertColumn()
+		: max_length(0), is_identity(false), is_nullable(true), has_default(false), precision(0), scale(0) {}
 
 	// Full constructor
 	MSSQLInsertColumn(const string &name, LogicalType duckdb_type, const string &mssql_type, bool is_identity,
```
oluies, line 136: **Low — `max_length` is initialized in the default constructor only.**

The full constructor leaves it indeterminate, and `BuildReadExpression` reads it to decide whether a `varchar` is MAX and how long the NVARCHAR cast should be. Nothing calls this constructor today, so it's latent — but a future caller gets a random cast length, and the failure would look like a collation or truncation bug rather than an uninitialized field.

**Fix:** add a `max_length` parameter to the full constructor, or at minimum initialize it (`max_length(0)`) there as well.

src/table_scan/table_scan.cpp
```diff
@@ -81,133 +81,20 @@ static idx_t ApplyClientFilters(ClientContext &context, MSSQLScanGlobalState &gl
...
-	}
-
-	if (NeedsNVarcharConversion(col, convert_varchar_max)) {
-		std::string nvarchar_len = GetNVarcharLength(col);
-		MSSQL_SCAN_DEBUG_LOG(2, "  NVARCHAR conversion: %s (%s, len=%d) → NVARCHAR(%s)", col_name.c_str(),
-							 col.sql_type_name.c_str(), col.max_length, nvarchar_len.c_str());
-		return "CAST(" + escaped_name + " AS NVARCHAR(" + nvarchar_len + ")) AS " + escaped_name;
+	std::string expr = MSSQLColumnInfo::BuildReadExpression(col_name, col.sql_type_name, col.max_length,
+															col.collation_name, convert_varchar_max);
+	if (expr != "[" + FilterEncoder::EscapeBracketIdentifier(col_name) + "]") {
```
oluies: **Low — a per-column string build and compare to gate a debug message.**

This rebuilds the plain escaped column name and compares it against the returned expression purely to decide whether to log, on the scan path, for every column. It also re-encodes the "unrewritten" form that `BuildReadExpression` now owns — so if that function ever changes how an unrewritten name is spelled, the log silently goes quiet or noisy.

Related: the flags `MSSQLColumnInfo` already computes (`is_geometry`, `is_cast_required`, `is_utf8`) are now unused on this path, with `BuildReadExpression` recomputing the same decisions from the type name and collation. They agree today, but that is two sources of truth for the same question — which is precisely what the `IsSpatialType` refactor earlier in this PR set out to remove.

**Fix:** build the escaped name inside the debug-log guard (or drop the comparison and log unconditionally at level 2). Longer term, have the `MSSQLColumnInfo` flags come from the same helpers `BuildReadExpression` uses.

oluies (dismissed):
Approving. Routing the `OUTPUT` list through `MSSQLColumnInfo::BuildReadExpression` is exactly right — one function, so the two lists cannot drift.

One Medium inline, and it's about the test rather than the fix: the `image` case doesn't actually exercise the rewrite it's named for, so this change currently ships without a guard for its own central claim. Two Lows besides.

| Severity | Where |
| --- | --- |
| Medium | `insert_returning_udt_columns.test` — the `image` case would pass with the fix removed |
| Low | `mssql_insert_config.hpp` — `max_length` uninitialized in the full constructor |
| Low | `table_scan.cpp` — per-column string rebuild to gate a debug log; flags now recomputed |


---
*roborev job 1645 (re-review of the new commits only), relayed via Claude Code.*

src/include/dml/insert/mssql_insert_config.hpp
```diff
@@ -124,14 +129,17 @@ struct MSSQLInsertColumn {
...
-					  bool is_nullable, bool has_default, const string &collation, uint8_t precision, uint8_t scale)
+					  bool is_nullable, bool has_default, const string &collation, uint8_t precision, uint8_t scale,
+					  int16_t max_length = 0)
```
oluies, line 138: **Low — 0 is defined, but it isn't a safe length.**

This closes the uninitialized read, but for a `varchar`/`char` column with a non-UTF-8 collation, `BuildReadExpression` → `NVarcharLength` returns `std::to_string(0)`, so the OUTPUT list gets:

```sql
CAST([c] AS NVARCHAR(0))
```

which SQL Server rejects. And because the parameter is *defaulted*, a future caller can still omit it without the compiler saying anything — then hits an error that reads like a SQL syntax or collation problem rather than a missing argument. The only real caller today (`mssql_catalog.cpp:694`) assigns the field directly, so nothing is broken now.

**Fix:** make the parameter required (drop the `= 0`) so every caller has to supply a length — or default it to `-1` (MAX), which at least produces a valid cast, and say so in the comment.

oluies (approved):
Re-approving. The `image` and whole-width cases now send their values through `OUTPUT INSERTED`, so they actually guard the rewrite — remove the fix and they fail, which is what was missing.

One Low: initializing `max_length` to 0 removes the undefined read but picks a value that can't produce a valid cast. Not blocking, and no caller hits it today.


---
*roborev job 1649 (re-review of `7361252` only), relayed via Claude Code.*

PR #355, opened 2026-09-16
fix: COPY silently dropped CLR UDT columns, and the metadata join that caused it was the slow one (#353)

### LINKED ISSUE #353: COPY silently drops CLR UDT columns from an existing target, and the join that causes it is also the slow one
## Symptom

`COPY … TO 'mssql://…' (FORMAT 'bcp', CREATE_TABLE false)` into a table that has a `geometry`, `geography` or `hierarchyid` column **drops that column without saying anything**.

Measured: a source supplying `id`, `g` (valid WKB) and `n` reported success, and `g` landed `NULL`. No error, no warning. Same for a `hierarchyid` column fed a valid path.

If the column is `NOT NULL` the load instead fails with

```
MSSQL: BCP failed: Cannot insert the value NULL into column 'g', table '…';
column does not allow nulls. INSERT fails.
```

which is baffling, because the user did supply a value.

## Cause

`TargetResolver::GetExistingTableColumnMetadata` builds the target metadata with

```sql
JOIN sys.types t ON c.system_type_id = t.user_type_id AND t.system_type_id = t.user_type_id
```

A CLR UDT column has `system_type_id` 240 and its own `user_type_id` (129 for geometry), and no `sys.types` row satisfies that condition for 240 — so **the column never appears in the result at all**. It is then absent from the `INSERT BULK` column list, and `BuildColumnMapping` has no target to map the source column onto.

Verified directly: that query against a table `(id int, g geometry, gg geography)` returns one row, `id`.

Two sites: `src/copy/target_resolver.cpp:826` and `:1158`. CTAS is unaffected — it builds from source types, not from this query.

## Fix, and it is faster than what is there now

Drop the join entirely and take the name from `TYPE_NAME`:

```sql
ISNULL(TYPE_NAME(c.system_type_id), TYPE_NAME(c.user_type_id)) AS type_name
```

Interleaved A/B on a 102-column table, 200 rounds each in one session:

| form | µs |
| --- | --- |
| current `JOIN sys.types` | 497 |
| `LEFT JOIN sys.types` | 1345 |
| `LEFT JOIN` + `TYPE_NAME()` fallback | 1263 |
| `LEFT JOIN` + a second `LEFT JOIN` for the declared name | 1353 |
| **no join at all** (the form above) | **115** |

The LEFT JOIN is what costs, `TYPE_NAME()` is free, and removing the join is **4.3× faster than today** while returning all 102 columns instead of 101.

Alias resolution — the reason the inner join exists — is preserved. On a table carrying a geometry, a hierarchyid and a column of an alias type over `varchar(50)`:

| column | name from the join-free form |
| --- | --- |
| `id` | `int` |
| `g` | `geometry` |
| `h` | `hierarchyid` |
| `a` (alias `cw_alias`) | **`varchar`** |
| `c1` | `nvarchar` |

The joined alias `t` is used for nothing but `t.name` at every site, so the join is simply dead once the expression replaces it.

## Behaviour once the column is visible

A bulk load still cannot carry a spatial value: the wire would want SQL Server's own Spatial Type Binary Format, not the OGC WKB a DuckDB `GEOMETRY` holds. So:

- a UDT column the source does **not** supply is dropped as today, by the existing issue #125 machinery for unmatched target columns;
- a UDT column the source **does** supply is **refused**, with a message naming the column, its type, and `INSERT` as the way — which does work since #352.

That turns silent data loss into a named refusal, which is all COPY can honestly offer without a WKB→internal serializer.

## Also in scope: the same substitution in the catalog queries

The three catalog metadata queries in `src/catalog/mssql_metadata_cache.cpp` never had this bug — they already use `LEFT JOIN … ISNULL(t.name, TYPE_NAME(c.user_type_id))` — but that is the **slow** form in the table above, and they sit on the path spec 076 optimised. Replacing `t.name` with `TYPE_NAME(c.system_type_id)` there makes their join dead too: same expression, one fewer join, no behaviour change.

## Acceptance

- A COPY whose source names a UDT column is refused with an actionable message, not silently dropped.
- A COPY whose source omits it still loads.
- An alias-typed column still resolves to its base type (`test/sql/catalog/udt_type_alias.test` must stay green).
- No `JOIN sys.types` left in either file.

Found while fixing #296.

VGSML:
Closes #353.

One fix, plus a revert of an attempt that did not work. #353's actual content is the second commit.

## The race this PR tried to fix is NOT fixed here — see #356

The first commit attempted it, the attempt did not work, and the third commit reverts it. Measured properly, 60 runs per build: **main 2/60**, this branch without the attempt 3/60, with it 2/60. Same ~4% noise, and the defect is **on main** — this branch neither causes nor worsens it. The earlier "0 in 30" claims in this PR were samples too small to mean anything at that rate.

What was learned is in #356, including four hypotheses that were measured and disproved, so nobody repeats them.

## 2. COPY silently dropped CLR UDT columns

A COPY into a table with a `geometry`, `geography` or `hierarchyid` column dropped that column without a word: a source supplying `id`, `g` (valid WKB) and `n` reported success and landed `g` as NULL. A NOT NULL target instead failed with *"Cannot insert the value NULL into column 'g'"* about a value the user had in fact supplied.

The metadata query joined `sys.types` on `system_type_id`, which no `sys.types` row satisfies for a CLR UDT (`system_type_id` 240), so the column never reached the resolver.

**The fix removes the join rather than widening it, and is faster than what was there.** Interleaved A/B on a 102-column table, 200 rounds each in one session:

| form | µs |
| --- | --- |
| current `JOIN sys.types` | 497 |
| `LEFT JOIN sys.types` | 1345 |
| `LEFT JOIN` + `TYPE_NAME()` fallback | 1263 |
| `LEFT JOIN` + a second `LEFT JOIN` | 1353 |
| **no join at all** | **115** |

So `ISNULL(TYPE_NAME(c.system_type_id), TYPE_NAME(c.user_type_id))` is **4.3× cheaper** than today and returns all 102 columns instead of 101. Alias resolution — the reason the inner join existed — is preserved: a column of an alias type over `varchar(50)` still answers `varchar`. Applied to **all eight** metadata queries, the four in the COPY resolver and the four catalog ones, which never had the bug but did have the slow form. That second half is the optimisation the issue asked for.

Behaviour once the column is visible: a UDT column **no** source feeds is left out of the load exactly as before; one a source **does** feed stops the load with a message naming the column, its type and `INSERT` as the way. Silent data loss becomes a named refusal, which is all COPY can offer without a WKB-to-internal serialiser.

Two things found while building it, both commented at the code:

- a `#temp` table **cannot** carry a user alias type — the server refuses the CREATE — so `TYPE_NAME` resolving in the current database rather than tempdb is safe on that path;
- the bare `catch (...)` around the target-metadata block swallows everything, **including a deliberate refusal thrown inside it**. The first version of this fix silently became a fallback to source-derived metadata, which declared a geometry column as `varbinary(max)` and let the server answer *"Invalid column type from bcp client"*. The check runs outside that block for exactly that reason.

## Verification

| | |
| --- | --- |
| integration suite | 195 cases, 6263 assertions, 5 env-gated skips |
| `udt_type_alias.test` | green — the real guard for the query change |

Depends on nothing; #352 touches `target_resolver.cpp` too, so whichever merges second rebases.


VGSML:
## Correction: commit 1 does not fix the race, and the "0 in 30" in its message is an unreliable sample

CI caught it on this branch — `sink_reads_own_catalog.test:130`, the same `Cannot execute: connection not in Idle state`. I then ran 60 local iterations instead of 30 and got **2 failures**. So the rate dropped from roughly 3/40 to 2/60, and my earlier zero was luck. Treat the claim in `a731b29`'s message as wrong; I would rather say that here than leave it standing.

## What the failure actually is, which I had backwards

The message comes from `MSSQLResultStream::Initialize` (`mssql_result_stream.cpp:145`) — it is a **scan** sending its SELECT and finding the connection busy, not the sink sending `INSERT BULK`. So guarding `OpenStream` protects the wrong side of the collision.

And a mutex cannot express the invariant here at all. The sink opens the BCP stream and holds the pinned connection in **Executing for the whole load**, not just for the send. Any scan that starts executing after that point finds it busy no matter what mutex either side takes. Holding the mutex across the whole stream is not an option either: the stream needs chunks from the very scan that would be waiting for it.

So the real requirement is the one spec 075 W3 states — every catalog scan of the sink's catalog **materialises before the sink opens its stream** — and the failure means something is escaping that, intermittently. The instrumentation I ran earlier shows the optimizer marking the right bind data and the scan reading `requires_materialization=1` on the same object, so the flag is not the hole; the hole is somewhere between that and "no scan is left running when the stream opens".

## What I suggest for this PR

Split it. Commit 2 — the UDT-column fix and the metadata-query change — is sound on its own and is what #353 asks for. Commit 1 should come out until the invariant is fixed properly rather than narrowed.

The catch: the metadata-query change is what made the race frequent enough to see, so shipping commit 2 alone raises the flake rate on a bug that is already in main. That is a real trade and it is the repo owner's call, not mine. The alternatives are to hold the whole PR until the race is understood, or to ship the UDT fix without the query optimisation, which measured clean at 0/30 but I am now less willing to call that proof.

CHANGELOG.md
```diff
@@ -7,6 +7,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `sys.types` row satisfies for a `geometry`, `geography` or `hierarchyid`
+  column, so the column never reached the resolver: a source column feeding one
+  was ignored and its values lost, and a NOT NULL target failed with "Cannot
+  insert the value NULL into column …" about a value the user had supplied. The
+  join is gone from all eight metadata queries — the name comes from
+  `ISNULL(TYPE_NAME(c.system_type_id), TYPE_NAME(c.user_type_id))`, which is
+  correct for UDTs and **4.3× cheaper** (497 µs → 115 µs on a 102-column table,
+  interleaved, 200 rounds) — and a source that feeds such a column is now
+  refused by name. A source that omits it still loads and the server fills it,
+  exactly as before.
```
oluies: **Medium — this entry describes code the PR reverts.**

The "the deferred `INSERT BULK` is guarded again" entry documents `a731b29`, which `8e7434a` reverts inside this same range. Nothing here ships that code, so the changelog promises a fix that doesn't exist — and it still quotes the *"3 in 40 … 0 in 30"* numbers that the revert message explicitly withdraws as luck, and still blames the metadata-query change for the timing shift.

**Fix:** delete the entry (the revert didn't touch `CHANGELOG.md`). Optionally replace it with a note that the pinned-connection race is still open at ~4% on main, now being fixed in #357.

src/copy/copy_function.cpp
```diff
@@ -491,6 +491,32 @@ unique_ptr<GlobalFunctionData> BCPCopyInitGlobal(ClientContext &context, Functio
 			CopyDebugLog(1, "BCPCopyInitGlobal: using source column metadata (table created/replaced)");
 		}
 
+		// Issue #353: a target column whose type the bulk wire cannot carry, still
```
oluies, line 531: **Medium — this refuses an all-NULL source column that used to load.**

The refusal fires for any mapped column with `bulk_unsupported` set, without checking whether the source column is all-NULL. Everywhere else in this file `ValidateExistingTableSchema` treats a typed constant NULL as *"the ordinary way to fill a column"* and admits it via `null_only_columns`.

So before this change:

```sql
COPY (SELECT id, NULL AS g, n FROM src) TO 'mssql://db/dbo/t' (FORMAT 'bcp', CREATE_TABLE false);
```

loaded, and `g` ended up NULL — which is the correct result, and the one the user asked for. It now fails at init with *"the bulk-load wire cannot carry"*. The same applies to `SELECT *`-shaped copies whose UDT columns happen to be NULL throughout. The error text also tells the user to leave the column out, which is a different query.

**Fix:** when a `bulk_unsupported` column is also `null_only_source`, drop it from the `INSERT BULK` list the way the unmatched path already does (the server fills NULL, or reports NOT NULL itself) and refuse only when the source can carry values. Alternatively refuse lazily in the encoder, on the first non-NULL. Either way, please add a `NULL AS g` case to `copy_udt_columns.test`.

src/copy/target_resolver.cpp
```diff
@@ -1050,6 +1051,31 @@ static uint16_t SQLServerTypeMaxLength(const string &type_name, int16_t max_leng
...
+//! rewrites them to `.STAsBinary()` and they arrive as WKB, which is why
+//! `IsKnownSQLServerType` admits them — but their WIRE form on the way back is
+//! SQL Server's own Spatial Type Binary Format, not those bytes, so a bulk load
+//! cannot send them. And the types with no decodable wire form at all
+//! (`sql_variant`, `hierarchyid`, any other CLR UDT) are exactly the ones
+//! `IsKnownSQLServerType` already rejects.
+//!
+//! The spatial half spells the two names out here rather than sharing a
+//! predicate, because the branch that introduces `MSSQLColumnInfo::IsSpatialType`
+//! (#352) is not merged yet; fold this into it on the rebase.
```
oluies, line 1096: **Low — this may refuse more than CLR UDTs.**

`IsKnownSQLServerType` also rejects system types that the old `sys.types` join *did* return — notably `sql_variant` and `timestamp`/`rowversion`. Those columns previously reached the VARCHAR/nvarchar fallback; now any source feeding them is refused at init.

For `sql_variant` an nvarchar bulk column would most likely have converted implicitly on the server, in which case this is an unannounced regression. The header comment says only "CLR UDTs and sql_variant", so the new `rowversion` refusal is documented nowhere — and #352 is concurrently making `rowversion` readable, which makes the round trip the interesting case.

**Fix:** check whether an nvarchar bulk load into `sql_variant` worked before. If it did, exempt it (or make the refused list explicit rather than deriving it from `IsKnownSQLServerType`). Either way document the `rowversion` case and add an existing-table COPY test covering `sql_variant`.

test/sql/copy/copy_udt_columns.test
```diff
@@ -0,0 +1,117 @@
...
+SELECT mssql_exec('udtc', 'IF OBJECT_ID(''dbo.UdtShapes'') IS NOT NULL DROP TABLE dbo.UdtShapes');
+
+statement ok
+SELECT mssql_exec('udtc', 'CREATE TABLE dbo.UdtShapes (id INT NOT NULL PRIMARY KEY, g GEOMETRY NULL, gg GEOGRAPHY NULL, h HIERARCHYID NULL, v SQL_VARIANT NULL, n NVARCHAR(10) NULL, d DECIMAL(9,3) NULL)');
+
+statement ok
+SELECT mssql_invalidate_cache('udtc');
+
+# --- the catalog reports the same types it always did. It never had this bug:
+# the catalog queries already used a LEFT JOIN with a TYPE_NAME fallback, so
```
oluies, line 187: **Low — stale comment.**

This says the catalog-query substitution will be guarded *"when it lands"*, but this same change already swaps the join out of all four catalog queries in `mssql_metadata_cache.cpp`. The header comment likewise scopes the change to "the COPY resolver's four queries" when it is in fact all eight.

**Fix:** update both comments to say the catalog queries changed here too — that is the half of the issue that was pure optimisation, and it deserves to be findable.

oluies (dismissed):
Approving. The join removal is well-argued and the measurements back it; turning silent data loss into a named refusal is the right call. Four comments inline — the CHANGELOG one and the all-NULL one are worth taking before merge.

**Summary of findings**

| Severity | Where |
| --- | --- |
| Medium | `CHANGELOG.md` — documents the `INSERT BULK` guard that `8e7434a` reverted |
| Medium | `src/copy/copy_function.cpp` — the refusal also rejects an all-NULL source column that used to load |
| Low | `src/copy/target_resolver.cpp` — `BulkWireCanCarry` also newly refuses `sql_variant` and `rowversion` |
| Low | `test/sql/copy/copy_udt_columns.test` — comment says the catalog-query change "will land"; it already did |


---
*Posted from a [roborev](https://github.com/) review (job 1640), relayed via Claude Code.*

src/copy/copy_function.cpp
```diff
@@ -603,6 +645,29 @@ void BCPCopySink(ExecutionContext &context, FunctionData &bind_data, GlobalFunct
 		throw IOException("MSSQL COPY: Previous error occurred: %s", gdata.error_message);
 	}
 
+	// Issue #353 review: the columns dropped from the load because the wire
+	// cannot carry them and bind admitted the pair for all-NULL only. They never
+	// reach the encoder, so its per-chunk guard cannot see them — check here, or
+	// a value would be dropped in the silence #353 was filed about.
+	for (const auto &dropped : gdata.null_only_dropped) {
```
oluies: **Medium — this moves the refusal from init to per-chunk, and a late value now partially loads.**

`IsTypeCompatible` never accepts a pair whose target type the bulk wire cannot carry, so *every* such column that a source feeds is marked `null_only_source` and dropped at init. The refusal therefore always lands here, on the first chunk that carries a value — not at init.

That matters once a value arrives after a flush boundary (`mssql_copy_flush_rows` = 102400 rows), or after another parallel writer's batch: in autocommit those rows are already committed. A source with `g` NULL for its first 200k rows and then a value **partially loads the table and then fails**. Before this change the same COPY loaded nothing at all.

Two consequences worth noting:

- the init-time loop at ~525-535 is now **unreachable** on the existing-table path;
- the new test comment "A pair bind CAN type-check is refused earlier, at init" describes a case that cannot occur.

**Fix — either:**

1. Drop the column only when the source type is DuckDB's `SQLNULL` (or otherwise provably all-NULL), and keep the init-time refusal for every other source type. `NULL AS g` still loads, and a typed source that merely happens to start with NULLs fails before any row is written; **or**
2. Keep the lazy refusal, but say plainly in the error text and the CHANGELOG that it can arrive after rows are committed, and delete or rewrite the now-dead init loop and the comment above.

I'd prefer (1), since #353 is a silent-data-loss issue and a partial write is the awkward middle ground — but (2) with honest documentation is a legitimate answer too.

src/copy/copy_function.cpp
```diff
@@ -603,6 +645,29 @@ void BCPCopySink(ExecutionContext &context, FunctionData &bind_data, GlobalFunct
...
+		if (dropped.source_index >= input.ColumnCount()) {
+			continue;
+		}
+		auto &vec = input.data[dropped.source_index];
+		UnifiedVectorFormat fmt;
+		vec.ToUnifiedFormat(input.size(), fmt);
```
oluies: **Low — deprecated overload.**

`vec.ToUnifiedFormat(input.size(), fmt)` is the two-argument form, which the pinned DuckDB marks `[[deprecated]]`. `PrepareColumnStates` right next to it already uses the single-argument form the 2.0 contract requires.

This adds a deprecation warning now and breaks the build under `-Werror`, or whenever the overload is removed from the 2.0 branch.

**Fix:** `vec.ToUnifiedFormat(fmt);`

src/copy/copy_function.cpp
```diff
@@ -439,6 +439,22 @@ unique_ptr<GlobalFunctionData> BCPCopyInitGlobal(ClientContext &context, Functio
...
 						// DEFAULT, computed) and must be left out of INSERT BULK entirely.
 						if (gstate->column_mapping[i] >= 0) {
+							// Issue #353 review: a column the wire cannot carry whose pair was
+							// admitted for the all-NULL case only. It cannot be declared (the
+							// VARCHAR fallback makes the server refuse the whole INSERT BULK),
+							// so drop it the way an unmatched column is dropped and let the
+							// server fill it — `NULL AS g` then loads, as it did before #353.
+							// Record it so the sink keeps checking it really is all NULL.
+							const auto &col = gstate->columns[i];
+							if (col.bulk_unsupported && col.null_only_source) {
```
oluies: **Low — this drop can feed the `catch (...)` trap you documented in the original commit.**

If the only source columns matching the target are null-only UDT columns — `COPY (SELECT NULL AS g) TO …` — then `mapped_columns` ends up empty and the "no source columns match" exception is thrown **inside** the `try`. The bare `catch (...)` swallows it and falls back to source-derived metadata, while `null_only_dropped` stays populated. The load then declares `g` from its source type against a geometry target, and the server answers with something unhelpful instead of our named message.

This is the same swallowing behaviour the commit message for the original fix called out — it just has a new way in.

**Fix:** throw the "nothing to load" error outside the `try`, or clear `null_only_dropped` in the catch. Best would be a specific message for "every matched column was dropped as null-only", which is a different user error from "no columns matched".

test/sql/copy/copy_udt_columns.test
```diff
@@ -0,0 +1,178 @@
...
+
+# --- and an all-NULL source for the UDT column loads too. `NULL AS g` is the
+# ordinary way to say "leave this column alone", and it worked before #353 —
+# by accident, because the join hid the column entirely. It has to keep
+# working: the column is dropped from the load exactly like an omitted one and
+# the server fills it. (It cannot be DECLARED instead: the type mapping gives
+# geometry the VARCHAR fallback, and an nvarchar column against a geometry
+# target makes the server refuse the whole INSERT BULK.)
+statement ok
+COPY (SELECT 3 AS id, NULL AS g, NULL AS h, 9 AS n) TO 'mssql://udtc/dbo/UdtCopyTgt' (FORMAT 'bcp', CREATE_TABLE false);
```
oluies, line 89: **Low — the new cases can't see the behaviour that's actually in question.**

Both the all-NULL case and the rowversion/sql_variant cases use single-row sources, so the value (when there is one) is always in the first chunk. The chunk-boundary case — where the refusal timing decides whether rows are already committed — is untested, and that is the whole substance of the Medium comment above.

**Fix:** add something like

```sql
COPY (SELECT i AS id, CASE WHEN i > 5000 THEN 'x' END AS h FROM range(10000) t(i))
  TO 'mssql://udtc/dbo/UdtCopyTgt' (FORMAT 'bcp', CREATE_TABLE false);
```

assert that it errors, **and** assert how many rows landed. That pins the partial-load behaviour down whichever way it is decided.

oluies (changes requested):
Requesting changes on one point — and it is one my own review caused, so apologies for the round trip.

My earlier comment offered two ways to fix the all-NULL refusal: drop the column when the source is provably all-NULL, or refuse lazily in the encoder. This took the second, and that trades a clean init-time failure for a **partial load**: a source whose first 200k rows are NULL and then carries a value now commits those rows and *then* fails. Before the change the same COPY loaded nothing. Given #353 is a data-loss issue, I don't think we want to land a different partial-write behaviour in the fix for it.

The other three are Low and can ride along.

| Severity | Where |
| --- | --- |
| **Medium** | `copy_function.cpp` — refusal moved to per-chunk, so a late value partially loads |
| Low | `copy_function.cpp:658` — deprecated two-argument `ToUnifiedFormat` |
| Low | `copy_function.cpp:449` — empty `mapped_columns` throw is swallowed by `catch (...)` |
| Low | `copy_udt_columns.test` — no case where the value arrives in a later chunk |

Happy to re-approve as soon as the first one is settled either way — including "documented as intended", if that's the call.


---
*roborev job 1644 (re-review of the new commits only), relayed via Claude Code.*

src/include/copy/target_resolver.hpp
```diff
@@ -123,6 +123,26 @@ struct BCPColumnMetadata {
 	// missing-column NullOnly path; a value in it is the type-mismatch error.
 	bool null_only_source = false;
```
oluies, line 129: **Low — the code now contradicts its own documentation.**

The new drop test rests on a bare `NULL` still being `SQLNULL` when COPY binds — you measured that, and say so in the commit message. But the comments documenting `null_only_columns` / `null_only_source` still assert the opposite: that a constant `NULL AS col` "reaches bind as INTEGER" and "cannot be told apart from real data here". They're unchanged by this diff, in `src/include/copy/bcp_config.hpp:129-137`, `src/copy/target_resolver.cpp:815-820`, here, and `src/copy/copy_function.cpp:414-417`.

The risk is concrete: a later maintainer who trusts those comments concludes the `SQLNULL` check is dead code and switches it back to `null_only_source` — which is precisely the change you measured as reintroducing #353's silent data loss. The new test would catch it, but only in the integration suite, which [doesn't run on PRs](https://github.com/hugr-lab/mssql-extension/issues/212).

**Fix:** correct those comments in this commit, while the measurement is fresh. Worth stating all three facts: a bare NULL arrives as `SQLNULL`; `null_only_columns` covers *typed* sources whose values the encoder checks per chunk; and a column whose target type the wire cannot carry is dropped at init only when its source is `SQLNULL`, refused at init otherwise.

src/copy/copy_function.cpp
```diff
@@ -491,6 +526,57 @@ unique_ptr<GlobalFunctionData> BCPCopyInitGlobal(ClientContext &context, Functio
...
+		// to precede the first non-NULL value.
+		for (const auto &col : gstate->columns) {
+			if (!col.bulk_unsupported) {
+				continue;
+			}
+			throw InvalidInputException(
+				"MSSQL COPY: column '%s' of table '%s' has type %s, which the bulk-load wire cannot "
+				"carry — its wire form is SQL Server's own, not the bytes a DuckDB value holds. "
+				"Use INSERT for this table, or leave the column out of the source (a constant "
+				"`NULL AS %s` is accepted and leaves it NULL).",
```
oluies: **Low — the hint is true only for an untyped NULL.**

This tells the user that `NULL AS <col>` is accepted, but that holds only while the value is still typed `SQLNULL`. Several ordinary sources give it a concrete type and hit this same refusal:

- `CAST(NULL AS VARCHAR) AS g`
- a column from a table or Parquet file that happens to be entirely NULL
- a NULL DuckDB has already typed by combining it with typed values

Someone in any of those cases follows the hint, writes what they reasonably believe is the same thing, and gets the identical error with nothing to explain the difference.

**Fix:** say it needs a literal, untyped `NULL AS <col>` and that a typed or cast NULL is refused — or drop the parenthetical and keep "leave the column out of the source", which always works.

oluies (dismissed):
Approving — the blocking Medium is resolved, and the fix is better than what I suggested.

Gating on the **source type** rather than on "provably all-NULL" is the right distinction, and measuring the counterfactual — gate on `null_only_source` instead and all 10000 rows land with the `hierarchyid` values silently gone, i.e. #353 reintroduced — is exactly the evidence that settles it. Pinning the row count rather than just the error means the test can tell those two outcomes apart. The `catch (...)` fix and the separate "every matched column was a constant NULL" message are both good catches beyond what I asked for.

Two Lows left, both about what the change *says* rather than what it does. Neither blocks.

| Severity | Where |
| --- | --- |
| Low | the `null_only_source` comments still assert the opposite of what this commit measured |
| Low | the refusal hint promises `NULL AS col` works, which is true only while it's untyped |


---
*roborev job 1647 (re-review of `c4ee03f` only), relayed via Claude Code.*

src/copy/copy_function.cpp
```diff
@@ -491,6 +528,59 @@ unique_ptr<GlobalFunctionData> BCPCopyInitGlobal(ClientContext &context, Functio
...
+		// to precede the first non-NULL value.
+		for (const auto &col : gstate->columns) {
+			if (!col.bulk_unsupported) {
+				continue;
+			}
+			throw InvalidInputException(
+				"MSSQL COPY: column '%s' of table '%s' has type %s, which the bulk-load wire cannot "
+				"carry — its wire form is SQL Server's own, not the bytes a DuckDB value holds. "
+				"Use INSERT for this table, or leave the column out of the source. A literal, "
+				"uncast `NULL AS %s` is also accepted and leaves the column NULL — but a TYPED "
```
oluies, line 560: **Low — "leaves the column NULL" isn't what happens for every column this refusal covers.**

The `typed vs uncast` half is now right. But the drop at ~474 removes the column from `INSERT BULK` entirely, so the server fills it exactly as it fills an *omitted* column — which is not always NULL:

- a **`rowversion`** column (which this refusal also covers, since `2ad15b7`) gets a **server-generated value**;
- a UDT or `sql_variant` column with a `DEFAULT` gets **the default**;
- a **`NOT NULL`** UDT column with no default makes the server fail with *"Cannot insert the value NULL…"* — the same baffling error #353 was filed about.

So someone following this hint for a `rowversion` or `NOT NULL` column gets a result the message didn't describe. Your own test comment at `copy_udt_columns.test:79-80` has it right: *"dropped from the load exactly like an omitted one and the server fills it"*.

**Fix:** say an uncast `NULL AS <col>` is treated like omitting the column, so the server fills it — NULL, its default, or a generated value — rather than promising NULL.

oluies (dismissed):
Re-approving. The comments now match what the branch measured, and adding the `CAST(NULL AS VARCHAR)` case to the test is more than the finding asked for — that's the exact trap the old hint would have walked someone into.

One Low left, in the same sentence: the hint is now correct about *which* NULLs are accepted, but still promises the wrong outcome for one class of column. Not blocking.


---
*roborev job 1648 (re-review of `15b13a9` only), relayed via Claude Code.*

oluies:
Both dependencies are now on `main`, so this is clear to rebase. The approval on `15b13a9` stands — nothing here needs another look from me once the conflict is resolved.

- **#357** merged as `7e2f7d9a` (the pinned-connection fix)
- **#352** merged as `1c980a40` (rowversion/JSON/geometry)

**On #357, since it bears on the revert in this PR.** Your call to revert `a731b29` and call the "0 in 30" luck was right, and #357's fix held up under test: #352 hit the race twice on CI against `b84e259` — once as `Connection closed while waiting for COLMETADATA` after 30 s, once as the canonical `Cannot execute: connection not in Idle state (current: Executing)` at `sink_reads_own_catalog.test:130`. After merging `main` into that branch the integration suite passed in 3m23s. So the two symptoms were one bug, and it is closed.

**The conflict.** Two files, and they are not the same kind of problem:

- `CHANGELOG.md` — both sides added entries at the top, mechanical.
- `src/copy/target_resolver.cpp` — **this is the one `2ad15b7` was written for.** #352 has now taught `IsKnownSQLServerType` about `timestamp`/`rowversion` so those columns are readable; your commit made `BulkWireCanCarry` name its refused types explicitly rather than derive them from that predicate, precisely so this merge would not silently re-admit them to the bulk wire. Resolving toward the derived form would put error 273 back in the middle of a load, and the only thing guarding it is `copy_udt_columns.test` — integration-only, which [does not run on PRs](https://github.com/hugr-lab/mssql-extension/issues/212). Worth running the copy suite locally after the rebase rather than trusting the green tick here.

I left the rebase to you rather than resolving it myself, for that reason.

Two Lows from the last review are still open and are fine to fold into the rebase or skip entirely:

- the refusal hint says an uncast `NULL AS <col>` "leaves the column NULL", but the drop makes the server fill it — so a `rowversion` gets a generated value, a `DEFAULT` column gets its default, and a `NOT NULL` UDT fails with the same message #353 was filed about. Your own test comment at `copy_udt_columns.test:79-80` has the accurate wording;
- `max_length`'s defaulted parameter in `mssql_insert_config.hpp` (that one is #352's, now on main).

---

PR #357, opened 2026-09-16
fix: pin a transaction's connection only after BEGIN has finished on it (#356)

### LINKED ISSUE #356: Intermittent: a scan finds the pinned connection busy inside a transaction — 'Cannot execute: connection not in Idle state'
`test/sql/transaction/sink_reads_own_catalog.test:130` fails intermittently, on CI and locally, with

```
IO Error: Failed to execute SQL batch: Cannot execute: connection not in Idle state (current: Executing)
```

The statement is a COPY reading from the same catalog it writes to, inside an explicit transaction:

```sql
BEGIN;
COPY (SELECT id, v FROM sk75.dbo.sk75_src) TO 'mssql://SK75/dbo/sk75_copy' (FORMAT 'bcp', CREATE_TABLE false);
COMMIT;
```

## Rate

60 runs of that file per build, same machine, same server, `--force-reload`:

| build | failures |
| --- | --- |
| **main** | **2 / 60** |
| a branch that rewrites the metadata queries | 3 / 60 |
| that branch plus an attempted guard | 2 / 60 |

About 4%, and **it is on main**. It was first seen as a one-off CI flake on #349 and has been dismissed as a flake since; it is not one.

A second symptom appears in the same runs and is probably the same defect seen from the other end: `IO Error: Connection closed while waiting for COLMETADATA` on the statement after `COMMIT`, preceded by `[MSSQL POOL] Closing connection in non-Idle state`.

## Where it is thrown

`MSSQLResultStream::Initialize` (`src/query/mssql_result_stream.cpp:145`) — a **scan** sending its SELECT and finding the connection already `Executing`. Not the sink sending `INSERT BULK`.

## What has been ruled out, with evidence

Please do not re-try these; each was measured and disproved.

- **"The scan is not marked for materialisation."** Instrumenting the bind-data pointer on both sides shows the optimizer marking `bd=0x…` for `sk75_src` and `TableScanInitGlobal` reading `mat=1` on that same pointer in the failing run. The flag is set and the scan acts on it.
- **"`Copy()` drops `requires_materialization`."** It does not. The field is on `MSSQLCatalogScanBindData` and its `Copy()` copies it.
- **"The COPY sink never takes `MaterializeMutex`."** It does, in `BCPCopyInitGlobal`, before acquiring the connection and for the rest of the init.
- **"Guarding the send fixes it."** Adding the mutex around `BulkLoadSession::OpenStream` changed the rate from 3/60 to 2/60, i.e. nothing. Reverted.
- **"The metadata queries are the trigger."** They are not: main fails at the same rate without them.

## Why a mutex looks like the wrong tool

The sink holds the pinned connection in `Executing` for the **whole load**, not for the send. Any scan that starts executing after the stream opens finds it busy whatever mutex either side holds. Holding the mutex across the whole stream is not available either: the stream is fed by the very scan that would be blocked on it, so that deadlocks by construction — and a plain `std::mutex` around the send already deadlocked once for a related reason (one thread taking it twice, init and deferred send on the same stack; the sample showed a single pipeline thread parked in `OpenStream` with no other holder).

So the thing to establish is not who locks what, but **who leaves the pinned connection in `Executing` at a moment when spec 075 W3 says everything should have drained** — and why it is timing-dependent when the materialisation decision is not.

## Reproduction

```bash
for i in $(seq 1 60); do
  build/release/test/unittest test/sql/transaction/sink_reads_own_catalog.test --force-reload \
    >/dev/null 2>&1 || echo "failed at $i"
done
```

A sample of 30 is **not** enough to conclude anything at this rate: a clean run of 30 happens about a third of the time even with the bug present. Two wrong conclusions in this investigation came from exactly that.

VGSML:
Closes #356.

Three commits, in the order they should be read: the analysis that found it, the instrument that identified it, then the fix and its test.

## The defect

A transaction's first statement could fail with

```
IO Error: Failed to execute SQL batch: Cannot execute: connection not in Idle state (current: Executing)
```

`ConnectionProvider::GetConnection` published the pinned connection **before** `BEGIN TRANSACTION` had finished on it:

```
161:  txn->SetPinnedConnection(conn);              // visible to every thread from here
165:  conn->ExecuteBatch("BEGIN TRANSACTION");     // connection -> Executing
      … receive the response, parse the ENVCHANGE descriptor …
210:  conn->TransitionState(Executing -> Idle);    // busy for this whole span
```

DuckDB initialises a plan's source and its sink on **different threads**, so the second one took the "a pinned connection already exists" early return and executed on a connection that was still mid-BEGIN. About **4%** for a statement that both reads from and writes to the same catalog inside a transaction — and when it struck it also leaked the connection out of the pool (`LEAKED active conn … quiescence contract violated`).

It had been seen once as a CI flake on #349 and dismissed as one. It was not.

## The fix, and why the obvious version is not enough

Moving the publication after the BEGIN is necessary and **insufficient**: two threads would then both miss the pinned connection, both acquire from the pool and both send BEGIN, leaving one pinned and the other leaked with an open transaction.

So the whole sequence — *is one pinned; if not, acquire, BEGIN, publish* — is one critical section per transaction. A second thread waits and then finds a connection that is pinned, begun and Idle. The lock order against the existing `connection_mutex_` is stated at the member, and the pool factory was checked for re-entrancy first: it builds from captured values and never returns to the provider, so holding the lock across `pool.Acquire` cannot self-deadlock.

## The instrument, which is what actually found it

The error said that *someone* held the connection and nothing about who. Five hypotheses were spent on that question. Now the connection records what took it to Executing and says so:

```
Cannot execute: connection not in Idle state (current: Executing);
it was taken to Executing by 'BEGIN TRANSACTION' on thread 4421914, this is thread 3397863
```

One line, and the defect was identified — after four wrong fixes had not. The naming is in the error itself, not behind a switch: it is built only on the failure path and "who has it" is as useful in production as it was here. `MSSQL_CONN_STATE=1` additionally logs one line per transition, on its own switch rather than a level of `MSSQL_DEBUG`, whose logging distorts exactly the timing a race needs.

## Verified from both sides

The new scenario in `test_concurrent_reads.cpp` runs 200 transactions of this shape:

| build | result |
| --- | --- |
| with the fix | 200 transactions, **0 failed** |
| with the fix reverted | 200 transactions, **2 failed**, naming `'BEGIN TRANSACTION'` as the holder, plus a leaked connection |

A test that passes on both builds proves nothing, so it was run against the reverted fix before being trusted. It loops rather than trying once, and the comment says why: at ~4% one iteration is meaningless while 200 make a clean run evidence rather than luck.

Also 250 runs of the sqllogictest that first showed it, `sink_reads_own_catalog.test`, in two independent batches — no failures, against 2/60 on main.

| | |
| --- | --- |
| integration suite | 194 cases, 6228 assertions, 5 env-gated skips |
| standalone C++ tests | 31/31 |

## On how this was investigated

The analysis document records five hypotheses that were measured and rejected, including one that was rejected **wrongly** and turned out to be the answer — its probe inserted a sleep *before* the BEGIN, which delays the publishing thread and masks the race instead of widening it. A probe that can only make a race rarer is not a test of it.

It also records the sample-size lesson: at a 4% rate a clean run of 30 happens about a third of the time, which is how two premature "fixed" claims were made earlier. Every number in the document and in these commits says how many runs it came from.


src/tds/tds_connection.cpp
```diff
@@ -56,6 +58,33 @@ static int GetMssqlDebugLevel() {
...
+		return env ? std::atoi(env) : 0;
+	}();
+	return level;
+}
+
+static uint64_t CurrentThreadTag() {
+	return static_cast<uint64_t>(std::hash<std::thread::id>{}(std::this_thread::get_id()) & 0xFFFFFF);
+}
+
+#define MSSQL_CONN_STATE_LOG(...)                   \
```
oluies, line 79: **Medium — this will fail CI.**

The new code doesn't pass clang-format, and the lint job runs `clang-format-14 --dry-run --Werror` over it:

- there is a **double blank line** before `namespace tds` (line 89) — `.clang-format` sets `MaxEmptyLinesToKeep: 1`;
- the two `MSSQL_CONN_STATE_LOG(...)` call sites (≈1264, ≈1275, ≈1282, ≈1342) are wrapped by hand, with argument packing and line lengths that differ from what clang-format-14 produces;
- the macro's backslash alignment here likely differs from the formatter's output too.

**Fix:** `clang-format-14 -i` over `src/tds/tds_connection.cpp`, `src/copy/bulk_load_session.cpp` and `src/connection/mssql_connection_provider.cpp`, then commit the result. (Note homebrew's default clang-format is not 14 and picks a different macro-continuation column — use `/opt/homebrew/opt/llvm@14/bin/clang-format` locally.)

src/connection/mssql_connection_provider.cpp
```diff
@@ -209,7 +214,12 @@ std::shared_ptr<tds::TdsConnection> ConnectionProvider::GetConnection(ClientCont
 	}
 
 	// Transition connection back to Idle (ExecuteBatch left it in Executing state)
-	conn->TransitionState(tds::ConnectionState::Executing, tds::ConnectionState::Idle);
+	conn->TransitionState(tds::ConnectionState::Executing, tds::ConnectionState::Idle, "transaction release");
```
oluies: **Low — the label names the wrong operation.**

This transition marks the *end of `BEGIN TRANSACTION`*, not a transaction release. The label exists precisely so that an `MSSQL_CONN_STATE=1` log can explain a race; here it will make a reader believe a COMMIT or ROLLBACK released the connection in the middle of the lazy pin — which is the opposite of what happened.

**Fix:** something like `"BEGIN TRANSACTION done"`.

specs/078-pinned-connection-race/analysis.md
```diff
@@ -0,0 +1,212 @@
...
+  in its init.
+- **"Guarding `BulkLoadSession::OpenStream` fixes it."** 3/60 → 2/60, i.e.
+  nothing.
+- **"The metadata-query rewrite is the trigger."** main fails at the same rate
+  without it.
+- **"The publish-before-BEGIN window is the race."** `ConnectionProvider::GetConnection`
+  does `SetPinnedConnection(conn)` at line 163 and only then sends
+  `BEGIN TRANSACTION` at line 168, so another thread calling `GetConnection`
+  in between gets the connection while BEGIN is in flight. Structurally it is a
+  real window and it fits every symptom. It is **not** the cause: widening it
```
oluies: **Low — the document now contradicts the code shipped in the same range.**

- The header (line 3-4) still says *"analysis only … No code changes"*, but this PR ships the fix.
- This bullet still lists publish-before-BEGIN as ruled out — *"It is **not** the cause"* — while §4.1 and §4.2 name exactly that as the root cause and describe the fix that landed.
- §5–§7 still say the state-transition logging *"should be built next"*; it is built, in this PR.

The practical cost: a later reader who follows the "do not re-try these" list is pointed away from the actual fix.

**Fix:** update the status line, annotate or strike this bullet so it forwards to §4.1, and mark §5–§7 superseded (or note that the instrument now exists).

src/include/catalog/mssql_transaction.hpp
```diff
@@ -21,6 +21,11 @@ class MSSQLTransactionManager;
 
 class MSSQLTransaction : public Transaction {
 public:
+	//! See pin_mutex_. Held by ConnectionProvider around the lazy pin.
+	mutex &PinMutex() const {
```
oluies: **Low — the documented lock order isn't enforced.**

The comment says `pin_mutex_` is taken before `connection_mutex_`, but nothing holds anyone to it: `PinMutex()` is public and so is `GetConnectionMutex()`, so a caller can take them in the opposite order and deadlock against `GetConnection`.

`std::mutex` is also non-recursive here, so if any future path under the pin lock (a pool acquire, a logging hook) re-enters `ConnectionProvider::GetConnection` for the same transaction, it deadlocks silently. That is the same shape as the recursive-mutex problem that was hit on the reverted guard in #355.

**Fix:** narrow the access — make `ConnectionProvider` a friend, or expose a scoped `LockForPin()` helper — and state the lock-order rule on `GetConnectionMutex()` as well, not only here.

oluies (dismissed):
Approving — the root cause and the fix look right, and the connection-holder instrumentation is a genuinely good addition. Four comments inline, one of which will fail CI as-is (clang-format), so please take that one before merge.

**Summary of findings**

| Severity | Where |
| --- | --- |
| Medium | `src/tds/tds_connection.cpp` — new code doesn't pass `clang-format-14`; the lint job will fail |
| Low | `mssql_connection_provider.cpp` — state-transition label names the wrong operation |
| Low | `specs/078-pinned-connection-race/analysis.md` — document contradicts the code shipped alongside it |
| Low | `mssql_transaction.hpp` — documented lock order is unenforced |


---
*Posted from a [roborev](https://github.com/) review (job 1639), relayed via Claude Code.*

.github/workflows/concurrency-tests.yml
```diff
@@ -4,7 +4,7 @@ name: Concurrency Tests
 # ClientContextState anchor fix under multi-threaded stress. The Makefile
 # target `test-concurrent-reads` is added in PR #127 but no CI workflow
 # invoked it — the PR author ran it locally and pasted soak results into
-# the PR. This job reproduces the full scenarios 1-8 sweep on every PR
+# the PR. This job reproduces the full scenarios 1-9 sweep on every PR
```
oluies, line 7: **Low — the new label is as inaccurate as the old one, just differently.**

The workflow now says "scenarios 1-9" (here, and at :36 and :213), but `test/cpp/test_concurrent_reads.cpp` has **no Scenario 4** — its comments number them 1, 2, 3, 5, 6, 7, 8, 9. So the job runs eight distinct scenarios, and "1-9" implies nine.

**Fix:** either renumber the new pin-race scenario as 4 in the test file and keep the "1-9" labels, or describe the job without a numeric range so it stops needing maintenance every time a scenario is added.

test/cpp/test_concurrent_reads.cpp
```diff
@@ -946,6 +1049,8 @@ int main() {
 		ok &= scenario_concurrent_mixed_reads(cfg, 4, 50);
 		ok &= scenario_concurrent_mixed_reads(cfg, 8, 25);
 		ok &= scenario_concurrent_attach(cfg, 4);
+		// Scenario 9 (issue #356): the lazy pin of a transaction's connection.
+		ok &= scenario_transaction_pin_race(cfg, 200);
```
oluies, line 1053: **Low — 200 iterations is not enough to make this a regression guard.**

`specs/078-pinned-connection-race/analysis.md` presents this scenario as the guard for #356: 0/200 failures with the fix, 2/200 with it reverted.

But 2/200 is a per-transaction failure rate of about 1%, and at that rate a 200-iteration run of the *reverted* code passes cleanly about 13% of the time:

```
0.99 ^ 200 ≈ 0.134
```

So CI would stay green on a regression of #356 roughly **one run in seven** — and this is the same class of small-sample reasoning that the revert commit on #355 correctly called out as luck.

**Fix:** raise the iteration count — about 1000 brings a clean run on the reverted code under 0.01%. Alternatively widen the window deterministically with a test hook placed *after* BEGIN is sent (the §4 correction explains why a sleep before BEGIN tests nothing). Either way please update the numbers quoted in the doc, since they are what a reader will trust.

specs/078-pinned-connection-race/analysis.md
```diff
@@ -0,0 +1,230 @@
+# Analysis 078 — who touches the pinned connection, and when (issue #356)
+
+**Status:** written as analysis-only, on `recon/356-pinned-connection-race`
+from `main` `b84e259`, after four wrong fixes that were each measured and
+withdrawn — the point was to stop guessing and map the actual flow first. It
+worked: §4.1 names the cause and the branch now ships the fix and the
+instrument alongside this document. Sections are left as they were written,
```
oluies, line 7: **Low — the status paragraph and the diff disagree.**

This says the sections "are left as they were written, with the corrections marked". The same diff rewrites the rejected-hypothesis bullet in §4 in place ("It was rejected because… That probe could only make the race rarer"), so that section was changed rather than left as written.

It matters because the stated aim of the document is to keep the original (wrong) reasoning visible so nobody re-walks it.

**Fix:** either keep the original sentence struck through with the correction below it, or change the status line to say the corrections were edited into the text.

oluies (approved):
Re-approving — all four findings from the previous round are properly fixed, and deleting `GetConnectionMutex()` outright once it turned out to have no callers is a better answer than the one I suggested.

Three Low comments on the new commits. None blocks the merge, but the second is worth acting on before this lands, because it decides whether the regression guard actually guards anything.

| Severity | Where |
| --- | --- |
| Low | `concurrency-tests.yml` — "scenarios 1-9" is inaccurate in a new way |
| Low | `test_concurrent_reads.cpp` — 200 iterations lets a regression pass ~1 run in 7 |
| Low | `analysis.md` — status line says sections were left as written; §4 was edited |


---
*roborev job 1643 (re-review of the new commits only), relayed via Claude Code.*

PR #359, opened 2026-09-17
test: a logical guard for the pinned-connection race, and the three Lows from #357's review (#356)

VGSML:
Follow-up to #357, which merged with three Low findings open. Two are naming; the third is the one this PR is really about.

## The regression test was statistical, and the arithmetic behind it was wrong

`scenario_transaction_pin_race` ran 200 COPY statements and its comment called a clean run evidence, citing a 4% rate. That rate is per *run of the SQL test*; the scenario's own measured rate is 2 in 200, about 1% per transaction — and at 1%, broken code passes 200 iterations clean **13%** of the time. CI would miss a regression roughly one run in seven. Raising the count does not repair a test of that kind, it only makes it slower: 1000 iterations measured at 15 minutes.

## The guard: test the API's contract, not the symptom

`test/cpp/test_transaction_pin.cpp` links the extension statically and calls `ConnectionProvider::GetConnection` directly, eight threads released from a barrier into one transaction's lazy pin. It asserts what the critical section promises:

- every caller gets the **same** connection;
- it is **Idle** when handed out — BEGIN has finished on it;
- a server transaction is open on it (`@@TRANCOUNT = 1` through the ordinary path);
- **one** pool connection is active during the transaction, **none** after ROLLBACK.

Two release patterns, because the two things the lock prevents show up on different timings: *all at once* (without the lock every caller passes "is there a pin yet?" before the first has acquired — the double acquire) and *staggered* (the rest arrive during the first caller's BEGIN round trip — the mid-BEGIN handout, which is #356 itself).

**Measured both ways, with no hook in production code.** Against the pre-fix provider, the all-at-once round fails immediately with 8 active connections and 7 leaked; the staggered round fails with every follower handed the connection in state `Executing`. With the fix: 20/20 rounds, under a millisecond each. Logical rather than statistical: the window is a network round trip, the callers' start skew is microseconds, so every caller is in the window every round.

The COPY loop stays at 20 iterations as the end-to-end smoke through DuckDB's own scheduling; its comment now says which test is the guard.

## The two naming Lows

- The job name loses its scenario range: there is no scenario 4, so "1-9" promised nine of eight. The TestDB comment in the same file was wrong in both halves (scenario 3 uses TestDB; 4 does not exist) and now states the truth without a range.
- `analysis.md`'s status line said the sections were left as written; the diff edited §4 in place. It now says so.

`make test-transaction-pin` — CI runs it against the debug tree the concurrency job already builds; locally on macOS use `PIN_TEST_BUILD=release`, because a debug (ASan) build deadlocks at process start on this platform.


oluies:
## roborev review — 2 findings (1 Medium, 1 Low)

Branch diff `1c980a4..588b92f` (3 commits) against `main`, roborev job 1655, agent `claude-code`.

### Medium — `Makefile`: `PIN_TEST_TRIPLET` and friends are expanded at parse time

`PIN_TEST_TRIPLET := $(shell ls $(PIN_TEST_VCPKG) ...)`, and the `PIN_TEST_VCPKG_LIB` / `PIN_TEST_LIBS` definitions that depend on it, use `:=`, so make evaluates them when it reads the Makefile — before the `$(PIN_TEST_BUILD)` prerequisite has run. On a tree where `build/<type>/vcpkg_installed` does not exist yet (fresh clone, or right after `make clean`), `make test-transaction-pin` computes an empty triplet, builds the tree, and then tries to link against `build/debug/vcpkg_installed//debug/lib/libssl.a`. The next run succeeds, so the failure looks random. CI only avoids it because an earlier step builds the debug tree first.

Note this sits next to what 588b92f fixed: that commit corrected *which* entry the `ls` picks, this is about *when* the `ls` runs.

*Fix:* define `PIN_TEST_TRIPLET`, `PIN_TEST_VCPKG_LIB` and `PIN_TEST_LIBS` with recursive `=` so they expand when the recipe runs, or compute the triplet inside the recipe. Also add `test-transaction-pin` to `.PHONY`.

### Low — `test/cpp/test_transaction_pin.cpp`: the staggered release uses a fixed 100 µs sleep

The staggered pattern assumes thread 0 is already inside its BEGIN round trip 100 µs after `go`, and nothing checks that — thread 0 still has to be scheduled, finish its yield loop and acquire from the pool. On a loaded 2-vCPU runner that may not be enough, and if the followers arrive first they simply behave like the all-at-once case. The fixed code still passes, but the pattern then stops checking for the mid-BEGIN handout it was added to catch, so for that defect the guard is statistical again — which is what this change set out to remove.

*Fix:* release the followers on observed state rather than a fixed sleep — poll (bounded) until the pool reports one active connection, or until `txn.HasPinnedConnection()` / the connection's state shows the BEGIN is in flight. Or count and report the rounds in which the followers actually overlapped the BEGIN.

VGSML:
Both taken, in one commit on top of a merge with `main`.

**Medium — parse-time expansion.** Right on both counts: `588b92f` fixed *which* entry the `ls` picks, this was *when* it ran. `PIN_TEST_VCPKG`, `PIN_TEST_TRIPLET`, `PIN_TEST_VCPKG_LIB` and `PIN_TEST_LIBS` are recursively expanded now (`=`), with the comment saying why, and `test-transaction-pin` is `.PHONY`. Checked with `make -n`: the link line resolves to `vcpkg_installed/arm64-osx/lib/libssl.a` at recipe time.

**Low — the 100 µs sleep.** Replaced by your first suggestion: the followers are released when the pool reports thread 0's connection active (`active_connections == before + 1`, bounded by 5 s) — under the fix that happens inside the critical section before BEGIN is sent (`LockForPin` → `Acquire` → `BEGIN` → `SetPinnedConnection`, checked in the provider). Whether the pin is already published at that moment says whether the followers overlapped the BEGIN; the test counts those rounds, prints `N of 10`, and fails on zero — so the guard is asserted against the window it exists for rather than hoped at. Three local runs: 10/10, 10/10, 9/10.

oluies:
## roborev review of 41ae89a — 2 findings (2 Low)

Follow-up commit only (`41ae89a`), roborev job 1657, agent `claude-code`. The Makefile fix is correct: the vcpkg paths now expand when the recipe runs. Releasing the followers on observed pool state addresses the earlier Low. Both remaining points are about whether the new `overlapped_begin` counter measures what it claims.

### Low — `test/cpp/test_transaction_pin.cpp`, staggered wait loop (~173–180): a timed-out wait counts as an overlap

When the 5 s deadline expires, the loop falls through without recording it. If thread 0 never got a connection by then (it was descheduled for a long time or `Acquire` was slow), `HasPinnedConnection()` is still false, so the round is counted in `overlapped_begin`. The "at least one overlap" check can then pass when nothing overlapped. A throw from thread 0 is caught later as an error. A slow `Acquire` that finishes after the deadline is not caught.

*Fix:* record whether the wait timed out. Count a round only when the condition was actually observed (`active >= before + 1 && !HasPinnedConnection()`), and consider logging rounds that timed out.

### Low — `overlapped_begin`: the signal depends on the ordering the test guards

"Active connection, no pin published" means "inside BEGIN" only for the fixed provider, which runs Acquire → BEGIN → `SetPinnedConnection`. The regressed provider pinned right after `Acquire`, before BEGIN, so on it this window is almost always closed. Against a regression the staggered rounds therefore record roughly zero overlaps. If the double-acquire and state checks happen not to fire on that run, the test fails with "no staggered round released its followers while the BEGIN was in flight". That message blames the test's timing rather than naming the regression. The count is also taken when the followers are released, not when they reach `LockForPin`, so it can overstate the real overlap.

*Fix:* tie the observation to connection state rather than to when the pin is published, e.g. check that thread 0's connection is `Executing` when the followers are released. At minimum, say in the failure message that a regressed pin-before-BEGIN ordering also produces zero.

VGSML:
Both taken.

**Timed-out wait counted as an overlap.** Right — the loop fell through with the pin unpublished and scored the round. Now the wait records whether thread 0's connection was actually observed active; a round counts only when it was **and** the BEGIN was in flight at that moment. Timed-out rounds are logged (`thread 0 had no pool connection after 5 s, followers released blind`) and reported beside the count.

**The signal depended on the ordering the test guards.** Tied to connection state now, per your suggestion: "BEGIN in flight" is *either* the pin not yet published (the fixed ordering, `LockForPin → Acquire → BEGIN → SetPinnedConnection`) *or* the pinned connection still `Executing` (the regressed ordering, pin before BEGIN) — so a regressed provider does not produce a zero count that blames timing. The zero-overlap failure message names both readings anyway, with the timed-out count, and says the double-acquire and state checks above are the ones that name a mid-BEGIN handout. Three local runs: `10 of 10 (0 timed out waiting for thread 0)`.

The count is still taken when the followers are *released*, not when they reach `LockForPin` — the comment says so; the next stop is the lock, and nothing between the release and it touches the connection.

oluies:
## roborev review of ce364e5 — Pass, no findings

Follow-up commit only (`ce364e5`), roborev job 1661, agent `claude-code`.

This addresses both Lows from the `41ae89a` review. A staggered round now counts as an overlap only when thread 0 was actually seen holding a pool connection with its BEGIN still in flight, and that state is recognised under either pin ordering: the pin not yet published, or the pinned connection still `Executing`. Rounds that reach the 5 s deadline are logged and reported rather than counted.

Nothing further from me on this PR.

Issue #363, opened 2026-09-17
vector(n) columns (SQL Server 2025) are unreadable through the catalog — declared BLOB, sent as varchar JSON; FLOAT[n] read/write + VECTOR_DISTANCE pushdown as its own spec after v0.3.0


VGSML:
## Today: a `vector(n)` column makes its table unreadable through the catalog

SQL Server 2025 (`mcr.microsoft.com/mssql/server:2025-latest`), `CREATE TABLE dbo.vec_probe (id int PRIMARY KEY, emb vector(3))`:

- `sys.columns` reports system type `varbinary`, user type `vector`, `max_length` 20 (8-byte header + 3 × float32), so the catalog declares the column **BLOB**;
- `sp_describe_first_result_set` reports **`varchar(max)`**: to a client without native vector support the server sends a JSON array (`[1.0000000e+000,0.0000000e+000,0.0000000e+000]`);
- `SELECT id, emb FROM m.dbo.vec_probe` fails: `MSSQL: column 'emb' is declared BLOB in the catalog but SQL Server returned TDS type 0xA7`.

Same class as #296 (`rowversion`): a system type that is not what the wire carries. Owner's decision 2026-09-17: **its own spec, after v0.3.0**.

## What the spec would cover

1. **Read** `vector(n)` as `FLOAT[n]` — parsing the JSON text the server sends today, or the native binary form behind the TDS `VECTORSUPPORT` feature extension (to investigate: whether the server then sends the 8-byte header + float32 payload, and what the LOGIN7 negotiation costs on servers without it).
2. **Write** a `FLOAT[n]` value (INSERT, UPDATE, COPY/BCP) as a JSON array literal or parameter cast to `vector(n)`; CTAS/COPY target type from a `FLOAT[n]` source column.
3. **Pushdown** (spec 079's writer): `array_cosine_distance` / `array_distance` / `array_inner_product` (and `array_negative_inner_product`) with a constant array → `VECTOR_DISTANCE('cosine' | 'euclidean' | 'dot', [col], CAST(@p AS vector(n)))`, so `ORDER BY array_cosine_distance(emb, $q) LIMIT k` runs as an exact kNN on the server (`SELECT TOP k … ORDER BY VECTOR_DISTANCE(…)`; measured to work in ORDER BY). No collation question on floats.
[...]

Issue #362, opened 2026-09-17
ORDER BY pushdown: the agreement test is "binary AND UTF-8-coded", not "_BIN2" — a code-page varchar under _BIN2 orders code-page bytes (spec 061 correction)


VGSML:
## The claim to correct

Spec 061 (`specs/061-collation-aware-order-pushdown/spec.md` § 2) says a binary collation — `_BIN` or `_BIN2` — orders by code point, which is DuckDB's order, and makes "the key is a string column under a `_BIN2` collation, or not a string at all" the test for pushing ORDER BY / TOP; its § 3 forces `COLLATE Latin1_General_BIN2` on the key when the column's collation is not binary. Both are wrong for `varchar`.

## Measured

`varchar(20) COLLATE SQL_Latin1_General_CP1_CI_AS` holding `z é Š a € A ab␣ ab`, SQL Server 2025 (docker):

| ordered by | result |
|---|---|
| the column's own collation | `€ Š a A ab␣ ab é z` — linguistic |
| `COLLATE Latin1_General_BIN2` (what 061 emits) | `A a ab␣ ab z € Š é` — **the code page's bytes**: € 0x80 < Š 0x8A < é 0xE9 |
| `COLLATE Latin1_General_100_BIN2_UTF8` | `A a ab␣ ab z é Š €` — **UTF-8 bytes** |
| DuckDB | `A a ab ab␣ z é Š €` |

`_BIN2` on a `varchar` column compares the bytes the value is stored in — the code page's — not code points; only a `_UTF8` binary collation transcodes first, after which byte order is code-point order on every plane. An `nvarchar` under `_BIN2` compares UTF-16 code units (exact through the BMP, the above-BMP divergence 061 already documents), and a UTF-8 collation does not change that: it applies to `varchar` only. The one residual on both is `ab` / `ab␣`, which the server calls equal (padding), so their order is a tie it breaks as it likes.

## The correction

[...]

Issue #361, opened 2026-09-17
Filter pushdown: a non-ASCII constant against a varchar column on a SQL_ collation loses the index seek (declared nvarchar)


VGSML:
## What happens

Since spec 076 the constants of a pushed filter travel as `sp_executesql` parameters declared from the column (`FilterEncoder::DeclarationForColumn`, `src/table_scan/filter_encoder.cpp`). A constant with any **non-ASCII** character is declared `nvarchar` whatever the column's collation. The reason is right in its own case (#321): a `varchar` variable takes the *database's* code page, so a Cyrillic constant against a UTF-8 column arrived as `?`.

On the installation-default collation family it costs the index seek. Measured on SQL Server 2025 (docker), table of 20 000 rows, nonclustered index on `a varchar(20)`, `SET STATISTICS PROFILE ON`:

| column collation | `sp_executesql` declaration | plan |
|---|---|---|
| `SQL_Latin1_General_CP1_CI_AS` (installation default) | `@p0 nvarchar(20)` = `N'ñu'` — what 076 emits | **Index Scan**, `CONVERT_IMPLICIT(nvarchar(20), a)` on the column side |
| `SQL_Latin1_General_CP1_CI_AS` | `@p0 varchar(20)` = `'ñu'` | **Index Seek** |
| `Latin1_General_CI_AS` (Windows collation) | `@p0 nvarchar(20)` = `N'ñu'` | Index Seek through `GetRangeThroughConvert` — unaffected |

So on a `SQL_` collation every filter whose constant carries an accented Latin-1 character (`'ñu'`, `'Müller'`, `'café'`) reads the whole index; on a Windows collation the optimizer derives a range through the convert and nothing is lost. The literal path (`mssql_scan_parameterize_filters = false`) has always spelled such constants `N'…'` and scans the same way.

## Remedy

Declare `varchar` when the constant is representable in **both** the column's and the database's code page — for the Latin-1 range on CP1252 that is a table lookup, and it is the case that matters — and `nvarchar` otherwise, exactly as now. The column's collation and the database collation are already in the metadata cache. Independent of the remote-pushdown work (spec 079); worth its own small PR with a plan assertion (`STATISTICS PROFILE` / `SHOWPLAN`) in the test, since a lost seek is invisible to a row-count test.

Found during the 065/067 revalidation, `specs/065-dml-pushdown-recon/revalidation-2026-09-17.md` § 8.2.

PR #364, opened 2026-09-17
spec: 079 SELECT pushdown + 080 DML pushdown on DuckDB 2.0's rewriter; 065/066/067 closed (spec-only)

### LINKED ISSUE #140: UPDATE fails on tables without primary key (Binder Error)
## Description

After attaching a SQL Server database using the mssql extension, `UPDATE` statements fail with a `Binder Error` if the target table has no primary key, even when the operation is semantically valid.

## Environment

- OS: Windows 11
- DuckDB version: 1.5.2
- SQL Server: (local, port 1433)

## Steps to Reproduce

1. Load and attach the extension:
   ```sql
   ATTACH 'Server=localhost,1433;Database=mydatabase;User Id=myuser;Password=mypassword' AS sqlserver (TYPE mssql);
   USE sqlserver;
   ```

2. Create a table **without** a primary key and insert some rows:
   ```sql
   CREATE TABLE test (id int);
   INSERT INTO test VALUES (1), (2);
   ```

3. Try to update a row:
   ```sql
   UPDATE test SET id = 3 WHERE id = 1;
   ```

## Expected Behavior

The `UPDATE` statement should execute successfully. SQL Server itself does not require a primary key to perform `UPDATE` operations.

## Actual Behavior

The query fails immediately with:

```
Binder Error: MSSQL: UPDATE/DELETE requires a table with a primary key. Table 'main.test' has no primary key.
```

## Notes

- The restriction is enforced by the extension's binder, not by SQL Server itself.
- Tables with a primary key are not affected.
- `SELECT` and `INSERT` work correctly regardless of the presence of a primary key.
- The error appears to originate in [`MSSQLTableEntry::BindUpdateConstraints`](https://github.com/hugr-lab/mssql-extension/blob/912816a1a1dee1fcc47352a72cea5b40777de3f5/src/catalog/mssql_table_entry.cpp#L223), which explicitly throws when no primary key is found.

VGSML:
Spec-only: no code, no tests.

## What to review — two specs

- **`specs/079-select-pushdown/spec.md`** — SELECT pushdown through DuckDB 2.0's `RemotePushdownOptimizer`: the T-SQL writer as the single vocabulary (`SupportsPushdown` is the writer's dry run, so a construct is supported iff it has an exact rendering — no second veto list), the construct table with the measured aggregate forms, `mssql_scan_params` as the execution vehicle (nothing new to run), the string rule (below), the scan path and `MSSQLOptimizer` kept as the fallback with a shared vocabulary, `mssql_remote_pushdown` default on, an agreement suite that runs every shape with the setting on and off.
- **`specs/080-dml-pushdown/spec.md`** — UPDATE / DELETE / INSERT … SELECT / CTAS as one server statement through the same writer, native string semantics, and the staged fallback for what the writer vetoes: 067's match-key ladder (PK → usable unique index via spec 077's `ChooseRowIdKey` → all columns NULL-safe) on a `##stage` table, plus the 066 remainder (scans feeding a DML materialise in a transaction, `defer_execution_` retires). Closes #140 fully.

## History, kept on purpose

- `specs/065-dml-pushdown-recon/revalidation-2026-09-17.md` — the research record both specs are written from: the rewriter mapped from the pinned source (§ 5, § 9.1), strings on the server measured (§ 8: collation order vs DuckDB's, which predicate forms keep the seek, trailing spaces and what strictness costs), aggregates and functions (§ 9.2), the decision trail. Issues #361, #362, #363 point into it.
- `specs/065-…/spec.md`, `066-…/spec.md`, `067-…/spec.md` (+ `research.md`, `join-agg-pushdown.md`) — the reconnaissance, marked **CLOSED** in their headers with a pointer to where each part went.

## Decisions recorded (owner, 2026-09-17)

- 065 is retired in favour of the core rewriter; SELECT first (079), DML second (080); `MSSQLOptimizer` and the scan's filter pushdown stay (the rewriter cannot reach a subquery under a non-pushable node, and a local view blocks it).
- DML through the rewriter is **native** — a SQL Server user expects SQL Server's DELETE.
- `vector` (SQL Server 2025) is its own spec after v0.3.0 — #363.
- Not 100 % pushdown: one writer, table-driven vocabulary, veto by default, fallback path.

## One thing to confirm in review — 079 D4, the string rule

Proposed: **sets are the server's** (`=`, `<>`, `IN`, `NOT IN`, `LIKE`, `NOT LIKE`, ranges, GROUP BY / DISTINCT / join keys — the column's collation and padding, no client re-check, the same in the scan path, the rewriter and DML); **orders are DuckDB's** (ORDER BY / MIN / MAX / window ORDER BY on a string key only under a binary, UTF-8-coded collation — #362 — else no pushdown). Measured on the shipped scan: `<>` and `NOT LIKE` are **already pushed natively** (`v <> 'ab'` → 2 rows on `ab, ab␣, AB, abc, x`, not DuckDB's 4); only `NOT IN` runs client-side, because the encoder has no `COMPARE_NOT_IN` case. Native on both sides is also the only rule under which `=` and `<>` partition a table. @oluies — you were in the original `<>` discussion, so this is asked here rather than assumed.

## Issues filed from the measurements

- #361 — spec 076 declares a non-ASCII constant `nvarchar` and loses the index seek on `SQL_` collations (§ 8.2).
- #362 — spec 061's agreement test is "binary AND UTF-8-coded", not `_BIN2` (§ 8.1).
- #363 — `vector(n)` columns are unreadable through the catalog today (§ 9.5).


oluies (commented):
Review of the two new specs, against the pinned DuckDB (`duckdb @ d673cf9a`), this repo's shipped source, and `duckdb/duckdb-mysql` — which implements the same `RemotePushdownOptimizer` contract and is the only shipping consumer of it, so it's useful as a precedent throughout.

The direction is sound and the research record is genuinely strong. The findings below are about places where a load-bearing claim doesn't match the source, or where a faithful implementation of the text would be defective.

**One blocking item:** pushed DML bypasses read-only enforcement (`080:55`). DuckDB's check is bind-time via `modified_databases`, and the rewrite turns the DML into a SELECT before the binder runs, so both DuckDB's guard and `CheckWriteAccess` are skipped. duckdb-mysql had to add this by hand in its own vehicle's bind.

**Worth settling before implementation:** the D5 agreement invariant vs. the scan's column rewrites (`079:153`, `079:89`); the NULL-ordering claim, which describes emulation that doesn't exist (`079:94`); the division-by-zero justification, which is wrong on the pin (`079:90` — verified: `select 1/0` returns `inf`, not NULL); the kill switch, which can't be read in `Supports` and must not gate `IS_REMOTE` (`079:157`, `079:172`); interim CTAS schema degradation in 079-alone (`079:101`); a missing veto for nested ORDER BY (`079:97`); and #350 being cited in the present tense while unmerged (`080:84`).

**Checked and found fine**, for the record, since these were the obvious places to worry:
- The count vehicle does **not** imply execute-at-bind — a static single-BIGINT shape needs no describe. Flagged only as "say so explicitly", because the adjacent `MSSQLScanBind` fallback is a trap.
- ORDER BY losing `LogicalOrder` is pre-existing and upstream-intended: `mssql_optimizer.cpp:509-510` already deletes it today, and duckdb-mysql ships the same property with tests asserting both ordered rows and no ORDER_BY operator.
- Materialisation-at-init without an interrupt check is pre-existing (spec 075) and doesn't fire for a solo pushed statement.
- `##stage` cleanup is inherited from `BulkLoadSession`'s destructor contract.
- Literal escaping is correctly routed through `sp_executesql` parameters.
- No CHANGELOG / DATAMODEL.md entry is expected here — every prior spec-only commit in this repo touches neither, and CLAUDE.md's settings table is populated when a setting ships in code.

**On the document set** (complexity, not correctness): the revalidation record's §9 restates 079/080's design nearly table-row-for-row, §5 and §9.1 explain the same mechanism twice, and §2 prescribes amendments to a spec this PR closes. Landing 065/066/067 already-CLOSED is deliberate per the description, so I'd leave them — but `join-agg-pushdown.md` carries a rule that now *contradicts* 079 (see the inline note), which is worth resolving rather than leaving as history.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+| `DELETE FROM t [USING u …] WHERE p` | `DELETE t FROM [s].[t] AS t [JOIN …] WHERE p` | same |
+| `INSERT INTO t [(cols)] SELECT …` (both sides remote) | `INSERT INTO [s].[t] ([cols]) SELECT …` | the SELECT through 079; an explicitly named identity column brackets the statement in `SET IDENTITY_INSERT` (077 W2); `INSERT … VALUES` stays on the shipped path (it carries no remote SELECT to win by) |
+| `INSERT INTO local SELECT … FROM remote` | the SELECT alone | the rewriter does this by itself (`push_select_only`); nothing to add |
+| `CREATE TABLE remote AS SELECT … FROM remote` | our CREATE (table kind, collation, lengths — the WITH options keep meaning), then the pushed `INSERT … SELECT` | not `SELECT INTO`; needs `EXECUTE_STATEMENT` claimed for CREATE TABLE AS only, and `SupportsPushdown(const SQLStatement &)` saying yes for that shape alone |
+| `RETURNING`, `MERGE INTO`, `ON CONFLICT` | — | veto; the shipped path handles what it handles today |
+
+Execution: `RemoteExecute` returns a ref to the DML form of the vehicle —
```
oluies: **Blocking — a `READ_ONLY` attach stops protecting the catalog.**

DuckDB's read-only enforcement is entirely bind-time: `ClientContext::CheckIfPreparedStatementIsExecutable` iterates `statement.properties.modified_databases` (`client_context.cpp:617-631`), and that set is populated only by `RegisterDBModify` in `bind_update.cpp:196` / `bind_delete.cpp:43` / `bind_insert.cpp:629` / `bind_create.cpp:320`.

`FinishPushdown` replaces the `UpdateStatement` with a `SelectStatement` **before binding** (`remote_pushdown_optimizer.cpp:1850-1870`), so those binders never run, `modified_databases` stays empty, and `StatementProperties::IsReadOnly()` reports the statement as read-only. The extension's own second guard is bypassed by the same mechanism: `MSSQLCatalog::CheckWriteAccess` is called only from `PlanInsert`/`PlanUpdate`/`PlanDelete`/`PlanCreateTableAs` (`src/catalog/mssql_catalog.cpp:619,741,750,789`), and D2 says those are reached only on the veto path. The CTAS row is covered too.

Net: `ATTACH … AS ms (READ_ONLY); UPDATE ms.t SET x = 1;` executes on the server.

The MySQL extension hit this and fixed it in exactly the layer this spec leaves open — `MySQLQueryBind` computes `tran_restrict_dml = transaction.GetAccessMode() == AccessMode::READ_ONLY` and throws `PermissionException` when the prepared statement yields no result fields (`mysql_scanner.cpp:312,327,355-360`), plus `START TRANSACTION READ ONLY` (`mysql_transaction.cpp:71`), with `test/sql/attach_read_only.test` pinning it.

Suggest: `SupportsPushdown` refuses every DML/CTAS node when `IsReadOnly()`, **and** a guard in the count vehicle; plus a W5 case.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+The scan's filter pushdown (`FilterEncoder`, `pushdown_complex_filter`) and
+`MSSQLOptimizer`'s ORDER BY / TOP pushdown are **kept**: they are the only
+mechanisms that reach a mixed-catalog plan, a subquery under a non-pushable
+node, or a local view — the rewriter cannot (§ 9.1). They walk bound trees,
+the writer walks parsed ones, so they cannot share a renderer; what they
+share is the vocabulary: **one** function table (`function_mapping.hpp`
+consulted by both, with `NOT IN` added so it stops being the scan's one
+strict exception), **one** collation predicate (`OrdersLikeDuckDB`, #362,
+consulted by `MSSQLOptimizer` and D4), **one** literal/parameter spelling
+(#361). A query cannot answer differently depending on which path took it.
```
oluies: **This invariant is not deliverable from the three things D5 lists as shared.**

The catalog scan rewrites the *column*, not just the function: `BuildColumnExpression` (`src/table_scan/table_scan.cpp:165-209`, called at `:318`, `:343`, `:394`) emits `CAST([c] AS NVARCHAR(n))` for a non-UTF-8 CHAR/VARCHAR/TEXT column and `NVARCHAR(MAX)` for `is_cast_required` types. `mssql_scan_params` applies neither — CLAUDE.md's own `mssql_warn_non_utf8_collation` row states it: a non-UTF-8 varchar "never reaches the binary kernel on the catalog path anyway (`NeedsNVarcharConversion` casts it server-side)", while a raw scan hands the bytes over verbatim.

`SQL_Latin1_General_CP1_CI_AS` is the SQL Server installation default, so this is the common configuration. And 079 *widens* it: `SELECT * FROM ms.t` becomes a whole-statement push, so every simple table read leaves the catalog-scan path — today you have to hand-write `mssql_scan()` to get the raw-byte path.

Note the writer can't just reuse the helper: `BuildColumnExpression` is `static` in `table_scan.cpp` and drives off `MSSQLColumnInfo` on **bound** input, while the writer walks parsed trees. It's a second implementation either way.

Suggest: add column rendering to the shared vocabulary, or veto in `SupportsPushdown(const TableRef &)`.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+literals instead, for plan comparison. Identifiers are bracket-quoted; the
+schema is the one the catalog resolved (the default schema is the catalog's,
+#322 when it lands).
+
+The vocabulary, from § 9.2 — each row is a rendering the agreement suite
+(W5) exercises:
+
+| construct | T-SQL | rule |
+|---|---|---|
+| FROM base table, remote view | `[schema].[name]` | one catalog; a view pushes, the server expands it |
```
oluies: **The base-table row drops four server-side column rewrites — one of them turns a working query into an error.**

`BuildColumnExpression` (`src/table_scan/table_scan.cpp:165-209`) unconditionally applies, on every catalog-scan SELECT list: geometry/geography → `.STAsBinary()`; `is_cast_required` (hierarchyid, sql_variant, CLR UDT) → `CAST(… AS NVARCHAR(MAX))`; and `NeedsNVarcharConversion` → `CAST(… AS NVARCHAR(n))`. Neither 079 nor 080 mentions any of them (grep for `NeedsNVarcharConversion|STAsBinary|is_cast_required|geometry|hierarchyid` over both: zero hits), and D2 specifies only "column refs".

Two consequences:
- non-ASCII values in a non-UTF-8 text column arrive as code-page bytes in a DuckDB VARCHAR;
- a `geometry`/`geography`/`hierarchyid`/`sql_variant` column makes the pushed statement **fail at bind** — `DescribedTypeToTdsMetadata` maps them to `TDS_TYPE_UDT`/`TDS_TYPE_SQL_VARIANT` (`src/mssql_functions.cpp:244-246`) and `TypeConverter::GetDuckDBType` throws `InvalidInputException("Unsupported SQL Server type …")` (`src/tds/encoding/type_converter.cpp:143-148`). That table reads fine today, and `RemoteExecute` cannot decline.

Both are forbidden by acceptance 3 and by W5's "run twice and compare".

(ntext/image are **fine** — decoded natively since spec 055, `type_converter.cpp:132-140`; the `table_scan.cpp:186` comment citing #197 is stale.)

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+| GROUP BY + aggregates | positional and alias references **expanded** (T-SQL has neither: errors 164 / 207); `COUNT_BIG(*)` / `COUNT_BIG(x)` / `COUNT_BIG(DISTINCT x)`; `SUM(CAST(int AS bigint))`; `AVG(CAST(x AS float))`; `MIN` / `MAX` (`bit` → `CAST(b AS tinyint)`; strings under D4); `STRING_AGG(CAST(x AS nvarchar(max)), <literal>) WITHIN GROUP (ORDER BY …)`; `STDEV` / `STDEVP` / `VAR` / `VARP`; `agg(x) FILTER (WHERE c)` → `agg(CASE WHEN c THEN x END)` | veto `first` / `any_value`, `arg_*`, `median` / `quantile` / `mode`, `approx_count_distinct` (a different estimator), `list` / `array_agg` / `histogram`, `bit_*`, `corr` / `covar_*` / `regr_*`, a non-literal `string_agg` separator, GROUP BY ALL / ROLLUP / CUBE / GROUPING SETS |
+| JOIN | INNER / LEFT / RIGHT / FULL / CROSS with ON; USING expanded to ON | the ON condition is any expression the writer renders — non-equi, OR, functions, subqueries — nothing join-specific; veto NATURAL, SEMI / ANTI, ASOF, POSITIONAL, LATERAL |
+| ORDER BY | `ORDER BY`, NULL placement emulated as `MSSQLOptimizer` does today (`CASE WHEN x IS NULL …` unless the column is NOT NULL) | string keys under D4 only |
```
oluies: **No such emulation exists — `MSSQLOptimizer` refuses this case, it does not emulate it.**

`IsNullOrderCompatible` (`src/table_scan/mssql_optimizer.cpp:271-284`) returns true only when the column is NOT NULL or the requested null order already matches SQL Server's, and `:315-320` `break`s out of the ORDER BY loop otherwise. `grep "CASE WHEN" src/` returns zero hits.

And the refused case is the common one: `default_null_order` defaults to `NULLS_LAST` (`duckdb/src/include/duckdb/main/settings.hpp:1006`) against SQL Server's ASC → NULLS FIRST, so a bare `ORDER BY x` on a nullable column is precisely what never pushes today. The repo already records this in a comment written to correct an earlier wrong claim — `src/table_scan/table_scan.cpp:1219-1221`.

W2 (line 181) lists "NULL-order emulation" as new work, which is right and contradicts this line.

This also undercuts W1's premise at line 174-176 ("the shapes `MSSQLOptimizer` handles today, so W5 can compare the two paths on the same statements"): for `ORDER BY <nullable>` there is nothing to compare against, because the old path declines it.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+schema is the one the catalog resolved (the default schema is the catalog's,
+#322 when it lands).
+
+The vocabulary, from § 9.2 — each row is a rendering the agreement suite
+(W5) exercises:
+
+| construct | T-SQL | rule |
+|---|---|---|
+| FROM base table, remote view | `[schema].[name]` | one catalog; a view pushes, the server expands it |
+| projections, WHERE, HAVING, ON | the expression writer | column refs, constants, `+ - *` (overflow errors on both sides), `/` → `CAST(a AS float) / NULLIF(b, 0)` (DuckDB: double, NULL on zero), `//` and `%` on integers, CASE, CAST to a mapped type (no TRY_CAST), COALESCE / NULLIF, BETWEEN, IN list, IS [NOT] NULL, AND / OR / NOT with parentheses (three-valued logic is the same), comparisons, LIKE with `[`, `%`, `_` escaped and an ESCAPE clause |
```
oluies: **"(DuckDB: double, NULL on zero)" — the second half is wrong on the pin, and the `NULLIF` guard therefore introduces a divergence rather than removing one.**

`null_on_division_by_zero` defaults **false** (`duckdb/src/include/duckdb/main/settings.hpp:1934`) and `ieee_floating_point_ops` defaults **true** (`:1619`). Verified on a build of the pinned submodule:

```
select 1/0;    -->  inf   (DOUBLE)
select 1//0;   -->  Invalid Input Error: Division by zero in expression (1 // 0)
select 1%0;    -->  Invalid Input Error: Division by zero in expression (1 % 0)
```

So `NULLIF(b, 0)` yields NULL where DuckDB yields `inf`. SQL Server's `float` has no infinity, so **no** T-SQL rendering can match — bare `/` raises 8134, `NULLIF` gives NULL. `NULLIF` is a defensible choice; the defect is presenting it as semantics-preserving, which means no divergence is recorded and W5 gets no exemption, so the agreement suite will fail on `a/0` with no note saying why.

The `//` and `%` half of the row is already correct as written — both throw, matching 8134 — so they rightly get no guard.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+the writer walks parsed ones, so they cannot share a renderer; what they
+share is the vocabulary: **one** function table (`function_mapping.hpp`
+consulted by both, with `NOT IN` added so it stops being the scan's one
+strict exception), **one** collation predicate (`OrdersLikeDuckDB`, #362,
+consulted by `MSSQLOptimizer` and D4), **one** literal/parameter spelling
+(#361). A query cannot answer differently depending on which path took it.
+
+### D6 — setting and switches
+
+`mssql_remote_pushdown` (BOOLEAN, default **true** once W5 is green): read
```
oluies: **"read in `Supports`, so a session can turn the rewriter off" is not implementable.**

`virtual bool Supports(RemoteCapability) const` (`duckdb/src/include/duckdb/catalog/catalog.hpp:399`) takes no `ClientContext`. `AddExtensionOption` defaults to `SetScope::SESSION` (`config.hpp:225-227`) and `PhysicalSet::SetExtensionVariable` routes it to the ClientContext (`physical_set.cpp:20-25`), while the context-free `DBConfig::TryGetCurrentSetting` reads the **database-level** store (`config.cpp:808-818`) — the pattern this repo already uses at `src/catalog/mssql_table_entry.cpp:40-45`.

So a plain `SET mssql_remote_pushdown = false` is invisible to `Supports` unless the option is registered `SetScope::GLOBAL`, and then it is global, not per-session.

The alternative named two lines below is also global-only: `DisabledOptimizersSetting` has only `SetGlobal`/`ResetGlobal` (`settings.hpp:1143-1153`). As written, 079 ships **no** per-session switch at all — worth saying plainly, since D6's stated rationale is per-session diagnosis.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+vacuous).
+
+## 2. Work
+
+### W1 — the writer, expressions and a single table
+
+`src/pushdown/mssql_sql_writer.{hpp,cpp}` (namespace `duckdb::mssql`,
+`SQLWriter`): expression rendering per the D2 table, column resolution
+against the metadata cache, parameter collection; `SupportsPushdown` on
+`MSSQLCatalog` as the dry run (D1); `Supports(IS_REMOTE | EXECUTE_QUERY_NODE)`
```
oluies, line 212: **Gate `EXECUTE_QUERY_NODE` only — never `IS_REMOTE`.**

`remote_catalog_count` is `atomic<CheckedInteger<idx_t, InternalException>>` (`database_manager.hpp:134`), incremented at ATTACH and decremented at DETACH from the *live* answer of `Supports(IS_REMOTE)` (`database_manager.cpp:244-252`, `337-339`), and `CheckedInteger::operator--` throws on underflow (`checked_integer.hpp:79-83`).

So with `IS_REMOTE` gated by a mutable setting:
- attach-true → set-false → DETACH: the decrement is skipped, the counter never returns to 0, and `Optimizer::OptimizeStatement` runs `RemotePushdownOptimizer` on **every statement in the instance forever** (`optimizer.cpp:220-226`);
- attach-false → set-true → DETACH: `InternalException` out of `DETACH`.

The counter is shared with any other remote extension (postgres/mysql) in the same instance, so the blast radius is process-wide. `IS_REMOTE` also drives unrelated accommodations (`CheckAmbiguousCatalogOrSchema`), so toggling it changes name resolution too.

Cheap fix, and the spec nearly has it: the rewriter consults `EXECUTE_QUERY_NODE` at every per-catalog decision (`remote_pushdown_optimizer.cpp:102, 872, 917, 1117`). Gating that alone loses nothing and leaves the counter alone.

Also: the default is "true once W5 is green", but W1–W6 ship as one PR — please state the merged default here.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+| ORDER BY | `ORDER BY`, NULL placement emulated as `MSSQLOptimizer` does today (`CASE WHEN x IS NULL …` unless the column is NOT NULL) | string keys under D4 only |
+| LIMIT / OFFSET | `TOP n`; `OFFSET … FETCH` (`ORDER BY (SELECT NULL)` when there is no ORDER BY — both sides are arbitrary then) | |
+| DISTINCT, set operations | `DISTINCT`; UNION / UNION ALL / EXCEPT / INTERSECT | veto DISTINCT ON, `EXCEPT ALL` / `INTERSECT ALL`, UNION BY NAME |
+| CTEs, subqueries | `WITH`; scalar / EXISTS / IN / quantified (`= ANY`) subqueries, correlated included | veto recursive CTEs (spec 080 material: UNION ALL only, no RECURSIVE keyword) |
+| window functions | ROW_NUMBER / RANK / DENSE_RANK / NTILE / LAG / LEAD / FIRST_VALUE / LAST_VALUE / aggregates OVER, PARTITION BY, ORDER BY, ROWS frames and the default RANGE frame | veto RANGE with offsets, GROUPS, EXCLUDE, QUALIFY |
+| `SELECT * EXCLUDE / REPLACE` | expanded from the catalog's column list | veto `COLUMNS(...)`, PIVOT / UNPIVOT, TABLESAMPLE, parameters |
+
+Types: the result's types are the **server's** for the pushed shape
```
oluies, line 101: **079-alone silently degrades the schema of a same-catalog CTAS — worth naming the interim window.**

`RewriteCreateInfo` pushes a CTAS's query whenever the target is not `SINGLE_REMOTE` (`remote_pushdown_optimizer.cpp:697-703`), and since D1 does not claim `EXECUTE_STATEMENT`, `ResolveDDLTarget` → `ResolveRemoteCatalog(…, EXECUTE_STATEMENT)` returns `Unknown` (`:158-161`) — so the second disjunct is unconditionally true and the query is always pushed.

`CREATE TABLE ms.dst AS SELECT * FROM ms.src` then reads `ms.src` through `mssql_scan_params`, whose bind types come from the describe (`src/mssql_functions.cpp:163-250`), never from `NativeDuckDBType()` — whose only call sites are `mssql_table_entry.cpp:51,128` and `mssql_catalog.cpp:689`, all on the catalog path. `MSSQL_VARCHAR(50)` becomes plain VARCHAR, and with `mssql_default_string_length = 0` the target is created `nvarchar(max)` with the database default collation. CLAUDE.md sells `mssql_catalog_native_types` as exactly what this breaks: "what lets a target created from an MSSQL source inherit its declared lengths and collations with no cast written by anyone."

No hook in 079 prevents it — returning `false` from `SupportsPushdown(const SQLStatement &)` lands in the same branch via `VerifyStatementSupport` (`:671-682`). Unlike `RewriteNode(InsertQueryNode)` (`:453-459`), CTAS has no node-level consult.

Either state the regression and its window (until 080 claims `EXECUTE_STATEMENT`), or hold the default off until 080 lands. No data loss — schema/indexability/collation only.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+| JOIN | INNER / LEFT / RIGHT / FULL / CROSS with ON; USING expanded to ON | the ON condition is any expression the writer renders — non-equi, OR, functions, subqueries — nothing join-specific; veto NATURAL, SEMI / ANTI, ASOF, POSITIONAL, LATERAL |
+| ORDER BY | `ORDER BY`, NULL placement emulated as `MSSQLOptimizer` does today (`CASE WHEN x IS NULL …` unless the column is NOT NULL) | string keys under D4 only |
+| LIMIT / OFFSET | `TOP n`; `OFFSET … FETCH` (`ORDER BY (SELECT NULL)` when there is no ORDER BY — both sides are arbitrary then) | |
+| DISTINCT, set operations | `DISTINCT`; UNION / UNION ALL / EXCEPT / INTERSECT | veto DISTINCT ON, `EXCEPT ALL` / `INTERSECT ALL`, UNION BY NAME |
+| CTEs, subqueries | `WITH`; scalar / EXISTS / IN / quantified (`= ANY`) subqueries, correlated included | veto recursive CTEs (spec 080 material: UNION ALL only, no RECURSIVE keyword) |
```
oluies, line 97: **Missing veto: a nested `ORDER BY` without `TOP`/`OFFSET`/`FOR XML` is rejected by SQL Server (error 1033).**

`SELECT * FROM (SELECT a FROM ms.t ORDER BY a) s` is legal DuckDB. The writer has an exact form for the subquery and an exact form for ORDER BY, so D1's dry-run rule answers yes — but the dry run validates per-node *renderability*, not cross-cutting positional grammar: rendering `ORDER BY a` into a scratch buffer succeeds regardless of where the string is later embedded.

This row explicitly lists scalar / EXISTS / IN / quantified subqueries as supported, and 1033 applies to those too, not just FROM-clause derived tables. Since `RemoteExecute` cannot decline, the result is a server syntax error at execution on a query that works today via the scan path.

§3 "Not proposed" only excludes the reverse case (a subquery under a non-pushable outer node), so this isn't covered. Add-a-veto-rule gap, not a flaw in the dry-run approach.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+the client path whole (research § 2: extra rows *written* are not
+harmless, so no relaxed pushdown of a DML predicate, ever).
+
+### D3 — the fallback: one path, keyed by the ladder
+
+The shipped rowid path generalised by 067's ladder, resolved per table at
+plan time:
+
+1. **Primary key** — today's join key.
+2. **A usable unique index** — spec 077's `ChooseRowIdKey`, already the
```
oluies: **#350 is open and unmerged — the body asserts its symbols in the present tense.**

`gh pr view 350` → `OPEN`, `mergedAt: null`, and `git grep ChooseRowIdKey|RowIdRefusal|discovery_error` at `3a48795` (main) returns nothing; `specs/077-identity-columns/` does not exist at either end of this range.

The header's "Depends on: #350" is right; "already the rowid source since #350" here, and "since #350 it records `discovery_error`" at line 151-152, contradict it. An implementer searching main for those symbols finds nothing.

Spec 079 line 20 in this same PR shows the phrasing that works: "`ChooseRowIdKey` is not needed here but the branch order is" — hedged, not asserted as shipped.

(This also means the W4 item at line 151-152 is *not* moot — it describes behavior that hasn't landed.)

specs/065-dml-pushdown-recon/spec.md
```diff
@@ -0,0 +1,162 @@
+# Spec 065 — Direct UPDATE/DELETE: one statement when everything pushes down
+
+**Status:** **CLOSED 2026-09-17 — superseded by spec 080 (`../080-dml-pushdown/spec.md`: its goal on DuckDB 2.0's rewriter, D3 and D5 carried over as W1/W4 there) and spec 079 (the writer). Kept as history.** Was: Draft, revalidated 2026-09-17 against `main` `3a48795` + #350 —
```
oluies: **The CLOSED pointer undercounts what actually carried over.**

This says only D3 and D5 went to 080 W1/W4. But 080 cites 065 **D2** at its line 48 ("`DEFAULT` in SET → veto … 065 D2.5") and 065 **D4** at line 114 ("stop refusing a keyless table at bind (065 D4)").

A reader trusting this line to learn "where each part went" — the stated purpose of the CLOSED markers — concludes D1/D2/D4 were dropped, when D2 and D4 reappear verbatim-cited in 080. Worth itemizing the way 058's status line does.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+   is unique in the staged set. A VOLATILE function in WHERE or SET on
+   rung 3 is refused by name ("add a unique index, or make the expression
+   deterministic").
+
+Delivery: rungs 1–2 keep the `VALUES`-join statements below the small-result
+threshold and stage above it; rung 3 always stages. The stage is a
+`##stage_<uuid>` scratch table filled by a `BulkLoadSession` (the writer
+spec 062 W0 built, `Adopt`ed onto the statement's own connection — the
+pinned one in a transaction), the DML is
+`UPDATE t SET t.c = s.c__new … FROM target t JOIN ##stage s ON <key>` /
```
oluies: **Rung 3's JOIN cannot seek, and 067 D3 puts the feeding scan on a *different* connection in autocommit.**

`IS NOT DISTINCT FROM` / `EXISTS (SELECT … INTERSECT SELECT …)` over every column is not SARGable, so each ~100k batch full-scans the target and escalates toward a table X lock.

Meanwhile `MaterializeSharedConnectionScans` returns early in autocommit — "Autocommit gives every scan its own pooled connection, so there is nothing to share and nothing to serialize" (`src/table_scan/mssql_optimizer.cpp:745-750`) — and 067 D3, which D3 here reuses wholesale, specifies the autocommit mode as "pipelined incremental … scan ∥ stage-fill ∥ DML **on separate connections**, ping-pong stages". So this is not stage-fully-then-join: the feeding scan holds S locks from a second session while the previous batch's DML takes X locks on the same table, batch after batch.

§4 names only LOB cost for rung 3. Worth a Risks callout, an isolation/TABLOCK decision (as `mssql_copy_tablock` needed on the write path), and a test that forces escalation. Blocking/timeouts are the likely shape; a strict 1205 cycle is plausible but unconfirmed.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+The shipped rowid path generalised by 067's ladder, resolved per table at
+plan time:
+
+1. **Primary key** — today's join key.
+2. **A usable unique index** — spec 077's `ChooseRowIdKey`, already the
+   rowid source since #350; its refusals (`RowIdRefusal`) name why a key
+   is unusable (datetime / sql_variant, #358; a cast-required type, #354).
+3. **All columns, NULL-safe** — the keyless base case (067 § 1's argument:
+   for a deterministic WHERE and SET, matching by value updates exactly
+   the set DuckDB would): `IS NOT DISTINCT FROM` on SQL Server 2022+, the
```
oluies, line 120: **§1's equivalence argument assumes both sides compare under the same semantics. They don't, in one reachable case.**

The argument — two indistinguishable rows either both satisfy the WHERE or neither does, since the predicate is a function of column values — holds when the WHERE is pushed to the scan (server semantics on both sides), which is the normal case and which 080 §0 states twice.

It breaks when the WHERE is one the scan's filter pushdown refuses and DuckDB evaluates it client-side under binary semantics, while the stage JOIN matches under a `_CI`/padding collation. `DELETE FROM keyless_t WHERE regexp_matches(v,'^ab$')` selects only `'ab'` client-side, but the JOIN also matches `'AB'` and `'ab␣'` — the revalidation's own §8.5 measures exactly that.

Native string semantics as the *contract* is settled (067:37-39, 080:17-19) and I'm not reopening it. This is the narrower gap: one sentence in rung 3 either admitting the case as in-contract, or refusing a rung-3 statement whose predicate is not fully pushed to the scan — the way the volatile-function guard already does.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+   `EXISTS (SELECT … INTERSECT SELECT …)` form below; the stage is
+   DISTINCT over the key; LOB columns excluded from the key when the rest
+   is unique in the staged set. A VOLATILE function in WHERE or SET on
+   rung 3 is refused by name ("add a unique index, or make the expression
+   deterministic").
+
+Delivery: rungs 1–2 keep the `VALUES`-join statements below the small-result
+threshold and stage above it; rung 3 always stages. The stage is a
+`##stage_<uuid>` scratch table filled by a `BulkLoadSession` (the writer
```
oluies: **Now that the stage is filled on the statement's own connection, `##` has lost its justification.**

066 D5 chose a global temp table because the stage was filled on a *second* connection ("it IS visible to the pinned connection"). D3 here `Adopt`s the `BulkLoadSession` onto the statement's own connection — the pinned one in a transaction — so a session-local `#stage` would do the same job and remove the cross-session visibility entirely.

Not a security finding as such: `##` targets are a deliberate, user-requested feature elsewhere in this codebase (`src/include/copy/load_policy.hpp:98,169` — "exactly why #189 asks for `##`, not `#`"), and the uuid is not guessable. Just worth one line saying which shape 080 intends, since the reason for `##` no longer applies.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+| `DELETE FROM t [USING u …] WHERE p` | `DELETE t FROM [s].[t] AS t [JOIN …] WHERE p` | same |
+| `INSERT INTO t [(cols)] SELECT …` (both sides remote) | `INSERT INTO [s].[t] ([cols]) SELECT …` | the SELECT through 079; an explicitly named identity column brackets the statement in `SET IDENTITY_INSERT` (077 W2); `INSERT … VALUES` stays on the shipped path (it carries no remote SELECT to win by) |
+| `INSERT INTO local SELECT … FROM remote` | the SELECT alone | the rewriter does this by itself (`push_select_only`); nothing to add |
+| `CREATE TABLE remote AS SELECT … FROM remote` | our CREATE (table kind, collation, lengths — the WITH options keep meaning), then the pushed `INSERT … SELECT` | not `SELECT INTO`; needs `EXECUTE_STATEMENT` claimed for CREATE TABLE AS only, and `SupportsPushdown(const SQLStatement &)` saying yes for that shape alone |
+| `RETURNING`, `MERGE INTO`, `ON CONFLICT` | — | veto; the shipped path handles what it handles today |
+
+Execution: `RemoteExecute` returns a ref to the DML form of the vehicle —
+`mssql_scan_params` with a **count result**: the statement runs through
```
oluies: **Worth stating explicitly that the count form must not reuse `MSSQLScanBind`'s shape discovery.**

I read this as a distinct vehicle whose result shape is statically a single BIGINT from the DONE token, which needs no describe — that's the right design and it's what duckdb-mysql does (`mysql_scanner.cpp:351-367`: PREPARE only, `call_return_type = StatementReturnType::NOTHING`, `suppress_dml_output`).

But the nearest existing bind is a trap an implementer could fall into by accident: `DescribeFirstResultSet` returns `shape.ok == false` for any statement with no result set (`src/mssql_functions.cpp:280-283`, "the statement returns no result set"), and the F1 fallback then **runs the statement at bind** (`:479-483`, `executed_at_bind = true`). Since the rewriter descends into an EXPLAIN's inner statement (`remote_pushdown_optimizer.cpp:249-251`), a vehicle inheriting that bind would make `EXPLAIN UPDATE ms.t …` perform the update.

One sentence here, plus a W5 assertion that `EXPLAIN` of a pushed DML changes no rows.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+|---|---|---|
+| `UPDATE t SET c = e, … [FROM u …] WHERE p` | `UPDATE t SET [c] = e, … FROM [s].[t] AS t [JOIN …] WHERE p` | every SET expression and the WHERE through 079's expression writer under the same rules; SET of a rowid-key column is refused as today; `DEFAULT` in SET → veto (server DEFAULT ≠ DuckDB's NULL default, 065 D2.5) |
+| `DELETE FROM t [USING u …] WHERE p` | `DELETE t FROM [s].[t] AS t [JOIN …] WHERE p` | same |
+| `INSERT INTO t [(cols)] SELECT …` (both sides remote) | `INSERT INTO [s].[t] ([cols]) SELECT …` | the SELECT through 079; an explicitly named identity column brackets the statement in `SET IDENTITY_INSERT` (077 W2); `INSERT … VALUES` stays on the shipped path (it carries no remote SELECT to win by) |
+| `INSERT INTO local SELECT … FROM remote` | the SELECT alone | the rewriter does this by itself (`push_select_only`); nothing to add |
+| `CREATE TABLE remote AS SELECT … FROM remote` | our CREATE (table kind, collation, lengths — the WITH options keep meaning), then the pushed `INSERT … SELECT` | not `SELECT INTO`; needs `EXECUTE_STATEMENT` claimed for CREATE TABLE AS only, and `SupportsPushdown(const SQLStatement &)` saying yes for that shape alone |
```
oluies, line 55: **Say that the statement-form ref is lazy, the way 079 D3 says it for the node form.**

`FinishPushdown` calls `RemoteExecute(context, statement)` **synchronously inside the optimizer** for DDL (`remote_pushdown_optimizer.cpp:1861-1870`), and the optimizer runs under EXPLAIN (`:249-251`) and PREPARE (`planner.cpp:225-243`). `RemoteExecute` returns a `TableRef` and nothing forces it to execute — duckdb-mysql's returns a bare `TableFunctionRef` and touches the server zero times at optimize time (`mysql_catalog.cpp:564-580`).

So this is safe if implemented lazily and broken if not, and 079 D3 states the laziness only for the `QueryNode` form. An implementation that performs the CREATE + load inside `RemoteExecute` would make `EXPLAIN CREATE TABLE ms.t AS SELECT …` create and fill the table — and the extension cannot veto on EXPLAIN-ness, since `SupportsPushdown` receives the inner `CreateStatement` stripped of its wrapper.

Also worth stating: whether the CREATE and the INSERT share one server transaction, or the table is dropped if the INSERT fails.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+### W3 — the ladder
+
+Rung 3 in `ChooseRowIdKey`'s caller (the plan-time resolution), the two
+NULL-safe join forms by server version, the volatile guard, the staged
+delivery (`##stage` + `Adopt` + batched JOIN DML), the DISTINCT stage and
+the LOB exclusion; bind-time refusal removed.
+
+### W4 — riding cleanups from 065 D5
+
+`mssql_dml_use_prepared` retired (loaded, read by nothing since inception);
```
oluies: **This is a breaking change, not a cleanup.**

The premise is right — `mssql_dml_use_prepared` is registered at `src/connection/mssql_settings.cpp:309-311`, read into `DMLConfig::use_prepared` at `src/dml/mssql_dml_config.cpp:40-41`, and `use_prepared` is referenced nowhere else.

But unregistering it makes `SET mssql_dml_use_prepared = true;` throw: `PhysicalSet::SetVariable` raises `InvalidInputException("Extension parameter %s was not found after autoloading")` — a hard error that kills the rest of a `.duckdbrc` or startup script, not a warning.

The repo's own precedent (spec 047) is deprecate-then-remove: the `mssql_open`/`mssql_close`/`mssql_ping` group was marked `[DEPRECATED]` and kept working. Suggest keeping the option registered as a documented no-op for one minor release, or at minimum calling out the breaking `SET` in the CHANGELOG.

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+The rewriter carries `MergeQueryNode`, and T-SQL has `MERGE`; the mapping is
+a D1 row when asked for. Until then MERGE keeps its bind-time PK requirement
+(DuckDB's MERGE binder needs rowid for match classification — upstream).
+
+## 2. Work
+
+### W1 — the DML writer and the count vehicle
+
+`UpdateQueryNode` / `DeleteQueryNode` / `InsertQueryNode` rendering in
+`SQLWriter`; the `IDENTITY_INSERT` bracket reused from 077 W2; the count
```
oluies, line 176: **`spec 062 W0` / `W1c` are not findable labels.**

`specs/062-insert-via-bcp/spec.md` uses D1–D5 decision numbering and has no W-numbered headings (`grep -n "^### " specs/062-insert-via-bcp/spec.md`). Those W labels exist only in that spec's commit trail (`db3a070`, `6e44027`) and were never backfilled into the frozen spec text. Same for the "spec 062 W1c" reference at line 22-23 and in 080 D1.

Newer specs (075/076/079/080) do use `### W1` headings consistently, so the convention is fine — 062 just predates it. Either cite 062's D-numbers or add the W headings there.

Separately, while you're in W1: cache and statistics invalidation after a pushed DML or CTAS isn't specified anywhere in 080. COPY and CTAS invalidate the row-count and statistics caches today; a pushed DML goes through no plan hook, so nothing triggers that and planner cardinalities go stale.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+|---|---|---|
+| FROM base table, remote view | `[schema].[name]` | one catalog; a view pushes, the server expands it |
+| projections, WHERE, HAVING, ON | the expression writer | column refs, constants, `+ - *` (overflow errors on both sides), `/` → `CAST(a AS float) / NULLIF(b, 0)` (DuckDB: double, NULL on zero), `//` and `%` on integers, CASE, CAST to a mapped type (no TRY_CAST), COALESCE / NULLIF, BETWEEN, IN list, IS [NOT] NULL, AND / OR / NOT with parentheses (three-valued logic is the same), comparisons, LIKE with `[`, `%`, `_` escaped and an ESCAPE clause |
+| scalar functions | one exact form per name | starts from the scan's `function_mapping.hpp` (`lower`, `upper`, `trim`, `ltrim`, `rtrim`, `year` … `second`, `+ - * %`, `negate`, `prefix` / `suffix` / `contains` → LIKE), grows one proven form at a time; vetoes as duckdb-mysql's and 061 § 4.5: `length` (`LEN` drops trailing spaces and counts UTF-16 units), `upper` / `lower` on non-ASCII (simple case mapping only — accepted as-is for the ASCII range: measured pushed today), `sqrt` / `ln` / `log` (server errors where DuckDB gives NaN), `power` / `exp` (overflow errors), `week` / `dayofweek` numbering |
```
oluies: **`negate` / `prefix` / `suffix` / `contains` are bound-tree names — the parsed tree never carries them.**

`function_mapping.cpp:35-37` has an explicit "LIKE pattern functions … Do not add them here" comment; `prefix`/`suffix`/`contains` are not in the table at all, they're handled by `EncodeLikePattern()`. And all four are names that post-bind optimizer *rules* assign: `like_optimizations.cpp:162-168` rewrites `LIKE 'x%'` → `prefix` on the bound plan, and `negate` is the bound catalog function name. The rewriter runs pre-bind (`planner.cpp:246` calls `OptimizeStatement` on the raw `unique_ptr<SQLStatement>`), where the same predicate is a `~~`/`!~~` `FunctionExpression` (`transform_expression.cpp:842`) and unary minus is `FunctionExpression("-")` with arity 1 (`:1587-1596`).

The design is fine — LIKE is already covered generically by the row above, and `mysql_sql_writer.cpp:267-268` shows the correct parsed-tree pattern (match `~~` directly, no bound-name table). Just drop or reword the parenthetical, and add an explicit arity-1 case for `-`, since the table's `-` entry is the binary one.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+| GROUP BY + aggregates | positional and alias references **expanded** (T-SQL has neither: errors 164 / 207); `COUNT_BIG(*)` / `COUNT_BIG(x)` / `COUNT_BIG(DISTINCT x)`; `SUM(CAST(int AS bigint))`; `AVG(CAST(x AS float))`; `MIN` / `MAX` (`bit` → `CAST(b AS tinyint)`; strings under D4); `STRING_AGG(CAST(x AS nvarchar(max)), <literal>) WITHIN GROUP (ORDER BY …)`; `STDEV` / `STDEVP` / `VAR` / `VARP`; `agg(x) FILTER (WHERE c)` → `agg(CASE WHEN c THEN x END)` | veto `first` / `any_value`, `arg_*`, `median` / `quantile` / `mode`, `approx_count_distinct` (a different estimator), `list` / `array_agg` / `histogram`, `bit_*`, `corr` / `covar_*` / `regr_*`, a non-literal `string_agg` separator, GROUP BY ALL / ROLLUP / CUBE / GROUPING SETS |
```
oluies: `COUNT(*) FILTER (WHERE c)` has no `x` to substitute, so the stated rewrite yields `COUNT_BIG(CASE WHEN c THEN END)`.

The row already distinguishes `COUNT_BIG(*)` from `COUNT_BIG(x)` for the non-FILTER case; add the third form — `COUNT_BIG(CASE WHEN c THEN 1 END)` — alongside the general rule.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+### D2 — the T-SQL writer
+
+Input: the parsed tree with the catalog name already stripped by the
+rewriter; output: the statement text and its parameter list. Every constant
+becomes an `sp_executesql` parameter `@pN`, declared **from the column** when
+the constant is compared with one (the 076 rule, with #361's fix: `varchar`
+when the constant is representable in both the column's and the database's
+code page, `nvarchar` otherwise) and from the value otherwise
+(`DeclarationForValue`); `mssql_scan_parameterize_filters = false` renders
+literals instead, for plan comparison. Identifiers are bracket-quoted; the
```
oluies, line 80: D5's shared-vocabulary list (function table, `OrdersLikeDuckDB`, literal spelling) doesn't include identifier quoting, which invites a fourth copy of the quoter.

Three exist today and all three correctly double `]` — `src/dml/insert/mssql_value_serializer.cpp:13-26`, `src/catalog/mssql_ddl_translator.cpp:95-109`, `src/table_scan/filter_encoder.cpp:178-188`. "Bracket-quoted" here is under-specified relative to them, and a `]` can reach the writer from a user alias (`SELECT 1 AS "x]y"`), not only from `sys.columns`.

duckdb-mysql routes every identifier through one shared helper (`mysql_utils.cpp:356-360`, used by the writer at `mysql_sql_writer.cpp:44-45`, including alias positions at `:728`). Adding "one identifier quoter" to D5/W4 is a one-line fix.

specs/065-dml-pushdown-recon/join-agg-pushdown.md
```diff
@@ -0,0 +1,166 @@
...
+- **Class 2 — engine mechanics (result types, overflow errors, truncation,
+  server-setting dependence).** These are wrong NUMBERS for a DuckDB query,
+  not "server semantics anyone expects" — wrappers are MANDATORY regardless
+  of the directive:
+
+| construct | mandatory form | verified |
+|---|---|---|
+| `count(*)`/`count(x)` | `COUNT_BIG(...)` (T-SQL COUNT is int) | doc |
+| `sum(int family)` | `SUM(CAST(x AS decimal(38,0)))` + client cast → HUGEINT; residual gap (10^38..1.7e38) physically unreachable | live: naked SUM(int) → error 8115 |
+| `avg(int/bigint/decimal)` | **decompose**: push `SUM(...)` + `COUNT_BIG(...)`, divide in DuckDB — bit-exact, sidesteps every T-SQL result-type rule at once | live: AVG(int) over {1,2} = **1** |
```
oluies, line 62: **This row and 079's D2 now specify contradictory forms for the same construct, both live in this PR.**

Here: decompose `avg` into `SUM(...) + COUNT_BIG(...)` and divide in DuckDB, "bit-exact". `specs/079-select-pushdown/spec.md:92`: `AVG(CAST(x AS float))`, which loses precision above 2^53 for bigint and decimal. §1 of this file likewise requires `SUM(CAST(x AS decimal(38,0)))` returning HUGEINT, where 079 chooses `SUM(CAST(int AS bigint))` and accepts the narrower type.

Either mark these rows superseded-by-079 with the reason the earlier forms were rejected, or adopt the decomposition in 079. As it stands an implementer has two authoritative-looking rules.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+
+### W4 — shared vocabulary
+
+`function_mapping.hpp` as the one table (the scan's encoder and the writer
+both consult it; `NOT IN` added), `OrdersLikeDuckDB` shared with
+`MSSQLOptimizer`, the #361 spelling shared with `DeclarationForColumn`.
+
+### W5 — the agreement suite
+
+`test/sql/pushdown/`: every D2 row as a statement run **twice** —
```
oluies, line 246: The suite as specified would miss the encoding divergence: it exempts "the D4 string cases", but D4 is about collation *order*, and the gap at line 89/153 is about *encoding*.

Worth stating the fixture requirements explicitly — a non-UTF-8 `varchar` column holding non-ASCII data, a `geometry` column, an `ntext` column — otherwise "run twice and compare" passes while the divergence ships. Same for a division-by-zero exemption (see line 90).

specs/080-dml-pushdown/spec.md
```diff
@@ -0,0 +1,220 @@
...
+### W4 — riding cleanups from 065 D5
+
+`mssql_dml_use_prepared` retired (loaded, read by nothing since inception);
+`EnsurePKLoaded` no longer degrades a discovery error to "no key" — since
+#350 it records `discovery_error` and the refusal names it, so with D3 a
+hiccup cannot silently change which path a statement takes.
+
+### W5 — tests
+
+- The bite test (065 § 4.1): a statement with an unmapped function in WHERE
```
oluies, line 212: Cases this list doesn't cover, from the findings above: a `READ_ONLY` attach refusing a pushed DML and CTAS; `EXPLAIN`/`PREPARE` of a pushed DML and of a pushed CTAS asserting nothing changed and no table was created; rung-3 lock escalation in autocommit with a concurrent reader; and a rung-3 statement whose predicate is not fully pushed to the scan.

specs/079-select-pushdown/spec.md
```diff
@@ -0,0 +1,266 @@
...
+nothing from us. `MSSQL_COUNTERS` gains a `remote_pushdown` statement counter
+(spec 063's lesson: an SQL-invisible path needs a counter or its suite goes
+vacuous).
+
+## 2. Work
+
+### W1 — the writer, expressions and a single table
+
+`src/pushdown/mssql_sql_writer.{hpp,cpp}` (namespace `duckdb::mssql`,
+`SQLWriter`): expression rendering per the D2 table, column resolution
```
oluies, line 210: "Column resolution against the metadata cache" — worth saying what happens on a cache miss.

Metadata is loaded on first touch (spec 076: one round trip per fresh table), so a dry run at optimize time for a table not yet loaded needs a connection, and inside a transaction that means the pinned one. Vetoing on a miss instead would silently disable pushdown for the first query against every table, which is probably worse. Either way it should be stated.

VGSML:
Thanks — every claim checked against the pin and the tree before acting, and all of them held (`1/0` → `inf` on our build, `IsNullOrderCompatible` refuses rather than emulates, `Supports` takes no context, `remote_catalog_count` is driven by the live `IS_REMOTE` answer, `RewriteCreateInfo` pushes the query with no hook to stop it). Fixes are in one commit (`84393d5`), one push.

**Blocking — READ_ONLY (080:55).** Two guards, as your mysql precedent: `SupportsPushdown` refuses every DML and CTAS node when the catalog `IsReadOnly()` (the statement then takes the shipped path, whose `Plan*` hook refuses it as today), **and** the count form's bind throws `PermissionException`, so no route around the first guard executes a write. W5 case added for both.

**079**
- *Column rewrites (153, 89).* The base-table row now renders every SELECT-list column through `MSSQLColumnInfo::BuildReadExpression` — since #352 that helper is already shared by the scan and INSERT's OUTPUT list and takes plain strings (type name, length, collation), which is exactly what the writer has from the metadata cache; `table_scan.cpp`'s `BuildColumnExpression` is the wrapper. `*` expands from the catalog's column list; predicates render the bare column. D5 lists it as the fourth shared thing, and the identifier quoter (80) as the fifth — three exist today, W4 makes them one.
- *NULL order (94).* Rewritten: the writer **emulates** (`CASE WHEN x IS NULL …`), it is new work in W2, shared back into `MSSQLOptimizer` in W4, and W1's premise now says the nullable shapes have no old-path counterpart.
- *`1/0` (90).* `NULLIF` kept as the least-wrong form (the server's float has no infinity) and recorded as a divergence with its own W5 exemption, not as semantics-preserving.
- *Switch (157, 172).* `mssql_remote_pushdown` is `SetScope::GLOBAL`, instance-wide, and the spec says there is no per-session switch; it gates `EXECUTE_QUERY_NODE` only, `IS_REMOTE` is a constant; default is stated as true at merge.
- *CTAS window (101).* Closed rather than named: W3 has the vehicle report native types (`MSSQL_VARCHAR(n)` with collation) from the describe's `system_type_name` + `collation_name` — the describe returns both, only the type name is read today — so a same-catalog CTAS keeps its lengths and collations and a pushed `SELECT *` has the catalog path's types. Interim window zero.
- *Nested ORDER BY (97)* vetoed without TOP/OFFSET (1033) as a node-writer rule; *parsed names (91)* — table keyed on `~~`/`!~~`, unary `-` arity 1, the bound names noted as never reaching the writer; *`COUNT(*) FILTER` (92)* third form added; *cache miss (170)* — loads on first touch, not a veto; *fixtures (197)* — non-UTF-8 varchar with non-ASCII data, geometry, ntext, sql_variant/hierarchyid, nullable ORDER BY key, same-catalog CTAS; exemptions listed.

**080**
- *Count form (56, 52).* Stated: static single-BIGINT shape, no describe, must not inherit `MSSQLScanBind` (the F1 fallback would run `EXPLAIN UPDATE`), lazy; the CTAS statement ref lazy too, CREATE + INSERT … SELECT in one server transaction so a failed load leaves no table. W5: `EXPLAIN`/`PREPARE` of a pushed DML and CTAS change nothing.
- *Rung 3 (101, 89).* Delivery is stage-fully-then-join — 067 D3's pipelined autocommit mode is explicitly not used, for the reason you give (a JOIN that cannot seek, escalating, against a second session's scan); risk callout and a concurrent-reader test added. A rung-3 statement whose predicate the scan does not push is refused by name, like the volatile guard — your regexp example is in the text.
- *`#stage` (98)* — session-local now, with the reason `##` no longer applies. *Invalidation (128)* — row-count/statistics after a pushed DML, catalog after CTAS, in W1. *062 labels* — cited by code symbol (`BulkLoadSession::Adopt`, `mssql::LoadTransaction`). *#350 (84)* — cited as open, present tense gone. *`mssql_dml_use_prepared` (150)* — deprecate-then-remove per the 047 precedent, CHANGELOG line.

**Document set.** 065's CLOSED line itemises D1–D6 (D2 and D4 do reappear in 080, as you found). `join-agg-pushdown.md`'s `sum`/`avg` rows carry a superseded-by-079 note with the reason (bigint overflow errors rather than silence; the AVG decomposition needs a client-side projection layer 079 D3 does not have; and its "DuckDB NULL on zero divisor" was wrong too). The rest of the record stays as history, as you suggested.

The 079 D4 string rule is unchanged and still the one thing to confirm.

oluies (approved):

PR #367, opened 2026-09-17
fix: NOT IN (and IN over an expression) reach the server — IN rendered as an operator expression (#366)

### LINKED ISSUE #366: Filter pushdown: NOT IN on a string column runs client-side (no COMPARE_NOT_IN in the encoder) while <>, NOT LIKE and IN are pushed — the one exception to the server-sets rule
## What happens

On the shipped scan every string predicate is pushed with the server's semantics — the column's collation and padding — except one. Measured on `SQL_Latin1_General_CP1_CI_AS`, five rows `ab`, `ab␣`, `AB`, `abc`, `x`:

| predicate | rows | whose answer |
|---|---|---|
| `v = 'ab'` | 3 | server |
| `v <> 'ab'` | 2 | server (DuckDB would say 4) |
| `v NOT LIKE 'ab%'` | 1 | server (DuckDB 2) |
| `v > 'ab'` | 2 | server (DuckDB 3) |
| `v NOT IN ('ab', 'x')` | **3** | **DuckDB** (the server would say 1) |

`NOT IN` is evaluated client-side because `FilterEncoder` has no `COMPARE_NOT_IN` case (`src/table_scan/filter_encoder.cpp`), not by policy: `IN` is encoded, `<>` is encoded. The result is the one inconsistency in the string contract — `v NOT IN ('ab')` and `v <> 'ab'` answer differently on the same data — and a full transfer for a predicate the server would seek.

## The fix

Add `COMPARE_NOT_IN` to the encoder as the negation of the `IN` form (`[v] NOT IN (@p1, @p2)`; the constants stay literal as `IN` lists do under spec 076), with the three-valued-logic case pinned: `NOT IN` with a NULL in the list yields no rows on both sides. Test beside the existing `IN` pushdown case, asserting the server's count (1 on the fixture above) and, under `MSSQL_DEBUG`, that the predicate reached the WHERE clause.

Decided in spec 079 D4 (PR #364, approved): sets are the server's for every string predicate in both paths; this is the last exception in the scan path. Spec 079 W4 lists it; it is small enough to go first as its own PR.

VGSML:
Closes #366.

## What was wrong

DuckDB binds `v NOT IN (...)` as `NOT (v IN (...))`. `FilterEncoder` encoded `IN` only as the table filter the combiner builds from a bare-column `IN` (`EncodeInFilter`); as a `BoundOperatorExpression` — which is what `NOT IN` always is, and what `IN` over an expression operand is — it fell through to "unsupported operator type" and ran client-side after a full transfer, under DuckDB's byte equality. Measured on a `SQL_Latin1_General_CP1_CI_AS` column holding `ab`, `ab␣`, `AB`, `abc`, `x`: `v NOT IN ('ab', 'x')` returned three rows where `v <> 'ab'` beside it returned two — the one exception to spec 079 D4's rule that string sets are the server's.

## The fix

`EncodeOperatorExpression` gains a `COMPARE_IN` / `COMPARE_NOT_IN` case: operand and list items through `EncodeValueExpression`, the list's constants declared from the operand's column exactly as a comparison's constant is (spec 076, `constant_peer`), so a `varchar` list keeps its seek. Rendered (`MSSQL_DEBUG=2`):

```
(NOT ([v] IN (@p0, @p1)))          -- v NOT IN ('ab', 'x')
(([n] + @p2) IN (@p3, @p4))        -- n + 1 IN (2, 3)
(NOT ([id] IN (@p5, @p6)))         -- rowid NOT IN (1, 2), rowid → the scalar PK on this path
```

A NULL in the list needs no special case: three-valued logic gives no row for `NOT IN` on both sides.

## Tests

`test/sql/catalog/not_in_pushdown.test` (26 assertions): the CI-collation case above (`abc` alone — DuckDB's answer would be `ab␣`, `AB`, `abc`), `IN` and `NOT IN` partitioning the table (3 + 2 = 5), `NOT IN ('ab', NULL)` → no rows, an integer list with a NULL column value, `n + 1 IN / NOT IN`, `rowid NOT IN`, and the same answers with `mssql_scan_parameterize_filters = false`. Local: catalog 42 cases / 1793 assertions, rowid 5, dml 8 — green.


oluies:
## roborev review — 2 findings (2 Low)

Branch diff `634b485..6350a74` (1 commit) against `main`, roborev job 1658, agent `claude-code`. No correctness problems found in the rendering itself. Both points are about what guards it.

### Low — `src/table_scan/filter_encoder.cpp:1165-1195`: the new `COMPARE_IN` / `COMPARE_NOT_IN` branch has no server-free test

The only coverage is `test/sql/catalog/not_in_pushdown.test`. It needs a live SQL Server, and the integration job is dispatch-only, so it does not run on ordinary PRs. `test/cpp/test_filter_encoder.cpp` already builds `BoundOperatorExpression` trees for `OPERATOR_NOT` without a server, but nothing asserts on the SQL this path produces. Four behaviours are unguarded, and a regression in any of them returns wrong rows:

- `NOT` over `COMPARE_IN` renders as `(NOT ([v] IN (@p1, @p2)))`.
- A `NULL` item stays a literal `NULL` and is not turned into a parameter.
- List constants are declared from the operand's column: a `varchar(10)` column gets `varchar(n)` parameters, not `nvarchar`.
- If any item cannot be encoded, the whole expression is unsupported. A partial list is never pushed.

*Fix:* add unit cases to `test_filter_encoder.cpp` for `COMPARE_IN` / `COMPARE_NOT_IN`, both bare and under `OPERATOR_NOT`, with a params set and `mssql_columns` wired. Assert on the SQL text and the declared types. Include one case with an item that cannot be encoded and assert `supported == false`.

### Low — `filter_encoder.cpp:1181-1192`: long lists are pushed without a cap

Each list item becomes one `sp_executesql` parameter until `MAX_PARAMS` runs out, and the items after that become literals. A very large `NOT IN` or expression-`IN` list therefore uses up the parameter budget for every other predicate in the same scan. With parameterisation off, it produces a statement of unbounded length. These queries used to run client-side. Now a long enough list can fail with a server error such as 8623/8632, where it used to succeed. `LegacyInFilter` has the same exposure already, but this change extends it to `NOT IN` and to expression operands.

*Fix:* cap the pushed list length, for example at a few hundred items or a fraction of `MAX_PARAMS`. Return `{"", false}` above the cap so a long list keeps its old client-side behaviour.

VGSML:
Both taken, one commit.

**Server-free coverage.** `TestInListOperatorExpressions` in `test_filter_encoder.cpp`, exactly the four behaviours you listed plus the cap: `NOT` over `COMPARE_IN` renders `(NOT ([name] IN (@p0, @p1)))` with both items declared `varchar(20)` from the operand's column; `COMPARE_NOT_IN` directly; an expression operand (`([id] + @p4) IN (@p5, @p6)`); a `NULL` item stays a literal `NULL` and registers no parameter; an item the encoder cannot render refuses the whole list and leaves the parameter set as it was.

**Length cap.** `MAX_PUSHED_IN_ITEMS = 256` on the operator path: above it the predicate is refused and DuckDB evaluates it, as every list did before this PR; the unit test pins 256 pushed / 257 refused with no parameter registered. `LegacyInFilter` is left as it is (a bare-column `IN` pushed without a cap is the shipped behaviour; changing it is its own decision). CHANGELOG says so.

oluies:
## roborev review of fbc3f6a — 2 findings (2 Low)

This review covers only the follow-up commit `fbc3f6a` (roborev job 1662, agent `claude-code`). It resolves both earlier Lows: the operator-expression path now has server-free unit cases for rendering, parameter declaration, NULL items, whole-list refusal, and the cap. What is left is about how far the cap reaches and how the CHANGELOG describes it.

### Low — `src/table_scan/filter_encoder.cpp:1180`: the cap applies per predicate, not per statement

`a NOT IN (<256>) AND b NOT IN (<256>) AND ...` still pushes every list. With parameterisation on, the statement keeps working, because `EncodeConstantValue` switches to literals once `SqlParamSet::MAX_PARAMS` is used up. But the commit message says the cap is there to prevent an unbounded batch when parameterisation is off, and 8623/8632, and several lists each under the cap can still add up to both.

*Fix:* apply the budget to the total number of list items in the enclosing search condition. Or, if a per-predicate cap is the intent, say so in the comment.

### Low — `CHANGELOG.md:22-24`: the entry reads as if every IN list is capped

"lists over 256 items stay client-side, as every list did before" is true only of the operator-expression path. A bare-column `IN` becomes a table filter, goes through `EncodeInFilter` (`filter_encoder.cpp:751`), is sent as literals, and is not capped.

*Fix:* reword it as "`NOT IN` lists, and `IN` over an expression, longer than 256 items stay client-side". Alternatively, apply the same cap in `EncodeInFilter` if it is meant to cover every IN list.

VGSML:
Both taken, in `b3cd0df` (on top of a merge with `main`, which now carries #368's encoder changes — the `IsAscii` removal and the cap constant sat in the same region).

**Per predicate, on purpose — and now said so.** The comment at the constant and at the check both state it: with parameterisation on, the statement's budget is `SqlParamSet::MAX_PARAMS` (the sink turns constants past it into literals), and with it off a batch of several capped lists is still bounded by their number; a bare-column `IN` is a table filter (`EncodeInFilter`) and is not capped, as it never was.

**CHANGELOG** reworded to what the cap covers: "`NOT IN` lists, and `IN` over an expression, longer than 256 items stay client-side, as they did before this change (…; the cap is per predicate); a bare-column `IN` is a table filter and is not capped, as it never was."

PR #368, opened 2026-09-17
fix: declare a non-ASCII constant varchar when both code pages hold it — the seek survives on SQL_ collations (#361)

### LINKED ISSUE #361: Filter pushdown: a non-ASCII constant against a varchar column on a SQL_ collation loses the index seek (declared nvarchar)

VGSML:
Closes #361.

## What was wrong

Spec 076 declares the constants of a pushed filter from the column they are compared with, and any **non-ASCII** constant as `nvarchar` whatever the column's collation — right for a UTF-8 column on a Latin-1 database (#321: a `varchar` parameter takes the *database's* code page, so a Cyrillic constant would arrive as `?`), and an **Index Scan** with `CONVERT_IMPLICIT` on the column for the installation default, `SQL_Latin1_General_CP1_CI_AS`, where `'ñu'`, `'Müller'`, `'café'` read the whole index (Windows collations seek through the convert and were unaffected). Measured in `specs/065-dml-pushdown-recon/revalidation-2026-09-17.md` § 8.2.

## The rule now

A non-ASCII constant is declared `varchar` when every character of it is representable in **both** the column's code page and the database's (the parameter takes the database's, the comparison converts to the column's), `nvarchar` otherwise — exactly the old behaviour for the cases that needed it.

- `mssql::CodePageOfCollation(name)` (new `src/catalog/mssql_code_page.{hpp,cpp}`): the `_CPnnn` token of a `SQL_` collation (`CP1` → 1252), the language family of a Windows collation (`Latin1_General`… → 1252, `Cyrillic_General`… → 1251, `Polish`… → 1250, Greek/Turkish/Hebrew/Arabic/Baltic/Vietnamese/Thai families, the double-byte families recognised as 932/936/949/950), `_UTF8` → 65001, Unicode-only or unknown → 0 ("cannot tell").
- `mssql::CodePageCanEncode(page, utf8)`: ASCII is representable everywhere, UTF-8 holds everything, the single-byte pages 874 and 1250–1258 answer from tables generated from Python's `codecs` (the defined code points of bytes 0x80–0xFF); double-byte and unknown pages answer false outside ASCII, so those constants stay on the `nvarchar` path they took before.
- `MSSQLColumnInfo` carries `code_page` and `database_code_page`, computed in the constructor from the two collation names it already receives.

## Measured on the docker server (cached plans read back through `mssql_scan`)

| statement | before | after |
|---|---|---|
| `v = 'ñu'`, `SQL_Latin1_General_CP1_CI_AS` | `(@p0 nvarchar(20))`, Index Scan | `(@p0 varchar(20))`, **Index Seek** |
| `w = 'ñu'`, `Latin1_General_CI_AS` | nvarchar, seek through the convert | varchar, seek |
| `u = 'ñu'`, `Latin1_General_100_BIN2_UTF8` | nvarchar, seek | varchar, seek |
| `u = 'ыы'`, UTF-8 column, 1252 database | nvarchar, seek | **nvarchar** (#321's rule), seek |

## Tests

- `test/sql/catalog/varchar_declaration_seek.test`: the rows for all four shapes, then the four plans asserted through `sys.dm_exec_sql_text` (the declaration is the statement text's prefix) and `sys.dm_exec_query_plan` (`PhysicalOp="Index Seek"`) — rows alone cannot tell a seek from a scan.
- `test/cpp/test_filter_encoder.cpp`: the two assertions that pinned the old rule replaced by the new one — `ü` → varchar on a 1252 and on a UTF-8 column, `ы` → nvarchar on both (the database's page decides), `ü` against a Cyrillic column → nvarchar (the column's page decides too), `ы` on a Cyrillic column of a Cyrillic database → varchar, a double-byte page → nvarchar; the resolver on `SQL_…_CP1`, `SQL_Ukrainian_CP1251`, `Latin1_General_100_BIN2`, `_UTF8`, `Cyrillic_General_100_CI_AS_SC`, `Polish`, `Indic_General` (0); `€` in 1252 and 1251, `ő` in 1250 only, above-BMP nowhere but UTF-8.
- Local: C++ tests pass, `annotated_max_string.test` (#321) pass, catalog 42 / ctas 11 / copy 29 cases green.

Not in scope: the literal path (`mssql_scan_parameterize_filters = false`) still spells a non-ASCII constant `N'…'`; it is the diagnostic mode and the issue records it as a fact rather than a defect.


VGSML:
Self-review round (`/code-review`, 11 findings, each checked against the tree) — fixes in `HEAD`:

- **The code page is the server's answer now.** `CAST(COLLATIONPROPERTY(c.collation_name, 'CodePage') AS INT)` rides with the four column-metadata queries and the database's page with the ATTACH-time collation query; the name-based resolver and its family table are gone (they were incomplete by construction — Georgian, Azeri_Cyrillic, Maltese, … would have stayed on `nvarchar` in silence). Two traps met on the way, both fixed and worth knowing: `COLLATIONPROPERTY` returns `sql_variant`, which the metadata reader does not parse, so the `CAST` is load-bearing; and a row callback that returns `false` sends an ATTENTION and cancels the query — the one-row collation query returns `true`.
- **`text` columns follow the same rule** — measured: a `varchar(max)` parameter turned `'ы%'` into `'?%'` and `LIKE` matched `'?abc'`; now `nvarchar(max)` when the page cannot hold the constant.
- **A visible behaviour change on `SQL_` collations, now in the CHANGELOG** — measured: `'ß' = 'ss'` is true under an `nvarchar` parameter (Windows Unicode rules) and false under `varchar` (the collation's own sort order, what a T-SQL `varchar` literal answers).
- Dead `IsAscii` removed; `docs/query-execution.md` states the declaration rule; `DATAMODEL.md` gains the invariant; the test names its precondition (database page 1252) and DETACHes.

Found alongside, out of scope, to be filed: `WHERE t = 'x'` on a `text` column through the scan fails with server error 402 (text is not comparable with `=`; `LIKE` works) — pre-existing since 076.

Local after the round: C++ 31 pass; catalog 42 / integration 12 / ctas 11 / regression 5 / copy 29 / rowid 5 / dml 8 / insert 17 / query 9 / transaction 9 / attach 4 / tds_connection 3 — green.

oluies:
## roborev review — 3 findings (1 Medium, 2 Low)

Branch diff `634b485..0469bab` (2 commits) against `main`, roborev job 1659, agent `claude-code`.

### Medium — `test/sql/catalog/varchar_declaration_seek.test`: the plan-cache assertions can pass on a stale plan

The four `sys.dm_exec_cached_plans` checks match any cached plan whose text looks like `(@p0 varchar(20))%VarcharDeclSeek%([v] = @p0)%`, not the plans this run produced. `DROP TABLE` does not reliably evict an ad-hoc or `sp_executesql` plan, and the test reuses the same table name and statement text on every run. On a long-lived dev or CI server, a plan left over from an earlier correct run can satisfy these checks even after a regression back to `nvarchar`, which is exactly the regression the test exists to catch.

The last assertion is fragile the other way too. It expects the optimizer to choose an Index Seek through a `CONVERT_IMPLICIT` on a UTF-8 column, which another server version or different statistics may not do.

*Fix:* clear the cache state first. Either `DBCC FREEPROCCACHE` (or free just the matching plan handles) before the probes, or add a per-run marker to the table name. For the `ы` / UTF-8 case, consider asserting only the declaration prefix, not the physical operator.

### Low — `src/catalog/mssql_metadata_cache.cpp`: a comment was separated from its function

`ParseCodePage` was inserted directly below `// A bit column as the simple-query layer renders it.`. That comment now sits above `ParseCodePage`, not above `FlagIsSet`, which it describes. Move `ParseCodePage` and its own comment above it.

### Low — `src/table_scan/filter_encoder.cpp` (varchar/text branches): the behaviour changes are not pinned by a live test

On `SQL_` collations this change alters results on purpose: `'ß' = 'ss'` used to be true and is now false, as the CHANGELOG says. No integration test locks that in. The live test also does not exercise the new `text` branch, which is where the `'ы%'` → `'?%'` false match was fixed. Both changes are covered only by the C++ tests, and only at the declaration-string level.

*Fix:* extend the integration test with (a) a `text` column holding a `'?…'` row and a `LIKE 'ы%'` filter that must not match it, and (b) a `SQL_Latin1_General_CP1_CI_AS` row holding `'ss'`, queried as `'ß'`.

VGSML:
All three taken, one commit.

**Stale plans (Medium).** Right — `DROP TABLE` leaves the `sp_executesql` plans behind and the text is identical run to run. The test now frees every cached plan whose text mentions the table (a cursor over `sys.dm_exec_cached_plans` + `DBCC FREEPROCCACHE(@handle)` through `mssql_exec`), asserts the count is **0**, runs the four statements again, and only then reads the cache — so the probes are this run's. The `ы` / UTF-8 probe asserts the declaration prefix only; the operator on a UTF-8 column is the optimizer's call.

**Behaviour changes pinned live (Low).** A `TEXT` column and a row `(20003, 'ss', 'ss', 'ss', '?abc')`: `v = 'ß'` on the `SQL_` column finds nothing under the varchar parameter, while `w = 'ß'` on the Windows collation finds it — the contrast the CHANGELOG describes; `t LIKE 'ы%'` matches nothing where the `varchar(max)` parameter matched `'?abc'`, with `t LIKE '?%'` as the control.

**Comment (Low).** `ParseCodePage` is above the bit-column comment again, which is `FlagIsSet`'s.

oluies:
## roborev review of e5a59aa — Pass, no findings

Follow-up commit only (`e5a59aa`), roborev job 1663, agent `claude-code`.

This addresses all three points from the last review:

- The plan-cache probes now free every cached plan mentioning `VarcharDeclSeek`, assert that none are left, and re-run the probe statements before reading the cache, so a plan from an earlier run can't satisfy them.
- New live rows pin the documented semantic change: `'ß'` no longer matches `'ss'` on the `SQL_` varchar column, and the Windows-collation column still matches. They also pin the `text` fix: `LIKE 'ы%'` does not match `'?abc'`.
- The `ParseCodePage` comment is back on its function.

Nothing further from me on this PR.

PR #370, opened 2026-09-18
fix(#369): copy a projected key column into the rowid slot through a reinterpreting view — a debug build asserted on MSSQL_VARCHAR(n) vs VARCHAR

### LINKED ISSUE #369: Debug build asserts in PopulateRowIdVector on a composite rowid key with a string column: STRUCT child is VARCHAR, the projected column is MSSQL_VARCHAR(n)
## What happens

Found by the adversarial review of #350 and reproduced there on `main` (`634b485`) as well as on the PR's head, so it is not spec 077's: in a **debug** build, `SELECT k1, k2, rowid FROM t` over a table with a **composite key that includes a string column** fails the internal assertion `source_p.GetType() == target.GetType()` in `PopulateRowIdVector` (`src/table_scan/table_scan.cpp`). `test/sql/catalog/first_touch_round_trips.test:69` hits it. Release builds do not assert and return the rows.

## Mechanism (to confirm)

`PopulateRowIdVector` fills the composite rowid `STRUCT` by `VectorOperations::Copy` from the projected PK column vectors into the STRUCT's children. The projected column carries the type the catalog reports — `MSSQL_VARCHAR(n)` / `MSSQL_NVARCHAR(n)` under `mssql_catalog_native_types = true` (spec 060) — while the rowid STRUCT's child was built from the column's plain `duckdb_type`, `VARCHAR`. Same physical vector, two `LogicalType`s that differ by the extension alias; the copy asserts on the mismatch. With `mssql_catalog_native_types = false` the types agree and the assertion cannot fire.

## Fix direction

Build the composite rowid's child types from the same reported type the scan uses (`NativeDuckDBType()` when native types are on), or copy the child through a reinterpreting reference rather than a typed `Copy` — the extension-type rule already in the repo is "extension types must agree everywhere; ReinterpretCast, not NopCast". Add the composite-string-key shape to a test that runs in the debug CI lane (the concurrency job builds debug), since release builds hide it.

VGSML:
Closes #369 — the debug-build assertion the adversarial review of #350 found on `main`.

## What was wrong

With `mssql_catalog_native_types` on (the default, spec 060) a projected string key column is `MSSQL_VARCHAR(n)` / `MSSQL_NVARCHAR(n)`, while the rowid — the scalar one and every child of a composite key's STRUCT — is typed from the column's plain `duckdb_type`. `PopulateRowIdVector` copied the projected vector into the rowid slot with a typed `VectorOperations::Copy`; a debug build refuses that (`source_p.GetType() == target.GetType()`, `vector_copy.cpp:20`), a release build performed it on the same bytes and returned the rows — which is why nothing noticed.

Reproduced on a sanitizer-free debug build (`DISABLE_SANITIZER=1 make debug`): `test/sql/catalog/first_touch_round_trips.test:69` (the reviewer's shape) and the new test both assert before the change and pass after.

## The fix

`CopyKeyColumn` at the three copy sites (two STRUCT children, the scalar rowid): a plain `Copy` when the logical types agree, otherwise a view of the source in the target's type (`Vector::Reinterpret` — the same bytes, no cast) and the copy through it. The extension-type rule the repo already follows: reinterpret, never a no-op cast.

## Tests

- `test/sql/rowid/rowid_native_types.test` (55 assertions): `typeof` of the column vs the rowid, the composite key with the string column projected beside rowid (the asserting shape), each key column projected alone, a scalar `nvarchar` key projected beside rowid, UPDATE/DELETE through those rowids, and the same shapes with native types off.
- **The rowid suite now runs against the debug build in the concurrency workflow** — a release build never says, so the class stays caught only there. Locally under debug: rowid 311 assertions / 9 files green; release: rowid 311, dml 258, catalog 1824 green.


VGSML:
Self-review round (`/code-review`, 7 findings, each checked) — fixes in `HEAD`:

- **The debug-lane step would have failed on its first run**: this job's TestDB is only `CREATE`d, never seeded by `docker/init/init.sql`, and five of the nine rowid files read init.sql's tables. `make test-rowid-debug` now runs the four self-contained files (`rowid_native_types`, `rowid_unique_index`, `rowid_key_types`, `rowid_discovery_failure_cached`) through the debug tree's sqllogictest driver, with the Makefile's exported `MSSQL_TESTDB_DSN` rather than a fourth inline copy of it. Verified locally against the debug tree: 55 / 68 / 26 / 24 assertions green.
- **The workflow's `paths` filter** did not include `src/table_scan/**`, so the lane would not have run for the very file the fix lives in; it does now, with `test/sql/rowid/**`.
- The test's alias `nt` collided with `native_types.test`'s; it is `rowid_nt`.

Considered and deferred, not lost: the altitude point that the rowid is the one DuckDB-visible type spec 060's "same variant everywhere" rule missed — typing `rowid_type` from the same native predicate would remove the reinterpreting copy at the source, at the cost of `typeof(rowid)` changing to carry `MSSQL_VARCHAR(n)` children and struct-literal filters against rowid needing the registered cast to apply per child. That is its own decision; the reinterpreting copy is correct at all three sites today and the debug lane guards a fourth. Whether an unrelated sanitizer report surfaces the first time a sqllogictest file runs under this job's ASan/UBSan build is something this PR's CI answers rather than pre-empts.

oluies:
## Review — 1 Low; the fix verified locally

Branch diff `a8bfa1e..5b29ea2` (2 commits). roborev job run with agent `claude-code`, plus a local check.

### Verified
I built `5b29ea2` debug against the pinned DuckDB (`62ee922d`) and ran it against the docker SQL Server:
- `first_touch_round_trips.test`, which asserted on `main`: 24 assertions, green.
- `test/sql/rowid/*`: 311 assertions in 9 files, green.
- `make test-rowid-debug`: all four files green, 55 / 68 / 26 / 24 assertions.
- The minimal repro from #369 (`SELECT k2, rowid` over a `(int, varchar(10))` key) now returns `k2 mssql_varchar(10)` next to `struct(k1 integer, k2 varchar)` instead of asserting.

**The approach is sound at this pin.** `Vector::Reinterpret` references the source buffer but keeps the view's own `type` member. A dictionary source is re-minted with the view's type, so dictionary vectors are handled too. The `D_ASSERT` on `InternalType()` catches any physical-type mismatch before the reinterpret. `CopyKeyColumn` covers all three typed copy sites in the scan, and no other `VectorOperations::Copy` remains in `src/`.

### Low — `src/table_scan/table_scan.cpp:636-637`: a comment was separated from its function
The new `CopyKeyColumn` and its comment were inserted below the existing `// Helper function to populate rowid vector from PK columns`. That line now sits above `CopyKeyColumn`, not above `PopulateRowIdVector`, which it describes. Move it down, or put `CopyKeyColumn` above it.

Nothing else from me. Running the self-contained rowid files in the debug lane is the right fix for "a release build never tells you".

VGSML:
Taken — the `// Helper function to populate rowid vector from PK columns` line is back above `PopulateRowIdVector` (`CopyKeyColumn` keeps its own comment). Thank you for building it debug against the pin and running the repro and the lane yourself.

PR #372, opened 2026-09-21
fix(#325): boolean ATTACH options accept a string value, as DuckLake's METADATA_PARAMETERS sends them

### LINKED ISSUE #325: Boolean ATTACH options cannot be set through DuckLake's METADATA_PARAMETERS: "Could not convert string 'true' to INT8"
DuckLake forwards `METADATA_PARAMETERS` to the inner `ATTACH` it builds for the metadata catalog, rendering each value with `Value::ToSQLString()` — so a value written in SQL as a string arrives at the mssql extension as a `VARCHAR`. String-valued options survive that; boolean ones do not.

## Repro

```sql
LOAD mssql; LOAD mssql_ducklake;

-- works
ATTACH 'ducklake:mssql:Server=…;Database=lake_meta;User Id=…;Password=…' AS lake
  (DATA_PATH '/tmp/lake', METADATA_PARAMETERS MAP{'schema_filter': 'dbo'});

-- fails
ATTACH 'ducklake:mssql:Server=…;Database=lake_meta;User Id=…;Password=…' AS lake
  (DATA_PATH '/tmp/lake', METADATA_PARAMETERS MAP{'lazy_validation': 'true'});
```

```
Invalid Input Error: Failed to attach DuckLake MetaData "__ducklake_metadata_lake" at path
+ "mssql:Server=…" Could not convert string 'true' to INT8
```

`MAP{...}` in DuckDB is `MAP(VARCHAR, VARCHAR)`, so there is no way to write a genuine BOOLEAN there — every value in that map is a string by construction. The option parser reads it with `GetValue<bool>()`, which for a `VARCHAR` goes through an integer cast and fails on `'true'`.

## Affected options

Every boolean one, as far as I can tell from the parser in `src/mssql_storage.cpp`: `lazy_validation` / `lazyvalidation`, `catalog`, `order_pushdown`. The string-valued ones (`schema_filter`, `table_filter`, `application_name`, `secret`) are fine — `schema_filter` is confirmed working through the same path.

## Why it matters

`METADATA_PARAMETERS` is the only channel a DuckLake user has to the inner mssql `ATTACH`, so today none of the boolean options are reachable for a `ducklake:mssql:` catalog. `lazy_validation` is the one I was after: it removes one TDS login from the attach (~0.15s of ~0.51s — see #324).

## Suggested fix

Accept the usual string spellings when the incoming `Value` is a `VARCHAR`: `true`/`false`, `1`/`0`, `yes`/`no`, case-insensitively, before falling back to `GetValue<bool>()`. `Value::DefaultCastAs(LogicalType::BOOLEAN)` does that conversion correctly for `'true'` where the integer path does not.

VGSML:
Closes #325.

## What was wrong

`lazy_validation`, `catalog` and `order_pushdown` were read with `Value::GetValue<bool>()`. On the 2.0 line that reads a `VARCHAR` through an `int8` cast, so `lazy_validation 'true'` failed:

```
Could not convert string 'true' to INT8
```

DuckLake's `METADATA_PARAMETERS` is a `MAP(VARCHAR, VARCHAR)`, and it forwards every value to the inner `ATTACH` as a quoted string. So none of the boolean options could be set for a `ducklake:mssql:` catalog.

## The fix

- A `VARCHAR` value goes through DuckDB's own string-to-boolean cast, the one `SET` uses: `true`/`false`, `t`/`f`, `yes`/`no`, `y`/`n`, `1`/`0`, any case.
- Any other string is refused, and the error names the option.
- A `BOOLEAN` value is read as before.
- There is no NULL branch. DuckDB's binder refuses a NULL for every `ATTACH` option before the extension sees it, and the test pins that.

## Test

`test/sql/attach/attach_boolean_string_options.test` needs no server. The host does not resolve, so an eager `ATTACH` fails and a lazy one succeeds. The test therefore checks that a value is honoured, not only that it parses. It covers:

- the string DuckLake sends, `lazy_validation 'true'`;
- `'YES'` through the `LazyValidation` alias, `'1'` and `'t'`;
- `'false'`, which turns eager validation back on;
- `catalog 'false'`, which really disables the catalog;
- a non-boolean string, refused with the option named;
- NULL, refused by DuckDB;
- a genuine `BOOLEAN`, which still works.

Verified to fail on the unfixed build with the issue's error. All six `test/sql/attach/` files pass against a local SQL Server.

Also a CHANGELOG entry, and one line in CLAUDE.md under the ATTACH options table. `DATAMODEL.md` is unaffected: no layer, invariant or flow changes.


VGSML:
## Self-review of a31a892: no changes needed

I checked each claim in the PR against the build:

- **The cast is DuckDB's own.** `'y'::BOOLEAN`, `'NO'::BOOLEAN`, `'1'::BOOLEAN` and `'T'::BOOLEAN` give true/false/true/true. `SET mssql_connection_cache = 'yes'` / `'n'` set true/false. The string path accepts exactly these spellings, so the comment and CHANGELOG are accurate when they say "the one `SET` uses".
- **The secret path never had this bug.** `CREATE SECRET … catalog 'false', use_encrypt 'no'` stores both as false, and an `ATTACH … SECRET s_bool` then refuses `mssql_refresh_cache()` with the catalog=false message. The secret reader also calls `GetValue<bool>()`, but DuckDB casts typed secret parameters (`LogicalType::BOOLEAN` in `mssql_secret.cpp`) at CREATE SECRET time. Only ATTACH options arrive untyped.
- **The value shape matches DuckLake's.** DuckLake renders each `METADATA_PARAMETERS` value with `Value::ToSQLString()`, so the inner ATTACH reads `lazy_validation 'true'`: a VARCHAR literal. That is the first statement of the test, and it failed on the unfixed build with `Could not convert string 'true' to INT8`.
- **No NULL branch.** DuckDB's binder refuses `catalog NULL` for every ATTACH option (`NULL is not supported as a valid option for ATTACH option`), and the test pins that. If DuckDB ever lets a NULL through, the test fails and says so.
- **CI runs the test.** `scripts/ci/integration_test.sh` selects every `test/sql/**/*.test` file except azure/fabric, by suffix against the registered list, on pull requests. The test needs no server, so it also passes without `MSSQL_*` variables.

**Accepted as is:**
- `order_pushdown 'no'` is checked only for parsing. Its effect needs a server and a plan to inspect, and the parser is shared with the other two options, which are checked by behaviour.
- The error text lists `true/false, yes/no, 1/0` and leaves out `t/f` and `y/n`. It is a hint, not the full grammar.

oluies (approved):
## Approved, with one Low

Branch diff `8eeaea8..a31a892` (1 commit), roborev job 1704, agent `claude-code`.

`lazy_validation`, `catalog` and `order_pushdown` now accept a string value. It goes through DuckDB's own string-to-boolean cast, and a value that isn't a boolean fails with an error naming the option. That is what DuckLake's `METADATA_PARAMETERS` pass-through needs (#325).

### Low — `test/sql/attach/attach_boolean_string_options.test:29-32`: `order_pushdown` is shown to be accepted, not honoured
The test passes `order_pushdown 'no'` but never checks that it has an effect. The ATTACH would succeed just the same if the string were parsed as `true`, or if the option were ignored. `lazy_validation` and `catalog` each get a check that tells their two values apart; `order_pushdown` doesn't.

*Fix:* in a server-backed test, tell `'true'` from `'false'` for `order_pushdown`, for example through the generated SQL or the `EXPLAIN` of an `ORDER BY` over an attached table. If that's not worth it, reword the comment so it doesn't claim `order_pushdown` is covered.

PR #373, opened 2026-09-21
fix(#351): make docker-up works on a fresh clone — .env optional, one default password, health read from the health field

### LINKED ISSUE #351: `make docker-up` fails on a fresh clone: compose requires a `.env` that nothing creates or documents
## What happens

On a fresh clone or a new git worktree, the documented first step of the test setup fails outright:

```
$ make docker-up
Starting SQL Server test container...
docker compose -f docker/docker-compose.yml up -d sqlserver
env file /path/to/repo/.env not found: stat /path/to/repo/.env: no such file or directory
make: *** [docker-up] Error 1
```

Nothing is started, and the message points at a file the repo never tells you to create.

## Why

- `docker/docker-compose.yml` declares `env_file: ../.env` on **both** services (`sqlserver` and `sqlserver-init`). In the Compose spec `env_file` is a hard requirement unless marked otherwise, so a missing file aborts the command.
- `.env` is gitignored (`.gitignore:48`), so it is never present in a fresh checkout.
- `.env.example` **is** tracked, but no doc instructs you to copy it. `docs/TESTING.md` says `make docker-up` in four places (lines 43, 100, 1155, 1508) and `CLAUDE.md` twice (lines 70, 76); none mention `.env`.
- The only `.env` reference in the docs is `docs/TESTING.md:144`, and it is about `make azure-test` passing variables through — not about the file being mandatory for `docker-up`.

So the documented path is broken for anyone who has not previously created the file by hand. It is invisible to regulars because `.env` persists locally, and invisible to CI if CI does not go through this target.

## Repro

```bash
git worktree add /tmp/wt HEAD   # or a fresh clone
cd /tmp/wt
make docker-up                  # fails; no container starts
```

## Suggested fixes

Any one of these; the first is the smallest:

1. Mark the file optional in `docker/docker-compose.yml` — Compose supports `env_file: [{path: ../.env, required: false}]`. All the variables already have `${VAR:-default}` fallbacks in the same file, so the defaults work without it.
2. Have `docker-up` create it: `@[ -f .env ] || cp .env.example .env` as the target's first line.
3. Document `cp .env.example .env` as step 0 in `docs/TESTING.md` and `CLAUDE.md`.

(1) plus (3) would be my preference — the compose defaults are already correct, so the file should not be load-bearing just to exist.

## Adjacent, not part of this issue

`MSSQL_TEST_PORT` defaults to `1433`. On a machine where any other project already publishes a SQL Server on 1433, the suite authenticates against that server instead and reports a wall of "Authentication failed for user 'sa'" that reads exactly like a code regression. Setting `MSSQL_TEST_PORT` in `.env` fixes it for both compose and make in one place — which is a good reason for `.env` to be easy to create.

VGSML:
Closes #351.

The documented first step of the test setup, `make docker-up`, had three defects.

## 1. `.env` was required but never present

`docker/docker-compose.yml` declared `env_file: ../.env` on both services. The Compose spec treats that as a hard requirement, and `.env` is gitignored, so a fresh clone failed before starting anything:

```
env file .../.env not found: stat .../.env: no such file or directory
```

It is now `required: false`, which needs Compose 2.24 or newer. Every variable the file reads already has a default.

## 2. Two different default passwords

Without `.env`, the SA password defaulted to `TestPassword1` and the healthcheck to `TPassw0rd!`. So compose run directly, as the file header documents, left the server unhealthy for good. Both now default to `TestPassword1`, as the Makefile does. Through `make` they already matched, because the Makefile exports its own default.

## 3. "unhealthy" was reported as ready

`make docker-up` detected health with `grep -q "healthy"`, which also matches `(unhealthy)`. A server whose healthcheck kept failing was reported ready, and the 120 s timeout fell through silently. The loop now reads `docker compose ps --format '{{.Health}}'`. It fails on `unhealthy` or on the timeout, and the message names the logs command. The warning in `integration-test` reads the same field.

## Docs

- `docs/TESTING.md` says `.env` is optional and how to create one.
- It also explains why the port is the usual reason to: another SQL Server on 1433 answers `Login failed for user 'sa'`, which looks like a regression but is not.
- `CLAUDE.md` gets one line about the same.

## Verification

All runs used a clean worktree with no `.env`, and their own project name, container names and port:

- Before the fix, `docker compose config` failed with the error above.
- `make docker-up` exited 0 in 18 s. Init seeded TestDB, and the extension read `LargeTable`'s 150000 rows with the default password.
- With the healthcheck overridden to `exit 1`, `make docker-up` exited 1 with the unhealthy message. The old loop printed "SQL Server is ready!" here.

Not in scope: `make` on a clone made without `--recursive` still fails earlier, on the missing `extension-ci-tools` submodule.


VGSML:
## Self-review of 3009f01 — no changes needed

- **CI does not use this compose file.** The integration job runs SQL Server as a service container, and the Kerberos and linux-ci stacks have their own compose files. So `required: false` sets the Compose ≥ 2.24 floor for local machines only. Compose 2.24 was released in January 2024, and I tested on 2.40.
- **The lint job passes.** yamllint with the lint job's ruleset (`relaxed` minus line-length/truthy/comments) is clean on `docker/docker-compose.yml`.
- **The loop checked both ways out.** On `healthy` it breaks with time remaining, so the timeout check after the loop does not fire. A container with no healthcheck reports an empty health string, so it waits and then fails on the timeout. The old loop proceeded silently in that case.
- **`docker-status` is unaffected.** It still addresses the container as `mssql-dev`, the fixed `container_name`.
- **No CHANGELOG line.** This is test infrastructure, and extension users never run it.

VGSML:
## Self-review, second pass: one Low, fixed in 747fd22

The long `env_file` form (`path:` / `required: false`) needs Docker Compose 2.24 or newer, and an older Compose cannot parse the file. `docs/TESTING.md` listed "Docker Compose" with no version. It now names v2.24+ and says why.

The earlier points still hold:
- CI does not use this compose file.
- The lint job's yamllint is clean.
- Both exits of the loop were exercised: `healthy` proceeds, `unhealthy` exits 1, and a missing health status waits and then fails on the timeout.

oluies (approved):
## Approved — no findings

Branch diff `8eeaea8..747fd22` (2 commits), roborev job 1705, agent `claude-code`.

`make docker-up` now works on a fresh clone:
- `.env` is optional, via the long-form `env_file` entry, which needs Compose 2.24+.
- The healthcheck's fallback password matches the container's default (`TestPassword1`).
- The wait loop reads the container's `Health` field instead of grepping for "healthy", which also matched "(unhealthy)". It now fails fast when the container turns unhealthy, or when it isn't healthy within 120 s.
- The docs match.

PR #374, opened 2026-09-21
feat(#371): document every mssql function in duckdb_functions() — description, examples, category, scalar parameter names

### LINKED ISSUE #371: Docs: make mssql's functions discoverable from duckdb_functions()
Hi! I'm working on helping AI agents and tools discover and correctly use DuckDB community extensions. There are 346 in the catalog and agents mostly can't find them.

This will be a multi-step effort – types, tables and other catalog objects need the same treatment – but functions are by far the biggest surface, so that's where I'm starting.

The reason is that an agent connected to a database can only learn what an extension does by querying `duckdb_functions()`. Your README and the `extended_description` in [community-extensions](https://github.com/duckdb/community-extensions) aren't reachable from a SQL connection, and are usually too large to fit in an agent's context anyway. Per-function metadata is – it's searchable one function at a time.

Here's what `mssql` exposes there today (14 scalar, 2 table):

| In `duckdb_functions()` | mssql |
|---|---|
| Functions registered | **16** |
| …with a `description` | 0 (0.0%) |
| …with an `examples` entry | 0 (0.0%) |
| …with real parameter names, not `col0` | 0 / 14 (0.0%) |

Catalog-wide that description figure is 29.6%, so this is an ecosystem gap rather than anything specific to `mssql`. It's mostly an API-shape problem: `RegisterFunction(ScalarFunction)`, the overload nearly everyone uses, has nowhere to put a description.

### The fix

```cpp
// before
loader.RegisterFunction(mssql_azure_auth_test_function);

// after
CreateScalarFunctionInfo info(mssql_azure_auth_test_function);
FunctionDescription desc;
desc.parameter_names = {"arg1"};
desc.description     = "…one sentence on what mssql_azure_auth_test does…";
desc.examples        = {"mssql_azure_auth_test(…)"};
desc.categories      = {"…"};
info.descriptions.push_back(desc);
loader.RegisterFunction(std::move(info));
```

`CreateScalarFunctionInfo`, `CreateAggregateFunctionInfo` and `CreateTableFunctionInfo` all work the same way. The definitions, if it's easier to read them directly: [`FunctionDescription`](https://github.com/duckdb/duckdb/blob/v1.5.5/src/include/duckdb/parser/parsed_data/create_function_info.hpp#L15-L26) and the [`RegisterFunction` overloads](https://github.com/duckdb/duckdb/blob/v1.5.5/src/include/duckdb/main/extension/extension_loader.hpp#L45-L57).

These numbers are from the 1.5.x line (1.5.5). I know a lot of you are getting ready for 2.0 – `FunctionDescription` and these overloads are unchanged there, so this can fold into that work rather than be a separate task.

**If you only do one thing, do parameter names** – they turn `f(col0, col1)` into a signature a caller can reason about.

This is also a good task to hand to a coding agent. The change is mechanical and repetitive, and an agent with your source in context can usually convert every registration in one pass.

<details>
<summary>A prompt that works</summary>

```
In this DuckDB extension, functions are registered with the bare
loader.RegisterFunction(<fn>) overload, which cannot carry documentation.

Convert each registration to the matching CreateScalarFunctionInfo,
CreateAggregateFunctionInfo or CreateTableFunctionInfo form and attach a
FunctionDescription with:

  - parameter_names: the real argument names, never col0/col1
  - description:     one sentence on what the function does
  - examples:        one runnable call. A bare expression for scalar and
                     aggregate functions; a full "SELECT * FROM f(...)"
                     statement for table functions, since a table function
                     used as a bare expression is a binder error.
  - categories:      a short tag or two

Also set info.on_conflict = OnCreateConflict::ALTER_ON_CONFLICT, because
that is what the bare RegisterFunction overload does internally and the
default on CreateInfo is ERROR_ON_CONFLICT.

Infer each description from the implementation and any existing README or
docs. Do not invent behaviour you cannot verify from the source - leave a
TODO instead.
```
</details>

<details>
<summary>A note on <code>examples</code> formatting</summary>

Core DuckDB writes scalar and aggregate examples as a bare expression – `date_trunc('hour', TIMESTAMPTZ '1992-09-20 20:38:40')` – and all 2,076 documented built-ins follow that. Table functions need a full statement (`SELECT * FROM my_func(…)`), since using one as a bare expression is a binder error. `spatial` uses full statements throughout with the result as a trailing comment, which is the most useful form for an agent.
</details>

<details>
<summary>16 functions have no `description` and 14 have placeholder parameter names – reproduce these numbers yourself</summary>

```sql
CREATE TABLE _before AS SELECT function_name, function_type FROM duckdb_functions();
LOAD mssql;
SELECT function_name, function_type, parameters, return_type,
       description IS NOT NULL AS documented,
       len(examples) > 0        AS has_example
FROM duckdb_functions()
WHERE (function_name, function_type) NOT IN (SELECT * FROM _before)
ORDER BY documented, function_name;
```
</details>

### Done well elsewhere

| Extension | Coverage | Source |
|---|---|---|
| `spatial` | 268/280 described, 270 with examples | [duckdb/duckdb-spatial](https://github.com/duckdb/duckdb-spatial) |
| `stochastic` | 444/444 described, all with examples | [query-farm/stochastic](https://github.com/query-farm/stochastic) |
| `anofox_tabular` | 157/159 described, 99% real parameter names | [DataZooDE/anofox-tabular](https://github.com/DataZooDE/anofox-tabular) |
| `inflector` | 30/30 described, with categories | [query-farm/inflector](https://github.com/query-farm/inflector) |

---
*From loading all 346 community extensions against DuckDB 1.5.5 and reading `duckdb_functions()`.*

– Rusty, [Query.Farm](https://query.farm)

VGSML:
Closes #371.

`duckdb_functions()` is all a client connected to the database can read about an extension's functions. Before this change none of ours had a description, and every parameter was listed as `col0`, `col1`, …

## What changed

All 14 functions (22 overloads) now carry a description, runnable examples and a category. Registration goes through one helper, `mssql::RegisterDocumentedFunction` in `src/mssql_function_docs.cpp`, which registers all of a function's overloads at once:

```cpp
mssql::RegisterDocumentedFunction(
    loader, {MSSQLExecScalarFunction::GetFunction()},
    {{"context", "sql"},
     "Runs a T-SQL statement or batch on an attached SQL Server database and returns the number of rows it affected.",
     {"mssql_exec('db', 'CREATE TABLE dbo.notes (id int PRIMARY KEY, body nvarchar(200))')"},
     {"query"}});
```

## Where the 2.0 line differs from the recipe in the issue

- **Scalar parameter names go into the function signature.** `duckdb_functions()` reads them from there, and uses `FunctionDescription::parameter_names` only when the signature has none.
- **The binder uses the same names for named arguments.** `mssql_exec(context := 'db', sql := '…')` now binds, and binder errors list `mssql_exec(context VARCHAR, "sql" VARCHAR)`. That makes the names API, and CLAUDE.md now says so.
- **`context` is used throughout.** The docs used `context_name`, `context` and `catalog` for the same argument. `architecture.md`, `query-execution.md` and `catalog-integration.md` now agree.
- **Table functions keep `col0`, `col1`, … for positional parameters.** DuckDB's `duckdb_functions()` names those by position whatever the registration says. Named parameters such as `prepared` show by name.
- **One registration per function, not per overload.** DuckDB applies a function's only description to every overload. With several descriptions it matches them by parameter types, and a description without types then reaches no overload that takes arguments.

## Examples are safe to run as written

An agent may run an example verbatim, so none of them delete or update data:
- `mssql_exec` and `mssql_exec_params` create a table and insert into it.
- `mssql_scan_params` reads system views, so it works on any server.

Every example was run against a server.

## Tests

- `test/sql/mssql_function_docs.test` needs no server.
  - It fails if any function lacks a description, example or category.
  - It fails if a scalar parameter is still named `col…`.
  - It checks that named arguments bind, and that an unknown name lists the real parameter names.
- The full SQL suite is green locally: 195 files, 17 skipped by their environment gates.


VGSML:
## Self-review of d63da95: three findings, fixed in 9affd5b

### Medium: `mssql_preload_catalog` took its optional schema as a vararg

- A vararg cannot be named, so `schema` did not appear in `duckdb_functions()`. It also could not be passed as `schema := …`.
- A positional argument after a named one is rejected, so `mssql_preload_catalog(context := 'db', 'dbo')` failed.
- A third argument was accepted and silently ignored: `mssql_preload_catalog('db', 'a', 'b')`.

**Fixed:** two overloads, `(context)` and `(context, schema)`, the same shape as `mssql_invalidate_cache`. A third argument is now refused. Both behaviours are pinned in `mssql_function_docs.test`. `preload_catalog.test` and `issue_317.test` pass against a server. CHANGELOG records the change, and the overload count is corrected from 22 to 23.

### Low: the `mssql_azure_auth_test` description was inaccurate

It said the function returns "the start of the token". `TruncateToken` actually returns the first 10 and last 3 characters plus the length. The description now says so.

### Low: CLAUDE.md overstated the helper's reach

It said "every function is registered through `RegisterDocumentedFunction`". The COPY function and the secret type are registered differently, so it now says "every scalar and table function".

### Checked, no change

- **Named arguments reach bind in declaration order.** `mssql_exec_params(params := …, statement := …, context := 'nope')` reaches bind with `context` resolved, and fails with the function's own "Unknown context 'nope'".
- **Named arguments do not change how a bound function is identified.** It is identified by name and argument types; the parameter names play no part, so plans and views are unaffected.
- **A short name list cannot slip through.** The helper's loop stops at the shorter of the signature and the name list, so a missing name would leave a `col…` behind. The test fails on any `col…` in a scalar mssql function.
- **Windows builds are dispatched.** MSVC and MinGW run in run 35597446576, because the SSPI file's registration compiles on every platform.

Found while running the examples: `mssql_preload_catalog`'s summary counts are wrong, filed as #375 (not in this PR).

oluies (approved):
## Approved — no findings

Branch diff `8eeaea8..9affd5b` (2 commits), roborev job 1706, agent `claude-code`.

- Every mssql scalar and table function now goes through `mssql::RegisterDocumentedFunction`, so `duckdb_functions()` shows a description, examples and a category for each one, and scalar parameters get real names.
- `mssql_preload_catalog`'s optional schema is now a second overload rather than a vararg.
- A test that needs no server pins all of it.

PR #377, opened 2026-09-21
fix(#376, #375): mssql_preload_catalog loads the schema list, marks only what it loaded, and reports it

### LINKED ISSUE #375: mssql_preload_catalog's summary mixes two meanings: "Preloaded schema 'dbo': 0 tables, 358 columns"
## What happens

The status string `mssql_preload_catalog()` returns counts some things as "new to the cache" and others as "loaded by this call". The mix also differs between the two paths. So any call except the very first one on a fresh ATTACH reports numbers that contradict each other. Measured on `main` against the docker `TestDB`:

```sql
SELECT mssql_preload_catalog('db');            -- Preloaded 5 schemas, 66 tables, 374 columns   (fresh ATTACH)
SELECT mssql_preload_catalog('db');            -- Preloaded 0 schemas, 66 tables, 374 columns
SELECT mssql_preload_catalog('db', 'dbo');     -- Preloaded schema 'dbo': 0 tables, 358 columns
SELECT mssql_invalidate_cache('db');           -- true
SELECT mssql_preload_catalog('db', 'dbo');     -- Preloaded schema 'dbo': 0 tables, 358 columns
```

"0 tables, 358 columns" is what an agent or a script reading the result sees after it has just loaded the whole of `dbo`. Invalidating first does not help, because invalidation marks entries stale and keeps them, so the tables are still not "new".

## Why

| path | schemas | tables | columns |
|---|---|---|---|
| whole catalog (`LoadAllSchemasMetadata`) | new to the cache | every table in the answer | every column in the answer |
| one schema (`BulkLoadAll` per-schema loop) | — (not reported) | new to the cache | every column in the answer |

The per-schema loop keeps the "new to the cache" rule on purpose. The comment next to `table_count++` says it is "what the preload status message has always reported". The whole-catalog path stopped doing that for tables when spec 071 W2 replaced its loop with one query, but still does it for schemas. Columns were never counted that way on either path.

## Fix direction

Report what this call loaded, on both paths and for every counter: the schemas covered, the tables and the columns in the answer. That is the only reading under which the numbers are stable and add up, and it is what "Preloaded …" says. Drop the "new to the cache" comparisons, and update the comment that defends them.

A test belongs next to `test/sql/catalog/preload_catalog.test`, which today only asserts `LIKE 'Preloaded%'`. It should:

- create a schema with a known number of tables and columns;
- preload it twice, once through each path;
- assert the exact numbers every time, including after an earlier preload.

Found while running the examples of #371's `duckdb_functions()` descriptions.

### LINKED ISSUE #376: mssql_preload_catalog never loads the schema list: after ('db', 'dbo') every other schema is gone; after ('db') the first query loads the whole catalog again
## What happens

Measured on `main` against the docker `TestDB`, which has five schemas: `dbo`, `test`, `My Schema`, `SELECT` and `schema"quote`.

**Per-schema preload hides every other schema for the rest of the session:**

```sql
ATTACH '…TestDB…' AS db (TYPE mssql);
SELECT mssql_preload_catalog('db', 'dbo');   -- Preloaded schema 'dbo': 61 tables, 358 columns
SELECT string_agg(schema_name, ',') FROM duckdb_schemas() WHERE database_name = 'db';   -- dbo
SELECT count(*) FROM db.test.test;
-- Catalog Error: Table with name "test.test" does not exist because schema "test" does not exist.
```

Without the preload, the same catalog lists all five schemas and the query returns 5 rows. Only `mssql_invalidate_cache('db')` brings the other schemas back.

**Whole-catalog preload is thrown away by the first catalog access.** With `MSSQL_DEBUG=1`, a fresh ATTACH followed by `mssql_preload_catalog('db')` and then `duckdb_tables()` logs this:

```
LoadAllSchemasMetadata — 5 schema(s), 66 table(s), 374 column(s) in ONE query    <- the preload
EnsureSchemasLoaded — loading schemas from SQL Server                            <- first catalog access: schemas_.clear()
LoadAllTableMetadata('dbo') — loading every schema in one query
LoadAllSchemasMetadata — 5 schema(s), 66 table(s), 374 column(s) in ONE query    <- the same load again
```

The call that exists to load the catalog up front costs one extra whole-catalog query instead of saving one.

## Why

Neither preload path loads the schema list (`SCHEMA_DISCOVERY_SQL`, which `EnsureSchemasLoaded` runs).

- **The per-schema path** (the `BulkLoadAll` loop) ends by setting `schemas_load_state_ = LOADED` while `schemas_` holds only the schema it was given. It also sets `tables_load_state` and every table's `columns_load_state` to LOADED on **every** schema in the cache. So if the schema list had been loaded before, the other schemas would instead turn up with empty table lists, or tables with no columns, marked complete.
- **The whole-catalog path** (`LoadAllSchemasMetadata`) publishes tables into `schemas_` but leaves `schemas_load_state_` untouched. The next `EnsureSchemasLoaded` sees NOT_LOADED and runs `schemas_.clear()`, and the whole load starts again.

## Fix direction

`BulkLoadAll` runs `EnsureSchemasLoaded` first, on both paths. That is one light query, and a no-op when the list is already loaded. After it:

- the schema list is actually loaded, so nothing needs to pretend it is;
- the per-schema path marks only what it loaded: the target schema's table list and the columns of the tables in its answer;
- a schema that is not in the list, because it is absent from the database or excluded by `schema_filter`, gets no phantom entry.

The summary counters of #375 change in the same place: once the list is loaded first, a "new to the cache" schema count is always 0. Both issues belong in one PR.

Found while fixing #375.

VGSML:
Closes #376. Closes #375.

## What was wrong

Neither preload path loaded the schema list. This caused two defects (#376), and a third (#375) is fixed in the same place.

**Per-schema preload hid every other schema for the session.** On a fresh ATTACH, `mssql_preload_catalog('db', 'dbo')` marked the schema list LOADED while it held only `dbo`:

```
SELECT count(*) FROM db.test.test;
-- Catalog Error: Table with name "test.test" does not exist because schema "test" does not exist.
```

The same code also marked every schema's table list and every table's columns LOADED. A schema that had one table looked up before the preload therefore came out holding that one table.

**Whole-catalog preload was thrown away.** `mssql_preload_catalog('db')` left the list NOT_LOADED. The first catalog access afterwards ran `EnsureSchemasLoaded`, which clears `schemas_`, and loaded the whole catalog again. With `MSSQL_DEBUG=1` the whole-catalog query appears twice.

**The summary counters meant different things (#375).** Some counted what was new to the cache, others what this call loaded. A second preload therefore said `Preloaded schema 'dbo': 0 tables, 358 columns`.

## The fix (`BulkLoadAll`)

- **The schema list loads first, on both paths.** `EnsureSchemasLoaded` is one light query, and a no-op when the list is already loaded.
- **The per-schema path marks only what it loaded.** That is the target schema's table list and the columns of the tables in its answer.
- **A schema the catalog does not show loads nothing.** Whether it is absent from the database or hidden by `schema_filter`, it gets no phantom entry.
- **Every counter reports what the call loaded.** The numbers are the same on every call. Once the list is loaded first, a "new to the cache" schema count would always be 0 anyway.
- **The dead schema-discovery branch of the old per-schema loop is removed.**

Ignoring whitespace the diff is +34 / −62 lines (`git diff -w`). The larger raw diff comes from removing one indentation level from the old loop body.

## Tests

`test/sql/regression/issue_375_376.test` has four parts. I ran each part on its own against the pre-fix build, and each failed on the assertion it targets:

1. **Exact counts** on a schema of its own, on both paths and across `mssql_invalidate_cache`.
2. **A per-schema preload on a fresh ATTACH keeps `dbo`,** and its tables stay readable.
3. **A schema with one table looked up before the preload** still loads its full table list afterwards.
4. **A whole preload is served from the cache.** `mssql_test_fail_metadata_after_rows = 1` makes any metadata query throw. Counting pool acquires cannot show this, because the scan path takes a connection before it finds everything already loaded, and sends nothing on it.

The full SQL suite is green locally (195 files, 17 skipped by their environment gates), and so are the C++ unit tests (31/31).

## Docs

- The preload section of `docs/catalog-integration.md` is rewritten. It still described the per-schema loop that spec 071 removed.
- CHANGELOG has an entry.
- `DATAMODEL.md` does not describe preload, so it is unchanged.


VGSML:
## Self-review of b244a20: no changes needed

- **Schema name case matches the rest of the catalog.** The catalog already resolves schema names case-sensitively: `db.DBO.TestSimplePK` fails with `schema "DBO" does not exist` without any preload. The old per-schema path handled `mssql_preload_catalog('db', 'DBO')` differently. It created a phantom `DBO` entry holding `dbo`'s tables, because the server query with `@s = N'DBO'` matches `dbo` under a case-insensitive collation. Now `DBO` is not in the list, so nothing is loaded and nothing is invented.
- **Filtered schemas never reach the whole-catalog counter.** A `schema_filter` that cannot be pushed down as `LIKE` still keeps hidden schemas out of it: the row callback of `LoadAllSchemasMetadata` checks `MatchesSchema` / `MatchesTable`, so `staged` holds only visible schemas.
- **An invalidation racing the preload is harmless.** If one lands between `EnsureSchemasLoaded` and taking `mutex_`:
  - the per-schema path finds no entry and reports 0;
  - the whole-catalog path publishes into a list marked NOT_LOADED, which the next lookup reloads.

  Both outcomes are wasted work at worst, never a cache that claims more than it holds.
- **The load state that used to be written at the end is still set.** The removed tail also set `state_`, `last_refresh_` and `schemas_last_refresh_`. `EnsureSchemasLoaded` sets all three now, and it runs first on both paths.
- **The retry reset still matches.** The per-schema reset lambda (PR #308) still rolls back `table_count` / `column_count` by this schema's share. With both counters now unconditional, the undo is symmetric.

oluies (dismissed):
## Approved, with three Lows

Branch diff `8eeaea8..b244a20` (1 commit), roborev job 1707, agent `claude-code`.

`mssql_preload_catalog` now loads the schema list first on both paths. It marks only the schema and tables it actually loaded, and reports what each call loaded, so repeated calls give the same numbers. That fixes #375 and #376.

### Low — `src/catalog/mssql_metadata_cache.cpp`, `BulkLoadAll` per-schema path: the schema match is now case-sensitive
The target schema used to go to the server as `@s` (`BuildExecuteSqlBatch(sql, "@s sysname", …)`, removed here). Under SQL Server's usual case-insensitive collation, `mssql_preload_catalog('db', 'DBO')` therefore loaded `dbo`. The new path is an exact-case `schemas_.find(schema_name)` on the in-memory map. It finds nothing, loads nothing, and reports `Preloaded schema 'DBO': 0 tables, 0 columns`, which looks the same as an empty schema. (I checked this against the diff: the `@s` parameter is gone and the lookup is `schemas_.find`.)

*Fix:* when the exact lookup misses, fall back to a case-insensitive match over `schemas_`. Failing that, say that the schema doesn't exist rather than report a zero-count success.

### Low — `BulkLoadAll`: a window between `EnsureSchemasLoaded(connection)` and taking `mutex_`
`EnsureSchemasLoaded` releases `mutex_` before `BulkLoadAll` takes it again. A concurrent invalidation or refresh in that gap can clear `schemas_`:
- The per-schema path then misses the schema and reports `0 tables, 0 columns`.
- The whole-catalog path re-creates schemas in `LoadAllSchemasMetadata` while `schemas_load_state_` stays `NOT_LOADED`. The next access then clears the cache and runs the full load again, which is the #376 symptom this PR fixes.

*Fix:* once the lock is held, check `schemas_load_state_ == LOADED` and retry if it isn't. Or add an `EnsureSchemasLoadedLocked` and call it under the same lock.

### Low — `test/sql/regression/issue_375_376.test`: no case where the target schema is already partly loaded
The merge-and-mark logic changed exactly there: it marks only the tables in the answer and keeps existing entries. The test checks that other schemas stay unmarked, but never that the target schema can hold old and new entries together.

*Fix:* look up `i375.t1` first (or `mssql_invalidate_cache('i375', 'i375', 't1')`), then run `mssql_preload_catalog('x', 'i375')`. Check that `duckdb_tables()` / `duckdb_columns()` show exactly 2 tables and 5 columns for `i375`.

VGSML:
Thanks! I addressed all three Lows in 66d4934, together with a merge of main (CHANGELOG and `catalog-integration.md` conflicts). The owner asked for an immediate merge, so these changes went in after your approval without a re-review. Please give them a look after the fact.

- **The lock window between loading the schema list and loading the metadata.** `BulkLoadAll` now takes `mutex_` once for both steps. The bodies moved into private `EnsureSchemasLoadedLocked` / `LoadAllSchemasMetadataLocked`, and the public entries lock and call them.
- **Schema name case.** The exact name is tried first, then the one schema that matches it ignoring case, so `'I375'` finds `i375` as `@s` used to on the server.
  - Several case-insensitive matches: the call asks for the exact name.
  - A schema that is absent or hidden by `schema_filter`: the call fails with the schema's name instead of reporting `0 tables, 0 columns`.
- **New cases in `issue_375_376.test`:**
  - a partly loaded target: one table is looked up first, and after the preload exactly 2 tables and 5 columns remain;
  - the case-insensitive match;
  - both refusals.

The full SQL suite (197 files) and the C++ unit tests (31) passed locally before the merge.

Your #372 Low, that `order_pushdown` was only shown to be accepted, will get its own small test-only PR. `EXPLAIN` tells the values apart: the Order By operator is gone with `true` and present with `false`.

PR #378, opened 2026-09-22
test(#325, review of #372): order_pushdown's string form is honoured, not only accepted

VGSML:
Follow-up to oluies' Low on #372. `attach_boolean_string_options.test` passes `order_pushdown 'no'` but never shows that the value has an effect. The test would pass just the same if the string were read as `true`, or if the option were ignored.

## What the test now checks

`test/sql/catalog/order_pushdown.test` checks the **plan** for both string forms, with the global `mssql_order_pushdown` off:

| ATTACH option | `EXPLAIN … ORDER BY id` |
|---|---|
| `order_pushdown 'yes'` | no Order By operator above the scan (pushed) |
| `order_pushdown 'no'` | Order By operator present |

The results are the same either way, and results are all the file's earlier sections compare.

The key column is `TestSimplePK.id`, the primary key from the seed. `order_pushdown_test.id` would not work: it is nullable, and a nullable ORDER BY key is never pushed, because DuckDB and SQL Server place NULLs differently. My first version used it and failed for exactly that reason.

The file passes against a local server (361 assertions). Test-only change.


VGSML:
## Self-review of the test commit: no changes needed

- **The two checks together prove both halves: the value is parsed, and it has an effect.**
  - If `'no'` were misread as true, the ORDER BY would be pushed and the `'no'` check (Order By present) would fail.
  - If the option were ignored, the global setting (off) would leave the Order By in place, and the `'yes'` check (Order By absent) would fail.
- **The positive check guards the negative one against a renamed operator.** If a DuckDB bump renamed the operator in the plan text, `<!REGEX>` on `'yes'` would pass without proving anything. The `<REGEX>` on `'no'` would fail first and point at the rename. The regex accepts both renderings, `ORDER_BY` and `Order By`.
- **The global setting at this point is known.** The section starts right after `SET mssql_order_pushdown = false` at line 147, so the ATTACH option is the only thing that can enable pushdown. Accepted as is, without an explicit `SET` of its own.
- **The seed table exists in CI.** `TestSimplePK` comes from `docker/init/init.sql`, and CI's `integration_test.sh` refuses to run without that seed. The integration job passed with the new checks.
- **Spelling coverage lives elsewhere.** Only `'yes'` and `'no'` are exercised here. The other spellings go through the same cast and are covered in `attach_boolean_string_options.test`.

oluies (approved):
## Approved — no findings

Branch diff `825488b..4f7e3f9`, roborev job 1710, agent `claude-code`.

This closes the Low from my review of #372. The string form of `order_pushdown` is now shown to change the plan, not only to be accepted: with `'yes'` there is no Order By above the scan, and with `'no'` it stays. The test uses the NOT NULL primary key of the seeded `TestSimplePK`, because a nullable ORDER BY key is never pushed down.

PR #379, opened 2026-09-22
feat(#322): default_schema — the schema unqualified names resolve against, from every configuration path

### LINKED ISSUE #322: Make the catalog's default schema configurable per connection (currently hard-coded dbo)
## What

`MSSQLCatalog::GetDefaultSchema()` returns a hard-coded `dbo`:

```cpp
optional<Identifier> MSSQLCatalog::GetDefaultSchema() const {
	return Identifier("dbo");
}
```

Shipped in **v0.2.5** (#129, PR #313) and being forward-ported to the 2.0 line in #314. It fixed the real bug — the base class answered `main`, a schema no SQL Server database has, so every unqualified name failed and users were creating a schema literally called `main` on the server to work around it. But `dbo` is a default, not a fact, and right now there is no way to say otherwise.

## Why it is not enough

`dbo` is the right default and wrong often enough to matter:

- a login whose SQL Server **default schema** is something else (`ALTER USER ... WITH DEFAULT_SCHEMA = sales`) — the extension will resolve unqualified names against `dbo` while the same login in SSMS resolves them against `sales`;
- databases organised per-tenant or per-domain (`sales`, `staging`, `raw`) where nothing of interest lives in `dbo` at all;
- an ATTACH already narrowed by `schema_filter` to a single schema that is not `dbo` — the default then points at a schema the catalog deliberately hides.

## What to do

Make it a **connection-level** option, since that is the scope it belongs to — one attached catalog, one default. Concretely:

- a `default_schema` field on `MSSQLConnectionInfo` (`src/include/mssql_storage.hpp`), next to `schema_filter` / `table_filter`, which are the same kind of per-ATTACH catalog-shape option;
- surfaced the way those are: **ATTACH option**, ADO.NET connection-string key, URI query parameter and `CREATE SECRET` parameter, with ATTACH winning — the precedence the README already documents for the filters;
- empty = keep today's behaviour (`dbo`), so nothing changes for anyone who does not set it.

Worth deciding while implementing, not now:

- whether an unset value should instead **ask the server** — `SELECT SCHEMA_NAME()` returns the login's own default schema, which is arguably the most correct answer. It costs a round trip at ATTACH, though the eager credential validation (spec 047 FR-011) already opens a connection there, so it may be free in practice;
- what to do when the named schema does not exist or is excluded by `schema_filter` — fail the ATTACH, or fall back to `dbo` with a warning.

## Not a regression

Nothing that works today breaks; this only removes a ceiling. Filed from a review conversation about #314, not from a user report.

VGSML:
Closes #322.

## What changes

`MSSQLCatalog::GetDefaultSchema()` used to return a hard-coded `dbo`. It now returns the `default_schema` option, or `dbo` when the option is unset, so nothing changes for anyone who does not set it. The option is read from the same places as `schema_filter`:

| Source | Key |
|---|---|
| ATTACH option | `default_schema` (wins over the others) |
| Connection string | `DefaultSchema` (also `Default Schema`, `default_schema`) |
| URI | `?default_schema=` |
| MSSQL secret | `default_schema` |

It applies to:
- a two-part name, `db.table`;
- `USE db`, which also sets `current_schema()`;
- CTAS into `db.table`;
- COPY into `db.table` or `mssql://db/table`. Both COPY forms used to put `dbo` in themselves. They now ask the catalog through `Catalog::GetDefaultSchema`.

The catalog's `default_schema_` member was set to `"dbo"` and never read. It now holds the option's value.

## The issue's open questions, decided

- **Unset means `dbo`, not `SCHEMA_NAME()`.**
  - SQL Server resolves an unqualified name through the login's default schema first and then `dbo`. DuckDB has only one default schema, so following the login's default would break unqualified references to `dbo` objects for such a login.
  - Asking the server would also cost a round trip at ATTACH, and would get no answer on the lazy-validation path, where no connection exists yet.
- **A default that `schema_filter` hides is refused at ATTACH.** The check needs no server.
- **A schema that does not exist is not probed at ATTACH.** The first unqualified use reports it: `schema "x" does not exist`.

## Tests

- **`test/sql/attach/attach_default_schema.test`** (no server):
  - the filter contradiction, passed as ATTACH options;
  - the same contradiction, passed as connection-string keys;
  - a default the filter allows, which is accepted.
- **`test/sql/catalog/default_schema_option.test`**:
  - every path in the table above;
  - the ATTACH option overriding the secret;
  - the unchanged default when the option is unset;
  - a schema that does not exist.
- **`default_schema.test` (#129)** is unchanged.
- **COPY into a two-part `db.#temp`** with a non-`dbo` default was checked by hand. SQL Server ignores the schema of a temp table.

The full SQL suite passes locally: 199 files, 17 skipped by their environment gates.

## Docs

- A "Default Schema" section in `docs/catalog-integration.md`.
- A row in the CLAUDE.md ATTACH-options table.
- A CHANGELOG entry.
- `DATAMODEL.md` needs no change: no layer, invariant or flow changes.


VGSML:
## Self-review of c92cdf3: no changes needed

- **An invalid `schema_filter` fails the same way as before.** The new ATTACH check runs `MSSQLCatalogFilter::SetSchemaFilter` on a filter that may come from the secret or the connection string, which nothing validated earlier. It validates first and throws the same `MSSQL schema_filter error: Invalid regex pattern …`, with or without `default_schema`; checked with `SchemaFilter=([bad`.
- **The spelling of the value matters at query time, not at ATTACH.** The ATTACH check matches the filter case-insensitively, because the filter regex is `icase`. The later lookup is case-sensitive, so `default_schema 'SALES'` passes a `^sales$` filter and then reports `schema "SALES" does not exist` on first use. That is why the docs and CLAUDE.md say to spell the name exactly as the server does.
- **An explicitly empty `default_schema ''` means `dbo`.** That also holds as an ATTACH override of a secret that sets one.
- **The COPY paths are safe when the target is not an MSSQL catalog.**
  - `CatalogDefaultSchema` catches only `CatalogException`, from a catalog that does not exist. The verification after it then reports the missing catalog by name, as before.
  - For a non-MSSQL catalog it returns that catalog's own default, such as `main`, and the existing check then refuses the target as "not an MSSQL catalog". Nothing new is reachable.
- **Temp-table targets are unaffected.** `db..#t` keeps its explicitly empty schema. A two-part `db.#t` now gets the configured default instead of `dbo`, and SQL Server ignores a temp table's schema. I checked this by hand with a non-`dbo` default.
- **No threading concern.** `default_schema_` is set in the constructor and only read after that.

oluies (dismissed):
## Approved, with three Lows

Branch diff `825488b..c92cdf3`, roborev job 1711, agent `claude-code`.

`default_schema` replaces the hard-coded `dbo`. It can come from an ATTACH option, a connection string, a URI or a secret, and it drives unqualified-name resolution, `USE`, CTAS and COPY target resolution. ATTACH refuses a `default_schema` that `schema_filter` hides.

### Low — `src/mssql_storage.cpp`: the `default_schema` / `schema_filter` check only runs for an explicit value
If `default_schema` is unset (so the default is `dbo`) and `schema_filter '^sales$'` hides `dbo`, ATTACH succeeds, and the first unqualified name fails. Writing `default_schema 'dbo'` explicitly, which is the same effective configuration, is refused at ATTACH. *Fix:* check the effective default (`default_schema.empty() ? "dbo" : default_schema`). If refusing a filter-only setup that hides `dbo` would break existing users, keep the explicit-only check and say so in the docs and the error text.

### Low — `src/mssql_storage.cpp` / `src/catalog/mssql_catalog.cpp:72`: the value is stored as given
Nothing trims or validates it. `default_schema ' i322'`, or a URI value that's still percent-encoded, is accepted and only surfaces later as `schema " i322" does not exist`. An explicit `''` quietly turns a secret's value back into `dbo`; that may be intended, but it's neither documented nor tested. *Fix:* trim whitespace, and either refuse an explicit empty value or document that it resets to `dbo`. Add a test for the clearing override.

### Low — `test/sql/catalog/default_schema_option.test`: two cases missing
- A two-part COPY target to a `#temp` table (`db.#tmp`) under a non-`dbo` default now resolves the schema to the custom default. SQL Server ignores the schema for `#temp` tables, so this is probably harmless, but nothing pins it.
- There's no case-mismatch case (`default_schema 'I322'`) to lock in the documented case-sensitive behaviour.

VGSML:
Thanks! All three Lows are addressed in the latest commit.

- **The filter check stays explicit-only, and that is now stated.** Refusing an unset `dbo` that the filter hides would break every existing ATTACH that pairs such a filter with fully qualified names. The code comment, `docs/catalog-integration.md` and CLAUDE.md now say so.
- **The value is trimmed, and an empty one means `dbo`.** That includes an explicit `default_schema ''` on the ATTACH, which clears a value set in a secret. This is documented and tested. A percent-encoded URI value was already decoded by the URI parser (`UrlDecode`), so no change was needed there.
- **Three new cases in `default_schema_option.test`:**
  - a two-part COPY into a `#temp` table under a non-`dbo` default;
  - a default spelled in a different case, `'I322'`, which fails with `schema "I322" does not exist`, as the docs say;
  - a value padded with whitespace, which is trimmed.

`default_schema_option.test` (56 assertions), `default_schema.test` and `attach_default_schema.test` pass locally.

oluies (approved):
Re-approving at `40db922`. All three Lows are addressed. The value is trimmed, and an explicit `''` means `dbo`, both documented and tested. The filter contradiction deliberately stays explicit-only, with the reason (it would break existing ATTACHes that pair such a filter with fully qualified names) now written in the code, the docs and CLAUDE.md. The test adds the `#temp` COPY target, the case-mismatch default and a whitespace-padded value.

Issue #380, opened 2026-09-22
COPY into a new table inside a transaction, then reading it in the same transaction, blocks until the metadata timeout


VGSML:
## What happens

```sql
SET mssql_metadata_timeout = 10;          -- default 300
ATTACH '…TestDB…' AS db (TYPE mssql);
BEGIN;
COPY (SELECT 1 AS v) TO 'db.dbo.hang_copy' (FORMAT 'bcp');   -- creates the table
SELECT count(*) FROM db.dbo.hang_copy;
-- IO Error: Metadata query failed: Query timeout    (after 10 s; 300 s by default)
```

Measured on `main` against the docker SQL Server. With the default `mssql_metadata_timeout` the `SELECT` hangs for five minutes before it fails.

The same shape through CTAS works and answers `1` at once:

```sql
BEGIN;
CREATE TABLE db.dbo.hang_ctas AS SELECT 1 AS v;
SELECT count(*) FROM db.dbo.hang_ctas;
```

## Why

- COPY's `CREATE TABLE` runs on the transaction's pinned connection and stays uncommitted, so it holds a schema-modification lock on the new table.
- The `SELECT` then needs the new table's metadata, and the metadata load runs on **another** pooled connection.
- On the server that request waits on `LCK_M_SCH_S`, blocked by the pinned session, until the metadata timeout ends it. It was observed in `sys.dm_exec_requests`.
- CTAS does not hit this because it registers the created table in the catalog itself, without a metadata query.

## Fix direction

Either of these would work:

- COPY registers the table it created in the catalog the way CTAS does, so reading it needs no metadata query.
- A metadata load inside a transaction runs on the pinned connection, which can see its own uncommitted DDL.

The first matches what CTAS already does.

Found while measuring bulk loads for #331, but it is independent of the isolation level: it reproduces with the default level.

PR #381, opened 2026-09-22
feat(#331): transaction_isolation — the isolation level of explicit transactions; put back after COMMIT/ROLLBACK

### LINKED ISSUE #331: ATTACH option for transaction isolation level, SNAPSHOT by default where the database allows it
## The problem

Several catalog scans in one DuckDB plan are several separate SELECTs to the server, materialised
independently on the pinned connection. Nothing holds a consistent read across them, so a query that
reads the **same table twice** can see two different states of it.

Found from `mssql_ducklake`, where DuckLake's commit-loop conflict check does exactly that — an outer
scan plus a `MAX(snapshot_id)` subquery over `ducklake_snapshot`. With `MSSQL_DEBUG=2`:

```
generated query = SELECT [snapshot_id], [schema_version], [next_catalog_id], [next_file_id] FROM [dbo].[ducklake_snapshot]   -> 12 rows
generated query = SELECT [snapshot_id] FROM [dbo].[ducklake_snapshot]                                                        -> 13 rows
```

Twelve and thirteen: another writer committed between the two scans. The subquery named a row the
outer scan never saw, the predicate matched nothing, and the caller crashed on the empty result.

Measured, four concurrent writers, ten commits each, fourteen rounds: **a writer lost in 6 of 14
rounds**. The same workload on DuckLake's postgres backend — another remote catalog, same DuckLake
code — lost **0 of 14**, because that scanner reads inside one transaction snapshot.

`mssql_ducklake` has worked around its own instance of this by rewriting that one query to read the
table once (hugr-lab/mssql-ducklake#15). The general case belongs here.

## The cure, demonstrated on this server

Two reads of one table inside one transaction, with an insert from another session in between:

| isolation level | first read / second read |
| --- | --- |
| `READ COMMITTED` | **3 / 4** — they disagree |
| `SNAPSHOT` | **4 / 4** — they agree |

Note this needs **transaction-level** snapshot. `READ_COMMITTED_SNAPSHOT` is not enough: it gives
statement-level consistency, and the two scans are two statements.

Also note it does not require knowing or fixing the order the scans run in — under `SNAPSHOT` every
read in the transaction sees the database as of the transaction's start, so the order stops
mattering.

## Proposal

An ATTACH option for the isolation level, defaulting to `SNAPSHOT` where the database allows it:

```sql
ATTACH '...' AS srv (TYPE mssql, transaction_isolation 'snapshot');   -- explicit
ATTACH '...' AS srv (TYPE mssql);                                     -- snapshot if supported, else today's behaviour
```

### Where it goes

One call site — `src/connection/mssql_connection_provider.cpp:170`, the lazy start of the SQL Server
transaction:

```cpp
// before
if (!conn->ExecuteBatch("BEGIN TRANSACTION")) {
// after
if (!conn->ExecuteBatch("SET TRANSACTION ISOLATION LEVEL SNAPSHOT; BEGIN TRANSACTION")) {
```

The `SET` has to precede `BEGIN`: the snapshot is taken at the first read after `BEGIN`, and the
level cannot be switched to `SNAPSHOT` inside an open transaction.

### The probe, at no cost

`ALLOW_SNAPSHOT_ISOLATION` must be `ON` for the database, so the level has to be chosen per database.

Where the failure lands, measured rather than assumed — `BEGIN` is **not** where it fails:

| under `SNAPSHOT`, setting OFF | result |
| --- | --- |
| `BEGIN TRANSACTION` | succeeds |
| first read of a **user** table | **error 3952** |
| read of `sys.objects` only | succeeds |

So "just try it and fall back" is a poor substitute for a probe: the error arrives on the first real
query inside an already-open transaction, which then has to be rolled back and replayed at another
isolation level. Cheaper to ask first.

There is already a query at attach that asks the database about itself —
`src/catalog/mssql_catalog.cpp:46`:

```cpp
"SELECT CAST(DATABASEPROPERTYEX(DB_NAME(), 'Collation') AS NVARCHAR(128)) AS db_collation";
```

One more column on that statement costs nothing — same round trip, same attach:

```sql
SELECT CAST(DATABASEPROPERTYEX(DB_NAME(), 'Collation') AS NVARCHAR(128)) AS db_collation,
       (SELECT snapshot_isolation_state FROM sys.databases WHERE database_id = DB_ID()) AS snapshot_isolation_state
```

Verified against the pinned test image (server 17.0.4075): the column follows
`ALTER DATABASE ... SET ALLOW_SNAPSHOT_ISOLATION ON/OFF` immediately, `0` ↔ `1`, with the collation
unchanged beside it.

It has to come from `sys.databases`: `DATABASEPROPERTYEX` has no property for this. All three
plausible names — `SnapshotIsolationState`, `IsSnapshotIsolationState`, `IsReadCommittedSnapshotOn` —
return NULL, including while the setting is ON.

**On permissions**, since `sys.databases` is often assumed to need them. Measured on the pinned image:
a login created with nothing but `CREATE LOGIN` + `CREATE USER` — no grants at all — still sees its
own database's row, and still sees it after `DENY VIEW ANY DATABASE`. Metadata visibility gives a
principal the row for a database it can connect to, so the ordinary case is covered.

That is box SQL Server, though, and it is exactly the kind of thing that differs on the platforms in
the open questions below. So the rule should be: a NULL or missing row means **unknown**, and unknown
falls back to today's behaviour rather than to `SNAPSHOT`. Missing permissions then cost nothing
except the improvement, and the explicit `transaction_isolation 'snapshot'` option stays available
for anyone who knows their database allows it.

### Not available at login

Worth stating so nobody looks for a cheaper source: the TDS login response does not carry this. The
parser here reads exactly one ENVCHANGE subtype, `0x08` BEGIN_TRANS (`src/tds/tds_token_parser.cpp`),
and MS-TDS defines no ENVCHANGE for the isolation level at all. Even if it did, it would report the
*session's* level, which is ours to set — not the *database's* `ALLOW_SNAPSHOT_ISOLATION`, which is
what decides whether setting it is allowed.

### Connection reset

Nothing to do: the isolation level is session state, and the pool already flags `RESET_CONNECTION`
on the next batch (`DEFAULT_RESET_CONNECTION = true`), which clears it.

With `mssql_reset_connection = false` (#189) the level will persist across pooled uses. That is not a
gap to close: switching the reset off is a deliberate act, and the setting exists precisely to hand
session state back to whoever asked for it — the isolation level joins the `##temp` table it was
added for.

## Open questions, not yet verified

- **Write conflicts.** These transactions also write. Under `SNAPSHOT`, updating a row another
  transaction changed since ours began returns error **3960** instead of blocking. Whether that is an
  improvement or a new failure mode depends on the caller retrying it — DuckLake's retry predicate
  matches on the substring `conflict`, and 3960 says "update conflict", so it would probably retry,
  but that is a guess and should be measured before this is turned on by default.
- Whether tempdb version-store pressure matters for catalog-sized workloads. Probably not, worth a
  look under the scale benchmark.

## Platform behaviour, from the documentation

Checked rather than guessed, because "SNAPSHOT by default" is only safe on the platforms that have
it. The gate already exists here — `EndpointType { OnPremises, AzureSQL, Fabric, Synapse }` in
`src/include/mssql_platform.hpp`, detected from the hostname — so this is a switch on something the
extension already knows, not new detection.

| platform | `EndpointType` | isolation there | what this change should do |
| --- | --- | --- | --- |
| SQL Server 2019+, Azure SQL MI | `OnPremises` | `SNAPSHOT` available once `ALLOW_SNAPSHOT_ISOLATION` is `ON`; `OFF` by default | probe, use `SNAPSHOT` when the state is 1 |
| Azure SQL Database, SQL database in Fabric | `AzureSQL` | RCSI is `ON` by default, `READ COMMITTED` is the default level; `SNAPSHOT` supported subject to the same `ALLOW_SNAPSHOT_ISOLATION` | same probe. RCSI alone does **not** fix this — it is statement-level, and the two scans are two statements |
| Fabric Warehouse / SQL analytics endpoint | `Fabric` | snapshot isolation enforced on **all** transactions; `SET TRANSACTION ISOLATION LEVEL` is **ignored** | send nothing — the bug cannot occur there |
| Synapse dedicated SQL pool, PDW | `Synapse` | grammar is `SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED` **only**; default `READ UNCOMMITTED`; RCSI only, set from `master` | must **not** send `SNAPSHOT` |

The statements this rests on:

- Fabric Warehouse — "Fabric Data Warehouse enforces snapshot isolation on all transactions." and
  "If you use T-SQL to change your isolation level, the change is ignored at query execution time and
  snapshot isolation is applied."
  ([Transactions in Fabric Data Warehouse](https://learn.microsoft.com/en-us/fabric/data-warehouse/transactions))
- Synapse — the documented syntax is literally `SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED`,
  and "The isolation level of the transactional support is default to READ UNCOMMITTED. You can
  change it to READ COMMITTED SNAPSHOT ISOLATION by turning ON the READ_COMMITTED_SNAPSHOT database
  option for a user SQL pool when connected to the master database."
  ([Use transactions in Azure Synapse Analytics SQL pool](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-develop-transactions),
  [SET TRANSACTION ISOLATION LEVEL](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql))
- Azure SQL Database and SQL database in Fabric — "`READ_COMMITTED_SNAPSHOT` `ON` is the default on
  Azure SQL Database and SQL database in Microsoft Fabric." (same page)
- The prerequisite and the ordering, which is why the `SET` goes before `BEGIN` — "The
  `ALLOW_SNAPSHOT_ISOLATION` database option must be set to `ON` before you can start a transaction
  that uses the `SNAPSHOT` isolation level", "A transaction can't be set to `SNAPSHOT` isolation
  level that started with another isolation level; doing so causes the transaction to abort", and "A
  transaction starts the first time it accesses data." (same page — and it matches what was measured
  above: `BEGIN` succeeds, the first user-table read is what fails.)

Two consequences worth carrying into the design:

- **Bulk load interacts with this.** "Optimized bulk load operations on heaps block queries that are
  running under the following isolation levels: `SNAPSHOT`, `READ UNCOMMITTED`, `READ COMMITTED`
  using row versioning. Conversely, queries that run under these isolation levels block optimized
  bulk load operations on heaps." This extension does BCP, and `mssql_ducklake`'s server-side commit
  bulk-loads staging tables inside the same transaction, so the two features meet. Worth measuring
  before `SNAPSHOT` becomes a default rather than an opt-in.
- **The write-conflict error differs by platform.** Box SQL Server raises 3960; Fabric Warehouse
  raises **24556** and **24706** ("Snapshot isolation transaction aborted due to update conflict"),
  at *table* granularity because it uses table-level locking, and the documentation's own advice is
  to retry. Any retry predicate should cover all three.

If a server-side discriminator is ever wanted instead of the hostname, `SERVERPROPERTY('EngineEdition')`
needs no permissions at all ("All users can query the server properties") and enumerates these
directly: 5 = SQL Database, 6 = Azure Synapse Analytics, 8 = Azure SQL Managed Instance, 11 = Synapse
serverless or Microsoft Fabric, 12 = SQL database in Fabric.

VGSML:
Closes #331. The default is **unchanged**. See "The default" below.

## The problem

The scans of one DuckDB transaction are separate statements on its pinned connection. Under the server's default READ COMMITTED, a row that another session commits between two scans is seen by one scan and not the other. SNAPSHOT reads the whole transaction as of its first access. Measured, with two reads inside one transaction and a committed insert from another session in between:

| level | first / second read |
|---|---|
| READ COMMITTED | 3 / 4 |
| SNAPSHOT | 3 / 3 |

## The option

`transaction_isolation` can be set as an ATTACH option (which wins), as `TransactionIsolation` in a connection string, as `transaction_isolation` in a URI, or as a secret parameter.

| value | what is sent |
|---|---|
| `default` / unset | nothing, as before |
| `read_uncommitted`, `read_committed`, `repeatable_read`, `serializable`, `snapshot` | `SET TRANSACTION ISOLATION LEVEL …` as its own statement before `BEGIN TRANSACTION` on the pinned connection |
| `auto` | SNAPSHOT when `sys.databases.snapshot_isolation_state = 1`; otherwise nothing, including when the state is unknown |

- **The SET is a separate statement.** Inside the BEGIN batch, a failed SET would let BEGIN run at the old level, and the reply scan stops at the ERROR token before it reaches the transaction descriptor.
- **The snapshot state is probed at ATTACH.** The query that reads the collation also reads `snapshot_isolation_state`, so there is no extra round trip.
- **An explicit `snapshot` on a database where it is OFF is refused at ATTACH.** Otherwise BEGIN would succeed and the first read of a user table would fail with error 3952.
- **Fabric Warehouse gets nothing.** It enforces snapshot isolation and ignores the SET.
- **Synapse gets nothing and refuses explicit levels other than `read_uncommitted`.** That is its only settable level, and it is already the default.

## Found and fixed: the isolation level survives the connection reset

SQL Server keeps a session's isolation level across the pool's `RESET_CONNECTION`. Measured: SNAPSHOT was still in force on the next autocommit statement, although the reset had run and had dropped the session's `#temp` table.

A transaction that set a level now sends `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` after COMMIT or ROLLBACK:

- **It is its own statement.** Appended to the ROLLBACK batch, it was lost when the server had already aborted the transaction on an update conflict (error 3960): that batch goes out with a stale transaction descriptor.
- **A connection whose level cannot be put back is closed** instead of being returned to the pool.

CLAUDE.md and DATAMODEL.md said the reset clears the level. Both are corrected. A `SET TRANSACTION ISOLATION LEVEL` sent through `mssql_exec()` has always leaked this way, and `docs/transactions.md` now says so.

## Measured under SNAPSHOT

- **Update conflicts.** Updating or deleting a row that another session changed since the transaction began fails with error 3960: `Snapshot isolation transaction aborted due to update conflict … Retry the transaction`. Under READ COMMITTED the same UPDATE simply re-reads the row. DuckLake's retry predicate matches the word "conflict".
- **Bulk loads.** COPY and `INSERT BULK` inside the transaction work, including TABLOCK on a heap, and their rows are visible within the transaction.
- **Readers during another session's TABLOCK heap load.** A SNAPSHOT reader is not blocked and sees the committed state. A READ COMMITTED reader waits for the load to commit.

## The default

The issue proposes `auto` as the default. I left the default unchanged.

- On SQL Server, `ALLOW_SNAPSHOT_ISOLATION` is OFF by default, so for most databases `auto` would change nothing.
- On Azure SQL Database it is ON by default. There `auto` would move every explicit transaction to SNAPSHOT and turn concurrent-update re-reads into 3960 errors that the caller must retry.

The consumer that wants the consistency, mssql_ducklake, can set `transaction_isolation 'auto'` itself. Switching the default to `auto` would be a one-line change.

## Tests

- **`test/sql/attach/attach_transaction_isolation.test`** needs no server. It checks:
  - value spellings;
  - that an invalid value is refused, naming the value;
  - that an unknown snapshot state is not treated as OFF.
- **`test/sql/transaction/transaction_isolation.test`**:
  - **On TestDB:** an explicit `snapshot` is refused, and `auto` falls back to the default level.
  - **With `serializable`:** it is in force inside the transaction, and READ COMMITTED is back on every pooled connection after COMMIT and after ROLLBACK.
  - **In a database of its own with snapshot allowed:**
    - two reads stay consistent across an insert committed by a second DuckDB connection (`con2`);
    - a concurrent update produces the 3960 conflict;
    - the level is restored after the server-aborted transaction.

With the restore disabled, the second test fails. The full SQL suite passes locally: 199 files, 17 skipped by their environment gates.

Unrelated, found while measuring: #380. COPY into a new table inside a transaction, followed by a read of that table in the same transaction, waits until the metadata timeout.


VGSML:
## Self-review of 87246d0: one finding, fixed in the follow-up commit; the rest checked

### Fixed: the snapshot probe could fail every ATTACH on a platform without it

`snapshot_isolation_state` was appended to the ATTACH collation query unconditionally, and any error in that query fails the ATTACH. If `sys.databases` on a platform lacks the column or the row, every ATTACH there would have failed, including for users who never set the option. Fabric Warehouse and Synapse are the candidates, and neither of them uses the answer.

The column is now added only for `snapshot`, which is refused at ATTACH when snapshot is OFF, or for `auto`, and never on Fabric or Synapse. Every other ATTACH sends the same query as before. The isolation tests pass, as do `varchar_declaration_seek.test` (which reads the code page from the same query) and `first_touch_round_trips.test`.

### Checked, no change

- **Cost.** An opted-in transaction costs two more round trips: the SET before BEGIN and the restore after COMMIT or ROLLBACK. Folding the SET into the BEGIN batch would mean teaching the hardened reply scan (`FindBeginTxnDescriptor`) to handle ERROR tokens. That is a possible follow-up. Anyone who does not set the option pays nothing.
- **Scope.** Autocommit DML, which opens its own server transaction through `LoadTransaction`, is untouched. The issue concerns the scans of one explicit transaction.
- **Reset ordering.** A pending `RESET_CONNECTION` on the connection goes out with the SET, and the SET takes effect after the reset. The level is always put back, even with `mssql_reset_connection = false`, because the extension set it.
- **Concurrency.** `snapshot_isolation_state_` is written once in `Initialize`, whose second call returns early (spec 076 W2), and it is read-only afterwards.
- **Error paths.** If the SET before BEGIN fails, the connection goes back to the pool at its old level and the error names the statement. If the restore fails, the connection is closed, and the pool drops closed connections at `Release`.

oluies (dismissed):
## Approved, with one Medium and two Lows

Branch diff `825488b..e8fa8e3`, roborev job 1712, agent `claude-code`.

`transaction_isolation` (ATTACH option, connection string, URI or secret) sends the level before `BEGIN` on the pinned connection. It auto-detects snapshot, validates the option at ATTACH, and puts READ COMMITTED back after COMMIT/ROLLBACK, because `RESET_CONNECTION` keeps the level.

### Medium — `src/connection/mssql_connection_provider.cpp`, `GetConnection`: a failed BEGIN can pool a session at the new level
The `SET TRANSACTION ISOLATION LEVEL …` now goes as its own statement ahead of `BEGIN`. If it succeeds and `BEGIN` then fails, the connection goes back through `pool.Release(conn)` without a `Close()`. That covers both `ExecuteBatch("BEGIN TRANSACTION")` returning false and the null-socket branch. The pooled session keeps SERIALIZABLE / SNAPSHOT / …, `RESET_CONNECTION` doesn't clear it, and later autocommit statements on that connection run at that level. This is the leak the PR exists to prevent. Only the `ReceiveMessage` failure path calls `conn->Close()` first. I checked this at `e8fa8e3`: lines ~192-203 release without closing.

Whether it bites depends on whether the pool discards a connection whose send failed. If `Release` already drops dead connections, what's left is the null-socket branch. The explicit fix is cheap either way.

*Fix:* once the isolation SET has succeeded, `Close()` before `Release` on every failure path, or send `TransactionIsolationRestoreStatement()` first.

### Low — `src/catalog/mssql_catalog.cpp`, `QueryDatabaseCollation`, `probe_snapshot` branch
The snapshot probe is appended to the collation query. If that subquery fails (a restricted principal, or a platform whose `sys.databases` differs), the whole query fails and the database collation and code page are lost too, not just the snapshot state. The surrounding `try` swallows the error, so the downgrade is silent. *Fix:* on failure, retry the plain `DATABASE_COLLATION_SQL` and leave `snapshot_isolation_state_ = -1`.

### Low — `test/sql/transaction/transaction_isolation.test`: no coverage of the paths that could leak
The tests cover the success paths and the 3960 rollback. Nothing covers a failure between SET and BEGIN, or a COMMIT failure, which is where a level can escape into the pool. If a test hook allows it, force a failure after SET, then assert through `ti_obs` that no pooled session is left at a level other than 2.

VGSML:
Thanks! I fixed the Medium and the first Low in the latest commit. I left the second Low without a test, for the reason below.

- **Medium: a failed BEGIN could pool a session at the new level.** Fixed. Once the isolation SET has succeeded, every BEGIN failure path now closes the connection before `Release`: `ExecuteBatch` returning false and the null-socket branch. The `ReceiveMessage` path already closed it. When no level is set, nothing changes.
- **Low: a failed probe took the collation and code page with it.** Fixed. If the query carrying the probe fails, the plain collation query runs again and the snapshot state stays unknown. For `auto`, unknown means "send nothing".
- **Low: no test for a failure between SET and BEGIN.** Not added. Nothing in SQL can make BEGIN fail after a successful SET: that happens only when the socket dies in between, and the tests have no hook for it. The protection is structural instead: every failure path closes the connection explicitly, and the pool drops closed connections at `Release`. The COMMIT-failure path keeps the connection pinned, so the later ROLLBACK runs the restore.

The isolation tests, `varchar_declaration_seek.test`, `first_touch_round_trips.test` and all of `test/sql/transaction/` pass locally.

oluies (dismissed):
Re-approving at `dbc3887`. The Medium is fixed: once the isolation SET has succeeded, every BEGIN failure path closes the connection before `Release` (`close_on_failure`), so a session carrying the level can't reach the pool. I checked that `TdsConnection::Close()` doesn't clear `last_error_`, so the thrown message still carries the reason. A failed snapshot probe now falls back to the plain collation query and leaves the snapshot state unknown, which `auto` reads as "send nothing".

VGSML:
Merged after a merge with main. #379 had landed beside it: both PRs add an option on the same four configuration paths, the same catalog member block, the same CLAUDE.md table and the same CHANGELOG section. The merge commit `1e776d5` only resolves those conflicts and keeps both sides everywhere. The full SQL suite passed on the merged tree (201 files).

PR #382, opened 2026-09-23
fix(#380): metadata inside a transaction loads on the pinned connection, into the transaction's own cache

### LINKED ISSUE #380: COPY into a new table inside a transaction, then reading it in the same transaction, blocks until the metadata timeout

VGSML:
Closes #380.

## The bug

```sql
BEGIN;
COPY (SELECT 1 AS v) TO 'mssql://db/dbo/new' (FORMAT 'bcp');
SELECT * FROM db.dbo.new;   -- waited out mssql_metadata_timeout (5 min)
```

The same happened after `mssql_exec` DDL. The catalog loaded the new table's metadata on a **pool** connection, and that connection waits on the schema lock of the transaction's uncommitted CREATE. With `mssql_connection_limit = 1` it was worse: the first table a transaction touched failed with `pool timed out (1 active of 1)`, because its metadata needed a second connection. The hang reproduces back to spec 076 (cc5652d).

## The fix

**Metadata inside a transaction** goes on the pinned connection. This covers a table's columns, a schema listing, the schema list and the rowid key. The loads land in `MSSQLTransactionMetadata`, a cache the `MSSQLTransaction` owns.

**The shared cache holds committed state only.** Inside a transaction it is read for names the transaction did not change, invalidated as before, and never written. I measured the naive version, which loads into the shared cache from the pinned connection, and it leaks in two ways:
- after ROLLBACK the table stayed bound as a phantom;
- a second connection found the uncommitted table in the cache while the transaction was open, then hung reading it.

At COMMIT and ROLLBACK the shared cache forgets every name the transaction changed (`NoteTransactionChange` / `ForgetTransactionChanges`). The pinned connection goes back to the pool.

**`mssql_refresh_cache()` and `mssql_preload_catalog()` are refused inside a transaction.** Both are bulk writes into the shared cache. `mssql_invalidate_cache()` is still allowed.

**`GetStorageInfo` inside a transaction** now uses the row count loaded with the entry. Before, it sent a DMV query per table on the pool. On a pool of one, `duckdb_tables()` took 204 s (68 tables × a 3 s acquire timeout); it now takes 0.12 s.

**Parallel writers never exceed `mssql_connection_limit − 1`.** This applies to COPY, CTAS and INSERT via BCP. The statement holds one connection itself, so an extra writer waits `mssql_acquire_timeout` for nothing. Measured: a CTAS of 300k rows in a transaction on a pool of two took 30 s before and takes 0.9 s now.

**A pool of one connection:**
- **CTAS in a transaction** runs whole on the pinned connection: the existence checks, the CREATE, and the rows as INSERT statements. ROLLBACK undoes it. Before, it failed with `Failed to acquire connection to check table existence`.
- **CTAS in autocommit** loads with INSERT statements. A bulk-load session adopts its connection in `GetGlobalSinkState`, before the source scan initialises. Statements take their connection at the first batch.
- **Materialisation.** The optimizer counts CTAS as a sink, like COPY and INSERT. On a pool of one it materialises the catalog's scans in autocommit too, so CTAS and INSERT … SELECT from the same catalog work.

**Existence checks for a CTAS in a transaction** go on the pinned connection whatever the pool size. Otherwise the check would hit the same lock wait as #380 itself.

**Unchanged (spec 057).** Catalog DDL and CTAS on a normal pool still autocommit outside the transaction. This is now documented, and pinned by a test for DROP TABLE.

## Found along the way

- **CTAS row count.** A CTAS on the INSERT path (`mssql_ctas_use_bcp = false`) reported its count without the last batch: 1000 for 1500 rows.
- **CTAS cleanup.** The cleanup DROP after a failed CTAS on the INSERT path now releases the INSERT's connection first. Before, the DROP waited on that connection's locks, and on a pool of one it waited for the connection itself.

## Not in this PR

UPDATE and DELETE in autocommit on a pool of one still cannot run. They are not counted as sinks, so their scan is not materialised. What they need is recorded in spec 080 W2 / W5 / acceptance 4. That includes a constraint on the staged path: on a pool of one it must take its connection after the source scan has returned it.

## Tests

- `transaction/transaction_metadata_pinned.test` covers:
  - COPY into a new table in a transaction, then read, listing and ROLLBACK;
  - `mssql_exec` DDL followed by `mssql_invalidate_cache` in a transaction;
  - refresh and preload refused;
  - COMMIT visible to a second connection;
  - DROP TABLE in a transaction.
- `transaction/transaction_single_connection_pool.test` covers a pool of one:
  - first touch, listing, and a rowid UPDATE in a transaction;
  - CTAS from `range()`, CTAS rolled back, CTAS from the same catalog;
  - CTAS and INSERT … SELECT in autocommit.
- `rowid/rowid_discovery_failure_cached.test` used a transaction on a pool of one to make discovery fail. That lever no longer fails discovery, so the test now has con2 hold the only connection.
- `test/cpp/test_load_policy.cpp` gains pool-limit cap cases and a pool dimension in the exhaustive invariant loop.

Local runs:
- integration: 215 test cases passed, 5 skipped;
- `make test`: 225 passed, 8 skipped.

Docs: DATAMODEL.md (new section "Inside an explicit transaction" and the writer cap), docs/transactions.md, CLAUDE.md, CHANGELOG.


VGSML:
## Self-review

I went through the whole diff after pushing. One finding needs a design decision. The rest are checks that came out clean.

### 1. Medium: transaction-heavy workloads never populate the shared cache

Inside a transaction a miss is loaded into the transaction's own cache, and nothing reaches the shared one. A workload that runs almost everything in explicit transactions therefore never populates the shared cache. DuckLake over an MSSQL metadata catalog is exactly that. Every transaction then pays one metadata round trip for every table it touches, where before this PR only the first one did (it loaded into the shared cache on a pool connection).

Measured on the local server, 50 transactions each reading two tables:

| | total |
|---|---|
| shared cache cold (only transactions ever run) | 0.215 s |
| shared cache warmed by autocommit reads | 0.090 s |

That is about 1.4 ms per table per transaction on localhost. Against Azure SQL it is one network round trip per table per transaction.

**Possible fix, not in this PR: promote a load to the shared cache when it is provably committed state.** Inside a transaction, a load that the pinned connection makes for a name the transaction has not changed describes committed state. The pinned connection sees other sessions' committed DDL, and its own uncommitted DDL only through names it changed. Under snapshot isolation the catalog views are not versioned. So such a load could also be written to the shared cache. Two gaps would have to close first:
- `mssql_exec` DDL in a transaction must always mark the transaction's own layer as changed, whatever `mssql_exec_invalidate_cache` says. Today it only notes the change with the setting on, so untracked uncommitted DDL could be promoted. The shared cache stays untouched by that marking, which is what the setting governs.
- No promotion under `transaction_isolation = read_uncommitted`, where catalog reads can see other sessions' uncommitted DDL.

I am leaving this for the author to decide rather than folding it in unasked.

### Checked, no finding

- **A prepared statement bound inside a transaction, executed after COMMIT** (its table entry lived in the transaction's layer): it runs correctly, including after an invalidation and inside a new transaction. The plan carries the scan's bind data, not the entry.
- **`ForgetTransactionMetadata` runs on all three exits:** commit failure, commit, and rollback. It needs no `ClientContext`, so rollback during shutdown is safe.
- **`NoteTransactionChange` is a no-op in autocommit.** The shared paths behave exactly as on main outside a transaction; the full suite is green.
- **Single-connection CTAS:**
  - the OR REPLACE drop goes through `RunDDL` on the pinned connection;
  - a DDL error releases once and then throws (the release is a no-op for the pinned connection anyway);
  - `drop_on_failure` is off, because ROLLBACK is the undo.
- **Writer cap at pool limit 0:** such a pool creates nothing, so one writer is the only meaningful answer. The exhaustive unit loop covers pools of 0, 1, 2 and 64.

VGSML:
Self-review finding 1 is split out as #383 and will be the next PR, per the author's decision. This PR keeps the agreed invariant: inside a transaction the shared cache is never written.

oluies (commented):
## Review — 1 High (verified), 2 Medium, 1 Low

Branch diff `4b8ea3b..4ef123c` (+1386/−82, 31 files), roborev job 1720, agent `claude-code`, plus a check of the High against the branch.

Not approving on the High alone — everything else here reads as a careful change, and the High is a missing lock rather than a wrong design.

### High — `src/dml/ctas/mssql_physical_ctas.cpp:43-96` (`GetGlobalSinkState`): CTAS's pinned-connection work at sink init is not serialised against the source scan

`GetGlobalSinkState` now issues batches on the **pinned** connection — `TableExists` → `ProbeExists`, `SchemaExists`, and `ExecuteDrop` / `ExecuteDDL` → `RunDDL` under `single_connection` — and it never takes `MSSQLCatalog::MaterializeMutex()`.

I checked this on the branch:
- `grep -c MaterializeMutex` is **0** in both `mssql_physical_ctas.cpp` and `mssql_ctas_executor.cpp`.
- COPY's sink init does take it, with a comment naming this exact race: "inside a transaction the source scans of this catalog drain under the catalog's MaterializeMutex, and DuckDB initialises this sink on another thread while they do. Wait for them here rather than find the pinned connection mid-stream; held for the rest of the init, because the CREATE TABLE below goes down the same connection."
- This commit adds `LOGICAL_CREATE_TABLE` to the sink set in the optimizer (`mssql_optimizer.cpp:743`), so the source scans of a CTAS now materialise at init — which is what puts the drain and these probes on the same connection at the same time.

*Scenario:* `SET threads=4; BEGIN; CREATE TABLE db.dbo.dst AS SELECT * FROM db.dbo.src;` — the sink's `SELECT 1 FROM INFORMATION_SCHEMA.TABLES …` runs against a connection in `Executing`, giving either a state-transition failure or interleaved TDS traffic. The new `transaction_single_connection_pool.test` CTAS case exercises the path but doesn't pin `SET threads`, so it can pass on scheduling luck.

*Fix:* in `GetGlobalSinkState`, after `ResolveConnectionMode`, take `std::unique_lock<std::mutex>(catalog_.MaterializeMutex())` when `gstate->state.in_transaction`, and hold it across the existence checks and the DDL, exactly as `BCPCopyInitGlobal` does.

### Medium — `src/catalog/mssql_table_entry.cpp:337-375` (`EnsurePKLoaded`): uncommitted DDL can be cached in the shared entry

In a transaction the rowid-key discovery now runs on the pinned connection, so it sees uncommitted DDL — but the answer is written to `pk_info_` on the **shared** entry and published through `pk_loaded_`. The comment justifies that with "a table it changed is its own entry", which holds only for changes recorded by `NoteTransactionChange`, and `mssql_exec` DDL records those only when `mssql_exec_invalidate_cache` is true (default **false**).

*Scenario:* `BEGIN; SELECT mssql_exec('db','CREATE UNIQUE INDEX ux ON dbo.t(id)'); DELETE FROM db.dbo.t WHERE id = 1; ROLLBACK;` — every session then plans a rowid over an index that does not exist, until someone invalidates. `DROP INDEX` in a transaction caches the mirror image: a permanent "no usable rowid key".

*Fix:* for an in-transaction discovery, keep the result in `MSSQLTransactionMetadata`, or leave `pk_loaded_` unpublished, so nothing learned from uncommitted state outlives the transaction.

### Medium — `src/catalog/mssql_catalog.cpp:498-512` (`SchemaListCache`): schema DDL never marks the transaction's cache

`SchemaListCache` guards the transaction's schema list with `metadata.IsAllChanged()`, but nothing sets that flag for schema DDL: `CreateSchema` (`:696`) and `DropSchema` (`:721`) invalidate only the shared `metadata_cache_`.

*Scenario:* `BEGIN; SELECT count(*) FROM duckdb_schemas() WHERE database_name='db'; CREATE SCHEMA db.s2; CREATE TABLE db.s2.t (id INTEGER);` — the last statement throws `Schema 's2' not found in MSSQL database`. Autocommit handles the same sequence.

*Fix:* call `NoteTransactionChange(context)` with an empty schema in both, next to the existing `InvalidateAll()`.

### Low — `src/include/copy/load_policy.hpp:186-189`: the reserved slot is subtracted even when nothing holds it

`pool_limit − 1` is applied unconditionally, but the reserved slot only exists when the statement holds a connection of the same catalog. `SET mssql_connection_limit = 4; COPY (SELECT * FROM 'data.parquet') TO 'mssql://db/dbo/t' (FORMAT 'bcp');` in autocommit gets 3 writers out of 4 available connections, on exactly the small-pool setups the cap is for.

*Fix:* subtract 1 only when the statement holds a catalog connection (in a transaction, or the plan scans the same catalog).

VGSML:
Thanks. All four findings reproduced, and all are fixed in `7a286bb`. Each fix has a regression test, and each test fails with its fix reverted. Local results: integration 215/215, C++ tests green.

**High: CTAS sink init vs. the source drain.** Confirmed. `GetGlobalSinkState` now takes `MaterializeMutex` when `in_transaction` and holds it across the existence checks and the DDL, exactly as `BCPCopyInitGlobal` does.
- Test: `transaction_metadata_pinned.test` runs three CTAS from the same catalog inside `BEGIN … COMMIT` under `SET threads = 4`.
- Without the lock it fails 5/5, on a state error on the pinned connection. With the lock it passes 5/5.
- The pool-of-one test now pins `SET threads = 4` as well.

**Medium: a rowid key from uncommitted `mssql_exec` DDL cached in the shared entry.** Confirmed, on your scenario. `mssql_exec` DDL inside a transaction now always marks the transaction's own layer. With `mssql_exec_invalidate_cache = true` it does that plus the shared invalidation, as before. With `false` it calls the new `MarkChangedLocally`: the transaction stops trusting the shared cache for everything, but nothing is added to what COMMIT/ROLLBACK make the shared cache forget, so the shared cache stays the caller's to invalidate, as the setting says. Lookups after the DDL therefore return entries of the transaction's own, and a key discovered on the pinned connection lands on those and dies with the transaction.
- Test: a table with no key; `CREATE UNIQUE INDEX` via `mssql_exec`; `DELETE` inside the transaction (1 row); `ROLLBACK`; then `DELETE` in autocommit is refused with no usable key.
- Without the fix, that last statement uses the rolled-back index.

**Medium: schema DDL does not mark the transaction.** Confirmed: `Schema 'tm380s' not found`, exactly as you described. `CreateSchema` and `DropSchema` now call `NoteTransactionChange(context)`.
- Test: your sequence, with `duckdb_schemas()` listed first.

**Low: `limit − 1` wastes a connection.** Agreed, and fixed at the root instead of with a condition on the cap. The stall came from `BulkLoadSession::TryStart` calling `Acquire()` with the default `mssql_acquire_timeout`. An extra writer is optional, so it now calls `Acquire(0)`: it takes an idle connection, or opens one while the pool is below its limit, or gets none and the thread shares the global writer. The cap is back to the pool limit.
- Re-measured: a 300k-row CTAS in a transaction on a pool of two takes 0.84 s. The pool is used whole whenever the statement holds none of it.
- `test_load_policy.cpp` is updated: pool 2 → 2, pool 3 → 3, and the exhaustive loop asserts `max_writers ≤ pool`.

**Not closed here, for the record.** Under `transaction_isolation = read_uncommitted`, a rowid-key discovery for a *shared* entry runs on the pinned connection and could see another session's uncommitted index. I don't have a cheap, sound fix. Changing the isolation level mid-transaction isn't one, and a pool connection brings back the #380 wait. It is narrow: an explicit dirty-read level, plus concurrent index DDL elsewhere, plus a table whose key this transaction is the first to discover. I'd rather file it than hack it. Say if you'd rather see it handled in this PR.

Docs are updated for all of the above: CLAUDE.md, DATAMODEL.md, CHANGELOG, docs/transactions.md.

oluies (commented):
## Review of `7a286bb` — the four fixes land; one Medium remains, in the same place

roborev job 1723, agent `claude-code`, plus checks against the branch.

**What I verified as fixed:**
- **High, CTAS sink init:** `GetGlobalSinkState` now takes `MaterializeMutex` and holds it through the init, the same shape COPY uses. Thank you for reproducing it 5/5 at `threads = 4` — that settles what my report could only argue.
- **Medium, uncommitted DDL in shared entries:** `MarkChangedLocally` makes `IsChanged` true for the whole transaction, and `MSSQLTableSet::GetEntryInTransaction` then skips step 2 entirely, so the transaction never binds a shared entry and cannot publish a rowid key learned from its own uncommitted DDL. Structurally right — the protection is upstream of `EnsurePKLoaded` rather than inside it.
- **Medium, schema DDL:** `NoteTransactionChange` in both `CreateSchema` and `DropSchema`.
- **Low, the writer cap:** back to the full `pool_limit`, with `Acquire(0)` and a fallback to the shared writer.

### Medium — `mssql_physical_ctas.cpp:55`: the guard is gated on the transaction, but autocommit materialises too

`MaterializeSharedConnectionScans` sets `requires_materialization` in **autocommit** as well, whenever `HasSingleConnectionPool(context, catalog)` (`mssql_optimizer.cpp:785`), and the scan side takes `MaterializeMutex` on exactly that flag (`table_scan.cpp:429-434`) — it does not test for a transaction. The CTAS guard tests `gstate->state.in_transaction`, so on a pool of one in autocommit the scan holds the only connection through its drain while `GetGlobalSinkState` → `TableExists` → `ProbeExists` → `pool.Acquire(-1)` runs unsynchronised. That is the same defect as the High, in the case your own new `one380_ac_ctas` block exercises — and with `SET threads = 4` and `mssql_acquire_timeout = 3` now in that file, it passes on the drain finishing inside 3 s rather than on an ordering guarantee.

*Fix:* gate on the optimizer's own condition, `if (gstate->state.in_transaction || catalog_.GetConnectionLimit() <= 1)`.

### Low — the lock is held over work it does not protect

For a CTAS the rest of the init includes **pool**-connection work: `ExecuteDDL` → `catalog->ExecuteDDL` when `!single_connection`, plus `InitializeBCP` / `ExecuteBCPInsert`, each a blocking `Acquire` (30 s default) and the latter holding its connection for the load. COPY's equivalent block only touches the pinned connection. On a contended pool this blocks every other materialising scan of the catalog, on any session, for work the mutex is not about. Unlocking before the pool-connection DDL and the BCP would keep the guarantee and drop the cost.

### Low — `Acquire(0)` counts as an acquire timeout

`timeout_ms == 0` takes the branch that does `stats_.acquire_timeout_count++` (`tds_connection_pool.cpp:259-265`), and `mssql_pool_stats` surfaces that counter. An extra writer declining is the normal outcome on a pool the statement is already using, so every parallel COPY/CTAS now reports timeouts that delayed nothing — the kind of misleading pool diagnostic #302 worked to remove. A non-counting try-acquire form would keep the stat honest.

### Low — `Claim::Unavailable` is still terminal, and the comments still say "blocking"

The header contract and the call site justify never asking again with "it re-attempts a blocking `Acquire()` on every chunk under pool pressure". With `Acquire(0)` a retry is cheap, yet one lost momentary race costs that thread its writer for the whole load — which partly gives back what raising the cap to `limit` won. Either distinguish "slot cap reached" (terminal) from "nothing free right now" (transient), or update the two comments.

---

Not approving yet only because of the Medium: it is the same race as the High, in the autocommit half. It looks like a one-line condition. Say the word if you would rather I approve as it stands and open the follow-ups as issues.

VGSML:
Thanks for the second round. All four are addressed in `7588e29`. Local results: integration 215/215, C++ tests green.

**Medium: the CTAS lock is gated on the transaction only.** Confirmed. The guard is now `in_transaction || GetConnectionLimit() <= 1`, the optimizer's own condition, the same flag the scan takes `MaterializeMutex` on.
- To be candid about coverage: this one is not pinned by a deterministic test. `one380_ac_ctas` (a pool of one, `threads = 4`) exercises the path, but without the fix it would still pass whenever the drain finishes inside `mssql_acquire_timeout`. I found no way to force the sink's init to run before the scan's. The fix rests on the condition matching the scan's, as you reasoned.

**Low: the lock is held over pool work.** Agreed. Both existence checks now run back to back under the lock; a table that exists answers the schema check without a second query. On a pool of more than one, the lock is released before the DROP, CREATE and bulk load, which all run on pool connections. On a pool of one it stays held through the DDL, because the DDL then shares the one connection with the scan (the bulk load is off there).

**Low: `Acquire(0)` counted as a timeout.** Added `ConnectionPool::TryAcquire()`. It never waits and is not counted in `acquire_timeout_count`, and the extra writers use it. A case in `test_pool_creation_failure.cpp` pins it: on a full pool `TryAcquire` returns at once and leaves the counter at 0, while `Acquire(0)` on the same pool still counts.

**Low: `Unavailable` terminal.** Split it. A new `Claim::Busy` means a slot was free but no connection was. It is transient: the slot is handed back, the thread shares the global writer and asks again on a later chunk, which is cheap now that the acquire never waits. `Unavailable` stays for a reached slot cap or a refused bulk load. COPY, CTAS and INSERT all handle `Busy`, and the header contract and the two stale "blocking Acquire" comments are rewritten.

Docs updated to match: CLAUDE.md, DATAMODEL.md, CHANGELOG.

oluies (approved):
## Approved

`7588e29` (plus the merge of main at `8abf7b4`). All four round-2 findings are fixed, and in two places better than the fix I had written locally — I dropped mine rather than push a duplicate.

**Verified in the branch:**

- **Medium, the guard's condition:** now `in_transaction || one_connection`, with `one_connection = GetConnectionLimit() <= 1` — the optimizer's own condition, so the autocommit pool-of-one case is covered.
- **Low, the lock's scope:** better than what I had. You keep the lock through the DDL **when the pool has one connection** (where the DDL does share that connection) and release it right after the two probes otherwise. I had released it after the DDL unconditionally, which holds it over the pool-connection DROP/CREATE for no reason. Asking `SchemaExists` only when the table did not already answer it is a nice extra.
- **Low, the counter:** `ConnectionPool::TryAcquire` as a real API through `AcquireImpl(0, failure, optional)` is the right shape — I had special-cased `timeout_ms == 0` inside `Acquire`, which leaves the next caller of `Acquire(0)` to rediscover the rule. `test_pool_creation_failure.cpp` pins it.
- **Low, the claim:** `Claim::Busy` is handled at all three call sites (COPY, CTAS, INSERT), with `Unavailable` left for a reached cap or a refused load.

**CI on this head:** Build, Integration Test, Concurrency stress, both C++ unit jobs and the docker stack all pass.

One note for whoever merges: `MaterializeSharedConnectionScans` now carries both changes — your autocommit single-connection branch and the `case_insensitive_map_t` key from #384 — and they compose, since `Catalog::GetCatalogEntry` is case-insensitive too.

oluies:
## roborev review — `4b8ea3b..8abf7b4` (all 4 commits on this branch)

Verdict: **Fail** — 8 findings (4 medium, 4 low).

Findings 1 and 4 are the same omission seen from two sides: the autocommit pool-of-one materialisation was wired for catalog scans + CTAS, but not for COPY-BCP sinks or raw `mssql_scan` sources — which is the shape `DATAMODEL.md` and the CHANGELOG now claim works.

---

### Medium

**1. `src/copy/copy_function.cpp:310` (gate at `:328-330`) — COPY BCP missed the pool-of-one fix**

The optimizer now sets `requires_materialization` in autocommit whenever the plan sinks into a single-connection catalog (`mssql_optimizer.cpp:799`), and `CollectSinkCatalogs` has always counted a `bcp` COPY as a sink. But `BCPCopyInitGlobal` acquires the pool's only connection *before* the mutex block and takes `MaterializeMutex` only `if (gstate->transaction_pinned)` — never for the new autocommit pool-of-one case. CTAS got both halves of the fix (`|| one_connection` on the lock, plus `use_bcp = false` so the rows don't hold the connection from init); COPY got neither.

On `mssql_connection_limit = 1`, autocommit:

```sql
COPY (SELECT … FROM db.dbo.src) TO 'mssql://db/dbo/dst' (FORMAT 'bcp');
```

COPY's `InitGlobal` and the scan's `InitGlobal` race for the one connection with no ordering, and whichever loses blocks for the full `mssql_acquire_timeout`. If COPY wins, it holds the connection for the whole load and the source scan can never run. `transaction_single_connection_pool.test` exercises CTAS and `INSERT … SELECT` on a pool of one, but no COPY — so the gap is untested.

*Suggested fix:* give `BCPCopyInitGlobal` the same condition CTAS uses — take `MaterializeMutex` when `gstate->transaction_pinned || mssql_catalog.GetConnectionLimit() <= 1`, and take it *before* `ConnectionProvider::GetConnection` — or, since a bulk load holds its connection from init, refuse/downgrade the BCP path on a pool of one the way `ResolveConnectionMode` does for CTAS. Add the COPY case to the single-connection test.

**2. `src/copy/bulk_load_session.cpp:147` — `TryAcquire` conflates "busy" with "dial failed"**

`TryAcquire()` does not only skip the wait: `AcquireImpl` still runs the `stats_.total_connections < config_.connection_limit` branch, which calls `CreateNewConnection` (a full TCP + TLS + LOGIN7 dial, bounded only by `mssql_connection_timeout`, default 30 s) with the pool mutex released. More importantly, a *creation failure* now returns `nullptr` → `Claim::Busy`, and `Busy` is explicitly non-terminal: `copy_function.cpp:750` and `mssql_physical_ctas.cpp:321` `break` without clearing `may_claim`. Previously the blocking `Acquire()` returning `nullptr` threw and produced `Claim::Unavailable`, which set `may_claim = false` and stopped the thread asking.

So against a server refusing new logins (expired token, server-side connection cap, firewall) every extra-writer thread re-dials once per `create_backoff_ms_` window (capped at 4 s) for the entire load, each dial blocking the sink thread for up to the connection timeout. The failure string is also discarded (`TryAcquire()` is called with no `failure` out-param), so none of this is visible.

*Suggested fix:* distinguish the two nulls — have `TryAcquire` report whether it failed to *create* (terminal → `Unavailable`, with the reason logged) versus found nothing free under the limit (transient → `Busy`). At minimum pass a `std::string *failure` and return `Unavailable` when it is non-empty.

**3. `src/mssql_functions.cpp:1046` — `MarkChangedLocally` wipes the transaction cache on a loose substring match**

`NoteTransactionChangeLocally` runs on the **default** setting (`mssql_exec_invalidate_cache = false`), and `MarkChangedLocally()` clears `entries_`, `absent_`, `listed_schemas_` and calls `cache_->Invalidate()` — i.e. it throws away the whole transaction's metadata cache and forces every subsequent bind in the transaction to reload on the pinned connection.

Its trigger is `ExecSqlMayChangeSchema`, a raw case-insensitive substring match over the whole statement text for `CREATE/DROP/ALTER/TRUNCATE/RENAME/EXEC` (`mssql_functions.cpp:949-958`). That matches `'UPDATE dbo.Orders SET status = ''created'''`, a column named `dropped_at`, `sp_executesql`, and every stored-procedure call. Before this change that branch did nothing at all when the setting was false, so the heuristic's false positives were free; now each one costs the transaction its cache, and a loop of `mssql_exec(…, 'EXEC p')` inside `BEGIN … COMMIT` re-wipes it on every call.

*Suggested fix:* scope the local mark to the names the statement plausibly touches, or gate it on a tighter test (leading keyword after whitespace/comments rather than `find()` anywhere), and make `MarkChangedLocally` set a "distrust shared" flag without dropping what the transaction already loaded itself — those entries were loaded on the pinned connection and are still valid.

**4. `src/table_scan/mssql_optimizer.cpp:799` + `src/mssql_functions.cpp:620-622` — raw `mssql_scan` is not materialised in autocommit pool-of-one**

The autocommit pool-of-one path only sets `requires_materialization` on `tally.catalog_scans`; raw `mssql_scan` / `mssql_scan_params` scans are counted into the tally but never flagged, and `MSSQLScanInitGlobal` materialises only `if (in_transaction)`. So

```sql
CREATE TABLE db.dbo.t AS SELECT … FROM mssql_scan('db', '…');
```

(or the COPY / INSERT equivalent) on `mssql_connection_limit = 1` in autocommit still streams the source on the pool's only connection while the sink asks for it, and waits out `mssql_acquire_timeout` — the exact shape `DATAMODEL.md` and the CHANGELOG now claim works ("in autocommit … a scan of the catalog it writes into is materialised first").

*Suggested fix:* materialise raw scans at `InitGlobal` when the catalog's pool limit is 1 as well as in a transaction — change the `in_transaction` test at `mssql_functions.cpp:620` to include `mssql_catalog.GetConnectionLimit() <= 1`, or propagate a flag from the optimizer as the catalog-scan path does. Either way, extend the single-connection test with a raw-scan source.

---

### Low

**5. `src/tds/tds_connection_pool.cpp:178` — optional probes inflate `acquire_count`**

`AcquireImpl` increments `stats_.acquire_count` before the `optional` flag is consulted, so every optional probe counts as an acquire. `TryStart` is called once per chunk per thread while `may_claim` is set, so a parallel COPY/CTAS/INSERT on a busy pool inflates `acquire_count` (surfaced by `mssql_pool_stats`, `mssql_diagnostic.cpp:149`) with attempts that were never real acquisitions, and skews any `acquire_wait_total_ms / acquire_count` reading. Same class of misleading diagnostic the `optional` flag was added to remove for `acquire_timeout_count`.

*Suggested fix:* guard the `acquire_count` increment with `if (!optional)` too, or add a separate `try_acquire_count`.

**6. `src/dml/ctas/mssql_ctas_executor.cpp:255-261` — BCP downgrade is broader than its rationale**

`ResolveConnectionMode` turns BCP off for *any* catalog with `connection_limit <= 1`, including autocommit and including plans whose source never touches this catalog (`CREATE TABLE db.dbo.t AS SELECT * FROM 'local.parquet'`). The stated reason — "a bulk load takes its connection before the source scan initialises" — only applies when the source *is* a scan of this catalog; with a local source there is no contention and the user silently loses the documented 2-10x BCP speedup by setting `mssql_connection_limit = 1`.

*Suggested fix:* restrict the downgrade to the case the comment describes — keep BCP unless the statement is inside a transaction, or the plan's source scans this catalog (the optimizer already computes that in `by_catalog` / `sink_catalogs` and could stash it on the bind data).

**7. `src/catalog/mssql_refresh_function.cpp:84`, `src/catalog/mssql_preload_catalog.cpp:89` — refusal is too broad**

The refusal tests `state.GetContext().transaction.IsAutoCommit()`, which is false for *any* open DuckDB transaction — including one that has never touched the MSSQL catalog being refreshed (or any MSSQL catalog at all). A script that wraps unrelated DuckDB work in `BEGIN … COMMIT` and calls `mssql_preload_catalog('db')` now gets a hard `InvalidInputException` where it previously worked correctly, since with no pinned connection there is no uncommitted state to publish.

*Suggested fix:* scope the check to the catalog actually being refreshed, e.g. refuse only when `ConnectionProvider::IsInTransaction(context, catalog)` (or the transaction has a pinned connection / recorded changes for it).

**8. `src/catalog/mssql_catalog.cpp:534` — `LookupSchema` fast path bypasses the transaction-aware cache selection**

`LookupSchema`'s fast path reads `metadata_cache_` (the shared cache) directly and returns before `SchemaListCache()` is ever consulted, so the new transaction-aware selection is bypassed on the hit path. A transaction that dropped a schema through `mssql_exec` (uncommitted, so the shared list still has it) gets a positive hit and a live `MSSQLSchemaEntry` back. The same split exists in `ScanSchemas`, which *does* go through the helper — the two now disagree about which cache is authoritative inside a transaction.

*Suggested fix:* route the fast path through `SchemaListCache(transaction.context.get())` as well, so there is one answer for "which schema list does this caller read".

---

<sub>Generated by roborev (job 1730, agent `claude-code`) and relayed by Claude Code. Findings are unverified against a running SQL Server — treat the reproductions as claims to check.</sub>

VGSML:
Thanks. All 8 findings are addressed in #388, which is opened from main because this PR is merged. Findings 1 and 4 reproduced on the server before the fix (`pool 'r1' timed out (1 active of 1, limit 1)` for COPY, and for a CTAS from `mssql_scan`). Finding 6 is covered by the same change as 1: the bulk load keeps BCP and takes its connection on the first chunk. On finding 3 I kept dropping what the transaction had already loaded, because DDL we cannot see through may have changed exactly those tables; the fix is a tighter, word-based detection. The details are in #388's description and its self-review.

PR #384, opened 2026-09-23
fix: the five open roborev findings that were still live on main

oluies:
Eighteen open failing roborev reviews (jobs 1634-1662), each finding checked against **main** rather than against the commit it was written on. Thirteen turned out to be answered already by later commits and were closed with that evidence. These five were still live.

## What was still live

**The scan tally keyed catalogs by case** (`mssql_optimizer.cpp`). Its two producers spell the key differently — a catalog scan carries `MSSQLCatalog::GetContextName()`, a raw `mssql_scan` the user's literal first argument — while DuckDB's catalog lookup, and the sink set right beside this map, are case-insensitive. Now `case_insensitive_map_t`.

**Honest about what this one is.** The review argued that a mixed-case alias splits the tally and lets two scans collide on the pinned connection. I built that shape and could not make it fail without the fix: a raw scan materialises at InitGlobal unconditionally inside a transaction (spec 075), so the two never hold the connection at once. Measured both ways on a debug build. It is a consistency fix, and `test/sql/transaction/transaction_mixed_case_alias_scans.test` says exactly that in its header instead of posing as the guard.

**Every bind-time connection failure of a `prepared := true` scan was reported as pool exhaustion** (`mssql_functions.cpp`). `GetConnection` also throws for a refused dial or a login failure — since #302 it carries the server's own words — so a wrong password answered "Raise `mssql_connection_limit`, or drop `prepared := true`", and the exception type was flattened to `IOException` on the way. Only a timeout gets that advice now.

**The prepared-session claim tested only the `in_use` flag**, not whether the session's connection is usable. `~MSSQLResultStream` closes a connection that a cancelled stream left out of `Idle`, and that connection belongs to the prepared session — so claiming it sent `EXEC sp_execute` down a `Disconnected` connection, the one case the ad-hoc fallback exists for. A session whose connection is not `Idle` now reads as a lost claim.

**The COPY refusal promised the wrong outcome** (`copy_function.cpp`). An uncast `NULL AS col` does not "leave the column NULL": the column is dropped from `INSERT BULK` exactly like an omitted one, so the **server** fills it — with NULL, with its `DEFAULT`, or with a generated value for a `rowversion`, and a `NOT NULL` column without a default fails there. The "but a TYPED null is not" clause is kept verbatim, because `copy_udt_columns.test` asserts that substring.

**`MSSQLInsertColumn::max_length` defaulted to 0** (`mssql_insert_config.hpp`). For a `char`/`varchar` column under a non-UTF-8 collation `NVarcharLength` renders `CAST([c] AS NVARCHAR(0))`, which SQL Server rejects with what reads like a syntax or collation error. The parameter is required now. No caller used the defaulted form.

## Not done, deliberately

Job 1634 also asked for a test that forces a duplicated plan node over a single prepared scan. That needs a hook the tree does not have, and a test that passes either way would be worse than none.

## Tests

copy 1129, query 644, insert 560, catalog 2022, rowid 311, dml 258, transaction 348, ctas 476 assertions green against the docker server on a debug build. clang-format-14 clean.


VGSML (commented):
## Review: 1 Medium (verified), 2 Low, 2 Nits

I read the full diff against main (`ab3de6d`). Thank you for checking each finding against main and for being candid about what the tally test does and does not guard. The COPY message and the required `max_length` are good as they are. The substring `copy_udt_columns.test` asserts ("a TYPED null is not") is still there.

### Medium: `mssql_functions.cpp`, the "exhausted" test also matches a dial timeout

`reason.find("timed out")` is true for the pool's exhaustion message, `pool 'x' timed out (N active of M, limit L)`. It is also true for a **creation failure caused by a TCP timeout**: the socket's `last_error_` is literally `Connection timed out` (`tds_socket.cpp:232`), and the factory and pool wrap it as `… could not create a connection: TCP connect to host:1433 failed: Connection timed out`.

I checked the text on a current build: `ATTACH 'Server=10.255.255.1,1433;…' (lazy_validation true, catalog false)`, `SET mssql_connection_timeout = 2`, and `mssql_scan` fails with an error matching `could not create a connection.*timed out`. With `prepared := true`, an unreachable host therefore still gets "Raise mssql_connection_limit". That is the one misattribution this change exists to remove, and an unreachable host is a common way to meet it.

*Fix:* match the pool's own exhaustion shape, not the words. `DescribeTimeoutLocked` always renders `' timed out (`, so `reason.find("' timed out (") != string::npos` separates the two. Alternatively, exclude `could not create a connection` explicitly.

*Test:* the reclassification has none, and both sides are easy to reach:
- negative: a lazy ATTACH to `10.255.255.1` with a short `mssql_connection_timeout`, then a `prepared := true` scan. The error names the timeout and not `mssql_connection_limit`;
- positive: `mssql_connection_limit = 1` with two prepared scans in one statement. This one does get the advice.

### Low: `reason` comes from `e.what()`

On the 2.0 line a DuckDB exception's `what()` is its JSON serialisation. `CreateNewConnection` already switched to `ErrorData(e).RawMessage()` for that reason: "the first version of this pasted `{"exception_type":"Connection",...}` into the user's error". `GetConnection` throws `IOException`, so the rewritten message embeds the JSON. Both the match and the embedded text should use `ErrorData(e).RawMessage()`. The line was already like this; it's being touched anyway.

### Low: `mssql_optimizer.cpp`, the comment states the collision as fact

The new comment says the split tally meant the two scans "then collided on the pinned connection — the #239 shape". The PR description and the new test's header both say the opposite, measured both ways: the raw scan materialises at InitGlobal inside a transaction, so they never hold the connection at once. The comment should say what the header says, a consistency fix, so the next reader doesn't go hunting for a crash that can't happen.

### Nit: the `InternalException` guard in `MSSQLScanInitGlobal`

By its own comment it is unreachable: "the two are set together today". The house rule on defensive branches is to name the case and prove it reachable, or leave it out. If the invariant needs protecting, a `D_ASSERT(!bind_data.execute_sql.empty() == …)` at bind time says it where the pair is set, instead of a runtime branch at init.

### Nit: the test file
- It uses the deprecated `'${MSSQL_TESTDB_DSN}'` form (the runner prints "Replacing deprecated"); new files use `'{MSSQL_TESTDB_DSN}'`.
- It sets up 17k rows for a test that, as the header says, passes on both builds. That is fine as an end-to-end pin, but the row count only matters for a guard. A few hundred would do and save CI time.

### Merge coordination

#382 also changes `MaterializeSharedConnectionScans`. It adds an autocommit branch for a pool of one connection that looks each tally key up with `Catalog::GetCatalogEntry`. That lookup is case-insensitive, so it composes with `case_insensitive_map_t`. Whichever of the two merges second needs a trivial rebase of that function.

Not covered and not asked for: the prepared-session `Idle` check has no test (a cancelled stream followed by a re-execution of the same prepared statement). Fine to leave as it is, since the change is small and obviously right.

VGSML:
@oluies, before merge could you address two items from the review above, each with a test? The rest (the optimizer comment and the nits) is up to you.

1. **Medium: the exhaustion test.** Match the pool's own shape, `reason.find("' timed out (")`, which `DescribeTimeoutLocked` always renders, instead of `"timed out"`. As it stands, a dial timeout to an unreachable host (`… could not create a connection: TCP connect … Connection timed out`) still gets the "Raise mssql_connection_limit" advice.
2. **Low: `ErrorData(e).RawMessage()` instead of `e.what()`**, both for the match and for the text embedded in the message, so no JSON ends up in the user's error.

Tests:
- **Negative:** a lazy ATTACH to `10.255.255.1` with `mssql_connection_timeout = 2` and `catalog false`, then a `prepared := true` scan. The error names the timeout and does **not** mention `mssql_connection_limit`.
- **Positive:** `mssql_connection_limit = 1`, with two `prepared := true` scans in one statement. The error **does** carry the advice.

Thanks!

oluies:
Pushed `df7b382`. Both requested items are done, each with the test you asked for, and the optional points too.

**Medium — the exhaustion test.** You were right, and the flaw was in my fix rather than in the finding. The match is now `' timed out ('`, the shape only `DescribeTimeoutLocked` renders.

**Low — `ErrorData(e).RawMessage()`** for both the match and the embedded text.

**Tests**, in `mssql_scan_prepared.test`, both measured against this build:

- **Negative:** a lazy ATTACH to `10.255.255.1` with `mssql_connection_timeout = 2`. The error is now `MSSQL: Failed to acquire connection: pool 'pr75_dead' could not create a connection: TCP connect to 10.255.255.1:1433 failed: Connection timed out` — no advice, no JSON. The assertion is **anchored at the start** of the message, because the advice replaces that prefix and RE2 has no negative lookahead, so a prefix match is what actually rules the advice out.
- **Positive:** `mssql_connection_limit = 1` with two prepared scans in one statement does get the advice. One thing worth recording: both settings have to be applied **before** the ATTACH, since the pool takes its limit then. My first version set them after, the pool stayed at 64, and the second bind simply succeeded — the test now says so.

**The optional points, all taken:**
- The optimizer comment no longer states the collision as fact; it says what the PR description and the test header say.
- The unreachable `InternalException` is gone, replaced by a `D_ASSERT` where `fallback_sql` and `execute_sql` are set.
- The test file uses `'{MSSQL_TESTDB_DSN}'` and a few hundred rows instead of 17k. The expected count moves with it, 4997 → 397 (computed, then confirmed against the server — the first trim left the old number and the suite caught it).

Not done, as you said it was fine to leave: no test for the prepared-session `Idle` check.

On merge coordination — agreed, and #382 is further along than this one. Merge yours first and I'll rebase `MaterializeSharedConnectionScans` here; the two changes to it compose, since `Catalog::GetCatalogEntry` is case-insensitive too.

**Tests:** query 657, transaction 348, copy 1129, insert 560, catalog 2022, rowid 311, dml 258, ctas 476 assertions green; clang-format-14 clean.

VGSML (approved):
Both items are addressed. The exhaustion test now matches the pool's own shape, and the reason comes from RawMessage. Both sides are pinned by tests (unreachable host: no advice; a pool of one with two prepared scans: the advice). The optional items are done too. Approving; merge once CI is green.

Issue #122, opened 2026-05-22
ATTACH to SQL Server can last for 30s on Windows with multiple network adapters


lucap-irion:
### Summary

On Windows, `TdsSocket::Connect()` gives the **full connection timeout** (default 30s) to each address returned by `getaddrinfo()` sequentially. When a hostname resolves to multiple addresses (common on multi-NIC Windows machines), and the first address(es) don't respond (SYN silently dropped — no RST), the connection hangs for the full timeout before trying the next address.

### Environment

- Windows 11/10 with multiple network adapters (Ethernet, Docker vEthernet, WSL, VPN)
- SQL Server listening on one specific adapter
- Hostname resolves via DNS to multiple IPv4 + IPv6 link-local addresses

### Steps to Reproduce

1. Have a Windows machine with multiple NICs (Docker Desktop, WSL2, or VPN installed)
2. SQL Server running locally, listening on one adapter (e.g., `192.168.1.x`)
3. `ATTACH 'Server=HOSTNAME,...' AS db (TYPE mssql)` — uses hostname, not IP

### Observed Behavior

`getaddrinfo(hostname, port, AF_UNSPEC)` returns 4-9 addresses (mix of IPv6 link-local and multiple IPv4 from different adapters). The current code in `TdsSocket::Connect()` iterates sequentially, giving each address the **full** `timeout_seconds * 1000` ms via `WaitForReady()`.

If the first address is unreachable (TCP SYN silently dropped — common for link-local IPv6 or addresses on inactive adapters), the connection takes `30s × N` before it reaches the working address.

Additionally, the issue is **intermittent** — the same address can succeed on one attempt and timeout on the next (likely due to TCP TIME_WAIT state or Windows Firewall throttling rapid reconnects).

### Expected Behavior

Connection should succeed within 1-2 seconds regardless of how many addresses DNS returns, matching the behavior of Microsoft.Data.SqlClient.

### Root Cause

In `src/tds/tds_socket.cpp`, `Connect()`:
```cpp
for (p = res; p != nullptr; p = p->ai_next) {
    // ... socket + connect ...
    WaitForReady(timeout_seconds * 1000, true);  // full timeout per address!
}
```
### Proposed Fix: Happy Eyeballs (RFC 8305)
Microsoft's own SqlClient (.NET) uses a staggered parallel connect approach since .NET 8:

1. Start non-blocking `connect()` to first address
1. After 250ms, if not yet connected, start connecting to next address in parallel
[...]

Issue #383, opened 2026-09-23
Loads made inside a transaction never reach the shared metadata cache: transaction-heavy workloads pay a metadata round trip per table per transaction


VGSML:
Follow-up to #380 / PR #382 (self-review finding 1).

## What happens

Since #382, a catalog miss inside an explicit transaction is loaded on the pinned connection into the transaction's own cache (`MSSQLTransactionMetadata`), and nothing of it reaches the shared cache. That is deliberate: the shared cache must hold committed state only.

The cost falls on workloads that run almost everything in explicit transactions. DuckLake over an MSSQL metadata catalog is one. For them the shared cache is never populated, so every transaction pays one metadata round trip for every table it touches. Before #382, only the first transaction did, because it loaded into the shared cache on a pool connection.

Measured on the local server, 50 transactions each reading two tables:

| | total |
|---|---|
| shared cache cold (only transactions ever run) | 0.215 s |
| shared cache warmed by autocommit reads | 0.090 s |

That is about 1.4 ms per table per transaction on localhost. Against Azure SQL it is one network round trip per table per transaction.

## Proposed fix: promote committed-state loads

A load the pinned connection makes for a name the transaction has not changed (`!MSSQLTransactionMetadata::IsChanged`, and not `IsAllChanged`) describes committed state:
- the pinned connection sees other sessions' committed DDL;
- it sees its own uncommitted DDL only through names the transaction changed;
- catalog views are not versioned under snapshot isolation.

So such a load can also be written to the shared cache: the entry, the table-list state and the rowid key. Two gaps must close first:

1. **`mssql_exec` DDL in a transaction always marks the transaction's own layer as changed** (`NoteTransactionChange`), whatever `mssql_exec_invalidate_cache` says. Today the change is noted only with the setting on, so untracked uncommitted DDL could be promoted. The marking leaves the shared cache untouched, and the shared cache is what the setting governs.
2. **No promotion under `transaction_isolation = read_uncommitted`**, where catalog reads can see other sessions' uncommitted DDL.

## Tests

- **Round trips:** N transactions reading the same tables cost one metadata load in total, not N. Assert it through `mssql_pool_stats` or the metadata-query fault-injection lever, as in #377.
- **No leak:**
  - a table created by `mssql_exec` in a transaction, with `mssql_exec_invalidate_cache = false` and read in that transaction, is never visible to a second connection or after ROLLBACK;
  - the same holds for a table created by COPY;
[...]

PR #385, opened 2026-09-23
spec(081): the dial charges the connection timeout once per DNS answer (triage of #122)

oluies:
Triage of #122, written up as a spec. **Documentation only — no behaviour change, no code touched.**

## The defect

`TdsSocket::Connect` walks the `getaddrinfo` list in sequence and hands each candidate the full budget:

```cpp
for (rp = result; rp != nullptr; rp = rp->ai_next) {   // tds_socket.cpp:179
    fd_ = socket(...); SetNonBlocking(true); connect(...);
    ...
    if (WaitForReady(true, timeout_seconds * 1000)) {  // :219  <-- full budget, per address
```

There is no overall deadline, so `timeout_seconds` is a per-candidate allowance rather than a ceiling on the attempt. A hostname with N answers and a blackholing first answer costs N × the timeout. The reporter sees 4–9 candidates from one hostname; at the default that is a 120–270s ceiling, and the 30s in the issue title is the cheapest shape of it.

It only bites when a candidate **drops** the SYN. A refusal costs microseconds and the loop moves on — which is why no CI run and no `127.0.0.1` test has ever seen this. A down adapter, a link-local IPv6 answer with no route, or a firewall set to DROP all produce the silent case, and a multi-NIC Windows box manufactures all three. `Microsoft.Data.SqlClient` has not behaved this way since .NET 8.

## What the spec proposes

- **W1** — staggered parallel connect per RFC 8305. 250 ms stagger, in-flight candidates are not abandoned, `getsockopt(SO_ERROR)` is the only success oracle, first socket to connect wins and the rest are closed. A working first address costs nothing, because no second socket is created before 250 ms. `timeout_seconds` finally bounds the whole attempt.
- **W2** — `AI_ADDRCONFIG` in `hints.ai_flags` (currently `0` at `tds_socket.cpp:164`), so the resolver stops returning families the host cannot route. One line; not sufficient alone and the spec says so.
- **W1.2** — one error naming the whole attempt. Today each failing candidate overwrites `last_error_`, so the message names whichever address happened to fail last with no sign that four others were tried. Part of why #122 was hard to diagnose from the outside.

## Three things worth review attention

**The spec claims no speed-up figure, deliberately.** §0 bounds it: nothing here is measured, because the failure needs a multi-NIC Windows host with a blackholing first answer and CI has none. What is verified is that a path exists which spends `N × timeout` where it should spend `timeout`. AC-1 requires one confirmation off real hardware before sign-off, and the reporter has been asked for it on #122.

**§3 rejects caching the winning address.** Tempting, and wrong — the reporter reports the failure as *intermittent*, so a cache would pin the client to an address that was good once, converting an intermittent stall into a permanent one.

**§5 is a retraction, kept on purpose.** The first pass of this triage claimed `mssql_connection_timeout` never reached the pool's dial, leaving users no workaround. That was read off a stale local `main`; spec 073 W3 had already fixed it. So `SET mssql_connection_timeout = 3` *is* a real mitigation in released builds, which makes #122 wrong rather than urgent — a priority call the next reader should not have to rediscover.

## The one open question

§4 leaves a `[NEEDS CLARIFICATION]` for `/speckit-plan`: whether the injectable candidate list is acceptable API surface on `TdsSocket` or should be file-local with a test-only declaration.

It exists because the test blackhole is the hard part. `test/cpp/test_login_routing_hops.cpp:743` already records the measurement: a bound-but-unlistening `127.0.0.1` port **drops** the SYN on macOS (`connect()` sits ~7.8s) but is **refused** on Linux — so the obvious trick exercises the wrong branch on the platform CI runs. Hence an injected address rather than one the kernel is hoped to supply, with case 2 skipping loudly when the environment cannot provide a blackhole. A test that silently runs the refusal branch while claiming to cover the timeout branch is worse than no test.

## Sequencing

Before #324 (parallel pool warm-up) — a warm-up that opens N connections up front multiplies whatever the dial costs.

## Verified

- Every code reference checked against `origin/main` @ `e6ab3a2`, after refetching; the stale-`main` read that produced the retracted finding is the reason each line number in the spec was re-confirmed against the fetched tree.


VGSML (commented):
## Review of spec 081: the diagnosis holds; parts of the surroundings have moved

The core is right. F1 (the full budget per candidate, with no overall deadline) is verified against current main. W1 as RFC 8305 with `SO_ERROR` as the only oracle, and §3's refusal to cache the winning address, are both correct. What needs updating is how the spec places itself relative to work that has since landed or is in flight: #382 (merged as `48c9e0d`) and #386, which implements #324 and #383. There are also two integration points that W1.2 will break unless the spec names them.

### 1. #324 is implemented, not future work, and it does not multiply the dial

The summary says 081 should land "before #324 … a warm-up that opens N connections up front multiplies whatever the dial costs". The implementation in #386 does the opposite for wall clock:
- `ConnectionPool::Prewarm` runs one login **per thread, concurrently**, so N connections cost about one dial, not N (measured: `min_connections 4` in 0.39 s against 0.20 s for one connection).
- It blocks the ATTACH until every login finishes, so today each thread can pay F1's `N_candidates × timeout`, and the ATTACH waits for the slowest thread. That is the same ceiling as a single dial, not a multiple of it.
- W1 fixes it transparently, because every prewarm login goes through the factory and so through `TdsSocket::Connect`.

Suggest: drop the sequencing claim. Add a line that `Prewarm` and `Adopt` (below) reach `Connect` through the pool factories, and that W1 needs no change there. If an ordering is still wanted, it is only this: prewarm N threads × staggered candidates is up to N × candidates sockets at once, which is worth a sentence under AC-4's fd accounting.

### 2. F3 ("ATTACH crosses the dial twice") is gone once #386 lands

#386 changes the four ATTACH validators (SQL, Azure AD, manual token, integrated) to **return** the connection they logged in, and `MSSQLCatalog::Initialize` hands it to the new `ConnectionPool::Adopt` as the pool's first connection. A plain ATTACH is now one dial and one login: 0.40 s → 0.20 s measured.
- With `min_connections > 0` there are two dial phases: the validation, then one parallel prewarm batch.
- The F3 line numbers (`mssql_storage.cpp:1837`) also move in that PR.

Suggest: rewrite F3 as "one dial for a plain ATTACH (#324); validation plus one parallel batch with `min_connections`", or drop it.

### 3. W1.2 collides with `MSSQLTranslateConnectionError`. This needs to be in the spec

Every ATTACH-time dial failure goes through `MSSQLTranslateConnectionError` (`src/mssql_storage.cpp`). That function classifies by **substring**, in this order:
1. `connection refused` / `econnrefused`;
2. **`resolve` / `host` / `enoent` / `name or service not known`**, which maps to "Cannot resolve hostname";
3. `timeout` / `timed out`.

The aggregated W1.2 message as specified names "the host, the number of candidates, and the distinct failures". That text:
- contains `host`, so an all-candidates-timed-out attempt is reported to the user as **"Cannot resolve hostname '…' – check server name"**;
- if any one candidate was refused while the others timed out, is reported as "connection refused" instead.

Either way the translated message says the opposite of what W1.2 set out to say. The spec should decide one of these:
- translate the aggregate in `TdsSocket` itself and have `MSSQLTranslateConnectionError` pass it through (for example, a structured outcome or a recognisable prefix it checks first);
- or pin the aggregate's wording so it avoids the classifier's keywords, with a unit case in `test_connection_error_translation.cpp` for each outcome mix (all timed out; mixed refused and timed out; all refused).

### 4. The pool's exhaustion shape is now an interface

Since #384, `BindDescribedScan` tells "pool exhausted" from "creation failed" by the pool's own rendering, `"' timed out ("` (`ConnectionPool::DescribeTimeoutLocked`). A creation failure reaches the user as `pool 'x' could not create a connection: TCP connect to host:port failed: <socket message>`. W1.2's aggregate becomes that `<socket message>`.
- It must never contain `' timed out (`, or an unreachable host is misread as pool exhaustion and given the "raise `mssql_connection_limit`" advice again.
- A one-line note plus a case in the same translation test would pin it.

### 5. Timeouts after #382

- **Every caller that dials** (the four factories, the validators, the prewarm threads, and the cleanup `DROP` of a failed CTAS) passes its budget through `Connect`, so W1's "whole-attempt deadline" semantics apply uniformly. That is worth one line.
- **Extra bulk-load writers** now use `ConnectionPool::TryAcquire`, which never waits, so a dial never runs *under* an acquire wait there. W1 changes nothing for them.
- **§5's workaround text** (`SET mssql_connection_timeout = 3`) is still correct, and still the right thing to post on #122 now.

### 6. `AI_ADDRCONFIG` has a known trap on loopback-only hosts

glibc does not count loopback as a "configured address" for `AI_ADDRCONFIG`. On a host or container whose only interface is `lo`, `getaddrinfo("localhost", …, AI_ADDRCONFIG)` can return **no** IPv4 or IPv6 answers (`EAI_NONAME` / `EAI_ADDRFAMILY`). Several CI and test shapes here dial `localhost`, among them the Kerberos stack's containers and docker-in-docker lanes.

Suggest W2 say: on `EAI_NONAME` / `EAI_ADDRFAMILY` with the flag set, resolve again without it. Alternatively, skip the flag for a numeric host or `localhost`. And add a test that dials `localhost` with the flag on.

### 7. Proposal: W3, the validation's first-hop address goes first for the ATTACH's own dials

The cheapest fix for the ATTACH-time cost is to stop re-discovering, a few milliseconds later, the address the validation has just found. #386 now performs, inside one ATTACH:
1. the validation dial;
2. `Adopt`, which turns that connection into the pool's first;
3. a parallel `Prewarm` of `min_connections − 1` more.

Step 3 re-resolves the host and re-walks the candidate list, so today every prewarm thread can pay F1 on its own.

**Proposed W3:**
- `Connect` records which candidate won the first hop. `TdsConnection` exposes it; `getpeername` is not usable for this (see the constraints below).
- The catalog hands that address to its factories as a preferred-first hint, for the dials the ATTACH makes itself.

**Why §3 does not forbid this.** §3 rightly refuses a *persistent* cache: the report says the failure is intermittent, so pinning to an address that was good once turns a sporadic stall into a permanent one. W3 is not that. It is:
- **a reorder, not a replacement.** The hinted address goes first and the rest of the `getaddrinfo` list stays behind it as fallback. Under W1 that costs nothing: if the hint answers, no second socket exists before 250 ms; if it died in the meantime, the stagger proceeds as usual. Even before W1 it removes F1 from the prewarm entirely.
- **scoped to the ATTACH.** The validation and the prewarm batch it immediately feeds use the hint. Pool refills minutes or hours later re-race the full list, exactly as §3 requires.

**Two constraints the spec should state:**
- **First hop only.** With login routing (spec 068: Azure SQL gateway redirects, Fabric), the adopted connection's *peer* is the routed node, not the gateway. Feeding that to the next connection's first dial would skip the gateway. The hint must be the address that won the **first** hop, recorded inside `Connect`, not read back with `getpeername` afterwards.
- **An IP-level hint, not a new host.** The host name stays what it was for SNI and certificate verification (spec 074, `HostNameInCertificate`). Only the order of the `sockaddr`s changes.

It reuses the seam §4 already needs. The injectable candidate list and "prefer this `sockaddr` first" are the same parameter, so W3 settles the `[NEEDS CLARIFICATION]` in the same direction: a file-local dial routine that takes an ordered candidate list, with `getaddrinfo` plus the optional hint producing it.

### 8. Line numbers

All references are pinned to `e6ab3a2`. #382 has since moved `mssql_catalog.cpp`: the factories and `Initialize` gained the #380/#383 code. #386 moves `mssql_storage.cpp` (the validators) and `tds_connection_pool.cpp` (`Prewarm`, `Adopt`, `TryAcquire`). Anchoring by function name (`TdsSocket::Connect`, the `hints` block, `WaitForReady`) instead of by line would keep the spec valid through both.

### Smaller

- **§4 `[NEEDS CLARIFICATION]` on the seam.** Prefer file-local plus a test-only declaration, as `test_login_routing_hops.cpp` does with its fakes: `TdsSocket`'s single-fd invariant (W1.1) is worth not widening even for tests.
- **AC-2's "± one stagger interval".** With in-flight candidates never abandoned, the deadline has to close *all* of them at `timeout_seconds`. Worth saying explicitly, so a late candidate cannot extend the attempt.
- **AC-4.** Also assert on the prewarm path (N threads × candidates), which is where a leaked candidate fd multiplies.

None of this changes the design. Items 3, 4 and 6 are the ones that would bite at implementation time. Item 7 is an addition to consider, not a correction.

PR #386, opened 2026-09-23
feat(#383, #324): warm the shared cache after a transaction; open the pool in parallel, adopt the validation connection, preload at ATTACH

### LINKED ISSUE #324: Pool warm-up: a way to open connections up front, in parallel (each TDS login is ~175ms and they are paid serially)

### LINKED ISSUE #383: Loads made inside a transaction never reach the shared metadata cache: transaction-heavy workloads pay a metadata round trip per table per transaction

VGSML:
Closes #383. Closes #324.

Two related "warm-up" changes, both on top of #382 (#380).

## #383: transactions fill the shared metadata cache again

Since #380, a catalog miss inside an explicit transaction is loaded on the pinned connection into the transaction's own cache. The shared cache is never written from inside a transaction; that is the invariant that stops uncommitted state from leaking. The cost falls on a workload that runs everything in transactions, such as DuckLake over an MSSQL metadata catalog: it loaded every table once per transaction.

**The design differs from the proposal written in #383.** Loads made during the transaction are not promoted. Instead, **after COMMIT or ROLLBACK** the tables the transaction loaded or changed are loaded into the shared cache on a **pool** connection (`MSSQLCatalog::WarmSharedCache`). That is committed state, read after the transaction ends, so a long transaction does not publish what it saw at its start. And the promotion's two preconditions stop mattering: untracked `mssql_exec` DDL, and `read_uncommitted`.
- Up to 8 tables of a schema are loaded one by one; more than that, and the schema is preloaded in one round trip (`BulkLoadAll`).
- It never waits for a connection (`Acquire(0)`), runs outside the transaction manager's lock, and swallows its failures, which leave the names invalidated as they already are.
- It is skipped on a pool of one connection.
- A transaction's lookup now also reads the shared **metadata** cache before going to the server (`MSSQLMetadataCache::TryGetLoadedTableMetadata`). A table found there is committed state, and its entry is published into the shared table set under the epoch guard.

Measured on a local server, 50 transactions each reading two tables:

| | total |
|---|---|
| shared cache cold, before | 0.215 s |
| shared cache cold, after | 0.100 s |
| shared cache already warm | 0.087 s |

## #324: open the pool in parallel, reuse the validation connection, preload at ATTACH

- **`mssql_min_connections` never opened anything.** It only kept idle connections from being reaped. `ConnectionPool::Prewarm` now opens them at ATTACH, with one login per thread. The slots are reserved under the lock, so a concurrent `Acquire` cannot overshoot the limit. A failure does not fail the ATTACH, because the eager validation has already vouched for the login. It is skipped under `lazy_validation`.
- **The validation connection becomes the pool's first connection.** The four ATTACH validators (SQL, Azure AD, manual token, integrated) now return the connection they logged in, and `Initialize` hands it to the new `ConnectionPool::Adopt`. It was logged in with the factory's own parameters, so a plain ATTACH is **one** login, not two.
- **New ATTACH options** `min_connections` and `preload`. DuckLake's `METADATA_PARAMETERS` forwards ATTACH options, not settings or function calls. Both options accept the string form it sends. `preload` is refused next to `lazy_validation true` or `catalog false`.

Measured on a local server:

| ATTACH | before | after |
|---|---|---|
| plain | 0.40 s | 0.20 s |
| `min_connections 4` | 0.54 s (4 logins in sequence ≈ 1 s) | 0.39 s |
| `min_connections 8` | 0.63 s | 0.54 s |

The "before" column for `min_connections` is this branch before the validation connection was adopted; on main the setting opens nothing.

## Tests

- **`transaction/transaction_metadata_warm.test`:**
  - a few tables loaded one by one;
  - more than 8 tables, so the schema is preloaded, and an untouched table of that schema is then in the shared cache;
  - after ROLLBACK the warm-up sees committed state;
  - a pool of one gets no warm-up.
  The lever is `mssql_test_fail_metadata_after_rows = 1`: a read that succeeds under it ran no metadata query. Verified by mutation: with the warm-up disabled, the test fails at the first check.
- **`attach/attach_pool_warmup.test`:**
  - `min_connections` in both the option and the string form, and the setting;
  - capped at the pool limit, and nothing opened under `lazy_validation`;
  - a negative value and a non-number are refused;
  - `preload` is checked with the same lever (a control run confirms the lever catches a load without `preload`), and both contradictions are refused.
- **`test/cpp/test_pool_prewarm.cpp`:** the logins really run concurrently (4 × 200 ms in about one), the limit caps them, and a partial failure is counted, gives its slot back, and hands back its reason.

Not covered: the adoption of the validation connection itself. Outside CI it is visible only in ATTACH time and `connections_created == 1`, which also held before. Counting logins server-side needs global counters that parallel tests disturb. `Adopt` has no C++ case either, because the fake connections are not Idle/alive.

Local runs: integration 218/218, C++ standalone tests all pass.

Docs: DATAMODEL.md (Layer 2 prewarm/adopt, Layer 4 warm-up), docs/transactions.md, CLAUDE.md, website (ATTACH options, settings), CHANGELOG.


VGSML:
## Self-review

I read the whole diff against main. Two findings are fixed in the commit above, along with two cosmetic ones. The rest checks out.

**Fixed**
- **`WarmSharedCache` used `Acquire(0)`.** After a transaction the pool being busy is the normal case, and `Acquire(0)` counts that as an acquire timeout in `mssql_pool_stats`. That is exactly what #382's review removed for the extra bulk-load writers. It now uses `ConnectionPool::TryAcquire`.
- **A shutdown rollback ran the warm-up.** When `RollbackTransaction` runs with its context already gone, DuckDB is tearing down, and a metadata round trip, possibly a login, would only delay it for a cache that is about to be destroyed. Skipped now.
- **Cosmetic:** the validators' comments named a method that does not exist (`MSSQLCatalog::AdoptConnection`; the real ones are `AdoptOnInitialize` → `ConnectionPool::Adopt`), and `IntegerAttachOption` had been inserted between `BooleanAttachOption` and its doc comment.

**Checked, no finding**
- **The warm-up runs outside `transaction_lock_`.** The touched set is copied before the transaction is erased, so nothing reads the dead transaction's metadata.
- **Step 2b in `GetEntryInTransaction`** (a table found in the shared metadata cache) runs only for a name the transaction has not changed. Since #382 that includes the local mark from untracked `mssql_exec` DDL. The entry it publishes is committed state, and the publish is epoch-guarded the same way `LoadSingleEntry`'s is. An `Absent` answer comes only from a loaded, unexpired shared table list, and that list has the catalog filters applied at load, so a filtered name is correctly absent.
- **Warm-up after ROLLBACK of a table the transaction created:** `GetTableMetadata` answers "does not exist", and the test asserts the table is gone for both connections.
- **`Prewarm`.** Slots are reserved under the lock before any dial and given back on failure. `CreateNewConnection` catches everything, so no exception escapes a worker thread. If a thread cannot be spawned, that connection is dialled inline. The factories were already called concurrently by refills.
- **`Adopt`** refuses a full pool, a shut-down pool, a connection that is not Idle and a dead one, and closes what it refuses. `Initialize` runs twice per ATTACH (spec 076), but the adopted connection is consumed on the first call.
- **Validation connection vs. pool factory:** the same packet size, UTF-8 request, TLS options, database and application name. After a login-routing hop both are on the routed node, as a factory connection would be.
- **`preload` after `Initialize`:** the collation is known by then, so `BulkLoadAll` reads code pages correctly.

**Known and stated in the description:** adopting the validation connection has no direct test. It is visible only in ATTACH time and `connections_created`, which was 1 before too.

src/mssql_storage.cpp
```diff
@@ -1393,9 +1409,12 @@ void ValidateConnection(MSSQLConnectionInfo &info, int timeout_seconds) {
...
 	}
 
-	// Close the test connection - it will be recreated by the pool
-	conn.Close();
-	MSSQL_STORAGE_DEBUG_LOG(1, "ValidateConnection: validation complete, test connection closed");
+	// Handed to the pool rather than closed (issue #324): it is logged in with
+	// the same parameters the pool factory uses, so the ATTACH's first pooled
+	// connection costs no second login.
+	auto validated = conn_holder;
+	MSSQL_STORAGE_DEBUG_LOG(1, "ValidateConnection: validation complete, connection kept for the pool");
```
oluies: **Blocker — the connection is now adopted after a drain that may not have drained.**

Eleven lines up: `socket->ReceiveMessage(response, 5000);` discards its return value, and `conn.TransitionState(Executing, Idle)` then runs *unconditionally*. `TdsSocket::ReceiveMessage` returns false on a poll timeout and leaves `connected_` true — so both `IsAlive()` and `GetState() == Idle` pass `Adopt`'s gate (`tds_connection_pool.cpp:410`).

Before this PR `conn.Close()` ran one line later, which made a half-read stream harmless. Now, `Encrypt` defaults on (and is mandatory on Azure SQL), so on a loaded server whose `SELECT 1` exceeds the hard-coded 5 s the connection enters `idle_connections_` with unread COLMETADATA/ROW/DONE still on the wire. The very next thing `Initialize` does is `QueryDatabaseCollation()` → `Acquire()` → `TryAcquireIdle()`, which pops exactly that connection and parses the stale `SELECT 1` response as its collation answer: a wrong `database_collation_` / code page for the whole session — which feeds the #361 varchar-parameter decisions — or a token desync that poisons that pooled connection for life.

Identical code at `:1237` (Azure) and `:1315` (manual token). Checking `ReceiveMessage`'s bool and closing on false restores the old safety.

src/tds/tds_connection_pool.cpp
```diff
@@ -398,6 +401,113 @@ void ConnectionPool::Release(std::shared_ptr<TdsConnection> conn) {
...
+	threads.reserve(to_create);
+	for (size_t i = 1; i < to_create; i++) {
+		try {
+			threads.emplace_back([this, &created, &errors, i]() { created[i] = CreateNewConnection(errors[i]); });
+		} catch (...) {
+			// No thread to be had: this one dials it after its own.
+			threads.emplace_back();
+		}
+	}
+	created[0] = CreateNewConnection(errors[0]);
```
oluies: **Blocker — an escaping exception here both terminates the process and leaks the reserved slots.**

`created[0] = CreateNewConnection(errors[0]);` runs outside any `try` while `threads` still holds up to `to_create - 1` joinable `std::thread`s. `CreateNewConnection` (`:555`) catches only `const std::exception &`, and its handler body — `error = ErrorData(e).RawMessage()` — can itself throw `std::bad_alloc`; a factory throw not derived from `std::exception` is not caught at all.

Two consequences:

1. Unwinding destroys joinable `std::thread`s → `std::terminate`, killing the DuckDB process during an ATTACH that should have returned an error. And the live workers are still writing `created[i]` / `errors[i]` into the vectors being unwound — a heap use-after-free.
2. The `stats_.total_connections += to_create` reservation at `:437` is never given back, so `stats_.total_connections >= config_.connection_limit` holds forever and `AcquireImpl`'s create branch (`:229`) is never taken again. Every later statement fails with `pool '...' timed out (0 active of N, limit N)` on a pool holding zero real connections.

The worker lambda at `:449` has the same terminate exposure on its own thread — where the single-threaded `AcquireImpl` path would have propagated the identical throw as an ordinary query error.

An RAII joiner over the spawn..join region plus a `catch (...)` in both the lambda and the inline slot fixes both.

src/catalog/mssql_table_set.cpp
```diff
@@ -487,6 +487,31 @@ optional_ptr<CatalogEntry> MSSQLTableSet::GetEntryInTransaction(ClientContext &c
...
 		}
+		// The shared METADATA cache: committed state, filled by an autocommit
+		// load, a preload, or the warm-up after an earlier transaction (issue
+		// #383). The entry built from it is committed state too, so it is
+		// published into entries_ -- epoch-guarded like LoadSingleEntry, so an
+		// invalidation racing this does not resurrect what it cleared.
+		const uint64_t epoch_at_start = invalidation_epoch_.load();
+		MSSQLTableMetadata cached_meta;
+		const auto cached = catalog.GetMetadataCache().TryGetLoadedTableMetadata(schema_name, name, cached_meta);
+		if (cached == MSSQLMetadataCache::CachedTableState::Absent) {
```
oluies: **Blocker — `CachedTableState::Absent` is a negative existence verdict the autocommit path does not honour, so the same SELECT succeeds outside a transaction and fails inside one.**

`TryGetLoadedTableMetadata` returns `Absent` whenever `schema.tables_load_state == LOADED` and the name is missing (`mssql_metadata_cache.cpp:1772`) — exactly the state this PR's own `preload` option leaves, since `LoadAllSchemasMetadataLocked` sets it for **every** schema (`:1042-1055`), and nothing populates `MSSQLTableSet::names_loaded_` / `is_fully_loaded_` from a preload (only `Scan` does, `:274`).

Repro:

```sql
ATTACH '...' AS db (TYPE mssql, preload true);
-- another session: CREATE TABLE dbo.NewT(id int)
SELECT * FROM db.dbo.NewT;              -- works
BEGIN; SELECT * FROM db.dbo.NewT; COMMIT;  -- Table with name NewT does not exist
```

Autocommit falls through to `LoadSingleEntry` → `GetTableMetadata`, which re-queries on a `schema.tables` miss regardless of `tables_load_state`. `IsTTLExpired` returns false for the default `mssql_catalog_cache_ttl = 0` (`mssql_metadata_cache.cpp:294`), so the verdict never expires — only `mssql_invalidate_cache()` clears it. The `WarmSharedCache` >8 branch creates the same state.

src/catalog/mssql_table_set.cpp
```diff
@@ -487,6 +487,31 @@ optional_ptr<CatalogEntry> MSSQLTableSet::GetEntryInTransaction(ClientContext &c
...
+			return nullptr;
+		}
+		if (cached == MSSQLMetadataCache::CachedTableState::Loaded) {
```
oluies: **Blocker — this entry carries `pk_loaded == false` on every bulk/preload-sourced path and is published into the SHARED `entries_`; `EnsurePKLoaded` then discovers on the transaction's PINNED connection and caches a *failure* permanently in that now-shared entry.**

Only `GetTableMetadata` sets `pk_loaded` (`mssql_metadata_cache.cpp:607`); the header states "The bulk paths do not" (`mssql_metadata_cache.hpp:90-91`).

After `ATTACH ... (preload true)` or a >8-table warm-up, `BEGIN; UPDATE db.dbo.t SET x=1 WHERE id=1;` takes this branch, gets `Loaded` with `pk_loaded=false`, and `entries_.emplace(name, entry)` puts it in the shared map. `EnsurePKLoaded` (`mssql_table_entry.cpp:345`) sees `in_transaction` and runs `RowIdKeyInfo::Discover` on the pinned connection; if that throws — metadata timeout, an already-errored transaction, a killed session — the catch at `:387` sets `pk_info_.exists = false` + `discovery_error` and then `pk_loaded_.store(true)`, deliberately permanent (#350).

That poisoned rowid key now lives in the shared entry, so every **other** session's UPDATE/DELETE on `dbo.t` is refused until `mssql_invalidate_cache()`. Pre-PR an in-transaction entry went to `MSSQLTransactionMetadata::AddEntry` and died with the transaction.

src/catalog/mssql_table_set.cpp
```diff
@@ -487,6 +487,31 @@ optional_ptr<CatalogEntry> MSSQLTableSet::GetEntryInTransaction(ClientContext &c
...
 			}
 		}
+		// The shared METADATA cache: committed state, filled by an autocommit
+		// load, a preload, or the warm-up after an earlier transaction (issue
+		// #383). The entry built from it is committed state too, so it is
+		// published into entries_ -- epoch-guarded like LoadSingleEntry, so an
+		// invalidation racing this does not resurrect what it cleared.
+		const uint64_t epoch_at_start = invalidation_epoch_.load();
+		MSSQLTableMetadata cached_meta;
+		const auto cached = catalog.GetMetadataCache().TryGetLoadedTableMetadata(schema_name, name, cached_meta);
```
oluies: **This branch returns without ever calling `catalog.EnsureCacheLoaded(context)`** — which both `LoadSingleEntry` and step 4 of this same function do — so the session's `mssql_catalog_cache_ttl` is never pushed onto the shared cache, and `TryGetLoadedTableMetadata`'s own `IsTTLExpired` then runs against a `ttl_seconds_` that can still be the constructor default of 0.

`MSSQLCatalog`'s constructor builds the cache with `int64_t cache_ttl = 0` (`mssql_catalog.cpp:91-92`), and only `EnsureCacheLoaded` / `RefreshCache` ever call `SetTTL`.

```sql
ATTACH '<dsn>' AS db (TYPE mssql, preload true);  -- PreloadAtAttach also skips EnsureCacheLoaded
SET mssql_catalog_cache_ttl = 5;
-- workload of BEGIN; SELECT * FROM db.dbo.t; COMMIT;  -- served entirely by this branch
```

`ttl_seconds_` is still 0, `IsTTLExpired` always returns false (`mssql_metadata_cache.cpp:294`), and the ATTACH-time metadata is served forever — a column added on the server is never picked up despite the 5 s TTL the user set. The same session in autocommit would have applied the TTL on its first `GetEntry`. This compounds the `Absent` comment below, whose verdict is gated on the same `ttl_seconds_`.

src/catalog/mssql_table_set.cpp
```diff
@@ -487,6 +487,31 @@ optional_ptr<CatalogEntry> MSSQLTableSet::GetEntryInTransaction(ClientContext &c
...
+					entry = result.first->second;
+				}
+			}
+			if (entry) {
+				return anchor(entry);
+			}
+		}
 	}
 
 	// 3. The table filter hides it.
```
oluies: **The new shared-cache read is placed BEFORE this filter check, and `GetTableMetadata` applies no filter — so a filter-hidden table can become visible inside a transaction and is then published into the shared `entries_` that autocommit's fast path serves.**

Autocommit `GetEntry` checks the filter at step 3 (`:73-83`) *before* ever loading. `GetEntryInTransaction`'s new block runs before the equivalent check here. And unlike `LoadAllSchemasMetadataLocked`, the per-schema bulk loader and `GetTableNames`, `MSSQLMetadataCache::GetTableMetadata` (`:442`) never consults `filter_->MatchesTable`.

Repro: ATTACH with `table_filter '^Sales'`; inside a transaction `CREATE TABLE db.dbo.Other(id int)` → `NoteTransactionChange('dbo','Other')`; at COMMIT `TouchedTables` includes it and `WarmSharedCache` calls `GetTableMetadata(conn,'dbo','Other')` (`mssql_catalog.cpp:1601`), landing it in the shared cache with `columns_load_state = LOADED`. A later `BEGIN; SELECT * FROM db.dbo.Other;` takes the new block, gets `Loaded`, builds an entry and emplaces it into the shared `entries_` at `:506` — after which autocommit's step-1 `entries_` hit serves it too, permanently bypassing the filter.

oluies (commented):
Review at max effort: 10 finder angles plus a gap sweep, with every finding re-verified by reading the post-change source. CI is green (17 pass / 6 skipped), so none of these is a build break — they are runtime and behaviour defects.

**Four I'd treat as blocking:** the adopted validation connection (`mssql_storage.cpp`), `Prewarm`'s terminate-and-leak path (`tds_connection_pool.cpp:455`), and the two `mssql_table_set.cpp` consequences of reading the shared cache inside a transaction — `Absent` as a negative existence verdict, and a PK-less entry published into the shared map where a discovery failure becomes permanent.

---

### Two themes worth naming

**1. The warm-up is opportunistic in intent but unconditional in cost.** It takes a connection (possibly *dialling* one), holds the catalog-wide cache mutex across a round trip, and re-reads metadata the transaction just read — all synchronously inside COMMIT. Filtering `touched` through this PR's own `TryGetLoadedTableMetadata` first, and calling `LoadAllTableMetadata` instead of `BulkLoadAll`, removes most of the `mssql_catalog.cpp:1582/1588/1590/1591/1601` comments at once.

**2. Three copied bodies each dropped a guard the original had.** `PreloadAtAttach` vs `mssql_preload_catalog.cpp:140-171`; `Adopt`/`Prewarm` vs `Release`'s tail; `Prewarm`'s failure bookkeeping vs `AcquireImpl`'s create bookkeeping. Sharing one implementation brings all three guards back for free.

---

### Below the cut — real, lower severity

- The `preload` + `lazy_validation` and `preload` + `catalog false` refusals at `mssql_storage.cpp:1915` sit *after* the eager validation has dialled — and on the Azure path after `BuildFedAuthExtension`, so a user can complete a device-code MFA prompt and *then* be told the options contradict each other. Every other option conflict is raised in the parse loop.
- `WarmSharedCache`'s single top-level `try` means one table's failure abandons every remaining table and schema; `by_schema` is built *outside* that try in a `noexcept` function (`bad_alloc` → terminate); and its catch logs `e.what()` where the house rule is `ErrorData(e).RawMessage()` — `tds_connection_pool.cpp:567` records fixing exactly this, since a DuckDB `what()` is JSON on the 2.0 line.
- `test/sql/catalog/first_touch_round_trips.test:23` now carries a false comment ("the eager validation connection is not pooled") and its `connections_created = 1` assertion passes for the opposite reason, so it no longer distinguishes one pool from two.
- `min_connections` is a one-shot at ATTACH; the reaper treats it as a retention floor but nothing refills it.
- `stats_.total_connections--` at `tds_connection_pool.cpp:471` can underflow if `Shutdown()` races `Prewarm`.
- `DATAMODEL.md:532` says the warm-up uses `Acquire(0)` where the code uses `TryAcquire()` — the code comment three lines above the call says the opposite in as many words. `Adopt` is missing from the pool class diagram although it changes connection ownership.
- `test_pool_prewarm.cpp:67`'s `ms < 2 * LOGIN_MS` wall-clock assertion duplicates the deterministic `max_in_flight` check and will flake on a loaded runner.

---

### Cleared, so they need not be re-walked

clang-format-14 is clean on all 13 changed C++ files; no CLAUDE.md naming or C++11-ODR rule is broken; the `MSSQLCatalog` member reordering is init-order-safe; `Initialize` re-entry is guarded; `threads[i-1]` indexing, `reserve`, and the `joinable()` fallback are correct; `Acquire(-1)` is the house convention; `anchor()` on the dirty-epoch path does not dangle; the `entries_` key spelling cannot diverge (both maps are case-sensitive and keyed identically); a concurrent invalidation cannot be overtaken by a warm-up write (the cache mutex is held across the round trip); tables the transaction created or dropped are correctly excluded by `IsChanged()`; and `Adopt` matches the pool factory on every login-affecting parameter.

The PR also incidentally fixes a pre-existing header/.cpp `ValidateConnection` overload mismatch.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+					schema.first, [&](const string &table, const MSSQLTableMetadata &meta) {
+						statistics_provider_->PreloadRowCount(schema.first, table, meta.approx_row_count);
+					});
+				MSSQL_CATALOG_DEBUG_LOG(1, "WarmSharedCache: preloaded schema '%s' (%llu tables)", schema.first.c_str(),
+										(unsigned long long)table_count);
+				continue;
+			}
+			for (const auto &table : schema.second) {
+				MSSQLTableMetadata meta;
+				if (metadata_cache_->GetTableMetadata(*connection, schema.first, table, meta)) {
```
oluies: **These round trips run inside `GetTableMetadata` / `BulkLoadAll`, both of which hold the catalog-wide `MSSQLMetadataCache::mutex_` across the network call — now on the COMMIT/ROLLBACK path, where no metadata query ran before.**

`GetTableMetadata` takes `std::lock_guard<std::mutex> lock(mutex_)` at `mssql_metadata_cache.cpp:447` and still holds it at the `sp_executesql` batch on `:508`; `BulkLoadAll` does the same at `:1079`.

Session A runs `BEGIN; UPDATE db.dbo.t SET ...; COMMIT;`. The server transaction has already committed, then `CommitTransaction` (`mssql_transaction.cpp:355`) calls `WarmSharedCache`, whose query blocks on a schema lock held by an unrelated SSMS `ALTER`. A's COMMIT does not return for up to `mssql_metadata_timeout` — 300 s by default, and `SET mssql_metadata_timeout = 0` is documented as "no timeout", i.e. forever. Meanwhile **every** other session's `GetTableMetadata` / `EnsureTablesLoaded` / `Invalidate*` / the new `TryGetLoadedTableMetadata` on that catalog blocks on the same `mutex_`.

Because the function is `noexcept` and logs only under `MSSQL_DEBUG`, the user sees a five-minute COMMIT and nothing else.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+		// worth counting in mssql_pool_stats (#382).
+		connection = connection_pool_->TryAcquire();
+		if (!connection) {
+			MSSQL_CATALOG_DEBUG_LOG(1, "WarmSharedCache: no idle connection, skipped");
+			return;
+		}
+		for (const auto &schema : by_schema) {
+			if (schema.second.size() > WARM_TABLES_PER_SCHEMA) {
+				idx_t schema_count = 0, table_count = 0, column_count = 0;
+				metadata_cache_->BulkLoadAll(*connection, schema.first, schema_count, table_count, column_count);
```
oluies: **`BulkLoadAll`'s per-schema publish REPLACES every table's `MSSQLTableMetadata` with a freshly built one — discarding the primary keys spec 076 W2 loaded in the same batch — for every table of the schema, including ones the transaction never touched.**

The staging loop builds `MSSQLTableMetadata table_meta;` from scratch (`mssql_metadata_cache.cpp:1200`) — the comment about seeding from the cache only consults the *local* `staged_tables` map, never `schemas_` — and publishes with `schema.tables[staged.first] = std::move(staged.second);` at `:1291`, so `pk_info` / `pk_loaded` are wiped.

A warm catalog where `dbo`'s 200 tables were each first-touched with their keys (one round trip each, spec 076); any transaction that touches 9 of them commits; `WarmSharedCache` runs `BulkLoadAll(conn, 'dbo')` and replaces all 200 entries with PK-less ones. Every subsequent UPDATE/DELETE/rowid scan on **any** of those 200 tables pays an extra `EnsurePKLoaded` round trip it was not paying before the commit — and it repeats on every qualifying commit.

The `<= 8` branch does not have this problem, because `GetTableMetadata` carries the key.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+		// its limit, or no warm-up at all. This runs on COMMIT's path. TryAcquire,
+		// not Acquire(0): a busy pool is the normal case here, not a timeout
+		// worth counting in mssql_pool_stats (#382).
+		connection = connection_pool_->TryAcquire();
+		if (!connection) {
+			MSSQL_CATALOG_DEBUG_LOG(1, "WarmSharedCache: no idle connection, skipped");
+			return;
+		}
+		for (const auto &schema : by_schema) {
+			if (schema.second.size() > WARM_TABLES_PER_SCHEMA) {
```
oluies: **`BulkLoadAll` has no already-loaded short-circuit (unlike `LoadAllTableMetadata`), so a transaction touching 9 tables of one schema re-runs the full per-schema `sys.objects` query on EVERY commit — including for tables it just DROPPED, which can never load.**

This goes straight to `BULK_METADATA_SCHEMA_SQL_TEMPLATE` (`mssql_metadata_cache.cpp:1128`); only `LoadAllTableMetadata` (`:661-690`) carries the "schema LOADED and every table's columns LOADED → return, no round trip" guard.

Since `touched` is `loaded ∪ changed` and merely-read tables are never invalidated, a **read-only** transaction over an already fully loaded schema pays the whole query at every COMMIT — the file's own comments measure a single schema listing at ~484 ms of server CPU at 200K objects, and it is paid under the cache-wide mutex (see the comment on line 1601). This is exactly the shape of the PR's target workload: a DuckLake metadata catalog has ~19 tables in one schema.

Worse on DROP: `MSSQLSchemaEntry::DropEntry` calls `NoteTransactionChange`, so `BEGIN; DROP TABLE db.stage.t1; ... DROP TABLE db.stage.t9; COMMIT;` — a compaction — triggers a full reload of the whole `stage` schema on the commit critical path, for tables that were just deleted. With ≤ 8 drops it is instead one wasted `SINGLE_TABLE_METADATA` round trip per dropped name, each returning false. Nothing is cached either way. It also loops `BulkLoadAll` per schema, reinstating the O(schemas × objects) pattern spec 071 W2 deleted.

Fix: call `metadata_cache_->LoadAllTableMetadata(*connection, schema.first)`, and drop `changes.tables` from the rollback/drop set.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+				idx_t schema_count = 0, table_count = 0, column_count = 0;
+				metadata_cache_->BulkLoadAll(*connection, schema.first, schema_count, table_count, column_count);
+				metadata_cache_->ForEachTableInSchema(
```
oluies: **This re-stamps `PreloadRowCount` for the WHOLE schema (not just the touched tables) and for pure cache hits, resetting `fetched_at` and the TTL-exempt `from_catalog_metadata` flag on entries `ForgetTransactionChanges` had just invalidated.**

`ForEachTableInSchema(schema.first, ...)` iterates every table of the schema, not `schema.second`, and `PreloadRowCount` (`mssql_statistics.cpp:82-91`) writes `fetched_at = now` + `from_catalog_metadata = true`.

Three concrete effects:

1. `BEGIN; INSERT INTO db.dbo.big SELECT ... -- 1M rows; COMMIT;` — `ForgetTransactionChanges` calls `statistics_provider_->InvalidateTable('dbo','big')` precisely so the next plan re-reads `sys.dm_db_partition_stats`, and this immediately re-seeds it with the metadata query's lagging `sys.partitions` figure.
2. On the listing path `IsCacheValid` returns true unconditionally for `from_catalog_metadata` (`mssql_statistics.cpp:142`), so `duckdb_tables()` / `SHOW ALL TABLES` is pinned to that stale count with **no** expiry.
3. In the `<= 8` branch (line 1601) `GetTableMetadata` may be a pure cache hit with no round trip at all and still republishes its OLD `approx_row_count` with a fresh timestamp — so for a table in a commit loop, `mssql_statistics_cache_ttl_seconds` can never elapse and the planner's cardinality is frozen indefinitely.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+	}
+	if (connection) {
+		try {
+			connection_pool_->Release(connection);
+		} catch (...) {
+		}
+	}
+}
+
+void MSSQLCatalog::PreloadAtAttach() {
```
oluies: **`PreloadAtAttach` copies `mssql_preload_catalog`'s body but drops TWO of its guards.**

`mssql_preload_catalog.cpp` does `EnsureCacheLoaded(client_context)` at `:143` and the `HasUsedAnyMSSQLCatalogInTransaction` refusal at `:131`; this does neither.

**(a) No `EnsureCacheLoaded`, so `mssql_metadata_timeout` is ignored.** It is the only caller of `SetMetadataTimeout(LoadMetadataTimeout(context))` besides `RefreshCache` (`mssql_catalog.cpp:1665/1702`), and the cache ctor's own comment calls its value a pre-ATTACH placeholder (`mssql_metadata_cache.cpp:393`, `DEFAULT_METADATA_TIMEOUT * 1000`). So `SET mssql_metadata_timeout = 5; ATTACH '...' AS db (TYPE mssql, preload true);` blocks the ATTACH for the compiled-in 300 s on a lock-blocked 200K-object catalog with no way to shorten it, and `0` ("no timeout") cannot be expressed. `SetTestFailAfterRows` is likewise unapplied, so the path cannot be fault-injected.

**(b) No #380 in-transaction refusal.** `BEGIN; CREATE TABLE a.dbo.t(id int);` (alias `a`, uncommitted, holding the schema lock on its pinned connection) then `ATTACH '<same DSN>' AS b (TYPE mssql, preload true);` runs the whole-catalog `BulkLoadAll` on b's own pool connection, which blocks on alias a's uncommitted DDL until `mssql_metadata_timeout` — the exact hang the scalar function refuses *by name* ("aliases of one database are independent catalogs and alias A's uncommitted DDL would hang a load on alias B").

`MSSQLAttach` has a `ClientContext` in scope at the call site (`mssql_storage.cpp:1934`); `PreloadAtAttach` simply does not take one. Both guards come back for free if the two entry points share one implementation.

src/catalog/mssql_catalog.cpp
```diff
@@ -1529,6 +1549,97 @@ void MSSQLCatalog::ForgetTransactionChanges(MSSQLTransactionMetadata &metadata)
...
+	for (const auto &key : tables) {
+		by_schema[key.first].push_back(key.second);
+	}
+	std::shared_ptr<tds::TdsConnection> connection;
+	try {
+		// Never wait: an idle connection, or a new one while the pool is below
+		// its limit, or no warm-up at all. This runs on COMMIT's path. TryAcquire,
+		// not Acquire(0): a busy pool is the normal case here, not a timeout
+		// worth counting in mssql_pool_stats (#382).
+		connection = connection_pool_->TryAcquire();
```
oluies: **`TryAcquire` is used here on the stated grounds that it "never waits", but `AcquireImpl`'s creation branch runs BEFORE the `timeout_ms == 0` bail — so a warm-up can perform a full TCP+TLS+LOGIN7.**

`AcquireImpl` tries `TryAcquireIdle`, then `if (stats_.total_connections < config_.connection_limit && ...) { lock.unlock(); conn = CreateNewConnection(error); }` at `:228-233`; the `if (timeout_ms == 0) return nullptr` bail is only reached afterwards at `:301`.

So with `SET mssql_connection_cache = false` (Release closes the pinned connection rather than pooling it), or after `RestoreIsolationLevel` closed it, every `BEGIN; SELECT ...; COMMIT;` dials a brand-new connection purely to warm a cache — bounded by `mssql_connection_timeout` (30 s, and per issue #122 N × that on a multi-homed Windows host). Including on `RollbackTransaction`, the error path where the server may already be unreachable and where, with nothing active, there is no `next_create_allowed_` backoff to space the retries.

It also takes the connection before checking whether anything is actually missing: the only early-out is `tables.empty()`. This PR's own `TryGetLoadedTableMetadata` answers "is this already Loaded?" with zero I/O and should filter `touched` before the pool is touched at all — which would also shrink the mutex-hold, the `BulkLoadAll` reload and the `PreloadRowCount` re-stamp flagged below.

src/tds/tds_connection_pool.cpp
```diff
@@ -398,6 +401,113 @@ void ConnectionPool::Release(std::shared_ptr<TdsConnection> conn) {
 	available_cv_.notify_one();
 }
 
+bool ConnectionPool::Adopt(std::shared_ptr<TdsConnection> conn) {
+	if (!conn) {
+		return false;
+	}
+	std::lock_guard<std::mutex> lock(pool_mutex_);
+	if (shutdown_flag_.load() || stats_.total_connections >= config_.connection_limit ||
+		conn->GetState() != ConnectionState::Idle || !conn->IsAlive()) {
```
oluies: **`Adopt` and `Prewarm` push straight into `idle_connections_` without `Release`'s `!config_.connection_cache` guard, so `mssql_connection_cache = false` no longer means an empty idle pool.**

`Release` at `:383` has `if (!config_.connection_cache || !conn->IsAlive()) { conn->Close(); stats_.connections_closed++; stats_.total_connections--; return; }` — the only read of `config_.connection_cache` in the codebase, and the documented meaning of the setting ("Enable connection pooling").

`Adopt`'s gate here is `shutdown_flag_ || total >= limit || GetState() != Idle || !IsAlive()`, and `Prewarm`'s publish loop (`:482-489`) checks nothing at all — it also skips `Adopt`'s `GetState() == Idle && IsAlive()` precondition entirely.

With

```sql
SET mssql_connection_cache = false;
SET mssql_min_connections = 4;
ATTACH '...' AS db (TYPE mssql);
```

the pool holds 5 idle server sessions (1 adopted + 4 prewarmed) on a configuration that asked for none, `AcquireImpl` reuses them via `TryAcquireIdle` with no cache check, and `CleanupThreadFunc`'s `to_keep = min_connections - active` (`:648`) additionally protects them from idle reaping. Session state the user disabled caching to avoid now bleeds between statements.

src/tds/tds_connection_pool.cpp
```diff
@@ -398,6 +401,113 @@ void ConnectionPool::Release(std::shared_ptr<TdsConnection> conn) {
...
+			ConnectionMetadata meta;
+			meta.connection = std::move(created[i]);
+			meta.connection_id = next_connection_id_++;
+			meta.last_released = std::chrono::steady_clock::now();
+			idle_connections_.push(std::move(meta));
+			stats_.idle_connections++;
+			stats_.connections_created++;
+			opened++;
+		}
+		if (opened > 0) {
```
oluies: **`Prewarm`'s failure bookkeeping diverges from `AcquireImpl`'s twice over.**

Compare `AcquireImpl:241-244` (success) and `:268-279` (failure), whose comment and CLAUDE.md's `mssql_acquire_timeout` row document the 250 ms → 4 s spacing as the fix for #302.

**(a) It never arms `next_create_allowed_` / the `create_backoff_ms_` doubling.** `ATTACH '...' AS db (TYPE mssql, min_connections 4, lazy_validation true)` against a server with an expired Azure AD token: all four logins fail, `creation_failures` is bumped and the reason recorded, but `next_create_allowed_` stays at `time_point{}` and `create_backoff_ms_` at its initial value — so once other connections are active, every later `Acquire` re-dials the dead server on every wakeup. That is the behaviour #302 removed.

**(b) The `opened > 0` branch clears `last_create_error_` even when some logins of the same call failed.** `min_connections 8` against a server that refuses some logins (resource governor, per-user connection cap): 6 open, 2 throw, `opened > 0` so `last_create_error_.clear()` runs, and `SELECT creation_failures, last_create_error, last_create_error_age_ms FROM mssql_pool_stats('db')` returns `2, '', -1` — the counter saying something broke and the two fields that exist to say *what* broke disagreeing with it. `MSSQLCatalog::Initialize` passes no `failure` out-param either, so the reason is lost entirely.

This contradicts the DATAMODEL.md line this PR adds ("a failure is recorded like any other"). `test_pool_prewarm.cpp`'s `TestPartialFailure` only checks the out-param, so it passes.

test/sql/transaction/transaction_metadata_warm.test
```diff
@@ -0,0 +1,201 @@
...
+
+query I
+SELECT count(*) FROM w383.dbo.w383_rb;
+----
+0
+
+statement ok
+ROLLBACK;
+
+statement error
```
oluies: **This section passes unchanged with `WarmSharedCache` deleted entirely** — it exercises `ForgetTransactionChanges` (pre-existing #380 behaviour), not the warm-up. So the PR's headline invariant has no test that can fail.

Delete both `catalog_.WarmSharedCache(touched)` calls from `mssql_transaction.cpp` (`:355`, `:434`) and re-run the file: after ROLLBACK, `ForgetTransactionChanges` already invalidates `dbo.w383_rb` in both the shared metadata cache and the table set, so the following `statement error ... does not exist` and its `con2` copy still pass.

The PR description's mutation check ("with the warm-up disabled, the test fails at the first check") holds only for the *first* section. Nothing in the file distinguishes "the warm-up loaded committed state" from "the warm-up loaded nothing" — there is no case where the warm-up must load a table whose COMMITTED shape differs from the shape the transaction saw. E.g. `ALTER TABLE ... ADD COLUMN` inside the transaction, rolled back, then asserting the shared cache / `DESCRIBE` lacks the column with the `mssql_test_fail_metadata_after_rows` lever armed.

VGSML:
Thanks. Everything is addressed in this PR, and the warm-up is redesigned rather than patched: your theme 1 was right, and the owner chose the shape below.

**The warm-up (#383), redesigned**
- **What:** only tables the transaction loaded or changed that the shared cache **lacks** afterwards. The check is `TryGetLoadedTableMetadata` and costs no I/O. Tables still held, names the table filter hides (both filters), and, after COMMIT, tables the transaction dropped are skipped before the pool is touched. The dropped set is new: `NoteTransactionChange(..., dropped)` from `DropEntry`, and a later re-CREATE of the same name clears it.
- **How much:** synchronous, on COMMIT's path, so it is kept small. At most `WARM_TABLES_LIMIT` = 5 tables, one by one through `GetTableMetadata`, which carries the key. More than that are left invalidated for lazy loading. The per-schema `BulkLoadAll` branch is gone, and with it the PK wipe, the re-run on every commit, the O(schemas × objects) loop and the drop-reload.
- **Connection:** `ConnectionPool::TryAcquireIdleOnly()`, which never logs in. With `connection_cache = false`, or after the pinned connection was closed, there is nothing idle and the warm-up is skipped. The same holds on ROLLBACK against a dead server.
- **Statistics:** `PreloadRowCount` is no longer called from the warm-up, so the counts `ForgetTransactionChanges` invalidated stay invalidated, and there is no re-stamp on cache hits.
- **Failures:** each table has its own `try`, and the loop stops only when the connection is left non-Idle. The whole body, the set building included, is inside the `noexcept` function's `try`. Logging uses `ErrorData(e).RawMessage()`.
- **Mutex / long COMMIT:** the remaining cost is at most 5 single-table loads under the cache mutex, the same cost an autocommit first touch pays. That is the owner's call: synchronous, but small.

**Step 2b (blockers 3 and 4, TTL, filter)**
- It now runs **after** the table-filter check and `EnsureCacheLoaded` (so the session TTL and metadata timeout apply).
- `TryGetLoadedTableMetadata` returns a bool. `Absent` is gone, so a name a loaded listing lacks falls through to the pinned load, as autocommit re-queries it.
- The entry built from the shared cache goes into the **transaction's own layer** (`metadata.AddEntry`), not `entries_`, so a rowid discovery that fails on the pinned connection dies with the transaction.

**Pool**
- **Validation drain (blocker 1):** the three sites share `DrainValidationQuery`. If `ReceiveMessage` returns false the connection is closed, and `Adopt` refuses a closed one. Validation itself still succeeds, as before.
- **`Prewarm` (blocker 2):**
  - the vectors are allocated before the reservation;
  - every dial goes through a `noexcept` wrapper that turns any throw, `std::exception` or not, into a failed login;
  - an RAII `Joiner` joins every started thread;
  - after `Shutdown` the counters are left alone rather than decremented (no underflow).
- **One way into the idle queue:** `Release`, `Adopt` and `Prewarm` all use `PoolIfReusableLocked` (Idle, alive, `connection_cache`, not shut down). `Prewarm` also returns 0 at once with caching off.
- **Bookkeeping:** `RecordCreateSuccessLocked` / `RecordCreateFailureLocked` are shared with `AcquireImpl`. In `Prewarm` success is applied first and failure last, so a partial failure arms the backoff and keeps `last_create_error`, and `creation_failures` agrees with it.
- **`min_connections` refill:** not added, deliberately. A refill from the cleanup thread would dial a dead server every few seconds (the backoff tops out at 4 s, a dial can take `connection_timeout`), and `Shutdown` joins that thread, so DETACH would wait on the attempt. `min_connections` is now documented as opened at ATTACH and kept from reaping, with lost connections re-opened on demand (DATAMODEL, pool layer).

**ATTACH**
- **Option contradictions:** `preload` + `lazy_validation` is refused right after the option loop. `preload` + `catalog false` is refused as soon as the secret, connection string and option have settled it. Both are before validation, so before any dial or MFA prompt.
- **One implementation:** `PreloadAtAttach` is gone. `MSSQLCatalog::Preload(context, schema, caller, …)` serves both `mssql_preload_catalog` and the ATTACH option, with the #380 refusal and `EnsureCacheLoaded`.

**Tests**
- `transaction_metadata_warm.test` checked by mutation:
  - no warm-up at all fails at the first section;
  - no warm-up on ROLLBACK only fails the new section: `ALTER … ADD extra` on the pinned connection, ROLLBACK, then under the lever `DESCRIBE` must succeed and show only `id`;
  - re-introducing the `Absent` verdict fails the new `preload` section (a table created behind the cache is found inside a transaction).
- Also new in that file: more than 5 missing leaves them to lazy loading (the lever fails the next read), a table dropped in the transaction, and a filter-hidden table created in a transaction staying hidden.
- `attach_pool_warmup.test`:
  - both contradictions are refused against an unreachable server;
  - `preload` honours the session's test lever;
  - `preload` is refused inside a transaction that used another alias;
  - `connection_cache = false` gives 0 idle / 0 total after `min_connections 3`.
- `test_pool_prewarm.cpp`: the factories now hand out a loopback-connected Idle connection, so the pool's Idle/alive gate is exercised. The wall-clock assertion is gone; overlap is asserted instead. New cases: a non-standard throw, caching off, `Adopt` of an unconnected connection, and `TryAcquireIdleOnly` never creating one.
- `first_touch_round_trips.test`: the comment now says what the count shows, and what it no longer distinguishes.

**Docs:** DATAMODEL (the pool class diagram now has `TryAcquire` / `TryAcquireIdleOnly` / `Adopt`, the one-way-in rule, the warm-up section and its flowchart), CHANGELOG and CLAUDE.md.

Fixed in `1565143`. Local: integration suite 218 cases / 7242 assertions green (5 skipped), `make test-cpp-run` 33/33.

PR #388, opened 2026-09-23
fix(#380): post-merge review of #382 — pool of one (COPY, raw scans, CTAS keeps BCP), TryAcquire failure is terminal, word-based DDL detection, scoped refresh/preload refusal

VGSML:
Follow-up to #382 (#380). It addresses all 8 findings of oluies' post-merge roborev review, https://github.com/hugr-lab/mssql-extension/pull/382#issuecomment-5795232960.

Findings 1 and 4 were **reproduced on the server before the fix**. On `mssql_connection_limit = 1`, in autocommit, both `COPY … FROM` the same catalog and `CREATE TABLE … AS SELECT * FROM mssql_scan(…)` failed with `pool 'r1' timed out (1 active of 1, limit 1)`.

| # | Finding | Fix |
|---|---|---|
| 1 (M) | COPY BCP missed the pool-of-one fix | COPY's init takes `MaterializeMutex` **before** its connection, uses it for the checks and DDL, then gives it back. The bulk-load session takes the connection on its **first chunk** (`BulkLoadSession::DeferAdoption` / `AdoptDeferred`), after the source has drained. |
| 6 (L) | CTAS dropped BCP on any pool of one | The same deferral replaces the downgrade, so CTAS keeps BCP whatever its source. |
| 4 (M) | Raw `mssql_scan` was not materialised in autocommit on a pool of one | It now materialises at init, or at bind for a statement the server cannot describe, when `in_transaction \|\| limit <= 1`, under the same mutex. |
| 2 (M) | `TryAcquire` conflated "busy" with "dial failed" | `TryAcquire(&failure, &creation_failed)`. A creation failure answers `Unavailable`, logged with the pool's reason, so the thread stops asking. Nothing free answers `Busy`. |
| 3 (M) | `MarkChangedLocally` fired on a substring | New `mssql::SqlMayChangeSchema` (`query/mssql_ddl_detect.hpp`, header-only): the keywords count only as whole words outside `'…'` / `N'…'` literals, `[…]` / `"…"` identifiers and comments (nested `/* */` included). The list is CREATE, DROP, ALTER, TRUNCATE, EXEC, EXECUTE and sp_rename. `EXEC` of a procedure still counts, because a procedure may run DDL. |
| 5 (L) | Optional probes inflated `acquire_count` | They are no longer counted. |
| 7 (L) | refresh / preload were refused in any DuckDB transaction | They are refused only while the transaction has a server transaction open on **that** catalog. The check is `MSSQLCatalog::HasOpenServerTransaction`, which uses `MetaTransaction::TryGetTransaction` so it never creates the catalog's transaction. |
| 8 (L) | `LookupSchema`'s fast path bypassed `SchemaListCache` | The fast path now goes through it, as `ScanSchemas` does. |

On finding 3's second suggestion, keeping what the transaction had already loaded: I deliberately kept the drop. DDL the extension cannot see through may have changed exactly those tables, so entries loaded before it can be stale.

## Tests

- **`transaction_single_connection_pool.test`, pool of one, autocommit:**
  - COPY from the same catalog;
  - CTAS and INSERT from a raw `mssql_scan`;
  - CTAS from a local source, which stays on BCP.
- **`transaction_metadata_pinned.test`:**
  - preload in a DuckDB transaction that has not used the catalog runs, and is refused once it has;
  - the refresh refusal's new wording;
  - `mssql_exec` with a `'created'` literal inside a transaction does not make the transaction distrust the shared cache. The lever is set before `BEGIN`, so a read that leaves the shared entry fails. Verified by mutation: the old substring check fails this case.
- **`test/cpp/test_ddl_detect.cpp`:** 21 cases covering literals, `[…]`, `"…"`, line and nested block comments, `dropped_at`, `executions`, `@exec` and `#create_tmp`.
- **`test/cpp/test_pool_creation_failure.cpp`:** `creation_failed` is set for a refused creation and not for a full pool, and `TryAcquire` leaves `acquire_count` alone.

Local runs: integration 216/216, C++ standalone tests all pass.

Docs: CLAUDE.md, DATAMODEL.md, docs/transactions.md and CHANGELOG, corrected where #382 described what this changes (CTAS on a pool of one "loads with INSERT statements"; "refused inside a transaction").


VGSML:
## Self-review

I read the whole diff after opening the PR. Two findings are fixed in the commit above; everything else checks out.

**Fixed**

1. **Regression: COPY into a `#temp` on a pool of one, in autocommit.** The deferred bulk-load connection gave the pool's only connection back between COPY's DDL and its load. The pool resets a connection before its next use, and that reset dropped the `#temp` the init had just created: `INSERT BULK failed: Invalid object name '#t388b'`, reproduced on the server. A temp target (`#` or `##`) now keeps its connection from init to the end of the load, as on any pool. After the statement, the table goes with the next reset, exactly as on a larger pool; a `#temp` meant to outlive the COPY is loaded inside a transaction, whose pinned connection is not reset. The regression is pinned in `transaction_single_connection_pool.test`.
2. **Under-detection in the new DDL scanner.** A batch that starts with `sp_executesql N'DROP TABLE t'`, with no `EXEC`, runs dynamic SQL whose text is a literal the scanner skips. The old substring match caught it through the `EXEC` inside `sp_executesql`. `SP_EXECUTESQL` is now in the keyword list, with a case in `test_ddl_detect.cpp`. `mssql_exec_params` still passes the caller's statement text, not its own wrapper, so it does not trip on it.

**Checked, no finding**

- **COPY's init now takes `MaterializeMutex` before `GetConnection`, in a transaction as well.** Inside a transaction `GetConnection` returns the pinned connection, and a scan draining under the mutex is using that same connection. Waiting for the drain first is the safer order. It cannot deadlock, because the scan side takes the mutex before any connection too.
- **The deferred adoption happens under `write_mutex` in COPY's sink and under the global sink mutex in CTAS.** On a pool of one the writer limit is 1, so every thread uses the shared session. By the first chunk the source's init, which materialises and releases, has already run.
- **A deferred session that never adopted (zero rows):** `Finish` is a no-op (`CloseStream` and `Commit` check the writer and the transaction, and `ReleaseConnection` checks the connection). CTAS's `FlushInserts` skips it, and COPY's finalize logs "no BCP stream".
- **CTAS on a pool of one, in a transaction:** still `single_connection` — rows as statements on the pinned connection, ROLLBACK undoes the table. Only autocommit defers.
- **`TryAcquire`'s `creation_failed`:** it is set wherever the call itself tried to create and failed, including the nothing-active branch that returns early. While another thread's failure has the pool in backoff, the call makes no attempt and answers `Busy`, so a thread makes at most one extra dial after the backoff before `Unavailable` stops it.
- **`HasOpenServerTransaction`:** it uses `TryGetTransaction`, so resolving the catalog in refresh or preload never creates the catalog's transaction. It keys on the pinned connection, and a transaction that answered everything from caches has none, so it has no server-side state for a refresh to race.

Local runs: integration 216/216, C++ standalone tests all pass.

oluies (approved):

oluies (approved):

